PR #2, opened 2023-03-22
refactor integrate qp2p minimal

b-zee:
- fix: allow logging via RUST_LOG
- refactor: make main itself async
- refactor: strip out public addr
- refactor: integrate qp2p and strip down

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,7 +34,7 @@ async fn start_node(my_addr: SocketAddr, peers_addrs: BTreeSet<SocketAddr>) {
         .map(|p| NetworkNode { addr: p })
         .collect();
 
-    let (sender, receiver) = Comm::new::<StableSetMsg>(my_addr, None).expect("Comms Failed");
+    let (sender, receiver) = Comm::new::<StableSetMsg>(my_addr).expect("Comms Failed");
 
     println!("Run stable set with peers {peers:?}");
     run_stable_set(sender, receiver, peers).await
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,7 +48,7 @@ fn start_node(my_addr: SocketAddr, peers_addrs: BTreeSet<SocketAddr>) {
 }
 
 fn main() {
-    tracing_log::env_logger::init();// TODO fix this later
+    tracing_subscriber::fmt::init();
 
     let (my_addr, peers_addr) = get_config();
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -5,8 +5,7 @@ use crate::comms::{Comm, NetworkNode};
 use crate::stableset::{run_stable_set, StableSetMsg};
 
 use std::collections::BTreeSet;
-use std::{env, fs, net::SocketAddr, time::Duration};
-use tokio::runtime::Runtime;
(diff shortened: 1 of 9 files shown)


src/main.rs
```diff
@@ -1,15 +1,183 @@
...
+use std::collections::{BTreeMap, BTreeSet};
+use std::sync::Arc;
+use std::{env, fs, net::SocketAddr};
 
 const PEERS_CONFIG_FILE: &str = "peers.json";
 
+mod net;
+
+#[derive(Clone)]
+struct MyComm {
```
grumbach, line 22: I think this network specific code should be in another module, not in main. Don't you think?

src/main.rs
```diff
@@ -1,15 +1,183 @@
...
+            None => return None,
+        };
+
+        // TODO: Abort because connection is probably broken.
+        let (mut send, recv) = connection.inner.open_bi().await.unwrap();
+
+        // TODO: Handle unhappy paths.
+        let _ = send.write_all(&data).await;
+        let _ = send.finish().await;
+        let data = recv.read_to_end(5000).await.unwrap();
```
grumbach, line 71: Is this blocking?

src/main.rs
```diff
@@ -27,30 +195,49 @@ fn get_config() -> (SocketAddr, BTreeSet<SocketAddr>) {
...
+        comm.establish_connection(p).await;
 
-fn main() {
-    tracing_log::env_logger::init();// TODO fix this later
+        tracing::trace!("sending `Ping` to `{p}`");
+        let msg = NetworkMsg::<StableSetMsg> {
+            id: MsgId::new(),
+            payload: StableSetMsg::Ping,
+        };
+        let abc = comm.send_bi(p, msg.to_bytes().unwrap()).await;
```
grumbach, line 229: Is this blocking until we get the response?

src/main.rs
```diff
@@ -27,30 +195,49 @@ fn get_config() -> (SocketAddr, BTreeSet<SocketAddr>) {
...
-    println!("Shutting Down Runtime for {}", my_addr);
-    rt.shutdown_timeout(Duration::from_secs(2));
-}
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::init();
+
+    let (addr, peers) = get_config();
+
+    let comm = start(addr).await;
```
grumbach, line 218: Good one!

grumbach (commented):
Big PR, impressive work here, lots of code, I'm not sure I understand it all so couple of questions:

> fix: allow logging via RUST_LOG

Nice one

> refactor: make main itself async

Why did we need main to be async? 

> refactor: strip out public addr

Why not?

> refactor: integrate qp2p and strip down

What's the difference with the previous integration? 
Nice one on the simplification of `start`. 
Do we still have the received msg queue that we can read from without blocking (Some, None)? 
Could you comment on the threads (spawns), not really sure what stays alive in the background, what dies? 

My focus is wether we can deadlock because we don't receive an expected message from a peer. 
Ideally, reading from a single mailbox (a select on all the connections) all the received messages would be my dream. And then we can start and build from here. I don't know if we're close to that?

PR #4, opened 2023-03-23
feat: stableset code in

grumbach:


src/stableset/mod.rs
```diff
@@ -1,66 +1,48 @@
...
-            .expect("ping failed");
 
-        println!("Checking responses...");
-        let respondants = receive(&mut receiver).await.expect("pong failed");
-        alive_peers.extend(respondants.clone());
-
-        for r in respondants {
-            not_alive_peers.remove(&r);
+                let elders = &membership.elders();
+                membership.on_msg(elders, myself, sender, stableset_msg);
```
b-zee, line 43: Can't we call `membership.elders()` within the method?

src/stableset/membership.rs
```diff
@@ -0,0 +1,204 @@
...
+
+const ELDER_COUNT: usize = 7;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct Membership {
+    pub stable_set: StableSet,
+}
+
+impl Membership {
+    pub fn new(genesis: &BTreeSet<NetworkNode>) -> Self {
```
b-zee, line 20: For me genesis means a single node, but in this case you refer to all the initial nodes?

PR #5, opened 2023-03-23
chore: create a basic workspace for the repo

joshuef:


grumbach (approved):
LGTM, 
One question though, why have a `safenode/src/lib.rs` _and_ a `safenode/src/main.rs`?

joshuef:
Hmm, habit? 

Atm, it's like the bin uses the node lib to run itself. That's the setup we have in sn_node, so I just copied it over. 
I think the idea is that you could build another `bin` impl using the lib eg.

PR #20, opened 2023-03-27
Refactor msg handling

joshuef:


safenode/src/stableset/mod.rs
```diff
@@ -6,10 +6,12 @@ use stable_set::{Elders, StableSet};
 pub use stableset_msg::StableSetMsg;
 
 use crate::{
-    comms::{Comm, CommEvent, MsgId, NetworkMsg, NetworkNode},
-    error::Result,
+    comms::{
+        send_on_stream, Comm, CommEvent, MsgId, MsgReceived, NetworkMsg, NetworkNode,
+        ResponseStream,
+    },
+    error::{Error, Result},
```
b-zee, line 13: We're not using `Error`.

```suggestion
    error::Result,
```

PR #30, opened 2023-04-03
feat: implement req/res

RolandSherwin:


safenode/src/network/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+                peer,
+                sender,
+            })
+            .await
+            .expect("Command receiver not to be dropped.");
+        receiver.await.expect("Sender not be dropped.")
+    }
+
+    /// Respond with the provided file content to the given request.
+    pub async fn respond_chunk(&mut self, file: Vec<u8>, channel: ResponseChannel<ChunkResponse>) {
```
bochaco, line 102: I guess this ha to be removed from the `Client`

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+        let swarm = {
+            // Create a Kademlia behaviour.
+            let mut cfg = KademliaConfig::default();
+            cfg.set_query_timeout(Duration::from_secs(5 * 60));
+            let kademlia =
+                Kademlia::with_config(local_peer_id, MemoryStore::new(local_peer_id), cfg);
+            let mdns = mdns::async_io::Behaviour::new(mdns::Config::default(), local_peer_id)?;
+            let behaviour = SafeNodeBehaviour {
+                request_response: request_response::Behaviour::new(
+                    ChunkStorageCodec(),
```
bochaco, line 152: Shouldn't we have a codec and protocol which embeds all type of storage data types rather than specific for chunks?

RolandSherwin: I've made changes to make the req/res generic now! :+1:

safenode/src/network/mod.rs
RolandSherwin: Renamed `Client` -> `NetworkApi` as the previous name was conflicting with our `Client` impl. The struct abstracts the underlying swarm.

safenode/Cargo.toml
```diff
@@ -14,18 +14,26 @@ name = "safenode"
...
 tokio = { version = "1.17.0", features = ["fs", "io-util", "macros", "rt", "time", "sync", "parking_lot"] }
 tracing = { version = "~0.1.26" }
-tracing-subscriber = "0.3.16"
+tracing-subscriber = {version= "0.3.16", features=["env-filter"]}
 tracing-appender = "~0.2.0"
 tracing-core = "0.1.30"
+walkdir = "2.3.1"
 file-rotate = "0.7.3"
-libp2p = { version="0.51", features = ["async-std", "dns", "quic", "kad", "mdns", "mplex", "noise", "macros", "tcp", "websocket", "yamux"] }
+libp2p = { version="0.51", features = ["async-std", "dns", "quic", "kad", "mdns", "mplex", "noise", "macros", "tcp", "websocket", "yamux", "request-response"] }
```
oetyng, line 33: Just noting that the deps are not ordered by name here. For laters.

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
-                cmd = receiver.recv().fuse() => {
-                    debug!("Cmd in: {cmd:?}");
-                    if let Some(SwarmCmd::Search(xor_name)) =  cmd {
-                        swarm.behaviour_mut().get_closest_peers_to_xorname(xor_name);
+            let event = match network_events.next().await {
+                Some(event) => event,
+                None => continue,
+            };
+            match event {
+                NetworkEvent::InboundSafeRequest { req, channel } => {
```
oetyng, line 79: A small note: Long time ago we cleared all references to `Safe` in the safe_network types, fns and varibles naming.. It was deemed superfluous. I.e. `InboundSafeRequest` would be better without `Safe`.

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
+                    // Reply with the content of the file on incoming requests.
+                    if let SafeRequest::GetChunk(xor_name) = req {
```
oetyng, line 81: Same with `SafeRequest` and `SafeResponse`.

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
-
-                event = swarm.select_next_some() => match event {
-                    SwarmEvent::NewListenAddr { address, .. } => {
-                        info!("Listening in {address:?}");
-                    },
-                    SwarmEvent::Behaviour(SafeNetBehaviour::Mdns(mdns::Event::Discovered(list))) => {
-                        for (peer_id, multiaddr) in list {
-                            info!("Node discovered: {multiaddr:?}");
-                            swarm.behaviour_mut().kademlia.add_address(&peer_id, multiaddr);
+                NetworkEvent::PeerDiscoverd => {
```
oetyng, line 92: Isn't it possible to include the peer discovered in this enum? That way we could perhaps add it to the dht without a sleep further down? Or is that all code under the hood that we can't / shouldn't / don't want to tweak?

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
+    // wait until we discover atleast one peer
+    peer_dicovered_rx.await?;
+    info!("Discovered a Peer");
+    // todo: sometimes, the node might query the network before it adds a peer to the DHT. The
+    // PeerDiscoverd event is triggered when it adds the peer to the DHT, but the op might fail and
+    // there is no way to confirm it since `RoutingUpdate` is private/no debug impl.
+    // Hence sleep for sometime before querying the network
+    thread::sleep(time::Duration::from_millis(100));
+
+    if let Some(files_path) = opt.upload_chunks {
```
oetyng, line 112: It was surprising to me that the variable `upload_chunks` containied a `files_path`.

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
+        for entry in WalkDir::new(files_path).into_iter().flatten() {
+            if entry.file_type().is_file() {
+                let file = fs::read(entry.path())?;
+                let chunk = Chunk::new(Bytes::from(file));
+                let xor_name = chunk.name();
+                info!(
+                    "Storing file {:?} with xorname: {xor_name:x}",
+                    entry.file_name()
+                );
+                storage.store(&chunk).await?;
```
oetyng, line 122: So, we have chunks in our file system, but we also keep them in some other type of `storage`?
What's the deal with that?

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
+                // store the name as key in the network
+                // Advertise oneself as a provider of the file on the DHT.
+                network_api.store_data(*xor_name).await?;
```
oetyng, line 125: And then we store data in the network?

Isn't it so that we do what the comment says, we register ourselves as a provider of the given key?

So... `register_as_provider_of`, was the first that popped into my head. Can surely be more succinctly expressed with a few thought cycles more.

safenode/src/bin/kadnode.rs
```diff
@@ -1,233 +1,178 @@
...
+    let _log_appender_guard = init_node_logging(&opt.log_dir)?;
+
+    let (mut network_api, mut network_events, network_event_loop) = EventLoop::new()?;
+    let temp_dir = TempDir::new()?;
+    let storage = DataStorage::new(&temp_dir);
+
+    // Spawn the network task for it to run in the background.
+    spawn(network_event_loop.run());
+
+    // Todo: Implement node bootstrapping to connect to peers from outside local network
```
joshuef, line 41: let's pull this out into a bootstrap func or something, and mark as unused for now

safenode/src/bin/kadnode.rs
joshuef: `PeerDiscoverd` -> `PeerDiscovered`

safenode/src/bin/kadnode.rs
joshuef, line 122: Why are we storing this file locally at our node?

safenode/src/bin/kadnode.rs
joshuef: This feels like the wrong way round from my understanding...

Im not sure we want to be using the `provider` stuff at all? We just want to send the data to the closest nodes, and they store it. Or?

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Clone)]
+/// API to interact with the underlying Swarm
+pub struct NetworkApi {
+    pub(super) sender: mpsc::Sender<CmdToSwarm>,
+}
+
+impl NetworkApi {
+    //  Listen for incoming connections on the given address.
+    pub async fn start_listening(&mut self, addr: Multiaddr) -> Result<()> {
+        let (sender, receiver) = oneshot::channel();
```
joshuef, line 30: sender and `self.sender` are confusing here. Can we name one of them a bit differently for clarity?

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    swarm: Swarm<SafeNodeBehaviour>,
+    command_receiver: mpsc::Receiver<CmdToSwarm>,
+    event_sender: mpsc::Sender<NetworkEvent>,
+    pending_dial: HashMap<PeerId, oneshot::Sender<Result<()>>>,
+    pending_start_providing: HashMap<QueryId, oneshot::Sender<Result<()>>>,
+    pending_get_providers: HashMap<QueryId, oneshot::Sender<HashSet<PeerId>>>,
+    pending_safe_requests: HashMap<RequestId, oneshot::Sender<Result<SafeResponse>>>,
+}
+
+impl EventLoop {
```
joshuef, line 61: I think `EventLoop` doesn't really capture the fact that we start up the connections in here.

Perhaps `NetworkSwarmLoop` ?

safenode/src/network/README.md
```diff
@@ -0,0 +1,10 @@
+### Some reference implementations
```
oetyng, line 1: Great docs.

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+        self.sender
+            .send(CmdToSwarm::StartListening { addr, sender })
+            .await?;
+        receiver.await?
+    }
+
+    /// Dial the given peer at the given address.
+    pub async fn dial(&mut self, peer_id: PeerId, peer_addr: Multiaddr) -> Result<()> {
```
oetyng, line 38: This is interesting. Can the network hold the same `PeerId` with different `MultiAddr`? (And vice versa?)

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+                peer_addr,
+                sender,
+            })
+            .await?;
+        receiver.await?
+    }
+
+    /// Advertise the local node as the provider of a given piece of data; The XorName of the data
+    /// is advertised to the nodes on the DHT
+    pub async fn store_data(&mut self, xor_name: XorName) -> Result<()> {
```
oetyng, line 52: Yeah, here we have that `register_as_holder_of` fn.

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+        let (sender, receiver) = oneshot::channel();
+        self.sender
+            .send(CmdToSwarm::StoreData { xor_name, sender })
```
oetyng, line 55: And this one, `CmdToSwarm::StoreData`, correspondingly as above.

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+            .await?;
+        receiver.await?
+    }
+
+    /// Find the providers for the given piece of data; The XorName is used to locate the nodes
+    /// that hold the data
+    pub async fn get_data_providers(&mut self, xor_name: XorName) -> Result<HashSet<PeerId>> {
```
oetyng, line 62: And here could be `get_holders`.

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+        let (sender, receiver) = oneshot::channel();
+        self.sender
+            .send(CmdToSwarm::GetDataProviders { xor_name, sender })
```
oetyng, line 65: And this, `CmdToSwarm::GetDataProviders`, correspondingly to above.

safenode/src/network/api.rs
```diff
@@ -0,0 +1,94 @@
...
+    ) -> Result<SafeResponse> {
+        let (sender, receiver) = oneshot::channel();
+        self.sender
+            .send(CmdToSwarm::SendSafeRequest { req, peer, sender })
+            .await?;
+        receiver.await?
+    }
+
+    /// Send a `SafeResponse` through the channel opened by the requester.
+    pub async fn send_safe_response(
```
oetyng, line 84: Also these, as previously mentioned.

safenode/src/network/event.rs
```diff
@@ -0,0 +1,172 @@
...
+#[behaviour(out_event = "SafeNodeEvent")]
+pub(super) struct SafeNodeBehaviour {
+    pub(super) request_response: request_response::Behaviour<SafeMsgCodec>,
+    pub(super) kademlia: Kademlia<MemoryStore>,
+    pub(super) mdns: mdns::async_io::Behaviour,
+}
+
+#[derive(Debug)]
+pub(super) enum SafeNodeEvent {
+    RequestResponse(request_response::Event<SafeRequest, SafeResponse>),
```
oetyng, line 34: request_response is actually not a very good name chosen by libp2p folks.
We should just map that to `Response` IMO. I mean, what other types of responses are there, other than to a `Request`? Unnecessary qualifier.

safenode/src/network/event.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Incoming `SafeRequest` from a peer
+    InboundSafeRequest {
+        /// Request
+        req: SafeRequest,
+        /// The channel to send the `SafeResponse` through
+        channel: ResponseChannel<SafeResponse>,
+    },
+    /// Emmited when we discover a peer.
+    /// might/might not be successfully added to the DHT; `RoutingUpdate` is private/no debug impl
+    PeerDiscoverd,
```
oetyng, line 69: Typo `PeerDiscoverd` => `PeerDiscovered`

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+            }
+            CmdToSwarm::GetDataProviders { xor_name, sender } => {
+                let query_id = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .get_providers(xor_name.0.to_vec().into());
+                let _ = self.pending_get_providers.insert(query_id, sender);
+            }
+            CmdToSwarm::SendSafeRequest { req, peer, sender } => {
```
oetyng, line 103: As per previously mention on Safe qualifier: `SendRequest`

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+                let request_id = self
+                    .swarm
+                    .behaviour_mut()
+                    .request_response
+                    .send_request(&peer, req);
+                let _ = self.pending_safe_requests.insert(request_id, sender);
+            }
+            CmdToSwarm::SendSafeResponse { resp, channel } => {
```
oetyng, line 111: As per previously mention on Safe qualifier: `SendResponse`

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+                    .get_providers(xor_name.0.to_vec().into());
+                let _ = self.pending_get_providers.insert(query_id, sender);
+            }
+            CmdToSwarm::SendSafeRequest { req, peer, sender } => {
+                let request_id = self
+                    .swarm
+                    .behaviour_mut()
+                    .request_response
+                    .send_request(&peer, req);
+                let _ = self.pending_safe_requests.insert(request_id, sender);
```
oetyng, line 109: As per previously mention on `Safe` qualifier: `pending_requests`

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+                        }
+                        Err(e) => {
+                            let _ = sender.send(Err(e.into()));
+                        }
+                    }
+                } else {
+                    warn!("Already dialing peer.");
+                }
+            }
+            CmdToSwarm::StoreData { xor_name, sender } => {
```
oetyng, line 87: Yeah, this doesn't contain the data, so `StoreData` confusing. `AnnounceHolding`, `RegisterAsHolder`, `RegisterAsProvider`, `StartProviding`etc. seems more accurate.

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+                peer_id,
+                peer_addr,
+                sender,
+            } => {
+                if let hash_map::Entry::Vacant(e) = self.pending_dial.entry(peer_id) {
+                    let _routing_update = self
+                        .swarm
+                        .behaviour_mut()
+                        .kademlia
+                        .add_address(&peer_id, peer_addr.clone());
```
oetyng, line 71: This is interesting. Why do we add the address before a dial? Asking juut to understand better what's happening.

safenode/src/network/command.rs
```diff
@@ -0,0 +1,123 @@
...
+    },
+    StoreData {
+        xor_name: XorName,
+        sender: oneshot::Sender<Result<()>>,
+    },
+    GetDataProviders {
+        xor_name: XorName,
+        sender: oneshot::Sender<HashSet<PeerId>>,
+    },
+    SendSafeRequest {
```
oetyng, line 41: Already commented other place.

safenode/src/network/event.rs
```diff
@@ -0,0 +1,172 @@
...
+                        sender
+                            .send(providers)
+                            .map_err(|_| Error::Other("Receiver not to be dropped".to_string()))?;
+
+                        // Finish the query. We are only interested in the first result.
+                        self.swarm
+                            .behaviour_mut()
+                            .kademlia
+                            .query_mut(&id)
+                            .ok_or(Error::Other("Query should be exist".to_string()))?
```
oetyng, line 118: "Query should be exist" => "Query should exist"

safenode/src/network/event.rs
```diff
@@ -0,0 +1,172 @@
...
+                }
+            },
+            SwarmEvent::NewListenAddr { address, .. } => {
+                let local_peer_id = *self.swarm.local_peer_id();
+                info!(
+                    "Local node is listening on {:?}",
+                    address.with(Protocol::P2p(local_peer_id.into()))
+                );
+            }
+            SwarmEvent::IncomingConnection { .. } => {}
```
oetyng, line 147: Just a note: I think this is something we could use for the bayesian function for suspicious behaviour detection, as David talked about.

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+};
+use tracing::warn;
+
+/// The main event loop recieves `SwarmEvents` from the network, `CmdToSwarm` from the upper layers and
+/// emmits `NetworkEvent` to the upper layers
+/// Also keeps track of the pending queries/requests and their channels. Once we recieve an event
+/// that is the outcome of a previously executed cmd, send a response to them via the stored channel.
+pub struct EventLoop {
+    swarm: Swarm<SafeNodeBehaviour>,
+    command_receiver: mpsc::Receiver<CmdToSwarm>,
```
oetyng, line 53: Bunch of naming nitpicks by me... :) ... but I'd name this `cmd_receiver` for stringency.

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    event_sender: mpsc::Sender<NetworkEvent>,
+    pending_dial: HashMap<PeerId, oneshot::Sender<Result<()>>>,
+    pending_start_providing: HashMap<QueryId, oneshot::Sender<Result<()>>>,
+    pending_get_providers: HashMap<QueryId, oneshot::Sender<HashSet<PeerId>>>,
+    pending_safe_requests: HashMap<RequestId, oneshot::Sender<Result<SafeResponse>>>,
```
oetyng, line 58: Small note: these three `pending_...` variables I think we can do something about as well.

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+impl EventLoop {
+    /// Creates the network components
+    /// - The `NetworkApi` to interact with the network layer from anywhere
+    ///   within your application.
+    ///
+    /// - The `NetworkEvent` receiver to get the events from the network layer.
+    ///
+    /// - The `EventLoop` taht drives the network.
```
oetyng, line 68: `taht` => `that`

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn new(// secret_key_seed: Option<u8>,
```
oetyng, line 69: // secret_key_seed: Option<u8>,
Is this something we might want to use later?

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+                    Default::default(),
+                ),
+                kademlia,
+                mdns,
+            };
+
+            let mut swarm =
+                SwarmBuilder::with_async_std_executor(transport, behaviour, local_peer_id).build();
+
+            // // Listen on all interfaces and whatever port the OS assigns.
```
oetyng, line 103: Double comment.

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            let addr = "/ip4/0.0.0.0/udp/0/quic-v1".parse().expect("addr okay");
+            let _listener_id = swarm.listen_on(addr).expect("listening failed");
```
oetyng, line 105: These two msgs are inconsequent.

Ln 104 describes what should be expected `"addr okay"` (I'd call it `"valid address"` btw).
Ln 105 describes what what we _don't_ want to happen.

safenode/src/network/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+            swarm
+        };
+
+        let (command_sender, command_receiver) = mpsc::channel(0);
```
oetyng, line 110: To keep the stringency line previously mentioned: `command_sender` => `cmd_sender`

safenode/src/bin/kadnode.rs
RolandSherwin: The `kadnode.rs` is just a prototype file for now, as we had discussed in slack :+1:

safenode/src/bin/kadnode.rs
RolandSherwin: Same as the earlier comment, the `kadnode.rs` is just a prototype file for now!

safenode/src/bin/kadnode.rs
RolandSherwin: Yes, this might not be what we want to do. Adding some todos here.

safenode/src/bin/kadnode.rs
RolandSherwin: The `kadnode.rs` is currently used for prototyping, these might be changed by Gab's work later…

safenode/src/network/command.rs
RolandSherwin: I think `add_address` just inserts the peer to the DHT, but does not actually connect to it. While `dial` connects to the peer. I've observed this pattern through the logs while fetching the chunks.

safenode/src/network/mod.rs
RolandSherwin: Sorry, those were some leftover from the example file

safenode/src/network/mod.rs
RolandSherwin: Changed it!

safenode/src/network/api.rs
RolandSherwin: Not sure, but should be possible, right? And this is what we're trying to prevent by having 1 socket addr per PeerId?

safenode/src/network/api.rs
RolandSherwin: The node should not go through the provider api to store/hold stuff, this will be worked on in the next PR. Added a todo for now.

safenode/src/network/event.rs
RolandSherwin: Not sure if I understand this correctly, since we have `Event::Message::Request` and `Event::Message::Response` which denote an incoming request/response?

safenode/src/network/command.rs
RolandSherwin: Will be worked on in a follow up PR as stated earlier.

safenode/src/network/command.rs
```diff
@@ -0,0 +1,126 @@
...
+            // 2. store data in them directly, not via provider
+            CmdToSwarm::StoreData { xor_name, sender } => {
+                let query_id = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .start_providing(xor_name.0.to_vec().into())?;
+                let _ = self.pending_start_providing.insert(query_id, sender);
+            }
+            CmdToSwarm::GetDataProviders { xor_name, sender } => {
```
jacderida, line 98: Is there possibly a name that's a bit more specific we could use rather than 'data provider'? This is the node or the peer that has the piece of data being requested? Are only some nodes/peers providers?

safenode/src/network/api.rs
```diff
@@ -0,0 +1,92 @@
...
+    SinkExt,
+};
+use libp2p::{request_response::ResponseChannel, Multiaddr, PeerId};
+use xor_name::XorName;
+
+use super::command::CmdToSwarm;
+
+#[derive(Clone)]
+/// API to interact with the underlying Swarm
+pub struct NetworkApi {
```
jacderida, line 23: Should `Api` be in the type/module names? Could it be called just `Network` and moved back to `mod`?

oetyng (approved):

safenode/src/network/command.rs
RolandSherwin: Sorry, `providers` are something that we should not use for chunks/dbcs, atleast that's what I gathered from David's presentation. I will work on everything related to `providers/data` in a follow-up PR.

joshuef (approved):

PR #32, opened 2023-04-03
Feat: registers chunks in mem storage

### LINKED ISSUE #38: Add in register storage
Add in the ability to store registers on the network

### LINKED ISSUE #40: Chunk storage
Pull in chunk storage from prev repo

bochaco:
Resolves #38 
Resolves #40


safenode/src/bin/kadnode.rs
```diff
@@ -54,11 +62,13 @@ async fn main() -> Result<()> {
             match event {
                 NetworkEvent::RequestReceived { req, channel } => {
                     // Reply with the content of the file on incoming requests.
-                    if let Request::Query(Query::GetChunk(xor_name)) = req {
-                        let addr = ChunkAddress(xor_name);
-                        let chunk = storage_clone.query(&addr).await.unwrap();
+                    if let Request::Query(Query::GetChunk(addr)) = req {
+                        let chunk = storage_clone.query_chunk(&addr).await.unwrap();
```
oetyng, line 66: `storage_clone.query_chunk(&addr)`
I think this is more consistent and suitable: `get_chunk`, as that is the specific name of the query (as in `GetChunk`, as is the name of that query variant).

safenode/src/bin/kadnode.rs
```diff
@@ -129,11 +140,55 @@ async fn main() -> Result<()> {
             .await
             .map_err(|_| eyre!("None of the providers returned file."))?
             .0;
-        if let Response::Query(QueryResponse::Chunk(chunk)) = resp {
+        if let Response::Query(QueryResponse::GetChunk(Ok(chunk))) = resp {
```
oetyng, line 143: Yep, good renaming.

safenode/src/storage/chunks.rs
```diff
@@ -6,181 +6,82 @@
...
-    ///
-    /// If the location specified already contains a `ChunkStorage`, it is simply used
-    ///
-    /// Used space of the dir is tracked
-    pub(super) fn new(path: &Path) -> Self {
-        Self {
-            file_store_path: path.join(CHUNKS_STORE_DIR_NAME),
-        }
+    #[allow(dead_code)]
+    pub(super) async fn addrs(&self) -> Vec<ChunkAddress> {
```
oetyng, line 38: I know this is since old. But maybe we can write out the term to use properly here?

safenode/src/storage/errors.rs
```diff
@@ -23,16 +26,47 @@ pub enum Error {
...
+    RegisterNotFound(RegisterAddress),
+    /// NetworkData error.
+    #[error("Network data error:: {0}")]
+    NetworkData(#[from] ProtocolError),
+    /// Data authority provided is invalid.
+    #[error("Provided PublicKey could not validate signature {0:?}")]
+    InvalidSignature(PublicKey),
+    /// Bincode error.
+    #[error("Bincode error:: {0}")]
+    Bincode(#[from] bincode::Error),
```
oetyng, line 40: Mm, this as well is since old.

But I was wondering if we should maybe stick to a single serialization protocol all over? Like MsgPack. And get that in from start, to avoid building up a backlog.

safenode/src/storage/errors.rs
```diff
@@ -23,16 +26,47 @@ pub enum Error {
...
+        doesn't match stored Register address: {reg_addr:?}"
+    )]
+    RegisterAddrMismatch {
+        cmd_dst_addr: RegisterAddress,
+        reg_addr: RegisterAddress,
+    },
+}
+
+// Convert storage error to messaging error message for sending over the network.
+impl From<Error> for ProtocolError {
```
oetyng, line 59: Now that we don't have these in separate crate, where messaging error was in sn_interface and we had the reversed dependency there so that sn_node knew about interface but not the other way around, we could skip this stringifying of errors.

I've always thought it is less than optimal. And now we could actually have these enum variants directly held in a higher level error.

safenode/src/storage/mod.rs
```diff
@@ -6,53 +6,68 @@
...
+        debug!("Replicating {data:?}");
+        match data {
+            ReplicatedData::Chunk(chunk) => self.chunks.store(chunk).await,
+            ReplicatedData::RegisterLog(data) => self.registers.update(data).await,
+            ReplicatedData::RegisterWrite(cmd) => self.registers.write(cmd).await,
+        }
+    }
+
+    /// Query the local store and return QueryResponse
+    pub async fn query(&self, query: &Query, requester: User) -> QueryResponse {
```
oetyng, line 64: Another thing since old that I have thought is a bit misplaced, is that we return a QueryResponse here. It is unnecessarily dragging in a higher level type to this lower level. If we could avoid that would be great.

safenode/src/storage/mod.rs
```diff
@@ -6,53 +6,68 @@
...
-// Example:
-// - with a xorname with starting bits `010001110110....`
-// - and a BIT_TREE_DEPTH of `6`
-// returns the path `ROOT_PATH/0/1/0/0/0/1`
-fn prefix_tree_path(root: &Path, xorname: XorName) -> PathBuf {
-    let bin = format!("{xorname:b}");
-    let prefix_dir_path: PathBuf = bin.chars().take(BIT_TREE_DEPTH).map(String::from).collect();
-    root.join(prefix_dir_path)
+    /// Store data in the local store
+    pub async fn store(&self, data: &ReplicatedData) -> Result<()> {
```
oetyng, line 54: Perhaps we could keep this one api for storing, and have the calling layer wrap to ReplicatedData?

Then store_chunk can be removed, which seems a bit non-stringent, considering we only use this api for register (and should be able to do with chunk as well).

safenode/src/storage/mod.rs
oetyng, line 64: Since we take a query here, maybe we should take a cmd above, for stringency?

That would also imply that we implement and keep spentbook/dbc spend storage in DataStorage struct.

But I guess that is implied if we take Query here, and will keep dbc requests as part of the Data reqquests.

safenode/src/storage/registers.rs
```diff
@@ -0,0 +1,735 @@
...
+    Error, Result,
+};
+
+use bincode::serialize;
+use tracing::{debug, error, info, trace, warn};
+
+/// Operations over the Register data type and its storage.
+#[derive(Clone, Default)]
+pub(super) struct RegisterStorage {
+    register_store: RegisterStore,
```
oetyng, line 31: Hmm.... do we need this additional layering? Could we perhaps remove it cleanly? Ideas?

safenode/src/storage/registers.rs
```diff
@@ -0,0 +1,735 @@
...
+                    Err(err) => {
+                        trace!("Editing Register failed {addr:?}: {err:?}");
+                        Err(err)
+                    }
+                }
+            }
+        }
+    }
+
+    // Gets stored register log from disk, trying to reconstruct the Register
```
oetyng, line 301: The docs say from disk here. Absolute nit pick, but I guess as it's now seen, we might just as well update it.

oetyng (approved):
Looks great generally.
Just had a few questions and a couple of possible changes I think we can make here already.

safenode/src/storage/errors.rs
joshuef: I'm going to pull this out as an issue to discuss as I think it'll need a bit of it.

PR #53, opened 2023-04-06
fix: store directly at the closest peers

### LINKED ISSUE #41: Store directly at nodes
Update the APIs to move away from kad provider, and simply get closest nodes

RolandSherwin:
closes #41


safenode/src/network/msg/mod.rs
```diff
@@ -7,8 +7,8 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 mod codec;
+pub use codec::{Cmd, CmdResponse, Query, QueryResponse, Request, Response};
```
maqi, line 10: the naming of `Cmd`, `Query` and `Request/Response` is bit overlapping.
If do want to divide the communications based on the usage, better give a more explicit naming.
and same if divide based on the targets (i.e. client-node, or among nodes)

safenode/src/node/api.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use super::error::Result;
+use crate::{
+    network::{Cmd, Query, Request},
+    storage::chunks::Chunk,
+};
+
+use super::Node;
+
+impl Node {
```
maqi, line 22: better have a name like `Vault` (or even just DataHolder) if it only handles the data storage duty?
`Node` means it also needs to handle the `network knowledge` and other duties ?

safenode/src/node/api.rs
```diff
@@ -0,0 +1,87 @@
...
+            let req = Request::Cmd(Cmd::StoreChunk(data.clone()));
+            let future = Box::pin(tokio::time::timeout(
+                Duration::from_secs(5),
+                self.network.send_request(req, node),
+            ));
+            sent_futures.push(future);
+        }
+
+        let mut responses = Vec::new();
+        while !sent_futures.is_empty() {
```
maqi, line 38: we may don't care of the response of `store data` ?
anyway, no harm to keep it unless it blocks the flow or affect the performance

safenode/src/node/api.rs
```diff
@@ -0,0 +1,87 @@
...
+            match select_all(sent_futures).await {
```
maqi, line 71: only expect one response shall be enough?

also, this seems a blocking call ?

safenode/src/node/api.rs
```diff
@@ -0,0 +1,87 @@
...
+                (Ok(res), _, remaining_futures) => {
+                    info!("got response for get_data: {res:?}");
```
maqi, line 73: loging the whole chunk could be lengthy.
unless the `Debug` formt of `res` only logs the name of the content?

safenode/src/node/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+            node_events_channel: node_events_channel.clone(),
+        };
+        let mut node_clone = node.clone();
+
+        // Spawn the network task for it to run in the background.
+        let _ = spawn(network_event_loop.run());
+
+        let _ = spawn(async move {
+            loop {
+                let event = match network_events.next().await {
```
maqi, line 60: shall be the `try_next` call here?

safenode/src/node/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+        &mut self,
+        request: Request,
+        response_channel: ResponseChannel<Response>,
+    ) -> Result<()> {
+        trace!("Handling request: {request:?}");
+        match request {
+            Request::Query(query) => match query {
+                Query::GetChunk(xor_name) => {
+                    let addr = ChunkAddress(xor_name);
+                    let chunk = self.storage.query(&addr).await?;
```
maqi, line 102: the self storage shall be just a read file from local disk.
so a `get` call instead of `query` shall be more precise ?

Also, I am wondering for the self storage, a direct read/write file shall be enough, no need to be `async` ?

maqi: also, the result of the data fetch may be an error.
in that case, the error shall be sent back as a response.
shall not just `unwrap` directly, which will result in a `terminated flow` and cause the request sender have to `time out`

safenode/src/node/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+                        Response::Query(QueryResponse::Chunk(chunk)),
+                        response_channel,
+                    )
+                    .await;
+                }
+                Query::GetDbc(_) => todo!(),
+            },
+            Request::Cmd(cmd) => match cmd {
+                Cmd::StoreChunk(chunk) => {
+                    self.storage.store(&chunk).await?;
```
maqi, line 115: same as query, the result of store shall be reported back.

safenode/src/network/msg/mod.rs
RolandSherwin: Do you think, `Action / ActionResult` and `Query / QueryResult` be a more apt name? The client should be at a higher level, so should we distinguish between them at this layer?

maqi: I think currently we can just stay with `Request/Response` or `Action/AcitonResult`, to distinguish the message direcion, shall be enough.

safenode/src/node/api.rs
RolandSherwin: Aye, we got `#[debug(skip)]` over the bytes.

safenode/src/node/mod.rs
maqi: right. but be careful that `recv` call is a blocking call.
which means may result in a `dead lock`

safenode/src/network/error.rs
```diff
@@ -36,9 +36,12 @@ pub enum Error {
     #[error("Kademlia Store error: {0}")]
     KademliaStoreError(#[from] kad::store::Error),
 
-    #[error("The mpsc::receiever has been dropped")]
-    ReceieverDropped(#[from] mpsc::SendError),
+    #[error("The mpsc::receiever for `NetworkEvent` has been dropped")]
+    NetworkEventReceieverDropped(#[from] mpsc::error::SendError<NetworkEvent>),
```
jacderida, line 40: There's a little typo here for both of these. `NetworkEventReceieverDropped` -> `NetworkEventReceiverDropped`.

safenode/src/network/event.rs
```diff
@@ -89,40 +86,24 @@ impl NetworkSwarmLoop {
...
                 }
-                _ => {}
+                KademliaEvent::RoutingUpdated { .. } => {
+                    self.event_sender.send(NetworkEvent::PeerAdded).await?;
+                }
+                KademliaEvent::InboundRequest { request } => {
+                    info!("got inboumd request: {request:?}");
+                }
+                todo => {
+                    panic!("KademliaEvent has not been implemented: {todo:?}");
```
bochaco, line 105: I wouldn't panic but log a warning, I think we may get an event and we don't want our node to crash just because we are not interested in such an event or because we still didn't handle it.

safenode/src/network/event.rs
```diff
@@ -168,7 +148,7 @@ impl NetworkSwarmLoop {
             }
             SwarmEvent::IncomingConnectionError { .. } => {}
             SwarmEvent::Dialing(peer_id) => info!("Dialing {peer_id}"),
-            e => panic!("{e:?}"),
+            e => panic!("SwarmEvent has not been implemented: {e:?}"),
```
bochaco, line 151: Sam as earlier, let's not panic but just log a msg

safenode/src/vault/api.rs
```diff
@@ -0,0 +1,106 @@
...
+            ReplicatedData::Chunk(chunk) => Cmd::StoreChunk(chunk.clone()),
+            ReplicatedData::RegisterWrite(cmd) => Cmd::Register(cmd.clone()),
+            ReplicatedData::RegisterLog(_) => todo!(),
+        };
+        // forward to the other closest nodes if we're seeing the data for the first time
+        // return early if we already have the data with us
+        match self.storage.store(&cmd).await {
+            CmdResponse::StoreChunk(Err(_)) => return Ok(()),
+            CmdResponse::CreateRegister(Err(_)) => return Ok(()),
+            CmdResponse::EditRegister(Err(_)) => return Ok(()),
```
bochaco, line 36: I think you could match them all with `CmdResponse::StoreChunk(Err(_)) |CmdResponse::CreateRegister(Err(_)) | CmdResponse::EditRegister(Err(_)) => return Ok(())`

PR #60, opened 2023-04-07
chore: various minor adjustments

jacderida:
While making an effort to understand the node start up and the different async tasks in use, I noticed small adjustments I thought I could make to perhaps improve clarity.

* Rename: `NetworkSwarmLoop` to `SwarmDriver`, which then provides the loop in its `run` function.
* Use GPT-4 to document `SwarmDriver` and its public functions. Did not need any adjustment.
* Rename some variables in `SwarmDriver::new` for extra clarity.
* Rename `Node::node_events_channel` to `Node::events_channel` since it's part of the `Node` struct.
* Use GPT-4 to document `Node` and its public functions. Did not need any adjustment.
* Removed comments that appeared to provide limited value.

### DIFF
--- a/safenode/src/bin/kadnode.rs
+++ b/safenode/src/bin/kadnode.rs
@@ -43,7 +43,6 @@ async fn main() -> Result<()> {
     let (node, node_events_channel) = Node::run().await?;
 
     let mut node_events_rx = node_events_channel.subscribe();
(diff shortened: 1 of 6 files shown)


oetyng (approved):

dirvine (commented):
Nice one. Clearer names and good documentation

PR #61, opened 2023-04-08
refactor(node): don't have client fn on nodes

oetyng:
This implements a `Client` and removes the client-specific logic from `Node`.


dirvine (approved):
Nice one. We need to check bincode use later to allow folk to write nodes/clients in other languages.

dirvine (approved):

PR #62, opened 2023-04-08
feat(dbcs): integrate to the system

oetyng:
The close group of a received `Spend`, calls the close groups of the parents of that `Spend`, and check that the returned parent `Spends` are valid. That is a prerequisite for storing the `Spend` sent by a client. (If any of them are not valid, that shall be communicated to those close groups.)

The Spend is then put in to `SpendStorage`, which validates them before storing them as a `valid Spend`. If a `double spend` is attempted, the `Spend` is permanently marked as `unspendable` (this shall also be communicated to the rest of the peers).
A peer will thus make sure to only return a `Spend` if it has been considered **valid** by the peer.

After this `Spend` has been stored, it is up to the client to query all
close nodes and verify that it is recognised by enough nodes. 
That then makes the spend valid.

- **NB 1**: More validations might be needed.
- **NB 2**: There are stubs to broadcast
a) Invalid parent spends, to each of their close groups.
b) Attempted double spends to our peers (i.e. this close group, responsible for this spend).
And so the sending, and handling of those msgs at recipient peers, is still to be done.

### DIFF
--- a/safenode/Cargo.toml
+++ b/safenode/Cargo.toml
@@ -28,7 +28,7 @@ rand = { version = "~0.8.5", features = ["small_rng"] }
 rmp-serde = "1.1.1"
 self_encryption = "~0.28.0"
 serde = {version = "1.0.133", features = [ "derive", "rc" ]}
-sn_dbc = { git = "https://github.com/oetyng/sn_dbc", branch = "remove-section-signatures", features = ["serdes"] }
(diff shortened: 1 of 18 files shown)


dirvine (approved):
Cool, marking this approved to have it merged and let us get the whole network up and running quickly

PR #64, opened 2023-04-10
feat(closegroup): use close group var

oetyng:
- This allows verification that we got enough nodes, according to our protocol.


safenode/src/network/mod.rs
```diff
@@ -41,6 +41,11 @@ use tokio::sync::{mpsc, oneshot};
 use tracing::warn;
 use xor_name::XorName;
 
+/// The maximum number of peers to return in a `GetClosestPeers` response.
+/// This is the group size used in safe network protocolm to be responsible for
```
maqi, line 45: protocolm  -> protocol

maqi (approved):

PR #68, opened 2023-04-11
feat(client): add file apis and self encryption

### LINKED ISSUE #70: Pull over self encryption and file apis from old repo.

oetyng:
This adds all file apis for chunking and storing files,
as well as retreiving and unpacking chunks.

Resolves #70.


safenode/src/client/chunks/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+use bytes::Bytes;
+use self_encryption::MIN_ENCRYPTABLE_BYTES;
+
+/// Data of size more than 0 bytes less than [`MIN_ENCRYPTABLE_BYTES`] bytes.
+///
+/// A `Spot` cannot be self-encrypted, thus is encrypted using the client encryption keys instead.
+#[allow(missing_debug_implementations)]
+#[derive(Clone)]
+pub(crate) struct SmallFile {
+    bytes: Bytes,
```
grumbach, line 24: `Spot` here in the comment is old nomenclature isn't it?

safenode/src/client/chunks/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/// Data of size larger than or equal to [`MIN_ENCRYPTABLE_BYTES`] bytes.
+///
+/// A `LargeFile` is spread across multiple chunks in the network.
+/// This is done using self-encryption, which produces at least 4 chunks (3 for the contents, 1 for the `DataMap`).
```
grumbach, line 30: > 4 chunks (3 for the contents, 1 for the `DataMap`)

Is this still true?

safenode/src/client/chunks/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+mod error;
+mod pac_man;
+
+pub(crate) use self::error::{Error, Result};
+pub(crate) use pac_man::{encrypt_large, to_chunk, DataMapLevel};
+
+use bytes::Bytes;
+use self_encryption::MIN_ENCRYPTABLE_BYTES;
```
grumbach, line 16: I was wondering if we could get rid of the SmallFile/BigFile distinction, and just padd smaller files with `0`s until we can encrypt them? How much is `MIN_ENCRYPTABLE_BYTES`? If it's small enough then maybe it's reasonable?
This would get rid of a lot of complexity too. This always bugged me somehow in sn.

safenode/src/client/file_apis.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        Ok(address)
+    }
+
+    /// Directly writes a [`LargeFile`] to the network in the
+    /// form of immutable self encrypted chunks, without any batching.
+    #[instrument(skip_all, level = "trace")]
+    async fn upload_large(&self, large: LargeFile, verify: bool) -> Result<ChunkAddress> {
+        let (head_address, all_chunks) = encrypt_large(large)?;
+        for next_batch in all_chunks.chunks(CHUNKS_BATCH_MAX_SIZE) {
```
grumbach, line 140: Here again, if we got rid of the Small/Large distinction we would simplify the code.

safenode/src/client/chunks/mod.rs
oetyng: Yeah definitely. Will fix!

grumbach (commented):
Looking good, just one question about the small big file distinction

safenode/src/client/chunks/mod.rs
oetyng: Yes.

safenode/src/client/chunks/mod.rs
oetyng: `MIN_ENCRYPTABLE_BYTES` is 3072. since min size of a chunk is 1024 and it requires at least 3 to be able to do self-encryption.

Yeah, we could perhaps pad smaller files. A question we can bring up right away in slack.

safenode/src/client/file_apis.rs
oetyng: Yep.

safenode/src/client/chunks/mod.rs
grumbach: Why don't we make 4096 size chunks? And fix this for good.

grumbach (approved):
🚢

PR #73, opened 2023-04-11
fix(network): replace generic Error types with more specific ones

bochaco:
BREAKING CHANGE: renaming `safeclient` binary to `safe`.

### DIFF
--- a/safenode/src/network/event.rs
+++ b/safenode/src/network/event.rs
@@ -85,7 +85,7 @@ impl SwarmDriver {
                 }
             }
             // handle Kademlia events
-            SwarmEvent::Behaviour(NodeEvent::Kademlia(event)) => match event {
+            SwarmEvent::Behaviour(NodeEvent::Kademlia(ref event)) => match event {
                 KademliaEvent::OutboundQueryProgressed {
                     id,
                     result: QueryResult::GetClosestPeers(Ok(closest_peers)),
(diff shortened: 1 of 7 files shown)


safenode/src/network/msg/mod.rs
```diff
@@ -44,7 +44,7 @@ impl SwarmDriver {
                     let _ = self
                         .pending_requests
                         .remove(&request_id)
-                        .ok_or(Error::Other("Request to still be pending".to_string()))?
+                        .ok_or(Error::ReceivedResponseDropped(request_id))?
                         .send(Ok(response));
```
RolandSherwin, line 48: The result after sending the response is ignored here, maybe it can be mapped to `.map_err(|_| Error::InternalMsgChannelDropped)?`

safenode/src/network/msg/mod.rs
```diff
@@ -54,7 +54,7 @@ impl SwarmDriver {
                 let _ = self
                     .pending_requests
                     .remove(&request_id)
-                    .ok_or(Error::Other("Request to still be pending.".to_string()))?
+                    .ok_or(Error::ReceivedResponseDropped(request_id))?
                     .send(Err(error.into()));
```
RolandSherwin, line 58: Same here, we can probably map the error to `InternalMsgChannelDropped` instead of ignoring the result

safenode/src/network/event.rs
```diff
@@ -104,9 +104,9 @@ impl SwarmDriver {
                         let new_peers: HashSet<PeerId> = closest_peers.peers.into_iter().collect();
```
RolandSherwin, line 104: Sorry, couldn't actually tag the line I wanted. Anways, in L116, we are just logging a trace msg if the `pending_get_closest_peers.remove()` returns `None`, maybe we can use `ReceivedResponseDropped` here as well?

bochaco: makes sense

RolandSherwin (approved):
LGTM!

PR #78, opened 2023-04-12
test: `closest_peers` + small fix

RolandSherwin:
- Also sets the `kad::replication_factor` that determines the a) The number of closer peers returned in response to a request, b) The number of closest peers to a key to search for in an iterative query. 
- It does not affect the maximum number of nodes in a bucket; this is set by `kad::K_VALUE`.


safenode/src/client/api.rs
```diff
@@ -248,7 +248,7 @@ impl Client {
     // If `get_all_responses` is false, we return the first successful response that we get.
     async fn send_and_get_responses(
         &self,
-        nodes: HashSet<PeerId>,
+        nodes: Vec<PeerId>,
```
oetyng, line 251: Seems it could be useful to ensure we don't have duplicates here?
Probably handled in lower layers (I assume). But still?

safenode/src/network/mod.rs
```diff
@@ -318,3 +326,109 @@ impl Network {
...
+        time::Duration,
+    };
+    use xor_name::XorName;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn closest() -> Result<()> {
+        let _ = init_node_logging(&None)?;
+        let mut networks_list = Vec::new();
+        let mut network_events_recievers = BTreeMap::new();
+        for _ in 1..25 {
```
RolandSherwin, line 352: One small observation that I had is that when the node count is > 28, the node never really sends out the `Cmd::GetClosestPeer` and the whole test freezes. There are 0 log lines printed as well. I suspect it has something to do with all the nodes running inside the same process?

safenode/src/client/api.rs
oetyng: Btw, was talking with Chris about node/peer distinction.

We've been trying to keep a distinction bewtween nodes and peers (I know this wasn't changed in this PR, but since this line is modified here).

So when we have peer ids, we should call them peers.

When we talk of our node, then we should call it node.

safenode/src/network/event.rs
```diff
@@ -107,7 +110,7 @@ impl SwarmDriver {
                     let new_peers: HashSet<PeerId> =
                         closest_peers.peers.clone().into_iter().collect();
                     current_closest.extend(new_peers);
-                    if current_closest.len() >= usize::from(K_VALUE) || step.last {
+                    if current_closest.len() >= CLOSE_GROUP_SIZE || step.last {
```
oetyng, line 113: I wonder about this...
If we haven't got all from the bucket, we can't really be sure we actually have _the closest_ ones. Only that we have the number of peers that we want to have _if they are the closest_.

safenode/src/client/api.rs
RolandSherwin: The result that we get from the `Swarm` is a `HashSet` which we then order it at the network layer. So the final value that we get here is ordered and contains unique peers. I thought preserving the order might be worth it.

safenode/src/network/event.rs
oetyng: What I mean is, that we're not ensured to get these results in any particular order. So when we have the n closest, those might just as well be the _farthest_ oners out of the bucket.

safenode/src/network/event.rs
```diff
@@ -107,7 +110,7 @@ impl SwarmDriver {
                     let new_peers: HashSet<PeerId> =
                         closest_peers.peers.clone().into_iter().collect();
                     current_closest.extend(new_peers);
```
oetyng, line 112: Is this check still necessary when we set replication factor for the kad cfg?
I don't know the connection there. How do those things relate (i.e. tthe kad cfg replication factor, and what is going on in this code block).

safenode/src/network/mod.rs
```diff
@@ -133,9 +126,18 @@ impl SwarmDriver {
...
-        // to outbound-only mode and don't listen on any address
+        // Kademlia configuration
+        let mut cfg = KademliaConfig::default();
+        // The `replication_factor` determines the a) The number of closer peers
+        // returned in response to a request b) The number of closest peers to a
+        // key to search for in an iterative query.
+        // It does not affect the maximum number of nodes in a bucket; this is set
+        // by `kad::K_VALUE`.
+        let replication_factor =
+            NonZeroUsize::new(CLOSE_GROUP_SIZE).ok_or(Error::NonZeroNumberExpected)?;
```
oetyng, line 137: I wonder if we could make the close group a `static` `NonZeroUsize` to avoid this `ok_or` check (which should always be unnecessary).

(it can't be made a `const` due to fn call)

Shouldn't botch the use of `CLOSE_GROUP_SIZE` in other places, I think?

safenode/src/network/mod.rs
```diff
@@ -234,6 +239,8 @@ fn restart_at_random(peer_id: &PeerId) {
 /// API to interact with the underlying Swarm
 pub struct Network {
     pub(super) swarm_cmd_sender: mpsc::Sender<SwarmCmd>,
+    #[allow(dead_code)]
```
oetyng, line 242: Why add it if dead code? I guess it is no longer dead, and this decorator missed to be removed?

safenode/src/network/mod.rs
oetyng: Hm, that's unfortunate.

safenode/src/network/mod.rs
```diff
@@ -318,3 +326,109 @@ impl Network {
...
+    use rand::thread_rng;
+    use std::{
+        collections::{BTreeMap, HashMap},
+        fmt,
+        time::Duration,
+    };
+    use xor_name::XorName;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn closest() -> Result<()> {
```
oetyng, line 348: I think this doesn't quite follow the test naming guidelines.. :thinking_face:

`closest_actually_are_the_closest` isn't a perfect name, just an example from the top of my head... but I think it shall be something like that.

safenode/src/network/event.rs
RolandSherwin: I have my doubts about this event as a whole tbh. When we use the `SwarmCmd::GetClosestPeers` to ask the swarm for the peers, the function reads that  "The result of the query is delivered in  `KademliaEvent::OutboundQueryCompleted{QueryResult::GetClosestPeers}` ". But there is no such `OutboundQueryCompleted`  variant inside `KademliaEvent`, instead we just have the `OutboundQueryProgressed` event?

So which is the truth here? Do the `Progressed` event actually means that we get the final result after querying the whole network? I don't see the `OutboundQueryProgressed` repeated as well, ie the network send one and only event back.

safenode/src/network/mod.rs
RolandSherwin: It is used inside the test, i.e, inside `cfg(test)` for now, so it still reports the field as dead_code. But I'm pretty sure storing the peer_id will be useful later down the line.

safenode/src/network/event.rs
RolandSherwin: Digging futher, the `KademliaEvent::OutboundQueryProgressed` is emitted by a function called `Kademlia::query_finished`.

maqi: There are flags within the returned result to indicate the status of the query and we shall define the `completion criterias` based on those.
There are TODO poins below mention that?

even with the changing of the kad cfg, this may still needed with a larger network for a more accurate result.

safenode/src/network/event.rs
RolandSherwin: >So when we have the n closest, those might just as well be the farthest ones out of the bucket, but we didn't yet receive the others, so we don't know.
The event `OutboundQueryProgressed` returns the final result, and there has been some incomplete docs/naming in libp2p? Added more details in the below comment.

maqi: Yes, I'd prefer stay with the current K_VALUE

safenode/src/node/api.rs
```diff
@@ -327,9 +324,10 @@ impl Node {
 
     async fn send_to_closest(&self, request: &Request) -> Result<Vec<Result<Response>>> {
         info!("Sending {:?} to the closest peers.", request.dst());
+        // todo: if `self` is present among the closest peers, the request should be routed to self?
```
oetyng, line 327: Could be.

It can be handled like that, or that we handle it directly. We have used both patterns. I'm on the fence on what is cleanest.

safenode/src/network/mod.rs
```diff
@@ -258,27 +265,28 @@ impl Network {
     }
 
     /// Find the closest peers to the given `XorName`
-    pub async fn get_closest_peers(&self, xor_name: XorName) -> Result<HashSet<PeerId>> {
+    /// If `client` is false, then include `self` among the `closest_peers`
+    pub async fn get_closest_peers(&self, xor_name: XorName, client: bool) -> Result<Vec<PeerId>> {
```
maqi, line 269: I think it will be clearer to have two pub functions then both call a priv fn with different flag, so lookes like:
```rust
pub async fn client_get_closest_peers(&self, xor_name: XorName) -> Result<Vec<PeerId>> {
    self.get_closest_peers(xor_name, true).await
}
pub async fn node_get_closest_peers(&self, xor_name: XorName) -> Result<Vec<PeerId>> {
    self.get_closest_peers(xor_name, false).await
}
async fn get_closest_peers(&self, xor_name: XorName, is_client: bool) -> Result<Vec<PeerId>> {
    ......
}
```

safenode/src/client/api.rs
oetyng: Yes, could absolutely be :+1:

safenode/src/network/event.rs
RolandSherwin: Reverting this change as per the Slack discussion

PR #82, opened 2023-04-12
feat: random query on peer added

maqi:
Fire a random get_closest_peers query when received a PeerAdded event.
This is necessary to allow nodes get a full knowledge of the network.
And allows the client upload chunk and get chunk back properly.


oetyng (approved):

oetyng:
Nice one! 🛳

PR #83, opened 2023-04-12
feat: implement wallet

oetyng:
An implementation of a local `Wallet` used by clients and nodes (the latter use them for their rewards).
There is one which is deposit only, and one which can also send tokens.

Later, a network `Wallet` store can be implemented thusly:
1. Chunk each `Dbc`, both spent and available.
2. For a semi-public `Wallet`:
     a. Store a register with address of your `PublicAddress`.
    Then push these ops:
     b. `self.address.encrypt(Deposit(ChunkAddress))`
     c. `self.address.encrypt(Spend(ChunkAddress))`
    And when the register has used `1023` entries:
     d. `self.address.encrypt(Extend(RegisterAddress))`
     ... which would occupy the last entry, and thus link to a new register.
 3. For a private `Wallet`:
     a. Store a register with address of `self.address.encrypt(self.address)`.
     ... then follow from b. in 2.
 4. Then, when a `Wallet` is to be loaded from the network:
     a. Get the `PublicAddress` from your secret.
     b. Fetch the register with address of either the plaintext of or the encrypted `PublicAddress`.
     c. Decrypt all entries and apply the ops to your `Wallet`, to get the current state of it.
     d. If there is another register linked at the end of this one, follow that link and repeat steps b., c. and d.

 We will already now pave for that, by mimicing that flow for the local storage of a `Wallet`.
 First though, a simpler local storage will be used. But after that a local register store can be implemented.

### DIFF
--- a/safenode/src/protocol/types/dbc_genesis.rs
+++ b/safenode/src/protocol/types/dbc_genesis.rs
@@ -1,0 +1,92 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use sn_dbc::{
+    rng, Dbc, DbcTransaction, Error as DbcError, Hash, InputHistory, MainKey, RevealedAmount,
+    RevealedInput, Token, TransactionBuilder,
+};
+
+use std::{fmt::Debug, result};
+use thiserror::Error;
+
+/// Amount of tokens to be owned by the Genesis DBC.
+/// At the inception of the Network a total supply of 4,294,967,295 whole tokens will be created.
+/// Each whole token can be subdivided 10^9 times,
+/// thus creating a total of 4,294,967,295,000,000,000 available units.
+pub const GENESIS_DBC_AMOUNT: u64 = u32::MAX as u64 * u64::pow(10, 9);
+
+/// A specialised `Result` type for types crate.
+pub type Result<T> = result::Result<T, Error>;
+
+/// Main error type for the crate.
+#[derive(Error, Debug, Clone)]
+// #[non_exhaustive]
+#[non_exhaustive]
+pub enum Error {
+    /// Error occurred when minting the Genesis DBC.
+    #[error("Genesis DBC error:: {0}")]
+    GenesisDbcError(String),
+    /// FailedToParseReason
+    #[error("Failed to parse reason: {0}")]
+    FailedToParseReason(#[from] DbcError),
+}
+
+/// Generate the genesis DBC.
+///
+/// Requires the initial section key to sign the share and a different secret key for the DBC.
+///
+/// The genesis DBC will be created using a different key from the initial section key. This is
+/// because the genesis DBC, along with its secret key, will be publicly available for auditing
+/// purposes. It needs to be a set rather than just a key because a spent proof share gets
+/// generated, which requires a key set. We can't use the same key for the genesis DBC and section
+/// because if the genesis DBC is publicly available, the secret key could be used to create a bad
+/// section that would be accepted by the network.
+#[allow(clippy::result_large_err)]
+pub fn create_genesis_dbc(genesis_main_key: &MainKey) -> Result<Dbc> {
+    let rng = &mut rng::thread_rng();
+
+    let dbc_id_src = genesis_main_key.random_dbc_id_src(rng);
+    let derived_key = genesis_main_key.derive_key(&dbc_id_src.derivation_index);
+    let revealed_amount = RevealedAmount::from_amount(GENESIS_DBC_AMOUNT, rng);
+
+    // Use the same key as the input and output of Genesis Tx.
+    // The src tx is empty as this is the first DBC.
+    let genesis_input = InputHistory {
+        input: RevealedInput::new(derived_key, revealed_amount),
+        input_src_tx: DbcTransaction {
+            inputs: vec![],
+            outputs: vec![],
+        },
+    };
+
+    let dbc_builder = TransactionBuilder::default()
+        .add_input(genesis_input)
+        .add_output(Token::from_nano(GENESIS_DBC_AMOUNT), dbc_id_src)
+        .build(Hash::default(), rng::thread_rng())
+        .map_err(|err| {
+            Error::GenesisDbcError(format!(
+                "Failed to build the DBC transaction for genesis DBC: {err}",
+            ))
+        })?;
+
+    // build the output DBCs
+    let output_dbcs = dbc_builder.build_without_verifying().map_err(|err| {
+        Error::GenesisDbcError(format!(
+            "DBC builder failed to create output genesis DBC: {err}",
+        ))
+    })?;
+
+    // just one output DBC is expected which is the genesis DBC
+    let (genesis_dbc, _) = output_dbcs.into_iter().next().ok_or_else(|| {
+        Error::GenesisDbcError(
+            "DBC builder (unexpectedly) contains an empty set of outputs.".to_string(),
+        )
+    })?;
+
+    Ok(genesis_dbc)
+}
--- a/safenode/src/protocol/types/dbc_genesis.rs
+++ b/safenode/src/protocol/types/dbc_genesis.rs
@@ -91,14 +91,6 @@ pub fn create_genesis_dbc(genesis_main_key: &MainKey) -> GenesisResult<Dbc> {
     Ok(genesis_dbc)
 }
 
-// Copyright 2023 MaidSafe.net limited.
-//
-// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
-// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
-// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-// KIND, either express or implied. Please review the Licences for the specific language governing
-// permissions and limitations relating to use of the SAFE Network Software.
-
 pub(super) type Result<T> = std::result::Result<T, Error>;
 
 /// Error type returned by the API
@@ -136,28 +128,11 @@ pub fn send_tokens(
     dbcs: Vec<(Dbc, DerivedKey)>,
     recipients: Vec<(Token, DbcIdSource)>,
     change_to: PublicAddress,
-    // reason: DbcReason,
-    // priority: SpendPriority,
 ) -> Result<ReissueOutputs> {
-    // let mut attempts = 0;
-    // let mut max_required = Token::zero();
-
     // We need to select the necessary number of dbcs from those that we were passed.
     // This will also account for any fees.
     let reissue_inputs = select_inputs(dbcs, recipients, change_to)?;
-
     reissue(reissue_inputs)
-    // // then we can reissue
-    // match reissue(reissue_inputs).await {
-    //     Ok(outputs) => return Ok(outputs),
-    //     Err(Error::FeeTooLow(required)) => {
-    //         max_required = required;
-    //         attempts += 1;
-    //     }
-    //     error => return error,
-    // }
-
-    // Err(Error::FeeTooLow(max_required))
 }
 
 /// The input dbcs will be spent on the network, and the resulting
@@ -173,10 +148,6 @@ fn reissue(
         input_dbcs,
         outputs,
         change: (change, change_to),
-        // #[cfg(not(feature = "data-network"))]
-        // fees_paid,
-        // #[cfg(not(feature = "data-network"))]
-        // all_fee_cipher_params,
     } = reissue_inputs;
 
     let mut inputs = vec![];
@@ -206,62 +177,13 @@ fn reissue(
         tx_builder = tx_builder.add_output(change, dbc_id_src);
     }
 
-    // let inputs_spent_proofs: BTreeSet<SpentProof> = input_dbcs
-    //     .iter()
-    //     .flat_map(|dbc| dbc.inputs_spent_proofs.clone())
-    //     .collect();
-    // let inputs_spent_transactions: BTreeSet<DbcTransaction> = input_dbcs
-    //     .iter()
-    //     .flat_map(|dbc| dbc.inputs_spent_transactions.clone())
-    //     .collect();
-
     // Finalize the tx builder to get the dbc builder.
     let dbc_builder = tx_builder
         .build(Hash::default(), &mut rng)
         .map_err(Error::DbcError)?;
 
-    // // Get fee outputs to generate the fee ciphers.
-    // #[cfg(not(feature = "data-network"))]
-    // let outputs = dbc_builder
-    //     .revealed_outputs
-    //     .iter()
-    //     .map(|output| (output.public_key, output.revealed_amount))
-    //     .collect();
-
     let inputs = dbc_builder.signed_spends();
     let spent_dbcs = inputs.iter().map(|spent| spent.dbc_id()).cloned().collect();
-
-    // // Spend all the input DBCs, collecting the spent proof shares for each of them
-    // for (public_key, tx) in inputs {
-    //     // // Generate the fee ciphers.
-    //     // #[cfg(not(feature = "data-network"))]
-    //     // let input_fee_ciphers = {
-    //     //     let fee_cipher_params = all_fee_cipher_params
-    //     //         .get(&public_key)
-    //     //         .ok_or(Error::DbcReissueError("Missing fee!".to_string()))?;
-    //     //     fee_ciphers(&outputs, fee_cipher_params)?
-    //     // };
-
-    //     let tx_hash = Hash::from(tx.hash());
-    //     // TODO: spend DBCs concurrently spawning tasks
-    //     let mut attempts = 0;
-    //     loop {
-    //         attempts += 1;
-    //         client
-    //             .spend_dbc(
-    //                 public_key,
-    //                 tx.clone(),
-    //                 reason,
-    //                 inputs_spent_proofs.clone(),
-    //                 inputs_spent_transactions.clone(),
-    //                 #[cfg(not(feature = "data-network"))]
-    //                 input_fee_ciphers.clone(),
-    //             )
-    //             .await?;
-
-    //         let signed_spend = client.get_signed_spend(public_key).await?;
-    //     }
-    // }
 
     // Perform verifications of input TX and spentproofs,
     // as well as building the output DBCs.
@@ -281,8 +203,6 @@ fn reissue(
         outputs: output_dbcs,
         change: change_dbc,
         spent_dbcs,
-        // #[cfg(not(feature = "data-network"))]
-        // fees_paid,
     })
 }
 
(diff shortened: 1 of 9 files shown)


safenode/src/protocol/types/dbc_genesis.rs
```diff
@@ -0,0 +1,92 @@
...
+use sn_dbc::{
+    rng, Dbc, DbcTransaction, Error as DbcError, Hash, InputHistory, MainKey, RevealedAmount,
+    RevealedInput, Token, TransactionBuilder,
+};
+
+use std::fmt::Debug;
+use thiserror::Error;
+
+/// Amount of tokens to be owned by the Genesis DBC.
+/// At the inception of the Network a total supply of 4,294,967,295 whole tokens will be created.
```
oetyng, line 18: This shall actually be only 30% of it. Will fix the comment.

safenode/src/protocol/types/dbc_genesis.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[error("Genesis DBC error:: {0}")]
+    GenesisDbcError(String),
+    /// FailedToParseReason
+    #[error("Failed to parse reason: {0}")]
+    FailedToParseReason(#[from] DbcError),
+}
+
+/// Generate the genesis DBC.
+///
+/// Requires the initial section key to sign the share and a different secret key for the DBC.
```
oetyng, line 41: Oops. Old docs. Will clean up.

safenode/src/protocol/types/transfers/error.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// Not enough was paid in fees for the Elders to process the spend.
+    #[error("Too low amount for the transfer. Highest required fee: {0:?}.")]
+    FeeTooLow(Token),
+    /// DbcError
+    #[error("DbcError: {0}")]
+    Dbcs(#[from] DbcError),
+    /// DbcReissueFailed
+    #[error("DbcReissueFailed: {0}")]
+    DbcReissueFailed(String),
+    /// Verification of DBC validly signed by a known section failed
```
oetyng, line 32: Doc update needed.
The error itself might be used later.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub amount: RevealedAmount,
+}
+
+/// A function for creating an offline transfer of tokens.
+/// This is done by creating new dbcs to the recipients (and a change dbc if any)
+/// by selecting from the available input dbcs, and creating the necessary
+/// spends to do so.
+///
+/// Those signed spends are found in each new dbc, and must be uploaded to the network
+/// for the transactions to take effect.
```
oetyng, line 58: Nit: Strictly, shall be transaction in singular.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    let mut created_dbcs: Vec<_> = dbc_builder
+        .build()
+        .map_err(Error::Dbcs)?
+        .into_iter()
+        .map(|(dbc, amount)| CreatedDbc { dbc, amount })
+        .collect();
+
+    let mut change_dbc = None;
+    created_dbcs.retain(|created| {
+        if created.dbc.public_address() == &change_to && change.as_nano() > 0 {
```
oetyng, line 134: This is not 100% solid in my mind, since we could have another output with sending to ourselves.

In the wallet (which calls this code) such an operation is skipped.

But anyway..

safenode/src/protocol/types/wallet/error.rs
```diff
@@ -0,0 +1,28 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use thiserror::Error;
+
+/// Specialisation of `std::Result`.
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Payment errors.
```
oetyng, line 14: Should rather call it Transfer errors, I think..

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+            .collect();
+        if to.is_empty() {
+            return Ok(vec![]);
+        }
+
+        let mut available_dbcs = vec![];
+        for dbc in self.wallet.available_dbcs.values() {
+            let derived_key = dbc
+                .derived_key(&self.wallet.key)
+                .expect("This dbc to be ours.");
```
oetyng, line 151: Oops, we need to skip here instead of the panic.

safenode/src/protocol/types/wallet/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+//! First though, a simpler local storage will be used. But after that a local register store can be implemented.
+
+mod error;
+mod local_store;
+mod network_store;
+
+pub use self::{
+    error::{Error, Result},
+    local_store::{LocalDepositWallet, LocalSendWallet},
+    // network_store::NetworkWallet,
```
b-zee, line 41: Remove this comment?

safenode/src/protocol/types/wallet/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Depending on the implementation of the send client, this may
+    /// also register the transaction with the network.
+    async fn send(&mut self, to: Vec<(Token, PublicAddress)>) -> Result<Vec<CreatedDbc>>;
+}
+
+/// A deposit wallet is a wallet that can receive tokens from other wallets.
+/// It can however not send tokens to other addresses.
+pub trait DepositWallet {
+    // /// Loads a wallet from the given path.
+    // fn load_from(path: &Path) -> Self;
```
b-zee, line 98: Remove this comment?

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    #[derive(Clone)]
+    struct MockSendClient {}
+
+    impl MockSendClient {
+        fn new() -> Self {
+            Self {}
+        }
+    }
```
b-zee, line 326: Perhaps use a unit struct?

```suggestion
    struct MockSendClient;
```

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+
+        let mut received_dbcs = dbcs
+            .into_iter()
+            .filter_map(|dbc| {
+                let id = dbc.id();
+                (!self.spent_dbcs.contains_key(&id)).then_some((id, dbc))
+            })
+            .filter_map(|(id, dbc)| dbc.derived_key(&self.key).is_ok().then_some((id, dbc)))
+            .collect();
+
```
b-zee, line 85: Should we also check here if the DBCs we deposit are spent already?

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        // If we've already combined input DBCs for the total output amount, then stop.
+        match change_amount.checked_sub(dbc_balance) {
+            Some(pending_output) => {
+                change_amount = pending_output;
+                if change_amount.as_nano() == 0 {
+                    break;
+                }
+            }
+            None => {
+                change_amount = Token::from_nano(dbc_balance.as_nano() - change_amount.as_nano());
```
maqi, line 202: bit confused here, when an overflow happens, this means the `dbc_balance` is greater than our left un-paid change,
which, the `change_aount` can be fulfill (i.e. set to 0 ?)

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn balance(&self) -> Token {
+        self.balance
+    }
+
+    fn new_dbc_address(&self) -> DbcIdSource {
+        self.key.random_dbc_id_src(&mut rand::thread_rng())
+    }
+
+    fn deposit(&mut self, dbcs: Vec<Dbc>) {
```
b-zee, line 72: Perhaps it's an idea to make a method that deposits a single DBC? This could return `Option<Dbc>`, moving the `Dbc` back to the caller if the DBC is not deposited. Then adjust the balance only by adding up the amount of the DBC just added.

Feel free to ignore, just an idea.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+mod error;
```
b-zee, line 9: Perhaps add module level documentation on what exactly a 'transfer' is conceptually?

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// The current balance of the wallet.
+    balance: Token,
+    /// These are dbcs we've owned, that have been
+    /// spent when sending tokens to other addresses.
+    spent_dbcs: BTreeMap<DbcId, Dbc>,
+    /// These are the dbcs we own that are not yet spent.
+    available_dbcs: BTreeMap<DbcId, Dbc>,
+    /// These are the dbcs we've created by
+    /// sending tokens to other addresses.
+    output_history: Vec<CreatedDbc>,
```
maqi, line 38: what's the diff to the `spent_dbcs` ?
according to the comment, seems they are the same?

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+            .available_dbcs
+            .iter()
+            .flat_map(|(_, dbc)| {
+                dbc.derived_key(&self.key)
+                    .map(|derived_key| (dbc, derived_key))
+            })
+            .flat_map(|(dbc, derived_key)| dbc.revealed_input(&derived_key))
+            .fold(0, |total, amount| total + amount.revealed_amount().value());
+
+        self.balance = Token::from_nano(new_balance);
```
maqi, line 98: why not just calculate a total amount from `received_dbcs` then:
`self.balance += total_balance_from_received_dbc` ?
which shall save certain computations?

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+
+        local_wallet.deposit(vec![genesis]);
+
+        assert_eq!(Token::zero(), local_wallet.balance());
+
+        Ok(())
+    }
+
+    /// --------------------------------
+    /// <-------> SendWallet <--------->
```
maqi, line 262: what does it mean for `SendWallet` ?
transfer the ownship of the wallet to other,
or send the whole balance of a wallet to other,
or wallet send some balance of it to other ?

safenode/src/protocol/types/wallet/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    async fn send(
+        &self,
+        dbcs: Vec<(Dbc, DerivedKey)>,
+        to: Vec<(Token, DbcIdSource)>,
+        change_to: PublicAddress,
+    ) -> Result<TransferDetails>;
+}
+
+/// A send wallet is a wallet that can send tokens to other addresses.
+/// It is also a deposit wallet, so it can receive tokens from other wallets.
```
maqi, line 72: if it can do both `send balance out` and `receive deposit in` , then calling it `SendWallet` is bit misleading ?

safenode/src/protocol/types/wallet/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn deposit(&mut self, dbcs: Vec<Dbc>);
+    /// Sends the given tokens to the given addresses.
+    /// Returns the new dbcs that were created.
+    /// Depending on the implementation of the send client, this may
+    /// also register the transaction with the network.
+    async fn send(&mut self, to: Vec<(Token, PublicAddress)>) -> Result<Vec<CreatedDbc>>;
+}
+
+/// A deposit wallet is a wallet that can receive tokens from other wallets.
+/// It can however not send tokens to other addresses.
```
maqi, line 95: If SendWallet can do the both, then why have a separate DepositWallet ?

safenode/src/protocol/types/wallet/network_store.rs
```diff
@@ -0,0 +1,7 @@
+// Copyright 2023 MaidSafe.net limited.
```
maqi, line 1: this is actually an empty file ?

maybe at least put some pseudo code or TODO comments in?

b-zee (approved):

safenode/src/protocol/types/dbc_genesis.rs
oetyng: The `GENESIS_DBC` will not be the full `u32::MAX`, but only 30% of it. 70% is to be issued later, when a solution to that has been found.
So, that's what needs an update here.

safenode/src/protocol/types/wallet/mod.rs
oetyng: Yeah, could do. That's the next iteration stuff.

safenode/src/protocol/types/wallet/mod.rs
oetyng: Maybe, that's the next iteration stuff. So it's meant to give a hint to people that that is what's coming next.

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+            available_dbcs: BTreeMap::new(),
+            output_history: vec![],
+        }
+    }
+
+    // /// Loads a serialized wallet from a path.
+    // fn load_from(path: &Path) -> Self {
+    //     Self {
+    //     }
+    // }
```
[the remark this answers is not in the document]
oetyng: The plan is to make it simple first. The simplest possible. So that we can see with the in-mem storage. The next simplest is just serializing it to file system. These steps are all described in `wallet/mod.rs`. The end goal ios to store the wallet entries in a Register, as is described in `wallet/mod.rs`. And so, one of the steps between serializing it to filesystem and storing it to a Register in the network, is to store it in a Register locally. Such a local Register store is something we will have for Clients anyway, as that is going to be the work flow (apply ops locally, sync with network intermittently).

safenode/src/protocol/types/wallet/local_store.rs
oetyng: Sure :+1:

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -0,0 +1,344 @@
...
+    #[tokio::test]
+    #[allow(clippy::result_large_err)]
+    async fn sending_decreases_balance() -> Result<()> {
+        let sender_main_key = MainKey::random();
+        let sender_dbc =
+            create_genesis_dbc(&sender_main_key).expect("Genesis creation to succeed.");
+
+        let client = MockSendClient::new();
+        let mut send_wallet: LocalSendWallet<MockSendClient> =
+            SendWallet::<MockSendClient>::new(sender_main_key, client);
```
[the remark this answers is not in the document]
oetyng: Will try it out again. It was complaining about this previously.

safenode/src/protocol/types/transfers/mod.rs
oetyng: Yeah, definitely.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    rng, Dbc, DbcIdSource, DerivedKey, Hash, InputHistory, PublicAddress, RevealedAmount,
+    RevealedInput, Token, TransactionBuilder,
+};
+
+use std::fmt::Debug;
+
+/// The input details to a transfer of tokens.
+#[derive(Debug)]
+pub struct Inputs {
+    /// The selected dbcs to spend, as to transfer their tokens to the recipients.
```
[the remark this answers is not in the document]
oetyng: Will see if I can clarify that part of the docs.

The meaning is the tokens that are to be transferred to the recipients, are taken from the selected dbcs, which are then spent in the process.

Input DBCS to Output DBCS can be `one-to-one`, `one-to-many`, `many-to-one` or `many-to-many`.
Here we take a list of each, and it's up to the client to populate them as to get one of the above.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub dbcs_to_spend: Vec<(Dbc, DerivedKey)>,
+    /// The amounts and dbc ids for the dbcs that will be created to hold the transferred tokens.
+    pub recipients: Vec<(Token, DbcIdSource)>,
+    /// Any surplus amount after spending the necessary input dbcs.
+    pub change: (Token, PublicAddress),
+}
+
+/// The token transfer results.
```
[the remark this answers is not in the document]
oetyng: Definitely :+1: I had that thought as well, tried with several different names and descriptions. A bit tricky this one. But will irk something out.

safenode/src/protocol/types/wallet/error.rs
```diff
@@ -0,0 +1,28 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    ///
+    #[error("Current balance: {balance}. Attempted spend: {attempted_spend}")]
+    NotEnoughBalance {
+        ///
```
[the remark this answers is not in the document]
oetyng: Yep, definitely.

safenode/src/protocol/types/wallet/local_store.rs
oetyng: I don't think the comment describes the same thing?
- `These are the dbcs we've created` -> Creating something means it didn't exist before.
- `These are dbcs we've owned, that have been spent` -> Spending a dbc means it is spent. It's used. Inert, dead.

So, the `output_history` field could probably be renamed to describe that better. But the coments are quite clear I'd say?

safenode/src/protocol/types/wallet/local_store.rs
oetyng: That's a way it could be done as well. Just consider the other approach slightly more solid, to not rely on intermediate state. And the computation cost is negligible.

safenode/src/protocol/types/wallet/local_store.rs
oetyng: It's described in the comments. A send wallet can send. A deposit wallet can only deposit.

safenode/src/protocol/types/wallet/mod.rs
oetyng: A wallet that can't be deposited to isn't really a wallet.

The distinguishing feature is if it can send or not.

I can remove the comment about it also being able to take deposits. It was a co-pilot line actually and I was lazy and just took it.

safenode/src/protocol/types/wallet/mod.rs
oetyng: Because you might want a wallet which can't send.

For example, the nodes only receiving rewards don't need the api to send tokens.

They can instantiate such wallets outside of the node software. Such as in a client.

safenode/src/protocol/types/transfers/mod.rs
oetyng: Yeah, I'm actually not entirely sure about this line myself. It is a copy paste from an older implementation.

safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        recipients,
+        change: (change, change_to),
+    } = send_inputs;
+
+    let mut inputs = vec![];
+    for (dbc, derived_key) in dbcs_to_spend {
+        let revealed_amount = match dbc.revealed_amount(&derived_key) {
+            Ok(amount) => amount,
+            Err(err) => {
+                warn!("Ignoring dbc, as it didn't have the correct derived key: {err}");
```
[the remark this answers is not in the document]
oetyng: Well, it's a private function and it is only called with inputs that are gotten from the `select_inputs` fn, which has already filtered out dbcs that we don't own (which won't get passed in there from the beginning, as we filter that at wallet level as well..).

Anyway, so that warn log line is practically unreachable.

We could move around some code to produce that `InputHistory` struct already in the `select_inputs` function, as we at the end of that function compare in and out amount, and return an error if we don't have enough to spend.

PR #86, opened 2023-04-13
feat(api): Register client API

### LINKED ISSUE #80: Create a Register API supporting resolution of concurrent ops made by the user

bochaco:
- Supports offline-first type of operations for users to work on Registers offline and syncing local changes with remote replicas on the network when they decide to.
- Public APIs for Register in offline mode are all sync, whilst those that work 'online', i.e. syncing right after each operation is made, are all `async`.
- Allow users to resolve/merge branches of Registers entries if their use case requires a single branch of values as content.

Resolves #80

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -168,7 +168,7 @@ jobs:
         timeout-minutes: 2
 
       - name: Start a client to carry out register actions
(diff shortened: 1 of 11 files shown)


dirvine (approved):
Nice work. 
Later some client examples for multisig would be helpful. We can have backup multisig (derived from master key) and also DKG (which will mean we need to get real time messaging in place)

PR #87, opened 2023-04-14
chore: update wallet and transfer docs

oetyng:
Also an attempt at better naming for wallet variable of created dbcs.
Still not entirely satisfactory though..

### DIFF
--- a/safenode/src/protocol/types/transfers/mod.rs
+++ b/safenode/src/protocol/types/transfers/mod.rs
@@ -6,6 +6,27 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+//! This module contains the functions for creating an offline transfer of tokens.
+//! This is done by emptying the input dbcs, thereby rendering them spent, and creating
+//! new dbcs to the recipients (and a change dbc if any) containing the transferred tokens.
+//! When a transfer is created, it is not yet registered on the network. The signed spends of
+//! the transfer is found in the new dbcs, and must be uploaded to the network to take effect.
+//! The peers will validate each signed spend they receive, before accepting it.
+//! Once enough peers have accepted all the spends of the transaction, and serve them upon request,
+//! the transfer is completed and globally recognised.
+//!
+//! The transfer is created by selecting from the available input dbcs, and creating the necessary
+//! spends to do so. The input dbcs are selected by the user, and the spends are created by this
+//! module. The user can select the input dbcs by specifying the amount of tokens they want to
+//! transfer, and the module will select the necessary dbcs to transfer that amount. The user can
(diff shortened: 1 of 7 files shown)


safenode/src/protocol/types/transfers/mod.rs
```diff
@@ -131,6 +155,8 @@ fn transfer(send_inputs: Inputs) -> Result<Outputs> {
 
     let mut change_dbc = None;
     created_dbcs.retain(|created| {
+        // This is not 100% solid in my mind, since we could have another output that is sending to ourselves.
+        // In the wallet (which calls this code) such an operation is skipped. But worth noting anyway.
         if created.dbc.public_address() == &change_to && change.as_nano() > 0 {
             change_dbc = Some(created.dbc.clone());
             false
```
grumbach, line 162: If any other DBCs are for us, we will remove them
```
             change_dbc = Some(created.dbc.clone());
             false
```
An overwrite the `change_dbc` variable. 
Effectively removing the DBC from both `created_dbcs` and `change_dbc`.

safenode/src/protocol/types/wallet/local_store.rs
```diff
@@ -146,10 +149,14 @@ impl<C: SendClient + Send + Sync + Clone> SendWallet<C> for LocalSendWallet<C> {
...
-            available_dbcs.push((dbc.clone(), derived_key));
+            if let Ok(derived_key) = dbc.derived_key(&self.wallet.key) {
+                available_dbcs.push((dbc.clone(), derived_key));
+            } else {
+                warn!(
+                    "Skipping DBC {:?} because we don't have the key to spend it",
+                    dbc.id()
+                );
+            }
         }
```
grumbach, line 160: Having DBCs that are not ours in our wallet should be a bug right? While having `expect` makes our code panic prone and we should indeed remove them, we should also note that this change potentially hides this bug from our eyes. 
Or is this a feature?

grumbach (changes requested):
The `change_dbc` bug is important IMO we should fix it.
One quick and dirty way to do it is to take the first DBC in our name and leave all the others in `created_dbcs` so we at least don't make DBCs disappear.

safenode/src/protocol/types/wallet/local_store.rs
oetyng: The wallet we have implemented does not add dbcs that are not for us.
We don't use expect and panics in production code, not for bugs either.

So, this is unreachable code for us. And it just needs some handling to compile.. so a warn log..

oetyng:
> The `change_dbc` bug is important IMO we should fix it. One quick and dirty way to do it is to take the first DBC in our name and leave all the others in `created_dbcs` so we at least don't make DBCs disappear.

Yeah this is old code. It's not possible to get there since all layers above prevent it. So it's not as issue.
But, yeah as I noted, it is not solid.

PR #93, opened 2023-04-14
feat tcp net

b-zee:
- feat(node): add behaviours with TCP transport
- feat(node): dial bootstrap nodes

### DIFF
--- a/safenode/src/bin/kadnode.rs
+++ b/safenode/src/bin/kadnode.rs
@@ -28,9 +28,14 @@ async fn main() -> Result<()> {
     let _log_appender_guard = init_node_logging(&opt.log_dir)?;
 
     let socket_addr = SocketAddr::new(opt.ip, opt.port);
+    let peers = parse_peer_multiaddres(&opt.peers)?;
 
     info!("Starting a node...");
     let (_node, node_events_channel) = Node::run(socket_addr).await?;
+
+    for (_peer_id, _addr) in peers {
+        // node.network.dial(peer_id, addr).await?;
+    }
 
     let mut node_events_rx = node_events_channel.subscribe();
     if let Ok(event) = node_events_rx.recv().await {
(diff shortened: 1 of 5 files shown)


safenode/src/bin/kadnode.rs
```diff
@@ -28,12 +27,33 @@ async fn main() -> Result<()> {
...
     info!("Starting a node...");
-    let (_node, node_events_channel) = Node::run(socket_addr).await?;
+    let (node, node_events_channel) = Node::run(socket_addr).await?;
 
     let mut node_events_rx = node_events_channel.subscribe();
-    if let Ok(event) = node_events_rx.recv().await {
+
+    for (peer_id, addr) in &peers {
+        // TODO: Do we want to wait for the dial?
+        if let Err(err) = node.network.dial(*peer_id, addr.clone()).await {
```
maqi, line 39: as here is the `initial` dials, I think here shall wait for the dial result.
However, as the reading in list may contain outdated peer, may better set a time out for it.
AND, the initial dial process may can be completed on the first (or first X) successful dials?

safenode/src/bin/kadnode.rs
```diff
@@ -28,12 +27,33 @@ async fn main() -> Result<()> {
     let _log_appender_guard = init_node_logging(&opt.log_dir)?;
 
     let socket_addr = SocketAddr::new(opt.ip, opt.port);
+    let peers = parse_peer_multiaddreses(&opt.peers)?;
 
     info!("Starting a node...");
-    let (_node, node_events_channel) = Node::run(socket_addr).await?;
+    let (node, node_events_channel) = Node::run(socket_addr).await?;
```
maqi, line 33: this conflicts with the current code base that having the returned `node` removed totally?

safenode/src/bin/kadnode.rs
```diff
@@ -28,12 +27,33 @@ async fn main() -> Result<()> {
...
+        };
+    }
+
+    loop {
+        let event = node_events_rx.recv().await;
+        let event = match event {
+            Ok(event) => event,
+            Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                tracing::error!("Node event channel closed!");
+                break;
```
maqi, line 50: in this case, shall the node to be considered as `dead` and carry out a restart or just panic?

safenode/src/bin/kadnode.rs
```diff
@@ -62,18 +82,32 @@ struct Opt {
     /// Defaults to 0.0.0.0, which will bind to all network interfaces.
     #[clap(long, default_value_t = IpAddr::V4(Ipv4Addr::UNSPECIFIED))]
     ip: IpAddr,
+
+    /// Bootstrap node to help us get connected to the network. Can be specified multiple times.
```
maqi, line 86: better have an example to show how the `peers` opt got used
btw. `MultiAddrs` seems a complex format to be used via CLI. maybe we don't need this opt at all, just read from an input file directly? like the previous network_contacts

safenode/src/node/mod.rs
```diff
@@ -21,7 +21,8 @@ use crate::{network::Network, storage::DataStorage};
 /// storage, and broadcasts node-related events.
 #[derive(Clone)]
 pub struct Node {
-    network: Network,
+    /// The `Network` instance.
+    pub network: Network,
```
maqi, line 25: better access via `access functions` instead of expose directly

safenode/src/bin/kadnode.rs
b-zee: The network contacts file could also be configured via a file. I understand this is a complex format, but I think it's reasonable for the time being. In my personal opinion, I am fine with leaving out the complexity of a config json.

safenode/src/network/mod.rs
```diff
@@ -129,24 +126,40 @@ impl SwarmDriver {
...
-        let transport = libp2p_quic::tokio::Transport::new(quic_config);
-        let transport = transport
-            .map(|(peer_id, muxer), _| (peer_id, StreamMuxerBox::new(muxer)))
+        let (relay_client_transport, relay_client) = libp2p::relay::client::new(peer_id);
+
+        // The transport is a combination of the relay and TCP.
+        let tcp_relay_transport = relay_client_transport
+            .or_transport(libp2p::tcp::tokio::Transport::new(
+                libp2p::tcp::Config::default().port_reuse(true),
+            ))
```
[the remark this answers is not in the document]
b-zee: At the moment, I can't get QUIC to work with NAT stuff. It seems (Rust) libp2p is still working on adding proper hole punch support for QUIC.

PR #95, opened 2023-04-14
feat(transfers): connect spends, fees and the msgs

### LINKED ISSUE #75: Pull in transfer priority queue from old repo

### LINKED ISSUE #94: Wire up fees, priority queue, and spends with msgs.

oetyng:
Resolves #94.


safenode/src/node/api.rs
```diff
@@ -113,118 +112,123 @@ impl Node {
...
-            }
-            Request::Query(query) => {
-                let resp = self.storage.read(&query, User::Anyone).await;
-                self.send_response(Response::Query(resp), response_channel)
-                    .await;
-            }
+        let response = match request {
+            Request::Cmd(cmd) => Response::Cmd(self.handle_cmd(cmd).await),
+            Request::Query(query) => Response::Query(self.handle_query(query).await),
             Request::Event(event) => {
```
jacderida, line 118: Hey, it looks like this doesn't relate to this PR directly, but I'm just curious, why would an `Event` be handled as a request? We have an event handler task, so just curious why that's not part of this? The doc comments say, "A fact sent to peers." Should it perhaps be called a `Fact`?

safenode/src/node/api.rs
```diff
@@ -113,118 +112,123 @@ impl Node {
...
+                CmdResponse::StoreChunk(resp)
             }
-            // This should be unreachable, as we only return protocol errors.
-            other => other?,
-        };
-
-        let response = match self
-            .storage
-            .write(&Cmd::Dbc {
+            Cmd::Register(cmd) => {
```
jacderida, line 171: Should these commands perhaps be called `WriteRegister` and `SpendDbc`?

safenode/src/node/api.rs
```diff
@@ -113,118 +112,123 @@ impl Node {
...
-            Err(super::Error::Protocol(error)) => {
-                // We drop spend attempts with invalid parents,
-                // and return an error to the client.
-                self.send_response(
-                    Response::Cmd(CmdResponse::Spend(Err(error))),
-                    response_channel,
-                )
-                .await;
-                return Ok(());
+    async fn handle_cmd(&mut self, cmd: Cmd) -> CmdResponse {
```
jacderida, line 165: Probably not something to be addressed in this PR, but it could be worth thinking about unit testing the command and event handlers. The storage and so on is something it would be possible to mock.

safenode/src/transfers/error.rs
```diff
@@ -0,0 +1,95 @@
...
+    SpendNotFound(DbcAddress),
+    /// A double spend attempt was detected.
+    #[error("A double spend attempt was detected. Incoming and existing spend are not the same: {new:?}. Existing: {existing:?}")]
+    DoubleSpendAttempt {
+        /// New spend that we received.
+        new: Box<SignedSpend>,
+        /// Existing spend of same id that we already have.
+        existing: Box<SignedSpend>,
+    },
+    /// We were notified about a double spend attempt, but they were for different dbcs.
```
jacderida, line 47: Would it be worth adding something to the doc comments to explain how this error could occur? The name is kind of begging the question, but it doesn't seem obvious why something you thought was a double spend would actually turn out not to be.

safenode/src/transfers/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+
+const STARTING_FEE: u64 = 4000; // 0.000004 SNT
+
+pub(super) struct Transfers {
+    our_name: XorName,
+    node_reward_key: MainKey,
+    spend_queue: SpendQ<SignedSpend>,
+    storage: SpendStorage,
+}
+
```
jacderida, line 34: Do the local wallet tests still provide all the test coverage for this code? I noticed those tests weren't changed in this PR.

safenode/src/transfers/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+        let signed_src_tx_hash = signed_spend.src_tx_hash();
+
+        if provided_src_tx_hash != signed_src_tx_hash {
+            return Err(Error::SignedSrcTxHashDoesNotMatchProvidedSrcTxHash {
+                signed_src_tx_hash,
+                provided_src_tx_hash,
+            });
+        }
+
+        // 2. Try extract the fee paid for this spend, and validate it.
```
jacderida, line 100: Minor style issue. If the code says `validate_fee`, the comment isn't providing much in addition. Personally, if you wanted to describe the step by step process, I think that would be better done in the doc comments and let the code here speak for itself.

safenode/src/node/api.rs
oetyng: Event is the proper and and everywhere used terminology.
Events are facts, that's how they are described. Something that happened, past tense.
(Check out event sourcing and CQRS, and you'll see).

Libp2p though has a strange use of cmds, queries and events.
Request/Response is what they (libp2p) call the vehicle for sending msgs between nodes.

Cmds, Queries and Events are the three things that a system - especially a messaging system - sends around. Doesn't matter if it's local or over the wire. The event handler task is for something else, network events. Not domain logic events.

Cmds - imperative; do this! Mutates.
Queries - read only.
Events - facts, it already happened.

So, when something happened at a node, we don't use a cmd to tell other nodes that it shall happen. And we don't ask (query) other nodes about it happening. We tell the world, that it happened.

And therefore we use a request, since that is the bearer of all intra-node messages.

safenode/src/transfers/error.rs
oetyng: It's just a path handled. When there are two items and we can't have type safety that ensures the dbc ids are the same.
Then we must validate: are the ids the same?

If they are not, then that is a path that must be handled some way, even though we don't have a path in our code to reach that, since the msg could be sent with malice.

It can be seen where the error is raised.

safenode/src/transfers/mod.rs
oetyng: The local wallet isn't part of this logic. This is node code for spends.
The closest a wallet is to this code is a layer up and next to it, as a nmode reveices rewards and need to put them somewhere.

SpendQueue has its tests that was added in this PR.

safenode/src/transfers/mod.rs
oetyng: Could be. But I go quite heavily on explaining step by step. I prefer that over assuming that code is self-explaining. Most people assume to much with that, as the code is known to them, with tons of implicit knowledge.

safenode/src/node/api.rs
oetyng: When we have many levels, we have been categorizing them at the higher level, providing the cmd description in the lower level. 

RegisterCmd has Create + Edit.

What category of cmd? Register. What type of register cmd? Edit.

Dbc has only one variant - so far - and could be SpendDbc there doesn't seem to be more variants to be added there.

safenode/src/transfers/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+    },
+    storage::SpendStorage,
+};
+
+use sn_dbc::{DbcId, DbcTransaction, MainKey, SignedSpend, Token};
+
+use std::collections::{BTreeMap, BTreeSet};
+use xor_name::XorName;
+
+const STARTING_FEE: u64 = 4000; // 0.000004 SNT
```
grumbach, line 26: These kind of hardcoded values should probably be in a config file right?

grumbach (commented):
couple notes

safenode/src/storage/chunks.rs
```diff
@@ -86,4 +64,30 @@ impl ChunkStorage {
...
+            .read()
+            .await
+            .iter()
+            .map(|(addr, _)| *addr)
+            .collect()
+    }
+
+    #[allow(dead_code)]
+    pub(super) async fn remove_chunk(&self, address: &ChunkAddress) -> Result<()> {
+        trace!("Removing Chunk: {address:?}");
```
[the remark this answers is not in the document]
oetyng: Because when a node responsibility-range shrinks so that chunks fall out of it, they can optionally remove chunks if they are getting close to the minimum required space.

safenode/src/transfers/mod.rs
oetyng: This is obviously temporary. This value is completely out of the blue.

safenode/src/storage/spends.rs
```diff
@@ -79,7 +72,26 @@ impl SpendStorage {
...
+        let size_of_new = std::mem::size_of_val(signed_spend);
+        let address = dbc_address(signed_spend.dbc_id());
+
+        let mut valid_spends = self.valid_spends.write().await;
+
+        let replaced = valid_spends.insert(address, signed_spend.clone());
+        if replaced.is_none() {
+            self.valid_spends_cache_size.increase(size_of_new);
+        }
+
```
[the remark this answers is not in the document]
oetyng: Yes, that is right.

There is a call to `self.validate` before, which checks for double spends. 

Just needs to update the fn signature to take `&mut self`, to avoid having something added in between that check, and this insert.

Good catch!

safenode/src/storage/chunks.rs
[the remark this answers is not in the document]
grumbach: Shouldn't they stop storing new chunks instead?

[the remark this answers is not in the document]
oetyng: No, because the new chunks _are_ within their responsibility range. The chunks that fell outside that range as their range shrinks, are _not_ their responsibility.

The former must be served, or they will be punished. The latter are not expected from them.

safenode/src/storage/spends.rs
```diff
@@ -79,7 +72,26 @@ impl SpendStorage {
...
+        let size_of_new = std::mem::size_of_val(signed_spend);
+        let address = dbc_address(signed_spend.dbc_id());
+
+        let mut valid_spends = self.valid_spends.write().await;
+
+        let replaced = valid_spends.insert(address, signed_spend.clone());
+        if replaced.is_none() {
+            self.valid_spends_cache_size.increase(size_of_new);
+        }
```
grumbach: Race condition enabing double spend between
`self.validate(signed_spend).await?;`
and 
`let replaced = valid_spends.insert(address, signed_spend.clone());`

safenode/src/storage/chunks.rs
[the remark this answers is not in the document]
oetyng: You know, when the network grows, every node has a smaller and smaller range that they are responsible for.

The first node is responsible for the entire range.

So, when network grows, the responsibility range shrinks.

Any chunk that falls outside the range, can be dropped.

grumbach (approved):
LGTM

safenode/src/transfers/error.rs
jacderida: Yeah, I saw the code where the error occurred and it still wasn't clear to me why this would occur.

safenode/src/node/api.rs
jacderida: Yeah, my first major project back in 2010 or so was a CQRS system with event sourcing. In a way this actually seems to be mixing read and write (both `Cmd` and `Query` is part of `Request`).

In our CQRS system, commands were on the left side and events were on the right, to be handled as a result of a command, which wrote to the event store.

To me, a request is making a demand of something, not a mechanism to make others aware that something happened. Unless you are making a request to ask what happened. Is that what's happening here? For me it just seems strange to see an `Event` as a request. So is the name here being influenced by libp2p?

oetyng: Request is the top level.

A request can be either a Cmd, a Query or an Event. 

It is unfortunate that it is called Request, because it limits it linguistically, when the functionality is that of a _msg_. A message is what it is. We send a msg.

> In our CQRS system, commands were on the left side and events were on the right, to be handled as a result of a command, which wrote to the event store.

Yep, I built a big CQRS system, with event sourcing.

Events were always the results of cmds.
But we don't quite have such a system now. We do direct modification with cmds, and not the functional style where a cmd produces events that are then _applied_ to current state, to reach next state.

Regardless. The problem here is that the naming of Request is unfortunate. We shouldn't limit ourselves because of thjat though. That is the making of libp2p.

safenode/src/transfers/error.rs
oetyng: Okay, well. I'll try to clarify that then.

safenode/src/node/api.rs
oetyng: (The `Cmd::Dbc` variant was renamed to `SpendDbc`.)

safenode/src/protocol/messages/spend.rs
```diff
@@ -0,0 +1,40 @@
...
+        dbc_id: DbcId,
+        /// The priority of the spend.
+        priority: SpendPriority,
+    },
+    /// Query for a `Spend` of a Dbc with at the given address.
+    GetDbcSpend(DbcAddress),
+}
+
+impl SpendQuery {
+    /// Returns the dst address for the request.
```
maqi, line 33: request -> query 
to be consistent with the enum name ?

safenode/src/protocol/types/error.rs
```diff
@@ -23,74 +25,38 @@ pub type Result<T> = result::Result<T, Error>;
...
-    InvalidSpendParent(BTreeSet<Box<SignedSpend>>),
-    /// An error from the sn_dbc crate.
-    #[error("Dbc Error {0}")]
-    Dbc(String),
-    /// Unexpected responses.
-    #[error("Unexpected responses")]
-    UnexpectedResponses,
+    /// Access denied for user
+    #[error("Access denied for user: {0:?}")]
+    AccessDenied(User),
```
maqi, line 59: Actually something like : `AccessDenied(User, ErrorReason)`
shall cover both the current AccessDenied and NoSuchUser, with more explicity?

safenode/src/storage/registers.rs
```diff
@@ -292,6 +271,27 @@ impl RegisterStorage {
...
         Ok(stored_reg)
     }
+
+    #[cfg(test)]
+    async fn addrs(&self) -> Vec<RegisterAddress> {
+        self.register_store.addrs().await
+    }
+
+    /// Used for replication of data to new nodes.
+    // Currently only used by the tests, to be used by replication logic
```
maqi, line 281: /// as well?
and here actually shall be a TODO point ?

safenode/src/storage/spends.rs
```diff
@@ -79,7 +72,36 @@ impl SpendStorage {
     /// If a double spend attempt is detected, a `DoubleSpendAttempt` error
     /// will be returned including all the `SignedSpends`, for
     /// broadcasting to the other nodes.
-    pub(super) async fn try_add(&self, signed_spend: &SignedSpend) -> Result<()> {
+    /// NOTE: The &mut self signature is necessary to prevent race conditions
```
maqi, line 75: you mean the step of `self.validate(signed_spend).await?;` ?

also, seems missing \` \` around `&mut self` ?

safenode/src/storage/spends.rs
```diff
@@ -79,7 +72,36 @@ impl SpendStorage {
...
+            self.validate(signed_spend).await?;
+        }
+
+        Ok(())
+    }
+
+    /// Validates a spend without adding it to the storage.
+    /// If it however is detected as a double spend, that fact is recorded immediately,
+    /// and an error returned.
+    /// NOTE: The &mut self signature is necessary to prevent race conditions
```
maqi, line 102: \`&mut self\`

safenode/src/storage/spends.rs
```diff
@@ -129,25 +150,20 @@ impl SpendStorage {
...
-
         Ok(())
     }
 
     /// When data is replicated to a new peer,
     /// it may contain double spends, and thus we need to add that here,
     /// so that we in the future can serve this info to Clients.
-    pub(super) async fn try_add_double(
-        &self,
+    /// NOTE: The &mut self signature is necessary to prevent race conditions
```
maqi, line 163: \`&mut self\`

safenode/src/transfers/error.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use crate::protocol::types::{address::DbcAddress, fees};
+
+use sn_dbc::{Error as DbcError, SignedSpend, Token};
+
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
+use thiserror::Error;
+
+/// The type returned by the `sn_routing` message handling methods.
```
maqi, line 17: hmm...  there is no longer `sn_routing` ?

safenode/src/transfers/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        // 4. Validate the parents of the spend.
+        // This also ensures that all parent's dst tx's are the same as the src tx of this spend.
+        validate_parent_spends(signed_spend.as_ref(), source_tx.as_ref(), parent_spends)?;
+
+        // This spend is valid and goes into the queue.
+        self.spend_queue.push(*signed_spend, paid_fee.as_nano());
+
+        // If the rate limit has elapsed..
+        if self.spend_queue.elapsed() {
```
maqi, line 114: this means the `spend_queue` got checked only where there is another spend got pushed to the spend_queue?
which will result in:
1, some spend got processed far late
2, the last pushed spend never got processed

safenode/src/network/mod.rs
```diff
@@ -50,6 +50,12 @@ use xor_name::XorName;
 /// an item in the network.
 pub(crate) const CLOSE_GROUP_SIZE: usize = 8;
 
+/// Majority of a given group (i.e. > 1/2).
+#[inline]
+pub const fn majority(group_size: usize) -> usize {
```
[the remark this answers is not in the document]
oetyng: Only used for group size for now, not terribly important now, we can discuss required num. It has been said that mnajoarity might be enough.

safenode/src/transfers/error.rs
oetyng: Old stuff following along from old repo.. (very old stuff... o_0)

safenode/src/transfers/mod.rs
oetyng: Yeah, exactly. But that's totally fine now. And if we donn't see plenty of requests, the network is basically dead.

safenode/src/node/api.rs
```diff
@@ -237,44 +241,16 @@ impl Node {
...
-        // Here we check that the spend that is attempted, was created in a valid tx.
-        let src_tx_validity = source_tx.verify(&known_parent_blinded_amounts);
-        if src_tx_validity.is_err() {
-            return Err(super::Error::Protocol(
-                ProtocolError::InvalidSourceTxProvided {
-                    signed_src_tx_hash: signed_spend.src_tx_hash(),
-                    provided_src_tx_hash: source_tx.hash(),
-                },
-            ));
+            let parent_spend = self.get_spend(parent_address).await?;
```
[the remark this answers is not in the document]
oetyng: Yes.

They must _all_ be valid.

safenode/src/protocol/messages/spend.rs
oetyng: Yes, good point.

safenode/src/transfers/mod.rs
maqi: but this doesn't make sense to rely one spend process on the next one.
as there will always be quiet period.
especially the spend request only sent to the close group?
which will make some nodes being quieter than other ?

oetyng: Yep, I thought of this as well. But it works. This is first iteration. Simplicity and get stuff up and running now.
As they say: "Make it work. Make it work as it should. Make it fast."

safenode/src/protocol/types/error.rs
oetyng: Yeah maybe so?
This is old code. Added it to an issue: https://github.com/maidsafe/stableset_net/issues/105

safenode/src/storage/registers.rs
oetyng: Uhm, not sure, this is old stuff pulled in from old repo.
Will have to check what you mean.

oetyng: Ah, I got it :+1:, added to follow up PR.

safenode/src/storage/spends.rs
oetyng: Clarified comments there, and added the ... what ever they are called :)

safenode/src/node/api.rs
```diff
@@ -113,118 +112,123 @@ impl Node {
...
+                        {
+                            match self.send_to_closest(&Request::Event(event)).await {
+                                Ok(_) => {}
+                                Err(err) => {
+                                    warn!("Failed to send double spend event to closest peers: {err:?}");
+                                }
+                            }
+                        }
+
+                        Err(ProtocolError::Transfers(
```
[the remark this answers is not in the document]
oetyng: Not sure what you are asking. But this error is very important, because it's what will have us tell all other close nodes that there was a double spend attempt, and they will update accordingly (after verifying it).

We can't really do much if we can't send it to them. We could add a retry. And a retry of the retry. And a retry on that.... There's no end to it.

The simple approach: If we can't send it now, we will have to rely on them seeing the exact same thing when they also got this request. Because, that's actually what will most likely happen in 99.9999...% of cases.
This notification to them is like a backup thing, to be _really_ sure.

safenode/src/node/api.rs
```diff
@@ -113,118 +112,123 @@ impl Node {
...
-                Response::Cmd(CmdResponse::Spend(Err(error))),
-                response_channel,
-            )
-            .await;
-            return Ok(());
+    async fn handle_query(&mut self, query: Query) -> QueryResponse {
+        match query {
+            Query::Register(query) => self.registers.read(&query, User::Anyone).await,
+            Query::GetChunk(address) => {
+                let resp = self.chunks.get(&address).await;
```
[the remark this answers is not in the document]
oetyng: Ah yes, I agree.

But I have plans for refactoring register a bit in the coming.

Iterations.

PR #96, opened 2023-04-15
chore: organize repo files

oetyng:
[chore: distinguish transfer modules](https://github.com/maidsafe/stableset_net/commit/0cddb77d1ee21bc55c022acacf7b2aef084c2dcc)

[chore: remove unnecessary indirection](https://github.com/maidsafe/stableset_net/commit/8478c02522c8e95021fc6d8f2dd46079edc41a6f) 

- The nesting doesn't serve any purpose, and is not very accurately
named.
- The contents are all directly parts of the protocol.


safenode/src/network_transfers/error.rs
```diff
@@ -0,0 +1,95 @@
...
+        existing: Box<SignedSpend>,
+    },
+    /// We were notified about a double spend attempt, but they were for different dbcs.
+    #[error("We were notified about a double spend attempt, but they were for different dbcs: {0:?}. Existing: {1:?}")]
+    NotADoubleSpendAttempt(Box<SignedSpend>, Box<SignedSpend>),
+    /// One or more parent spends of a requested spend had a different dst tx hash than the signed spend src tx hash.
+    #[error(
+        "The signed spend src tx ({signed_src_tx_hash:?}) did not match the provided source tx's hash: {provided_src_tx_hash:?}"
+    )]
+    SignedSrcTxHashDoesNotMatchProvidedSrcTxHash {
```
b00f, line 54: I suggest to define a `Custom` error type that message can be formatted.

oetyng: Good that you pointed out that one. I don't like the name of that error.. But it will have to be improved.

What custom error is it you propose? The error we're looking at is a custom error.

b00f: Custom or Genera error, that can have formatted string.  like
```rust
General(String)
```

That can be used like:
```rust
return Err(Error::General(format!("The signed spend src tx ({signed_src_tx_hash:?}) did not match the provided source tx's hash: {provided_src_tx_hash:?}"))
```

oetyng: Alright, yeah ok I understand.

We've been trying to avoid that actually, and design specific errors for each and every situation requiring an error.

oetyng: That error name I got, with not too much time spent on it, is horrendous though :)

oetyng: I'm testing with a rename to `TxTrailMismatch`, and let the msg for it explain closer.

grumbach (approved):
Good one getting rid of `types`! 🚢

safenode/src/network_transfers/error.rs
oetyng: Fixed it in follow up PR.

PR #98, opened 2023-04-16
feat: extend kadclient to up-/download files

### LINKED ISSUE #100: Have client upload files instead of chunks.
Use the files api, and make sure it is encapsulated.

oetyng:
Resolves #100.

It can now upload and download entire files, instead of small chunks.

Additionally, the files apis are encapsulated in their own struct, as
to not bloat the client api.

### DIFF
--- a/safenode/src/bin/kadclient.rs
+++ b/safenode/src/bin/kadclient.rs
@@ -7,9 +7,9 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use safenode::{
-    client::{Client, ClientEvent, Error as ClientError},
+    client::{Client, ClientEvent, Error as ClientError, Files},
     log::init_node_logging,
-    protocol::{address::ChunkAddress, chunk::Chunk},
+    protocol::address::ChunkAddress,
 };
 
 use bls::SecretKey;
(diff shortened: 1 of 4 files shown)


safenode/src/bin/kadclient.rs
```diff
@@ -101,11 +98,11 @@ async fn main() -> Result<()> {
         }
 
         for xorname in chunks_to_fetch.iter() {
-            println!("Fetching chunk {xorname:?}");
-            match client.get_chunk(ChunkAddress::new(*xorname)).await {
-                Ok(chunk) => info!("Successfully got chunk {}!", chunk.name()),
+            println!("Downloading file {xorname:?}");
```
maqi, line 101: this is a chunk downloaded, not the whole file

safenode/src/bin/kadclient.rs
```diff
@@ -101,11 +98,11 @@ async fn main() -> Result<()> {
...
+            match file_api.read_bytes(ChunkAddress::new(*xorname)).await {
+                Ok(bytes) => info!("Successfully got file {xorname} of {} bytes!", bytes.len()),
```
maqi, line 103: same comment as above, this is just a chunk part of the file fetched

safenode/src/bin/kadclient.rs
oetyng: No, @maqi , check the innards of `read_bytes`. It gives a whole file.

It takes a chunk name, checks if it is a head chunk asked for (the one holding the data map, or the final one if data map chunkling resulted in more than one chunk).

grumbach (approved):
Nice one, files are a much better API than Chunks!
LGTM

PR #99, opened 2023-04-16
feat(client): integrate wallet

### LINKED ISSUE #89: Store wallet to drive.

oetyng:
Resolves #89.

[feat(wallet): store and load from disk](https://github.com/maidsafe/stableset_net/commit/7ba6c521c43807592a765ff7c6387567b8acf3ff) 
As a temporary solution, the serialized wallet can be stored to disk.
Next the wallet ops will be stored to disk as a Register.
 
[chore(wallet): use load_from in tests](https://github.com/maidsafe/stableset_net/commit/e159431eb88d4d55d762643e36b2b2d208744379) 
This auto-generates a new mainkey.
 
[feat(client): instantiate wallet in client](https://github.com/maidsafe/stableset_net/commit/a57ea01925f7d69fac4cba43b4b28c49b649f354)

### DIFF
--- a/safenode/src/client/mod.rs
+++ b/safenode/src/client/mod.rs
@@ -12,24 +12,24 @@ mod error;
 mod event;
 mod file_apis;
 mod register;
+mod wallet;
 
 pub use self::{
     error::Error,
     event::{ClientEvent, ClientEventsReceiver},
     file_apis::Files,
     register::{Register, RegisterOffline},
+    wallet::WalletClient,
 };
 
 use self::event::ClientEventsChannel;
 
 use crate::network::Network;
 
-use bls::SecretKey;
-
 /// Client API implementation to store and get data.
 #[derive(Clone)]
 pub struct Client {
     network: Network,
     events_channel: ClientEventsChannel,
-    signer: SecretKey,
+    signer: bls::SecretKey,
 }
--- a/safenode/src/client/wallet.rs
+++ b/safenode/src/client/wallet.rs
@@ -1,0 +1,50 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use sn_dbc::{Dbc, DbcIdSource, DerivedKey, PublicAddress, Token};
+
+use crate::protocol::{
+    offline_transfers::{create_transfer, Outputs as TransferDetails},
+    wallet::{Result, SendClient, SendWallet},
+};
+
+use super::Client;
+
+/// A wallet client can be used to send and
+/// receive tokens to/from other wallets.
+pub struct WalletClient<W: SendWallet> {
+    client: Client,
+    wallet: W,
+}
+
+impl<W: SendWallet> WalletClient<W> {
+    /// Create a new wallet client.
+    pub fn new(client: Client, wallet: W) -> Self {
+        Self { client, wallet }
+    }
+
+    /// Send tokens to another wallet.
+    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<()> {
+        let _dbcs = self.wallet.send(vec![(amount, to)], &self.client).await?;
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl SendClient for Client {
+    async fn send(
+        &self,
+        dbcs: Vec<(Dbc, DerivedKey)>,
+        to: Vec<(Token, DbcIdSource)>,
+        change_to: PublicAddress,
+    ) -> Result<TransferDetails> {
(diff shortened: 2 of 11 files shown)


safenode/src/client/wallet.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+impl<W: SendWallet> WalletClient<W> {
+    /// Create a new wallet client.
+    pub fn new(client: Client, wallet: W) -> Self {
+        Self { client, wallet }
+    }
+
+    /// Send tokens to another wallet.
+    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<()> {
```
b00f, line 32: After sending, usually it is not bad to print transaction ID or some information to be track by user. Here we don't return anything. Is it OK?

oetyng: Yeah, maybe so.

My vision here is rather that a UI uses the api to get the data to show in the UI, rather than relying on log lines.

b00f: > Tx ids is way too low level for a user.

Probably some users might be interested to check the id. In case of any issue on the production, This information probably are so valuable that time.

oetyng: Yeah, you're right, some log lines can be useful.

safenode/src/client/mod.rs
```diff
@@ -12,24 +12,24 @@ mod error;
...
 
-use bls::SecretKey;
-
 /// Client API implementation to store and get data.
 #[derive(Clone)]
 pub struct Client {
     network: Network,
     events_channel: ClientEventsChannel,
-    signer: SecretKey,
+    signer: bls::SecretKey,
```
b00f, line 34: Not a good name for SecretKet. Signer usually refer to a wrapper over secret_key.

safenode/src/protocol/wallet/keys.rs
```diff
@@ -0,0 +1,84 @@
...
+use tokio::fs;
+
+/// Filename for storing the node's reward (BLS hex-encoded) main key.
+const MAIN_KEY_FILENAME: &str = "main_key";
+/// Filename for storing the node's reward (BLS hex-encoded) public address.
+const PUBLIC_ADDRESS_FILENAME: &str = "public_address";
+
+/// Writes the public address and main key (hex-encoded) to different locations at disk.
+pub(super) async fn store_new_keypair(root_dir: &Path, main_key: &MainKey) -> Result<()> {
+    let secret_key_path = root_dir.join(MAIN_KEY_FILENAME);
```
b00f, line 24: Not bad if we can have a BOM for the file, like: <2 byte version> + <Type of Key>
Type of key for BLS can set to 12381

safenode/src/protocol/wallet/mod.rs
```diff
@@ -75,42 +77,51 @@ pub trait SendClient {
...
-    fn balance(&self) -> Token;
+pub trait DepositWallet: Wallet {
     /// Used to generate a new dbc id for receiving tokens.
     fn new_dbc_address(&self) -> DbcIdSource;
     /// Will only deposit those that are actually accessible by this wallet.
     fn deposit(&mut self, dbcs: Vec<Dbc>);
 }
+
+#[derive(serde::Serialize, serde::Deserialize)]
+pub(super) struct KeyLessWallet {
```
b00f, line 113: Question: KeylessWallet is a DepositWallet?

oetyng: It's just a keyless wallet. It can be used for either deposits or sending.

Desposits, where you can actually read the balance, requires a key. And sending requires a key.

safenode/src/protocol/wallet/keys.rs
oetyng: Yeah, secret key handling is a part of the master document in discourse. All that we have done in that area so far has been rather temporary.

If you have an idea for a more long term solution, please do a write up on discourse and add the link to the pinned master document wiki (it has a section about secret key handling).

safenode/src/protocol/wallet/mod.rs
b00f: What is the purpose behind KeyLessWallet?

safenode/src/client/mod.rs
b00f, line 34: `Signer` usually refers to a wrapper around the secret key that can sign and more.
SecretKey is not signer, you might do encryption, or ...
Probably `sec_key` is a better name here.

b00f (approved):

safenode/src/client/mod.rs
oetyng: Absoutely, totally agree. This one was an existing name, but I have actually already a change for this in work I'm doing now.

safenode/src/protocol/wallet/mod.rs
oetyng: Right now just to have a separate structure that can be serialized to disk without the key.

Later we will be serializing ops that are appended to a log (the register), and then hydrtate the state by replaying the ops.

But in either case, the secret key is stored in a separate way.

grumbach (approved):
LGTM

safenode/src/protocol/wallet/keys.rs
b00f: The topic about [key-store](https://maidsafe.discourse.team/t/code-design-topic-keystore-format/758) format in discourse.

PR #101, opened 2023-04-17
fix: adding nat relay behaviour

b00f:
In this PR, a relay Behavior added to the node's Behaviors. It's toggle and can be disabled.

Fixing: #49 

TODO:
How to Enable/Disable relay Behavior?


b00f:
> I'm not sure this makes sense to add on itself, without the `relay::client` stuff.

Yes, right. Probably we need kind of configuration to toggle on/off relay client/server. Or other way?

PR #103, opened 2023-04-17
CI: undertake daily checks

maqi:
carry out a daily run of network unit tests and e2d tests.


.github/workflows/daily_run.yml
```diff
@@ -0,0 +1,96 @@
+name: DailyRun
```
jacderida, line 1: Could we call the workflow "Full Network Tests"?

jacderida (changes requested):

PR #111, opened 2023-04-18
feat: add online transfer logic

### LINKED ISSUE #113: Add online transfer logic with fees

oetyng:
Resolves #113.

***

This includes the paying of fees in a transfer, which the `offline` transfer logic does not yet include.

**NB 1**: Some duplication of types in the `online` transfers mod. This will be improved upon in the followup PRs.

**NB 2**: Also, the inclusion of fees in a `semi-offline` transfer is still to be worked on (the scenario when a shopkeeper receives a begun transaction, and connects to network to get the required fees, and then extends the transaction with the fees, as to then upload it all.)

**NB 3**: The transaction build logic in the two files of `offline`/`online` can have further code re-use, but it's a bit tricky and simpler to separate these for now.

### DIFF
--- a/safenode/src/network_transfers/mod.rs
+++ b/safenode/src/network_transfers/mod.rs
@@ -11,6 +11,7 @@ mod error;
 pub(crate) use self::error::{Error, Result};
 
 use crate::{
+    node::NodeId,
     protocol::{
         address::DbcAddress,
         fees::{FeeCiphers, RequiredFee, SpendPriority, SpendQ},
@@ -21,12 +22,11 @@ use crate::{
 use sn_dbc::{DbcId, DbcTransaction, MainKey, SignedSpend, Token};
 
 use std::collections::{BTreeMap, BTreeSet};
-use xor_name::XorName;
 
 const STARTING_FEE: u64 = 4000; // 0.000004 SNT
 
 pub(super) struct Transfers {
-    our_name: XorName,
+    node_id: NodeId,
     node_reward_key: MainKey,
     spend_queue: SpendQ<SignedSpend>,
     storage: SpendStorage,
@@ -34,9 +34,9 @@ pub(super) struct Transfers {
 
 impl Transfers {
     /// Create a new instance of `Transfers`.
-    pub(crate) fn new(our_name: XorName, node_reward_key: MainKey) -> Self {
+    pub(crate) fn new(node_id: NodeId, node_reward_key: MainKey) -> Self {
         Self {
-            our_name,
+            node_id,
             node_reward_key,
             spend_queue: SpendQ::with_fee(STARTING_FEE),
             storage: SpendStorage::new(),
@@ -49,12 +49,19 @@ impl Transfers {
     }
 
     /// Get the required fee for the specified spend priority.
-    pub(crate) fn get_required_fee(&self, dbc_id: DbcId, priority: SpendPriority) -> RequiredFee {
+    pub(crate) fn get_required_fee(
+        &self,
+        dbc_id: DbcId,
+        priority: SpendPriority,
+    ) -> (NodeId, RequiredFee) {
         let amount = self.current_fee(priority);
 
         debug!("Returned amount for priority {priority:?}: {amount}");
 
-        RequiredFee::new(Token::from_nano(amount), dbc_id, &self.node_reward_key)
+        let required_fee =
+            RequiredFee::new(Token::from_nano(amount), dbc_id, &self.node_reward_key);
+
+        (self.node_id, required_fee)
     }
 
     /// Get the current fee for the specified spend priority.
@@ -81,7 +88,7 @@ impl Transfers {
         &mut self,
         signed_spend: Box<SignedSpend>,
         source_tx: Box<DbcTransaction>,
-        fee_ciphers: BTreeMap<XorName, FeeCiphers>,
+        fee_ciphers: BTreeMap<NodeId, FeeCiphers>,
         parent_spends: BTreeSet<SignedSpend>,
     ) -> Result<()> {
         // 1. Validate the tx hash.
@@ -126,9 +133,9 @@ impl Transfers {
     fn validate_fee(
         &self,
         tx: &DbcTransaction,
-        fee_ciphers: BTreeMap<XorName, FeeCiphers>,
+        fee_ciphers: BTreeMap<NodeId, FeeCiphers>,
     ) -> Result<Token> {
-        let fee_paid = decipher_fee(&self.node_reward_key, tx, self.our_name, fee_ciphers)?;
+        let fee_paid = decipher_fee(&self.node_reward_key, tx, self.node_id, fee_ciphers)?;
 
         let spend_q_snapshot = self.spend_queue.snapshot();
         let spend_q_stats = spend_q_snapshot.stats();
@@ -188,10 +195,10 @@ fn validate_parent_spends(
 fn decipher_fee(
     node_reward_key: &MainKey,
     tx: &DbcTransaction,
-    our_name: XorName,
-    fee_ciphers: BTreeMap<XorName, FeeCiphers>,
+    node_id: NodeId,
+    fee_ciphers: BTreeMap<NodeId, FeeCiphers>,
 ) -> Result<Token> {
-    let fee_ciphers = fee_ciphers.get(&our_name).ok_or(Error::MissingFee)?;
+    let fee_ciphers = fee_ciphers.get(&node_id).ok_or(Error::MissingFee)?;
     let (dbc_id, revealed_amount) = fee_ciphers.decrypt(node_reward_key)?;
     let output_proof = match tx.outputs.iter().find(|proof| proof.dbc_id() == &dbc_id) {
         Some(proof) => proof,
--- a/safenode/src/protocol/mod.rs
+++ b/safenode/src/protocol/mod.rs
@@ -23,6 +23,8 @@ pub mod error;
 pub mod fees;
 /// Token transfers.
 pub mod offline_transfers;
+/// Token transfers.
+pub mod online_transfers;
 /// Register type.
 pub mod register;
 /// A wallet for network tokens.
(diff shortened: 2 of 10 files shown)


safenode/src/network_transfers/mod.rs
```diff
@@ -49,12 +49,19 @@ impl Transfers {
     }
 
     /// Get the required fee for the specified spend priority.
-    pub(crate) fn get_required_fee(&self, dbc_id: DbcId, priority: SpendPriority) -> RequiredFee {
+    pub(crate) fn get_required_fee(
+        &self,
+        dbc_id: DbcId,
+        priority: SpendPriority,
+    ) -> (NodeId, RequiredFee) {
```
maqi, line 56: this then means the `RequiredFee` is charged by this node, right?

but our `node_reward_key` is already included within the `RequiredFee`, which can be accessed

safenode/src/protocol/mod.rs
```diff
@@ -23,6 +23,8 @@ pub mod error;
 pub mod fees;
 /// Token transfers.
 pub mod offline_transfers;
+/// Token transfers.
```
maqi, line 26: better to expand the comments of `offline_transfers` and `online_transfers` to explain little bit what's the different is?

safenode/src/protocol/online_transfers/error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use thiserror::Error;
+
+pub(crate) type Result<T> = std::result::Result<T, Error>;
+
+/// Error type returned by the API
+#[derive(Debug, Error)]
+#[allow(clippy::large_enum_variant)]
+#[non_exhaustive]
+pub enum Error {
```
maqi, line 19: better rename this to `OnlineTransferError` and the one for offline to be `OfflineTransferError`
bearing the same enum name and having most entries the same naming for both may cause trouble/confusion for tracing.

safenode/src/protocol/online_transfers/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+
+use super::{
+    fees::{RequiredFee, SpendPriority},
+    messages::{Query, QueryResponse, Request, Response, SpendQuery},
+};
+
+/// The input details necessary to
+/// carry out a transfer of tokens.
+#[derive(Debug)]
+pub struct Inputs {
```
maqi, line 33: same comment as to the above `Error` , also to the following `Outputs` , `CreatedDbc`
as the main parts remains the same, better at least give them a different naming for ease of tracing.

trying to inherit will be better if possible

safenode/src/protocol/online_transfers/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+        // Input amount increases with the amount of the dbc.
+        total_input_amount = total_input_amount.checked_add(dbc_balance)
+            .ok_or_else(|| {
+                Error::DbcReissueFailed(
+                    "Overflow occurred while increasing total input amount while trying to cover the output DBCs."
+                    .to_string(),
+            )
+            })?;
+
+        // ---------------- fee part start ----------------
```
maqi, line 221: this part is about output, not `fee` ?

safenode/src/protocol/online_transfers/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+/// The tokens of the input dbcs will be transfered to the
+/// new dbcs (and a change dbc if any), which are returned from this function.
+/// This does not register the transaction in the network.
+/// To do that, the `signed_spends` of each new dbc, has to be uploaded
+/// to the network. When those same signed spends can be retrieved from
+/// enough peers in the network, the transaction will be completed.
+#[allow(clippy::result_large_err)]
+fn crate_transfer_with(selected_inputs: Inputs) -> Result<Outputs> {
```
maqi, line 283: crate -> create ?

safenode/src/protocol/online_transfers/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+    let responses = client
+        .send_to_closest(request)
+        .await
+        .map_err(|e| Error::CouldNotGetFees(e.to_string()))?;
+
+    // We just want to receive at least a majority of results, we don't care about any errors
+    // so we log them, but return whatever results we get. If not enough for upper layer, it will error there.
+    let results: BTreeMap<_, _> = responses
+        .into_iter()
+        .flat_map(|res| {
```
maqi, line 362: this actually can be covered by the following `filter_map` ?

safenode/src/protocol/online_transfers/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+    let request = Request::Query(Query::Spend(SpendQuery::GetFees {
+        dbc_id,
+        priority: SpendPriority::Normal,
+    }));
+    let responses = client
+        .send_to_closest(request)
+        .await
+        .map_err(|e| Error::CouldNotGetFees(e.to_string()))?;
+
+    // We just want to receive at least a majority of results, we don't care about any errors
```
maqi, line 358: shall an error to be raised in case of just received less than majority valid fees?

safenode/src/network_transfers/mod.rs
oetyng: > this then means the RequiredFee is charged by this node, right?
Yes.

> but our node_reward_key is already included within the RequiredFee, which can be accessed
We need the node id when the fee ciphers are sent in to the nodes. This is how each node can find their set of fee ciphers within a spend request.

safenode/src/protocol/mod.rs
oetyng: These are merged in the follow up PR.

> NB 1: Some duplication of types in the online transfers mod. This will be improved upon in the followup PRs.

oetyng: These are merged into one in the follow up PR.

> NB 1: Some duplication of types in the online transfers mod. This will be improved upon in the followup PRs.

safenode/src/protocol/online_transfers/error.rs
oetyng: These are merged into one in the follow up PR.

> NB 1: Some duplication of types in the online transfers mod. This will be improved upon in the followup PRs.

safenode/src/protocol/online_transfers/mod.rs
oetyng: These are merged into one in the follow up PR.

> NB 1: Some duplication of types in the online transfers mod. This will be improved upon in the followup PRs.

safenode/src/protocol/online_transfers/mod.rs
oetyng: > total_output_amount.checked_add(fee_per_input)

It is about fees.

safenode/src/protocol/online_transfers/mod.rs
oetyng: That is handled in thae calling code above. And we don't need to raise an error for that. We just skip that input dbcs and try with the next.

If we in the end didn't have enough tokens in the successful dbcs, _then_  we error.

safenode/src/protocol/online_transfers/mod.rs
oetyng: Then we can't log for the specific type of err. See the logs are different:

flat_map:
`warn!("Error when querying for fees: {error}");`

filter_map:
`warn!("Unexpected response to fee query: {other:?}");`

safenode/src/protocol/online_transfers/mod.rs
maqi: first, this `fee part` is duplicated with the above `fee part`
second, `total_output_amount.checked_add(fee_per_input)` is more about `output compose` instead of `fee calculation`
hence, maybe rename this part to be `output composing part` ?

safenode/src/protocol/online_transfers/mod.rs
maqi: yeah, I understand the `caller` can carry out the mojority check.
However, if the `majority` is always required, then it will be more safe/convenient to make the check included in this `get` func, so that no extra work need to be undertaken outside it?

safenode/src/protocol/online_transfers/mod.rs
oetyng: It is not duplicated. 

It's separated because this part was feature gated before (and might be still, we haven't decided on that yet).

This code part was only run when fees were used. Thus it is a `fee part`.

The other ones are about calculating total output _without_ fees.
But that is not important to distinguish, since that is never feature gated.

The fees though may be, therefore these parts which are specifically for fees, are marked here as `fee part start`and `fee part end`.

safenode/src/protocol/online_transfers/mod.rs
oetyng: The first flat map removes connection/network type of errors.

The first filter map removes non-query responses (and logs that as an unexpected response, which it is).

The second filter map identifies an error at node within handling of get fees, and logs that. It also identifies an unexpected query response type, and logs that.

safenode/src/protocol/online_transfers/mod.rs
oetyng: That is actually not more convenient. And there is no extra work.

We are using the `required_responses` variable for additional checks further down after ln 148, see line 165 to ln 170.

```
let max_invalid_fees = num_responses - required_responses;
            if invalid_fees.len() > max_invalid_fees {
                let valid_responses = num_responses - invalid_fees.len();
                warn!("Not enough valid fees received from nodes to spend the input. Found: {valid_responses}, needed: {required_responses}", );
                continue;
            }
```

oetyng: Get fees gets what it can get. Caller decides if it is enough or not.

It makes more sense to keep that type of validation (whether enough or not) on the one and same level, instead of sprinkling it out over several different functions.

PR #125, opened 2023-04-19
feat(node): dial peers on startup

b-zee:
We dial optional peers on startup that will get added to our routing
table et al. This will cause our node to get booted by specifying a
bootstrap node address.

### DIFF
--- a/safenode/src/bin/kadnode.rs
+++ b/safenode/src/bin/kadnode.rs
@@ -8,7 +8,6 @@
 
 use safenode::{
     log::init_node_logging,
-    network::Network,
     node::{Node, NodeEvent},
 };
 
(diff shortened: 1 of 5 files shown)


safenode/src/bin/kadnode.rs
```diff
@@ -28,12 +27,25 @@ async fn main() -> Result<()> {
...
+                break;
+            }
+            Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
+                tracing::warn!("Skipped {n} node events!");
+                continue;
+            }
+        };
         match event {
             NodeEvent::ConnectedToNetwork => {
                 info!("Connected to the Network");
```
grumbach, line 51: Looks like we only ever break from this infinite loop when `Err(tokio::sync::broadcast::error::RecvError::Closed)`, only to block forever in a sleep loop below.
On the other hand, when we have `NodeEvent::ConnectedToNetwork` we don't break out.
I have trouble understanding what this code is doing. Am I missing something?

oetyng (approved):

oetyng (approved):
LGTM!

oetyng (changes requested):
Was a bit trigger happy there.
Let's wait for Anselmes question to be cleared!

safenode/src/bin/kadnode.rs
b-zee: Previously, we were just sleeping forever and listening for `ConnectedToNetwork` once. I just wanted to make it explicit and log when we encountered an issue. But yes, we should aim for a proper handling and shutdown. I just didn't want to change any functionality at the moment, because I assumed the sleep was there for a reason to keep whatever is running, running.

oetyng (approved):

PR #126, opened 2023-04-19
Log when a peer disconnects

grumbach:
This is the starting point of offline node detection


oetyng (approved):

maqi (approved):
Just a note that the `ConnectionClosed` could be due to the `idle_time_out` got reached.
This means the connection is idle may just because of `no query/data transmitted on it`. 
i.e. it not necessarily mean the peer is dead.

PR #129, opened 2023-04-19
feat: impl simple cli for wallet ops, sending

### LINKED ISSUE #128: Add cli interface for Wallet send op.
Add a simple `send` api for tokens.

oetyng:
Resolves #128.

[feat: impl simple cli for wallet ops, sending](https://github.com/maidsafe/safe_network/commit/258624fdca16e4bafef77ede54b26ca3a4fbc925)

- Adds send operation.
- NB: Does not yet store the created dbcs, for giving them to the
recipients out of band.

### DIFF
--- a/safenode/src/bin/kadclient.rs
+++ b/safenode/src/bin/kadclient.rs
@@ -148,35 +148,32 @@ async fn wallet(opt: &Opt, client: &Client) -> Result<()> {
     if let Some(hex) = &opt.send_to {
         let address = parse_public_address(hex)?;
 
-        if let Some(amount) = opt
-            .send_amount
-            .clone()
-            .map(|amount| parse_tokens_amount(&amount))
-        {
-            if amount.as_nano() > 0 {
-                let mut wallet_client = WalletClient::new(client.clone(), wallet);
-                match wallet_client.send(amount, address).await {
-                    Ok(_new_dbcs) => {
-                        info!("Sent {amount:?} to {address:?}");
-                        println!("Sent {amount:?} to {address:?}");
-                        let wallet = wallet_client.into_wallet();
-                        let new_balance = wallet.balance();
+        let amount_str = match &opt.send_amount {
+            Some(amount_str) => amount_str,
+            None => panic!("An amount is expected when an address has been specified!"),
+        };
+        use std::str::FromStr;
+        let amount = Token::from_str(amount_str)?;
 
-                        if let Err(err) = wallet.store().await {
-                            warn!("Failed to store wallet: {err:?}");
-                            println!("Failed to store wallet: {err:?}");
-                        } else {
-                            info!("Successfully stored wallet with new balance {new_balance:?}.");
-                            println!(
-                                "Successfully stored wallet with new balance {new_balance:?}."
-                            );
-                        }
-                    }
-                    Err(err) => {
-                        warn!("Failed to send {amount:?} to {address:?} due to {err:?}.");
-                        println!("Failed to send {amount:?} to {address:?} due to {err:?}.");
-                    }
+        let mut wallet_client = WalletClient::new(client.clone(), wallet);
+        match wallet_client.send(amount, address).await {
+            Ok(_new_dbcs) => {
+                info!("Sent {amount:?} to {address:?}");
+                println!("Sent {amount:?} to {address:?}");
+                let wallet = wallet_client.into_wallet();
+                let new_balance = wallet.balance();
+
+                if let Err(err) = wallet.store().await {
+                    warn!("Failed to store wallet: {err:?}");
+                    println!("Failed to store wallet: {err:?}");
+                } else {
+                    info!("Successfully stored wallet with new balance {new_balance:?}.");
+                    println!("Successfully stored wallet with new balance {new_balance:?}.");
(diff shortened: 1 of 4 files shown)


safenode/src/client/wallet.rs
```diff
@@ -30,9 +32,15 @@ impl<W: SendWallet> WalletClient<W> {
     }
 
     /// Send tokens to another wallet.
-    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<()> {
-        let _dbcs = self.wallet.send(vec![(amount, to)], &self.client).await?;
-        Ok(())
+    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<Vec<CreatedDbc>> {
+        let dbcs = self.wallet.send(vec![(amount, to)], &self.client).await?;
+        // self.wallet.store().await;
```
b-zee: Any future ideas for this commented line?

safenode/src/bin/kadclient.rs
```diff
@@ -138,6 +145,42 @@ async fn wallet(opt: &Opt) -> Result<()> {
         }
     }
 
+    if let Some(hex) = &opt.send_to {
+        let address = parse_public_address(hex)?;
+
+        if let Some(amount) = opt
```
b-zee: Not sure if you're planning on erroring if the `send_amount` isn't given? I'm fine with this for now, but the user won't get to know if he specifies `send_to` but forgets the amount.

safenode/src/bin/kadclient.rs
```diff
@@ -138,6 +145,42 @@ async fn wallet(opt: &Opt) -> Result<()> {
...
+            .send_amount
+            .clone()
+            .map(|amount| parse_tokens_amount(&amount))
+        {
+            if amount.as_nano() > 0 {
```
b-zee: Same here perpaps, show user an error or message if they specify `0`. I am also in favour of handling such cases with an early return to make sure we keep the code indentation low.

safenode/src/bin/kadclient.rs
```diff
@@ -149,6 +192,33 @@ async fn get_client_dir() -> Result<PathBuf> {
...
+                println!("Invalid amount to send: '{amount_str}': {msg}.");
+            }
+            other_err => {
+                warn!("Invalid amount to send: '{amount_str}': {other_err:?}.");
+                println!("Invalid amount to send: '{amount_str}': {other_err:?}.");
+            }
+        },
+    }
+
+    Token::from_nano(0)
```
b-zee: Isn't it perhaps better to just straight error instead of defaulting to 0? I know the reason gets printed, but I'm not sure it's good to randomly default like this.

safenode/src/protocol/wallet/keys.rs
```diff
@@ -19,7 +19,15 @@ const MAIN_KEY_FILENAME: &str = "main_key";
 /// Filename for storing the node's reward (BLS hex-encoded) public address.
 const PUBLIC_ADDRESS_FILENAME: &str = "public_address";
 
+/// Parse a public address from a hex-encoded string.
+#[allow(clippy::result_large_err)]
```
b-zee, line 23: Is there a reason the `Error` grew so big?

safenode/src/client/wallet.rs
oetyng: aah... yeah... need to fix, let me see if I planned it for now or follow up..

safenode/src/bin/kadclient.rs
oetyng: Good point, I can update that now already.

safenode/src/bin/kadclient.rs
oetyng: Ah all that is fixed in the next PR (with less indentation, I'm a stickler for that as well ;) )

safenode/src/protocol/wallet/keys.rs
oetyng: It's something (from a while back) but didn't delve into it. Deferred.

safenode/src/bin/kadclient.rs
oetyng: Probably good to check the next one to see what stuff from here, is fixed there.

safenode/src/client/wallet.rs
oetyng: Ah, it was a remainder. I had solved that already :)

oetyng: (removing the line)

safenode/src/bin/kadclient.rs
oetyng: Will get some conflicts now with next PR but ye.

safenode/src/bin/kadclient.rs
oetyng: fixed

safenode/src/bin/kadclient.rs
oetyng: fixed

safenode/src/client/wallet.rs
oetyng: fixed

b-zee (approved):

PR #132, opened 2023-04-19
refactor(cli): use subcmds

### LINKED ISSUE #133: Use subcmds for cli

oetyng:
Resolves #133.

Implements subcmds for cli, but also includes some changes to the use of files and registers.


safenode/src/bin/cli/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Parser, Debug)]
+#[clap(name = "safeclient cli")]
+pub(super) enum Opt {
+    #[clap(name = "cfg", subcommand)]
+    Cfg(cfg::CfgCmds),
+    #[clap(name = "wallet", subcommand)]
+    /// Manage wallets on the SAFE Network
+    Wallet(wallet::WalletCmds),
+    #[clap(name = "files", subcommand)]
+    /// Manage files on the SAFE Network
```
b-zee, line 36: This should be 'register' I assume.

oetyng: Yep! Had it, but good catch.

safenode/src/bin/cli/cfg.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use clap::Parser;
+use eyre::Result;
+use std::path::PathBuf;
+
+#[derive(Parser, Debug)]
+pub enum CfgCmds {
+    /// Set the location of the log files.
+    #[clap(name = "log-dir")]
+    Logs {
```
jacderida: What does this command actually do? It doesn't seem like it's setting a permanent location for a logging directory? Also, do we actually need this for the client? I get setting a log dir for the node, but is it necessary for the client? If it was, it may be better to have `--log-dir` as a global option.

safenode/src/bin/cli/files.rs
```diff
@@ -0,0 +1,112 @@
...
+use tracing::info;
+use walkdir::WalkDir;
+use xor_name::XorName;
+
+#[derive(Parser, Debug)]
+pub enum FilesCmds {
+    Upload {
+        /// The location of the files to upload.
+        #[clap(name = "files-path")]
+        files_path: PathBuf,
```
jacderida: Is this the location of a directory? You can add this to clarify:
`#[clap(name = "files-path", value_name="DIRECTORY")]`

I would also suggest renaming the argument to `path`. The `files` part isn't strictly necessary, since it's already localised to a `files upload` command.

safenode/src/bin/cli/files.rs
```diff
@@ -0,0 +1,112 @@
...
+    },
+    Download {
+        /// The location of the file names stored
+        /// when uploading files.
+        #[clap(name = "file-names-path")]
```
jacderida: I would clarify the name of this argument and how it makes the command behave. Are you going to be supplying many different file paths? Or should it be the location of one directory, and all the files in that directory would be downloaded? It would be worth elaborating the doc comments there to describe the behaviour. If appropriate I would also consider renaming it to just `path`, since again, it is localised to a `files download` command.

safenode/src/bin/cli/register.rs
```diff
@@ -0,0 +1,118 @@
...
+pub enum RegisterCmds {
+    /// Create a new register with the given pet name.
+    Create {
+        #[clap(name = "name")]
+        name: String,
+    },
+    Edit {
+        #[clap(name = "name")]
+        name: String,
+        ///
```
jacderida: We should get into the habit early of properly documenting every command to make the CLI look professional.

safenode/src/bin/cli/wallet.rs
```diff
@@ -0,0 +1,148 @@
...
+use eyre::Result;
+use std::path::PathBuf;
+use tokio::fs;
+use tracing::{info, warn};
+use walkdir::WalkDir;
+
+#[derive(Parser, Debug)]
+pub enum WalletCmds {
+    Deposit {
+        /// Tries to load a hex encoded `Dbc` from the
```
jacderida, line 26: Why do you use such a small line length for these doc comments?

safenode/src/bin/cli/wallet.rs
```diff
@@ -0,0 +1,148 @@
...
+        /// given path and deposit it to the wallet.
+        #[clap(name = "dbc-dir")]
+        dbc_dir: PathBuf,
```
jacderida, line 29: Why does this point to a directory rather than a file? Is it going to deposit all the files in the directory?

safenode/src/bin/cli/wallet.rs
```diff
@@ -0,0 +1,148 @@
...
+        /// The location of the wallet file.
+        #[clap(name = "wallet-dir")]
+        wallet_dir: PathBuf,
```
jacderida: Does the user really need control of this directory? Can't we just store this somewhere for them, like under `~/.safe`?

safenode/src/bin/cli/cfg.rs
oetyng: Yeah, IDK, I just tried to not remove any functionality.
If you think we can move it one level, then I'll fix that.

safenode/src/bin/cli/files.rs
oetyng: Great, thanks. Will fix.

safenode/src/bin/cli/files.rs
oetyng: Okay, yeah.

So this will load a new file that we created, where we stored all file addresses that we uploaded.

So it will load that file, deserialize into a vec and iterate through it to download the previously uploaded files.

safenode/src/bin/cli/register.rs
oetyng: Definitely. I will fix.

safenode/src/bin/cli/wallet.rs
jacderida: Incidentally, I also want to change the location of `~/.safe`, to `~/.config/safe`, which is what other apps do, but that's a separate issue.

safenode/src/bin/cli/wallet.rs
oetyng: Yeah I could probably extend them. I try to keep them not too much longer than the field/fn signature etc that they document. That way if people keep tabs where they want to minimize width to only see what's necessary, they won’t cut out lots of docs.

Probably shaped from how I work, I've got so many tabs open so I don’t want lots of "white space" in a tab.

safenode/src/bin/cli/wallet.rs
oetyng: Yep.
One file per dbc.

Fits well with the faucet plan, where they download a dbc hex to a dir.

So they can keep downloading to that dir, and deposit as they go along.

(It's idempotent so they don't need to remove already deposited ones.)

safenode/src/bin/cli/wallet.rs
oetyng: Yep, definitely, I had that as default before.

But even better to remove this arg, less clutter is better and the .safe is where all this is stored.

Will fix

oetyng: Aha, saw the second comment now.

Yeah sounds good. Align with standard practises.

safenode/src/bin/cli/cfg.rs
oetyng: I will do this in a follow up PR, as a proper refactor of that in isolation.
Here we just maintain functionality.

safenode/src/bin/cli/files.rs
oetyng: File names path is something different than download path.

So it's where the file names are found (they are serialized into a file that was created in the upload dir of files that were uploaded, all this is temporary, it's just one iteration on top of what was even more temporary before it..).

So, leaving that description out I feel will be confusing. I.e. we are not specifying a download path here.

safenode/src/bin/kadclient.rs
```diff
@@ -6,75 +6,30 @@
...
+use eyre::Result;
+use tracing::info;
 
 #[tokio::main]
 async fn main() -> Result<()> {
     let opt = Opt::parse();
-    let _log_appender_guard = init_node_logging(&opt.log_dir)?;
+
+    if let SubCmd::Cfg(cmds) = &opt.cmd {
+        cfg_cmds(cmds).await?;
```
b-zee: This is also called later on in the `match` statement.

oetyng: Yep. I had to have it the first thing we do, since logs need to be enabled from start.

But then I didn't want to not handle it further down. Shall I wild card that one below? That's less desireable I feel? (Wild carding in general, *potential* double handle is better then?)

b-zee (approved):

safenode/src/bin/kadclient.rs
oetyng: Fixed!

PR #136, opened 2023-04-20
[testing] test: client CLI confirming dead node gone in closest

joshuef:


maqi:
work moved to https://github.com/maidsafe/safe_network/pull/134

PR #141, opened 2023-04-20
chore: remove file logs from client cli

oetyng:
- Instead do something like enable -vvvv and pipe it into a log file.


jacderida (approved):

oetyng:
> AI-Generated Summary: This pull request removes file logging from the client CLI and makes the following changes:
> * Instead of file logging, allows using `-vvvv` option and piping the output into a log file

This isn't exactly true. There is no code here that allows for using `-vvvv` or any of its variants.
But the suggestion is that we'd implement such if we wanted to.

PR #145, opened 2023-04-21
ci: remove merge run on main

joshuef:
Now we have merge queue again, we're running tests twice if we do this.

This simplifies that


grumbach (approved):
🚢

PR #149, opened 2023-04-21
feat: detect dead peer

maqi:
This PR contains work to detect dead node properly and fire the report only once.

### DIFF
--- a/safenode/src/network/event.rs
+++ b/safenode/src/network/event.rs
@@ -167,8 +167,24 @@ impl SwarmDriver {
                     }
                 }
             }
-            SwarmEvent::ConnectionClosed { peer_id, .. } => {
-                info!("Peer {} has gone offline", peer_id);
+            SwarmEvent::ConnectionClosed {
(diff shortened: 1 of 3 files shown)


safenode/src/network/event.rs
```diff
@@ -167,8 +167,24 @@ impl SwarmDriver {
                     }
                 }
             }
-            SwarmEvent::ConnectionClosed { peer_id, .. } => {
-                info!("Peer {} has gone offline", peer_id);
+            SwarmEvent::ConnectionClosed {
+                peer_id, endpoint, ..
+            } => {
+                // The connection closed due to no msg flow show as `endpoint::receiver`
```
oetyng, line 173: Is there nothing we need to do with this?
A ping or so perhaps?

maqi: libp2p itself will carry out a dail attempt by itself.
under the current config, seems we don't need to do anything so far for idle timed out connections.

bochaco (approved):

PR #150, opened 2023-04-21
feat: file system storage

### LINKED ISSUE #156: Implement file system storage.
Pull in the logic from old repo, and implement it for the new spend storage.

oetyng:
Resolves #156.

This implements file system storage for chunks, registers and spends.

### DIFF
--- a/safenode/src/domain/node_transfers/mod.rs
+++ b/safenode/src/domain/node_transfers/mod.rs
@@ -23,7 +23,10 @@ use crate::{
 
 use sn_dbc::{DbcId, DbcTransaction, MainKey, SignedSpend, Token};
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    path::Path,
+};
 
 /// This is an arbitrary number.
 /// The supply/demand dynamics enabled by the
@@ -41,18 +44,18 @@ pub(crate) struct Transfers {
 
 impl Transfers {
     /// Create a new instance of `Transfers`.
-    pub(crate) fn new(node_id: NodeId, node_reward_key: MainKey) -> Self {
+    pub(crate) fn new(node_id: NodeId, node_reward_key: MainKey, root_dir: &Path) -> Self {
         Self {
             node_id,
             node_reward_key,
             spend_queue: SpendQ::with_fee(STARTING_FEE),
-            storage: SpendStorage::new(),
+            storage: SpendStorage::new(root_dir),
         }
     }
 
     /// Get Spend from local store.
     pub(crate) async fn get(&self, address: DbcAddress) -> Result<SignedSpend> {
-        self.storage.get(address).await
+        Ok(self.storage.get(&address).await?)
     }
 
     /// Get the required fee for the specified spend priority.
@@ -84,7 +87,7 @@ impl Transfers {
         a_spend: &SignedSpend,
         b_spend: &SignedSpend,
     ) -> Result<()> {
-        self.storage.try_add_double(a_spend, b_spend).await
+        Ok(self.storage.try_add_double(a_spend, b_spend).await?)
     }
 
     /// Tries to add a new spend to the queue.
@@ -130,7 +133,7 @@ impl Transfers {
             // NB: This works for now. We can look at
             // a timeout backstop in coming iterations.
             if let Some((signed_spend, _)) = self.spend_queue.pop() {
-                return self.storage.try_add(&signed_spend).await;
+                return Ok(self.storage.try_add(&signed_spend).await?);
             }
         }
 
--- a/safenode/src/domain/storage/address/register.rs
+++ b/safenode/src/domain/storage/address/register.rs
@@ -21,18 +21,29 @@ pub struct RegisterAddress {
 }
 
 impl RegisterAddress {
-    /// Constructs a new `RegisterAddress` given `name` and `tag`.
+    /// Construct a new `RegisterAddress` given `name` and `tag`.
     pub fn new(name: XorName, tag: u64) -> Self {
         Self { name, tag }
     }
 
-    /// Returns the name.
+    /// Return the identifier of the register.
+    /// This is used to locate the register on the network.
+    pub fn id(&self) -> XorName {
+        let mut bytes = [0u8; 40];
+        let name_bytes = self.name.0;
+        let tag_bytes = self.tag.to_be_bytes();
+        bytes.copy_from_slice(&name_bytes[0..32]);
+        bytes.copy_from_slice(&tag_bytes[0..8]);
+        XorName::from_content(&bytes)
+    }
+
+    /// Return the name.
     /// This is not a unique identifier.
     pub fn name(&self) -> &XorName {
         &self.name
     }
 
-    /// Returns the tag.
+    /// Return the tag.
     pub fn tag(&self) -> u64 {
         self.tag
     }
--- a/safenode/src/domain/storage/address/register.rs
+++ b/safenode/src/domain/storage/address/register.rs
@@ -29,11 +29,9 @@ impl RegisterAddress {
     /// Return the identifier of the register.
     /// This is used to locate the register on the network.
     pub fn id(&self) -> XorName {
-        let mut bytes = [0u8; 40];
-        let name_bytes = self.name.0;
-        let tag_bytes = self.tag.to_be_bytes();
-        bytes.copy_from_slice(&name_bytes[0..32]);
-        bytes.copy_from_slice(&tag_bytes[0..8]);
+        let mut bytes = vec![];
+        bytes.extend_from_slice(&self.name.0);
+        bytes.extend_from_slice(&self.tag.to_be_bytes());
         XorName::from_content(&bytes)
     }
 
--- a/safenode/src/domain/storage/chunks/mod.rs
+++ b/safenode/src/domain/storage/chunks/mod.rs
@@ -27,7 +27,7 @@ use tokio::{
     io::AsyncWriteExt,
 };
 use tracing::info;
-use xor_name::XorName;
+use xor_name::{XorName, XOR_NAME_LEN};
 
 const CHUNKS_STORE_DIR_NAME: &str = "chunks";
 
@@ -58,11 +58,11 @@ impl ChunkStorage {
     fn chunk_filepath_to_address(path: &Path) -> Result<ChunkAddress> {
         let filename = path
             .file_name()
-            .ok_or_else(|| Error::NoFilename(path.to_path_buf()))?
+            .ok_or_else(|| Error::PathIsNotAFile(path.to_path_buf()))?
             .to_str()
             .ok_or_else(|| Error::InvalidFilename(path.to_path_buf()))?;
 
-        let xorname = XorName(<[u8; 32]>::from_hex(filename)?);
+        let xorname = XorName(<[u8; XOR_NAME_LEN]>::from_hex(filename)?);
         Ok(ChunkAddress::new(xorname))
     }
 
--- a/safenode/src/domain/storage/chunks/mod.rs
+++ b/safenode/src/domain/storage/chunks/mod.rs
@@ -10,86 +10,271 @@ mod chunk;
 
 pub use self::chunk::Chunk;
 
-use super::{ChunkAddress, Error, Result};
+use super::{
+    error::{Error, Result},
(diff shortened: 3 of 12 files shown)


safenode/src/domain/storage/chunks/mod.rs
```diff
@@ -10,86 +10,271 @@ mod chunk;
...
 
 impl ChunkStorage {
-    pub(crate) fn new() -> Self {
-        let capacity =
-            NonZeroUsize::new(CHUNKS_CACHE_SIZE).expect("Failed to create in-memory Chunk storage");
+    /// Creates a new `ChunkStorage` at the specified root location
+    ///
+    /// If the location specified already contains a `ChunkStorage`, it is simply used
+    ///
+    /// Used space of the dir is tracked
```
oetyng: Fixing this old comment.

safenode/src/domain/storage/chunks/mod.rs
```diff
@@ -10,86 +10,271 @@ mod chunk;
...
+
+    fn chunk_addr_to_filepath(&self, addr: &ChunkAddress) -> Result<PathBuf> {
+        let xorname = *addr.name();
+        let path = prefix_tree_path(&self.file_store_path, xorname);
+        let filename = hex::encode(xorname);
+        Ok(path.join(filename))
+    }
+
+    #[allow(unused)]
+    pub(super) async fn remove(&self, address: &ChunkAddress) -> Result<()> {
```
oetyng, line 78: Adding comments here to clarify when this is to be used.

safenode/src/domain/storage/spends.rs
```diff
@@ -6,58 +6,71 @@
...
+        match read(file_path).await {
+            Ok(bytes) => {
+                let spend: SignedSpend = deserialize(&bytes)?;
+                if address == &dbc_address(spend.dbc_id()) {
+                    Ok(spend)
+                } else {
+                    // This can happen if the content read is empty, or incomplete,
+                    // possibly due to an issue with the OS synchronising to disk,
+                    // resulting in a mismatch with recreated address of the Chunk.
+                    Err(Error::SpendNotFound(*address))
```
oetyng, line 65: Fixing this `Chunk` in comment.

safenode/src/domain/storage/registers/mod.rs
```diff
@@ -367,6 +457,34 @@ impl RegisterStorage {
             op_log: stored_reg.op_log,
         })
     }
+
+    #[cfg(test)]
```
RolandSherwin, line 461: Can we make this function available everywhere by removing the `cfg` flag? This might be useful during the data replication flow.

safenode/src/domain/node_transfers/mod.rs
```diff
@@ -41,18 +44,18 @@ pub(crate) struct Transfers {
...
             spend_queue: SpendQ::with_fee(STARTING_FEE),
-            storage: SpendStorage::new(),
+            storage: SpendStorage::new(root_dir),
         }
     }
 
     /// Get Spend from local store.
     pub(crate) async fn get(&self, address: DbcAddress) -> Result<SignedSpend> {
-        self.storage.get(address).await
+        Ok(self.storage.get(&address).await?)
```
maqi, line 58: is the unwrap then wrap necessary?
same to the belows

safenode/src/domain/storage/address/register.rs
```diff
@@ -21,18 +21,27 @@ pub struct RegisterAddress {
...
-    /// Constructs a new `RegisterAddress` given `name` and `tag`.
+    /// Construct a new `RegisterAddress` given `name` and `tag`.
     pub fn new(name: XorName, tag: u64) -> Self {
         Self { name, tag }
     }
 
-    /// Returns the name.
+    /// Return the identifier of the register.
+    /// This is used to locate the register on the network.
+    pub fn id(&self) -> XorName {
```
maqi, line 31: hmm... 
I feel it's bit confusing now, both `id()` and `name()` returns with `XorName` 
I can understand that `id` is `hash(name+tag)` to be unique
probably not expose `name()` (to avoid mis-use) 
or to make `id()` type safe
or give `name()` a more explicit naming, say `service_name` or `content_name` or etc?

safenode/src/domain/storage/chunks/mod.rs
```diff
@@ -10,86 +10,269 @@ mod chunk;
...
+    }
+
+    fn chunk_filepath_to_address(path: &Path) -> Result<ChunkAddress> {
+        let filename = path
+            .file_name()
+            .ok_or_else(|| Error::NoFilename(path.to_path_buf()))?
+            .to_str()
+            .ok_or_else(|| Error::InvalidFilename(path.to_path_buf()))?;
+
+        let xorname = XorName(<[u8; 32]>::from_hex(filename)?);
```
maqi: better use XOR_NAME_LEN instead of 32.

btw. would like to confirm that the filename of a chunk is in hex, which will be 2*XOR_NAME_LEN , and it will be ok to be parsed here, right?

safenode/src/domain/storage/chunks/mod.rs
```diff
@@ -10,86 +10,269 @@ mod chunk;
...
+        Ok(ChunkAddress::new(xorname))
+    }
+
+    fn chunk_addr_to_filepath(&self, addr: &ChunkAddress) -> Result<PathBuf> {
+        let xorname = *addr.name();
+        let path = prefix_tree_path(&self.file_store_path, xorname);
+        let filename = hex::encode(xorname);
```
maqi, line 72: better specify `lowercase` or `uppercase`

safenode/src/domain/storage/error.rs
```diff
@@ -25,6 +27,12 @@ pub enum Error {
     /// Chunk not found.
     #[error("Chunk not found: {0:?}")]
     ChunkNotFound(ChunkAddress),
+    /// No filename found
+    #[error("Path contains no file name: {0}")]
+    NoFilename(PathBuf),
```
maqi: it's not filename but NoFileInDir ?

safenode/src/domain/storage/registers/mod.rs
```diff
@@ -11,128 +11,95 @@ pub mod register;
...
+        self.write_log_to_disk(&vec![cmd.clone()], &stored_reg.op_log_path)
             .await
     }
 
+    /// This is to be used when a node is shrinking the address range it is responsible for.
+    #[allow(dead_code)]
+    pub(super) async fn remove(&self, address: &RegisterAddress) -> Result<()> {
+        trace!("Removing Register: {address:?}");
+        let filepath = self.address_to_filepath(address)?;
+        remove_file(filepath).await?;
```
maqi, line 99: will `file not find` be counted as an error?
As I understand, it shall not?

safenode/src/domain/storage/registers/mod.rs
```diff
@@ -152,28 +119,94 @@ impl RegisterStorage {
...
+
+        for cmd in log {
+            if let Err(err) = self.write_register_cmd(cmd, path).await {
+                error!("Failed to write Register cmd {cmd:?} to disk: {err:?}");
+                last_err = Some(err);
             }
         }
+
+        if let Some(err) = last_err {
+            Err(err)
```
maqi, line 153: I do think here we shall fail the whole (remove the whole path, if there is per path for each log? otherwise to remove all created cmd_files?) in case of having ANY of error?

safenode/src/domain/storage/spends.rs
```diff
@@ -98,25 +122,25 @@ impl SpendStorage {
...
-                    double_spends.insert(address, (existing.clone(), signed_spend.clone()));
+                // We don't error if the double spend failed, as we rather want to
+                // announce the double spend attempt to close group. TODO: how to handle the error then?
+                let _ = self.try_store_double_spend(&existing, signed_spend).await;
 
                 // The spend is now permanently removed from the valid spends.
-                let _removed = valid_spends.remove(&address);
+                // We don't error if the remove failed, as we rather want to
+                // announce the double spend attempt to close group.
+                // The double spend will still be detected o querying for the spend.
```
maqi: o querying ->  by querying ?

safenode/src/domain/storage/spends.rs
```diff
@@ -165,20 +189,88 @@ impl SpendStorage {
...
+    async fn try_get_double_spend(
+        &self,
+        address: &DbcAddress,
+    ) -> Result<(SignedSpend, SignedSpend)> {
+        trace!("Getting double spend: {address:?}");
+        let file_path = self.address_to_filepath(address, &self.double_spends_path)?;
+        match read(file_path).await {
+            Ok(bytes) => {
+                let (a_spend, b_spend): (SignedSpend, SignedSpend) = deserialize(&bytes)?;
+                // They should have the same dbc id, so we can use either. TODO: Or should we check both? What if they are different?
```
maqi: line break for the TODO point

safenode/src/domain/storage/registers/mod.rs
oetyng: Absolutely, and it has been used like that before. So yep!

safenode/src/domain/node_transfers/mod.rs
oetyng: Yes it is, that's how it maps from one error type to another (the returned type has a variant with #[from].

safenode/src/domain/storage/address/register.rs
oetyng: Yeah, it’s just that it’s so old the use of name, and I was thinking bout a rename, but better to think it through, and it’s not really about the move to use file system, so I wanted the PR to not do such changes.

safenode/src/domain/storage/registers/mod.rs
oetyng: If you wanted it removed, and it wasn't there, then that's not an error. You got what you wanted, for it to not be there :)

safenode/src/domain/storage/registers/mod.rs
oetyng: Perhaps, but this is just moving over the code we had. Better to do refactors and improvements in another step.

safenode/src/domain/storage/chunks/mod.rs
oetyng: I don't know what you mean here.

maqi: Ah, not important, I thought the hex::encode has a lower_case / upper_case specification.
however seems not, so just forgot it.

safenode/src/domain/storage/error.rs
oetyng: Ahum, will check. It's the old code, but if isn't making sense then better fix it now.

safenode/src/domain/storage/chunks/mod.rs
oetyng: Yeah old code again. I can put it in there.

oetyng: I asssume it will be ok to parse it here, since we've had this code for a long time in the other repo.

safenode/src/domain/storage/registers/mod.rs
oetyng: But I mean, just remove it when you implement the replication!

safenode/src/domain/storage/error.rs
oetyng: The error is used here:

```
let filename = path
            .file_name()
            .ok_or_else(|| Error::NoFilename(path.to_path_buf()))?
```

It seems it expects that the path is to a file, this for it to have a filename, but then it isn't a path to a file. So not exactly that there is no file in a dir.

But I could clarify by renaming to... say ExpectedPathToBeAFile. PathNotAFile, ShouldBeAFile, or something like that..

oetyng: I went with `PathIsNotAFile`, how about that?

safenode/src/domain/storage/chunks/mod.rs
oetyng: Fixed XOR_NAME_LEN .

safenode/src/domain/storage/spends.rs
oetyng: fixed

safenode/src/domain/storage/spends.rs
oetyng: fixed

jacderida (approved):

PR #155, opened 2023-04-21
WIP: data replication 

RolandSherwin:
- the basic flow is working, need to flush things out

### DIFF
--- a/safenode/src/domain/client_transfers/online.rs
+++ b/safenode/src/domain/client_transfers/online.rs
@@ -13,7 +13,7 @@ use crate::{
     domain::fees::{FeeCiphers, RequiredFee, SpendPriority},
     network::close_group_majority,
     node::NodeId,
-    protocol::messages::{Query, QueryResponse, Request, Response, SpendQuery},
+    protocol::messages::{DataRequest, DataResponse, Query, QueryResponse, SpendQuery},
 };
 
 use sn_dbc::{
@@ -167,7 +167,7 @@ async fn select_inputs(
         };
         // ---------------- fee part end ----------------
 
-        // Add this dbc as input to be spent.
+        // Add this dbc Requestas input to be spent.
         dbcs_to_spend.push((dbc, derived_key));
 
         // Input amount increases with the amount of the dbc.
@@ -356,12 +356,12 @@ fn create_transfer_with(selected_inputs: Inputs) -> Result<Outputs> {
 }
 
 async fn get_fees(dbc_id: DbcId, client: &Client) -> Result<BTreeMap<NodeId, RequiredFee>> {
-    let request = Request::Query(Query::Spend(SpendQuery::GetFees {
+    let request = DataRequest::Query(Query::Spend(SpendQuery::GetFees {
         dbc_id,
         priority: SpendPriority::Normal,
     }));
     let responses = client
-        .send_to_closest(request)
+        .send_data_req_to_closest(request)
         .await
         .map_err(|e| Error::CouldNotGetFees(e.to_string()))?;
 
@@ -376,7 +376,7 @@ async fn get_fees(dbc_id: DbcId, client: &Client) -> Result<BTreeMap<NodeId, Req
             res
         })
         .filter_map(|resp| match resp {
-            Response::Query(resp) => Some(resp),
+            DataResponse::Query(resp) => Some(resp),
             other => {
                 warn!("Unexpected response to fee query: {other:?}");
                 None
--- a/safenode/src/domain/storage/registers/mod.rs
+++ b/safenode/src/domain/storage/registers/mod.rs
@@ -10,9 +10,7 @@
 pub mod register;
 
 use self::register::{Action, EntryHash, Register, User};
-
 use super::{prefix_tree_path, Error, RegisterAddress, Result};
-
 use crate::{
     domain::storage::list_files_in,
     protocol::{
@@ -23,7 +21,6 @@ use crate::{
         },
     },
 };
-
 use bincode::serialize;
 use std::path::{Path, PathBuf};
 use tokio::{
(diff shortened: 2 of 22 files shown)


safenode/src/domain/client_transfers/online.rs
```diff
@@ -167,7 +167,7 @@ async fn select_inputs(
         };
         // ---------------- fee part end ----------------
 
-        // Add this dbc as input to be spent.
+        // Add this dbc Requestas input to be spent.
```
maqi, line 170: request as

safenode/src/domain/storage/registers/mod.rs
```diff
@@ -91,6 +88,33 @@ impl RegisterStorage {
...
+        let iter = list_files_in(&self.file_store_path)
+            .into_iter()
+            .filter_map(|e| e.parent().map(|parent| (parent.to_path_buf(), e.clone())));
+
+        let mut addrs = BTreeMap::new();
+        for (parent, op_file) in iter {
+            if let Entry::Vacant(vacant) = addrs.entry(parent) {
+                if let Ok(Ok(cmd)) = read(op_file)
+                    .await
+                    .map(|serialized_data| deserialize::<RegisterCmd>(&serialized_data))
```
maqi, line 106: wondering shall the deserialized register got returned as well for replication ?
to avoid second round of read ?

safenode/src/network/mod.rs
```diff
@@ -283,54 +281,102 @@ impl Network {
...
     }
 
-    /// Returns the closest peers to the given `XorName`, sorted by their distance to the xor_name.
+    /// Get the closest peers to the given key by querying the network. The result is
+    /// sorted by their distance to the key.
+    ///
     /// If `client` is false, then include `self` among the `closest_peers`
-    async fn get_closest_peers(&self, xor_name: XorName, client: bool) -> Result<Vec<PeerId>> {
+    /// Use `client_get_closest_peers` or `node_get_closest_peers` for convenience.
+    pub async fn get_closest_peers(&self, key: Vec<u8>, client: bool) -> Result<Vec<PeerId>> {
```
maqi, line 308: to avoid mis-use, better not to expose this `pub` .
try to access via client_get_closest or node_get_closest always

safenode/src/network/mod.rs
```diff
@@ -283,54 +281,102 @@ impl Network {
...
+    }
 
-        if CLOSE_GROUP_SIZE > closest_peers.len() {
-            warn!("Not enough peers in the k-bucket to satisfy the request");
-            return Err(Error::NotEnoughPeers);
+    /// Get the closest peers to the given key from the local DHT. The result is
+    /// sorted by their distance to the key.
+    ///
+    /// If `client` is false, then include `self` among the `closest_peers`
+    pub async fn get_closest_local_peers(&self, key: Vec<u8>, client: bool) -> Result<Vec<PeerId>> {
```
maqi, line 329: try not to expose this as Pub directly?

safenode/src/node/replication.rs
```diff
@@ -0,0 +1,155 @@
...
+use libp2p::PeerId;
+use std::{collections::HashSet, time::Duration};
+use tokio::spawn;
+
+pub(crate) async fn send_out_data_addrs(
+    network: Network,
+    chunk_storage: ChunkStorage,
+    _register_storage: RegisterStorage,
+) {
+    tokio::time::sleep(Duration::from_secs(10)).await;
```
maqi, line 27: don't think we need this `periodic` style check
will discuss it in detail in slack

safenode/src/node/replication.rs
```diff
@@ -0,0 +1,155 @@
...
+                    addr,
+                ))))
+            }
+            DataAddress::Spend(_) => todo!(),
+        };
+        let network = network.clone();
+        let chunk_storage = chunk_storage.clone();
+        let _register_storage = register_storage.clone();
+        let _handle = spawn(async move {
+            if let Ok(responses) = network.node_send_to_closest(&req).await {
```
maqi, line 92: this will send the req to close_group_peers, which is for sure a waste of resources and could trigger an avalanch that slow down the network a lot.
same we can discuss it further in the slack

safenode/src/node/replication.rs
```diff
@@ -0,0 +1,155 @@
...
+                        // todo
+                        // register_storage.update(register).await.ok();
+                    };
+                }
+            }
+        });
+    }
+}
+
+async fn check_for_group_change(
```
maqi, line 113: we don't need to carry out an explicit periodic check.
just carry out replication on `node add` & `dead node` shall be enough

safenode/src/network/mod.rs
oetyng: Good point. We don't want the situation where we can access lots of apis that we don't need to and shouldn't call.
Only the minimum required.

safenode/src/network/mod.rs
oetyng: Good point. Same as above.

safenode/src/node/replication.rs
oetyng: Good observations.

maqi:
Following is for a pseudo code suggestion for the replication flow.
```rust
when newly_node_added {
    // Kad config insert node on connection established, hence the result shall contain the new node
    let closest_peers = self.kad.get_closest_peers_from_local(self.peer_id);
	self.replication(closest_peers, closest_peers, node_added);
	
	// To avoid holding no-longer-closest data, we shall carry out a clean-up when there is new node added
	self.clean_up_data()
}


when dead_node_first_detected {
    // Not carry out `remove_address` yet, the `get_closest_peers_from_local` shall give a result
    // containing the dead_node (if it is among the closest).
    // In case from_local won't give that, we can still cmp distance with the last closest peer to figure out
    // whether and whom is the replcament of the dead node.
    let prev_closest_peers = self.kad.get_closest_peers_from_local(self.peer_id);
    self.kad.remove_address(node_dead);
    // This is after the `remove_address` , hence the dead node shall got removed
    let cur_closest_peers = self.kad.get_closest_peers_from_local(self.peer_id); 
	self.replication(cur_closest_peers, prev_closest_peers, node_dead);
}


fn replication(self, cur_closest_peers, prev_closest_peers, node_churned) {
    let target = if cur_closest_peers.contains(node_churned) {
		// This is about a newly added node
		node_churned
	} else if prev_closest_peers.contains(node_churned) {
        // The dead node was among the closest
        node_appears_in_cur_closest_but_not_in_pre_closest
	} else {
        //The dead node was among the closest
    };
	self.send_data_to_target(target)
}

fn send_data_to_target(self, target) {
    // we don't need to wait for resp in this case, hence can spawn directly
	let data_to_be_replicated = self.data_to_be_replicated();
	spawn {
	    data_to_be_replicated.iter().for_each(|entry| self.send_to_peer(target));
	}
}

fn data_to_be_replicated(self) -> Vec<_> {
    // Here we can have two models:
	// Model 1: as current, we send out the addrses only, and let the target ask for what it missing
	//          but this is then a req/res communication, may require handler and cann't just spawn without wait? 
	// Model 2: directly send data we had to the target.
	//          this doesn't require wait on resp, hence no handler required and can be spawed directly
	//          We can optimize it by only send random 1/4 (given CLOSE_GROUP_SIZE is 8)         
}
```

PR #157, opened 2023-04-21
test: cover spend storage with unit tests

### LINKED ISSUE #158: Cover SpendStorage with unit tests.

oetyng:
Resolves #158.

### DIFF
--- a/safenode/src/domain/storage/spends.rs
+++ b/safenode/src/domain/storage/spends.rs
@@ -117,24 +117,23 @@ impl SpendStorage {
     /// and double spent attempts to be missed (as the validation and adding
     /// could otherwise happen in parallel in different threads.)
     pub(crate) async fn validate(&mut self, signed_spend: &SignedSpend) -> Result<()> {
-        if self.is_unspendable(signed_spend.dbc_id()).await {
-            return Ok(()); // Already unspendable, so we don't care about this spend.
-        }
-
         let address = dbc_address(signed_spend.dbc_id());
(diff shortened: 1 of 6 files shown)


safenode/src/domain/storage/spends.rs
```diff
@@ -281,14 +281,194 @@ impl Display for SpendStorage {
...
+        let dbc = split(&src_dbc, &key, 1).expect("Split to succeed.");
+        let (dbc, _) = &dbc[0];
+        let spend = dbc.signed_spends.last().expect("Should contain a spend.");
+
+        // Adding the exact same spend is idempotent.
+        storage
+            .try_add(spend)
+            .await
+            .expect("First spend should be added.");
+        storage
```
jacderida, line 337: Is this operation idempotent because it doesn't produce an error? Or is there any state we need to verify to make sure something wasn't changed?

jacderida (approved):

safenode/src/domain/storage/spends.rs
oetyng: Adding the exact same spend again and again, doesn't change anything. Because it can only exist or not exist.

oetyng: If we however tried to add a spend - with same id as an existing spend - but something else in it is not the same, then we would have an error.

Depending on what is different, that is just an error returned to the client, or if something that amounts to what would qualify as a double spend attempt, then the spend is killed immediately (i.e. it is marked as a double spend attempt).

PR #160, opened 2023-04-22
fix: do not panic in cli

oetyng:
- There is no need for it. Print what did/did not happen and exit.

### DIFF
--- a/safenode/src/bin/cli/files.rs
+++ b/safenode/src/bin/cli/files.rs
@@ -63,7 +63,7 @@ async fn upload_files(files_path: PathBuf, file_api: &Files) -> Result<()> {
                     chunks_to_fetch.push(*address.name());
                 }
                 Err(error) => {
-                    panic!(
+                    println!(
                         "Did not store file {file_name:?} to all nodes in the close group! {error}"
                     )
                 }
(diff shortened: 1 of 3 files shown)


safenode/src/bin/cli/files.rs
```diff
@@ -103,7 +103,7 @@ async fn download_files(file_names_dir: PathBuf, file_api: &Files) -> Result<()>
                         println!("Successfully got file {xorname} of {} bytes!", bytes.len());
                     }
                     Err(error) => {
-                        panic!("Did not get file {xorname:?} from the network! {error}")
+                        println!("Did not get file {xorname:?} from the network! {error}")
```
dirvine, line 106: Should this also retry the Get and also above retry the store. 
Perhaps argument for just printing could not Get, but I wonder should store try again and only fail after 3 attempts or something similar.

safenode/src/bin/cli/wallet.rs
```diff
@@ -97,7 +97,8 @@ async fn send(amount: String, to: String, client: &Client) -> Result<()> {
     use std::str::FromStr;
     let amount = Token::from_str(&amount)?;
     if amount.as_nano() == 0 {
-        panic!("This should be unreachable. An amount is expected when sending.");
+        println!("Invalid format or zero amount passed in. Nothing sent.");
```
dirvine, line 100: This one looks good though, no harm and no cost.

safenode/src/bin/cli/files.rs
dirvine: Actually this is fine, Error back to app and let the app decide.

dirvine (approved):

PR #163, opened 2023-04-23
feat: store created dbcs as hex to file (and load for deposit)

### LINKED ISSUE #161: Print created dbcs to file as hex.

oetyng:
Resolves #161 and #162.

When a users sends tokens, a client will now print the created dbcs to the `created_dbcs` dir in wallet dir.
This allows a client to send these files to the dbc recipients, for recipients to deposit the dbcs to their wallets.
(The received dbcs are placed in the `received_dbcs`dir, and user then just calls deposit on the client, and it loads them into the wallet.)

### DIFF
--- a/safenode/src/bin/cli/wallet.rs
+++ b/safenode/src/bin/cli/wallet.rs
@@ -19,8 +19,12 @@ use std::path::Path;
 
 #[derive(Parser, Debug)]
 pub enum WalletCmds {
-    /// Tries to load any hex encoded `Dbc`s from the `received_dbcs`
+    /// Deposit `Dbc`s to the local wallet.
+    /// Tries to load any `Dbc`s from the `received_dbcs`
     /// path in the wallet dir, and deposit it to the wallet.
+    /// The user has to manually place received dbc files to
+    /// that dir, for example by choosing that path when downloading
+    /// the dbc file from email or browser.
     Deposit,
     Send {
         /// This shall be the number of nanos to send.
--- a/safenode/src/bin/cli/wallet.rs
+++ b/safenode/src/bin/cli/wallet.rs
@@ -15,16 +15,13 @@ use sn_dbc::Token;
 
 use clap::Parser;
 use eyre::Result;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 #[derive(Parser, Debug)]
 pub enum WalletCmds {
-    Deposit {
-        /// Tries to load a hex encoded `Dbc` from the
-        /// given path and deposit it to the wallet.
-        #[clap(name = "dbc-dir")]
-        dbc_dir: PathBuf,
-    },
+    /// Tries to load any hex encoded `Dbc`s from the `received_dbcs`
+    /// path in the wallet dir, and deposit it to the wallet.
+    Deposit,
     Send {
         /// This shall be the number of nanos to send.
         /// Necessary if the `send_to` argument has been given.
@@ -38,13 +35,13 @@ pub enum WalletCmds {
 
 pub(crate) async fn wallet_cmds(cmds: WalletCmds, client: &Client, root_dir: &Path) -> Result<()> {
     match cmds {
-        WalletCmds::Deposit { dbc_dir } => deposit(dbc_dir, root_dir).await?,
+        WalletCmds::Deposit => deposit(root_dir).await?,
         WalletCmds::Send { amount, to } => send(amount, to, client, root_dir).await?,
     }
     Ok(())
 }
 
-async fn deposit(_dbc_dir: PathBuf, root_dir: &Path) -> Result<()> {
+async fn deposit(root_dir: &Path) -> Result<()> {
     let mut wallet = LocalWallet::load_from(root_dir).await?;
 
     let previous_balance = wallet.balance();
--- a/safenode/src/bin/cli/wallet.rs
+++ b/safenode/src/bin/cli/wallet.rs
@@ -8,16 +8,14 @@
 
 use safenode::{
     client::{Client, WalletClient},
-    domain::wallet::{parse_public_address, DepositWallet, LocalWallet, Wallet},
+    domain::wallet::{parse_public_address, LocalWallet, Wallet},
 };
 
-use sn_dbc::{Dbc, Token};
+use sn_dbc::Token;
 
 use clap::Parser;
 use eyre::Result;
 use std::path::{Path, PathBuf};
-use tokio::fs;
-use walkdir::WalkDir;
 
 #[derive(Parser, Debug)]
 pub enum WalletCmds {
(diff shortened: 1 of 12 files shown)


safenode/src/bin/cli/wallet.rs
```diff
@@ -8,25 +8,20 @@
...
 
 #[derive(Parser, Debug)]
 pub enum WalletCmds {
-    Deposit {
-        /// Tries to load a hex encoded `Dbc` from the
-        /// given path and deposit it to the wallet.
-        #[clap(name = "dbc-dir")]
-        dbc_dir: PathBuf,
-    },
+    /// Tries to load any hex encoded `Dbc`s from the `received_dbcs`
```
jacderida: I think it would be helpful to change this to "Deposit any DBCs from the local wallet and store them on the network". I don't think we need to leak details to the user about directories and hex encoding.

safenode/src/domain/wallet/local_store.rs
```diff
@@ -496,6 +519,72 @@ mod tests {
...
+        let sender_root_dir = sender_root_dir.path().to_path_buf();
+
+        let mut sender = LocalWallet::load_from(&sender_root_dir).await?;
+        let sender_dbc = create_genesis_dbc(&sender.key).expect("Genesis creation to succeed.");
+        sender.deposit(vec![sender_dbc]);
+
+        let send_amount = 100;
+
+        // Send to a new address.
+        let recipient_root_dir = create_temp_dir()?;
```
jacderida: You could consider using the `assert_fs` crate for working with and creating with temporary directories. It's really cool and useful. It drops any directories when they go out of scope, whereas this code would leave them around after the test runs.

safenode/src/domain/wallet/local_store.rs
```diff
@@ -496,6 +519,72 @@ mod tests {
...
+
+        tokio::fs::create_dir_all(&received_dbc_dir).await?;
+        let received_dbc_file = received_dbc_dir.join(&dbc_id_file_name);
+
+        // Move the created dbc to the recipient's received_dbcs dir.
+        tokio::fs::rename(created_dbc_file, &received_dbc_file).await?;
+
+        assert_eq!(0, recipient.wallet.balance().as_nano());
+
+        recipient.try_load_deposits().await?;
```
jacderida, line 571: I wasn't sure exactly, is this the method you're looking to cover with this test? You could consider putting that in the name of the test.

Also, I can see the setup might be a bit tedious, but it could be worth maybe having two or three DBCs, since I see it will attempt to load all of them. There was also a code path where there were no deposits. Is that worth having its own test case?

safenode/src/domain/wallet/local_store.rs
```diff
@@ -22,49 +24,69 @@ use std::{
...
     pub async fn store(&self) -> Result<()> {
-        store_wallet(&self.root_dir, &self.wallet).await
+        store_wallet(&self.wallet_dir, &self.wallet).await
+    }
+
+    /// Stores the given dbc to the `created dbcs dir` in the wallet dir.
+    /// Each recipient has their own dir, containing all dbcs for them.
+    /// These can then be sent to the recipients out of band, over any channel preferred.
+    pub async fn store_created_dbc(&mut self, dbc: Dbc) -> Result<()> {
+        store_created_dbcs(vec![dbc], &self.wallet_dir).await
```
jacderida, line 50: Just to help me understand, are the wallet operations all local? So when you run the `wallet send` command as a user, you're not actually sending anything to someone else, you just create a local directory for the recipient where the DBC is saved? So how do you actually perform the send for real?

oetyng: In the units its all local like you describe, because we have not injected a connected client, just a mock.

But the implementation in the safe client injects an actual connected client, so it gets uploaded to network on send 👍

safenode/src/bin/cli/wallet.rs
oetyng: (Deposited dbcs are not checked for validity + (registering if not yet existing there) in network yet though. TBD.)

oetyng: And the details that the dbcs need to be placed in `received_dbcs` is necessary I think, since the user still has to receive the dbcs out of band, and place them somewhere manually - which is in that dir.

safenode/src/domain/wallet/local_store.rs
oetyng: Mm, yes, but not only. It's testing `store_created_dbc` and `try_load_deposits`. I can break them out into two tests. 

(But the `try_load_deposits` need the first one to run anyway, and using the `try_load_deposits` is a good way to confirm that `store_created_dbc` works as expected and does produce something that is properly working in when `try_load_deposits` is run. )

(Btw do we name tests after the functionality/behaviour we want the system to have/display, or individual fn names?)

oetyng: I renamed to `store_created_dbc_gives_file_that_try_load_deposits_can_use`. A bit long, but should be very clear at least :)

safenode/src/domain/wallet/local_store.rs
oetyng: Done!

safenode/src/bin/cli/wallet.rs
oetyng: I removed the `hex` part and added a first more user friendly informational text, and then an explanation of the necessary steps to be taken by a user.

safenode/src/client/wallet.rs
```diff
@@ -34,9 +34,15 @@ impl<W: SendWallet> WalletClient<W> {
...
+    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<Dbc> {
         let dbcs = self.wallet.send(vec![(amount, to)], &self.client).await?;
-        Ok(dbcs)
+        if let Some(info) = dbcs.into_iter().next() {
+            Ok(info.dbc)
+        } else {
+            Err(Error::CouldNotSendTokens(
+                "No DBCs were returned from the wallet.".to_string(),
+            ))
+        }
```
b-zee:
```suggestion
        match &dbcs[..] {
            [info, ..] => Ok(info.dbc.clone()),
            [] => Err(Error::CouldNotSendTokens(
                "No DBCs were returned from the wallet.".into(),
            )),
        }
```

safenode/src/client/wallet.rs
```diff
@@ -34,9 +34,15 @@ impl<W: SendWallet> WalletClient<W> {
     }
 
     /// Send tokens to another wallet.
-    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<Vec<CreatedDbc>> {
+    pub async fn send(&mut self, amount: Token, to: PublicAddress) -> Result<Dbc> {
```
b-zee, line 37: What is the reason behind changing this interface/signature?

oetyng: The `CreatedDbc` struct contains `RevealedAmount`, which is useful to know what the amount in the dbc actually was (as it's blinded once created, and only the recipient will be able to see it).
But this `RevealedAmount` type is a bit low level to be part of that interface, and isn't very useful at that level. It should be part of the wallet ledger I think, where it will be automatically accounted for in a user friendly manner (and in later versions, they could export it from wallet api if wishing so).

safenode/src/client/wallet.rs
oetyng: Done.

safenode/src/client/wallet.rs
b-zee: What is the reason behind changing this interface?

b-zee (approved):

safenode/src/domain/wallet/local_store.rs
oetyng: Ah, I think I now read your question more as it was intended..

The actual dbcs... yes how do they reach the recipient. It's exactly as you describe, it goes to the created_dbcs directory in the client wallet path.

The user has to deliver it out of band to the recipient. On paper, on a thumb drive, in email, on whatsapp.  What ever suits :) 
But when we have some messaging in the network, it would be a way that the user could use by default, I would think :)

PR #167, opened 2023-04-23
feat: init network with genesis

### LINKED ISSUE #166: Initialize network with genesis.
All nodes initialize their spend storage with the genesis spend.

oetyng:
Resolves #166.

All nodes initialize their spend storage with the genesis spend.

### DIFF
--- a/safenode/src/domain/fees/required_fee.rs
+++ b/safenode/src/domain/fees/required_fee.rs
@@ -8,7 +8,7 @@
 
 use super::{Error, RequiredFeeContent, Result};
 
-use sn_dbc::{DbcId, Hash, MainKey, Signature, Token};
+use sn_dbc::{Hash, Signature};
 
 use serde::{Deserialize, Serialize};
 use tiny_keccak::{Hasher, Sha3};
@@ -26,9 +26,7 @@ pub struct RequiredFee {
 impl RequiredFee {
     /// Instantiate RequiredFee by encrypting the amount to the id of the dbc to spend, and signing
     /// it all with the Node reward main key.
-    pub fn new(amount: Token, dbc_id: DbcId, reward_main_key: &MainKey) -> Self {
-        let content = RequiredFeeContent::new(amount, dbc_id, reward_main_key.public_address());
-        let reward_address_sig = reward_main_key.sign(&content.to_bytes());
+    pub fn new(content: RequiredFeeContent, reward_address_sig: Signature) -> Self {
         Self {
             content,
             reward_address_sig,
--- a/safenode/src/domain/fees/required_fee.rs
+++ b/safenode/src/domain/fees/required_fee.rs
@@ -24,8 +24,8 @@ pub struct RequiredFee {
 }
 
 impl RequiredFee {
-    /// Instantiate RequiredFee by encrypting the amount to the id of the dbc to spend, and signing
-    /// it all with the Node reward main key.
+    /// The required fee consists of the content where the amount is encrypting to the id of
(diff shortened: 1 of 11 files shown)


safenode/src/domain/fees/required_fee.rs
```diff
@@ -26,9 +26,7 @@ pub struct RequiredFee {
 impl RequiredFee {
     /// Instantiate RequiredFee by encrypting the amount to the id of the dbc to spend, and signing
```
oetyng, line 27: Oops, doc update needed.

safenode/src/domain/fees/required_fee.rs
```diff
@@ -26,9 +26,7 @@ pub struct RequiredFee {
...
     /// it all with the Node reward main key.
```
oetyng: Oops, doc update needed.

oetyng: fixed

safenode/src/domain/storage/spends.rs
```diff
@@ -199,6 +182,34 @@ impl SpendStorage {
         Ok(())
     }
 
+    async fn add(&mut self, signed_spend: &SignedSpend) -> Result<()> {
+        let addr = dbc_address(signed_spend.dbc_id());
+        let filepath = self.address_to_filepath(&addr, &self.valid_spends_path)?;
+
+        if filepath.exists() {
+            self.validate(signed_spend).await?;
```
maqi, line 190: in case there is existing file record, just validate but no `add` ?
i.e. this means the spend has been added, right ?
and this doesn't count as an error ?

safenode/src/domain/storage/spends.rs
```diff
@@ -199,6 +182,34 @@ impl SpendStorage {
...
+            return Ok(());
+        }
+
+        // Store the spend to local file system.
```
maqi, line 194: does it need to be stored to network ?

safenode/src/domain/wallet/mod.rs
```diff
@@ -86,6 +90,16 @@ pub trait Wallet {
     fn balance(&self) -> Token;
 }
 
+/// A wallet has that can sign msgs.
```
maqi: A wallet shall be able to sign msgs ?

safenode/src/domain/storage/spends.rs
oetyng: If there is an existing record, that means we already have a spend for that dbc.

However, since `fn add` is only called after `fn validate` it means the spend snuck in _right between validating and adding_. 

But! Since both `fn validate` and ` fn add` take `&mut` reference to self, that race isn't actually possible, and it's more or less a double security as comment state, in the unlikely event that an uncautious developer removes the `&mut` reference.

If that was to happen (`&mut` removed), and a spend for same dbc snucks in right between the calls, it means that we validate the newly added spend, and it _will_ detect a double spend.
And if it wasn't a double spend, just the exact same spend coming in again, well, we don't need to add it again (i.e. the functionality of try add is idempotent), so we just return `Ok(())`.

safenode/src/domain/storage/spends.rs
oetyng: This is the node storage in the network. So, by storing it here, the spend is stored in the network (and considered so by the world when majority of close group has it).

safenode/src/domain/wallet/mod.rs
oetyng: Fixed.

maqi (approved):

PR #172, opened 2023-04-24
feat: initial setup of KademliaConfig for provider usage

joshuef:


safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
         resp: Response,
         channel: ResponseChannel<Response>,
     },
+    RegisterProvidedData {
+        record: Record,
+    },
+    /// Get data from the kademlia store
+    GetProvidedData {
```
oetyng: Do we need to qualify it as Provided data? What other data do we have there?

safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
...
+                        sender
+                            .send(crate::protocol::messages::QueryResponse::GetChunk(Ok(
+                                Chunk::new(Bytes::from(data.value.clone())),
+                            )))
+                            .map_err(|_| Error::InternalMsgChannelDropped)?;
+                    }
+                    None => {
+                        // empty response here just to test out
+                        sender
+                            .send(crate::protocol::messages::QueryResponse::GetChunk(Err(
```
oetyng: Bit more readable to move the long qualification to use statenent at top of doc.

safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
...
+                                ProtocolError::ProvideRecordNotFound,
```
oetyng: RecordNotFound could be enough.

(Otherwise, Provided, d at the end missing).

safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
...
+            SwarmCmd::RegisterProvidedData { record } => {
+                debug!("RECORDstarttttt: {:?}", record.key);
+                let _ = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .start_providing(record.key.clone())?;
+
+                debug!("RECORDingg: {:?}", record.key);
+                // TODO: when do we remove records. Do we need to?
```
oetyng: When our range of responsibility has shrunk (network grows) and data we hold fall outside of it, and we are running low on space.

We would need to if we don’t want to restart due to not holding what we should.

safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
...
+            }
+            SwarmCmd::RegisterProvidedData { record } => {
+                debug!("RECORDstarttttt: {:?}", record.key);
+                let _ = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .start_providing(record.key.clone())?;
+
+                debug!("RECORDingg: {:?}", record.key);
```
oetyng: This log messages need some updates.

safenode/src/network/mod.rs
```diff
@@ -141,9 +163,23 @@ impl SwarmDriver {
             .map(|(peer_id, muxer), _| (peer_id, StreamMuxerBox::new(muxer)))
             .boxed();
 
+        // Configure the memory store to be able to hold large records
```
oetyng: large + r

safenode/src/network/mod.rs
```diff
@@ -338,6 +335,24 @@ impl Network {
...
+        self.send_swarm_cmd(SwarmCmd::GetProvidedData { key, sender })
+            .await?;
+
+        receiver
+            .await
+            .map_err(|_e| Error::InternalMsgChannelDropped)
+    }
+
+    /// Register self as Provider for Data
+    pub async fn regigster_as_provider_for_record(&self, record: Record) -> Result<()> {
```
oetyng: regig

safenode/src/node/api.rs
```diff
@@ -186,12 +195,22 @@ impl Node {
...
+                    key: RecordKey::new(chunk.address().name()),
+                    value: chunk.value().to_vec(),
+                    publisher: None,
+                    expires: None,
+                };
+
+                match self.network.regigster_as_provider_for_record(record).await {
+                    Ok(()) => CmdResponse::StoreChunk(Ok(())),
+                    Err(err) => {
+                        error!("Failed to register chunk as provider: {err:?}");
```
oetyng: The chunk is not being registered as provider. The node is.

safenode/src/node/api.rs
```diff
@@ -186,12 +195,22 @@ impl Node {
...
+                        CmdResponse::StoreChunk(Err(ProtocolError::ProvideRecordNotSaved))
```
oetyng: RecordNotSaved should be enough (or for stringency, Stored, since that's what we say in other places).

Otherwise, missing r at end of Provide.

safenode/src/protocol/error.rs
```diff
@@ -19,6 +19,15 @@ pub type Result<T> = result::Result<T, Error>;
 #[derive(Error, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[non_exhaustive]
 pub enum Error {
+    /// There was an internal error while processing the request.
+    #[error("There was an internal error while processing the request")]
+    InternalProcessing(String),
+    /// We failed to send the provider record into the sawrm command handler
```
oetyng: *provided 
*swarm

.github/workflows/merge.yml
```diff
@@ -331,7 +374,7 @@ jobs:
       - name: Upload Node Logs
         uses: actions/upload-artifact@main
         with:
-          name: sn_node_logs_churn_${{matrix.os}}
+          name: sn_node_logs_e2e_${{matrix.os}}
```
maqi: this is a `churn` log?
also need to be differiate with the above `e2e` logs ?

.github/workflows/merge.yml
```diff
@@ -331,7 +374,7 @@ jobs:
...
           path: log_files.tar.gz
         if: failure()
-        continue-on-error: true
+        continue-on-error: true
```
maqi: missing a line break as the end of file.
there is a red circile stop sign shown in the PR file diff

resources/scripts/network_churning.sh
```diff
@@ -12,15 +12,16 @@ nodes_count=$(ls $log_dir | wc -l)
...
     ((count++))
+    target_port=$((12000 + $count))
+
     echo Iteration $count
-    echo Restarting node $count
-    cargo run --release --example safenode_rpc_client -- "127.0.0.1:1200$count" restart 5000
-	sleep 5
+    echo Restarting node on port $target_port
+    cargo run --release --example safenode_rpc_client -- "127.0.0.1:$target_port" restart 1
+	sleep 1
```
maqi, line 26: 1s interval is too aggressive I think.
In case of a dead_node, a `get_clsosest_peers` will take longer time to complete, due to the `dial and wait for outbound error on timed_out` , which may take 5-10s longer.
And replication of data will then take longer to complete.

safenode/src/domain/storage/chunks/mod.rs
```diff
@@ -6,273 +6,68 @@
...
-        // Sync OS data to disk to reduce the chances of
-        // concurrent reading failing by reading an empty/incomplete file.
-        file.sync_data().await?;
-
-        trace!("Stored new chunk {addr:?}");
-
-        Ok(())
+    /// Returns size of this chunk after serialisation.
+    pub fn serialised_size(&self) -> usize {
+        self.value.len()
```
maqi, line 57: this is now same to the `payload_size` ?
is Bytes serialization always equal to Bytes itself?

safenode/src/network/cmd.rs
```diff
@@ -43,11 +50,58 @@ pub enum SwarmCmd {
...
+                                ProtocolError::ProvideRecordNotFound,
+                            )))
+                            .map_err(|_| Error::InternalMsgChannelDropped)?;
+
+                        warn!("No data found in kademlia store for key: {:?}", key);
+                    }
+                }
+            }
+            SwarmCmd::RegisterProvidedData { record } => {
+                debug!("RECORDstarttttt: {:?}", record.key);
```
maqi: correct the wording before merge

safenode/src/network/cmd.rs
joshuef: Indeed, this was more of a "how much of this does kad handle Q". I'll rephrase.

safenode/src/network/cmd.rs
joshuef: 🔪

.github/workflows/merge.yml
joshuef: good spot in my copypasta there

resources/scripts/network_churning.sh
joshuef: i think we can dial this in as we zone in on our desired replication method. At the moment this works, so I'm inclined to leave this for now.

safenode/src/domain/storage/chunks/mod.rs
joshuef: I'm not sure. This was just moved in as I removed the ChunkStorage which was 99% of this file prior. 

The methods in this file have not changed.

safenode/src/network/cmd.rs
joshuef: killed it 👍

safenode/src/network/cmd.rs
```diff
@@ -58,6 +77,38 @@ pub struct SwarmLocalState {
...
+                    .kademlia
+                    .start_providing(record.key.clone())?;
+
+                // TODO: Does Kad automatically remove irrelevant records?
+                // Or do we need to manage this manually as we move about the network?
+                let _ = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .put_record(record, libp2p::kad::Quorum::All)?;
```
RolandSherwin, line 110: Maybe we could remove the provider stuff here and just roll with the PUT/GET flow for now. My understanding is that, to fetch data using the provider API, we should first make a call to `kad::get_providers(key)` and use then contact the peers (providers) for our data. But currently we’re calling `get_closest_peer(key)` and then requesting the data from them directly. And preventing the “provider” nodes from being of any help.

Also, the PUT/GET flow using `.put_record()` is the one responsible for the actual "data" replication. So it might be better to just start with that and add in providers later.

safenode/src/network/cmd.rs
```diff
@@ -58,6 +77,38 @@ pub struct SwarmLocalState {
 impl SwarmDriver {
     pub(crate) fn handle_cmd(&mut self, cmd: SwarmCmd) -> Result<(), Error> {
         match cmd {
+            SwarmCmd::GetData { key, sender } => {
+                match self.swarm.behaviour_mut().kademlia.store_mut().get(&key) {
```
RolandSherwin, line 81: Related to the comment on the `RegisterProvidedData` cmd,
If we're going with just the PUT/GET flow, it might be better to refactor the `GET` flow here. As, calling `kademlia.get_record(key)` starts an iterative query through the network till it finds the data. And I think we can remove `get_closest_peers` that happens prior to a call to this (via `send_to_closest_peers`). 

The only difference between both of the approaches is that, the iterative `get_record()` call short circuits when it finds the first node that holds the data and the data is returned from that node. Whereas, the current flow queries the network till it finds all the `closest_peers`, then request from every single one of them.

safenode/src/network/mod.rs
```diff
@@ -85,8 +89,27 @@ impl SwarmDriver {
...
+        let _ = cfg.set_replication_factor(
+            NonZeroUsize::new(CLOSE_GROUP_SIZE).ok_or_else(|| Error::InvalidCloseGroupSize)?,
+        );
         let _ = cfg.set_query_timeout(Duration::from_secs(5 * 60));
 
+        // Require iterative queries to use disjoint paths for increased resiliency in the presence of potentially adversarial nodes.
+        let _ = cfg.disjoint_query_paths(true);
+
+        // Provider records never expire
+        let _ = cfg.set_provider_record_ttl(None);
```
RolandSherwin, line 111: Think we might also have to set, `cfg.set_record_ttl(None);` this would prevent the "normal" records from expiring.

safenode/src/network/cmd.rs
RolandSherwin: - `start_providing(record.key.clone())?` - produces `ProviderRecords` which are replicated among the `closest_peers(key)`.
- `.put_record(record, libp2p::kad::Quorum::All)?` - produces `Records` which are replicated among the `closest_peers(key)`

maqi: according to the comment at https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/behaviour.rs#L888-L909
seems here is using `both` push and pull models.
which, though not incorrect, but may be bit wasteful

safenode/src/domain/storage/chunks/mod.rs
grumbach: Why don't we serialize it and `len()` it to know?

safenode/src/network/mod.rs
```diff
@@ -139,9 +162,23 @@ impl SwarmDriver {
...
             .boxed();
 
+        // Configures the memory store to be able to hold larger
+        // records than by default
+        let memory_store_cfg = MemoryStoreConfig {
+            max_value_bytes: 1024 * 1024,
+            // What is the difference between these two?
+            max_provided_keys: 1000,
+            max_records: 1000,
+            ..Default::default()
```
grumbach, line 172: As per the docs in the lib's src:
```rs
/// The maximum number of records.
max_records: usize,

/// The maximum number of provider records for which the local node is the provider.
max_provided_keys: usize,
```

A `max_provided_keys` smaller than `max_records` would mean our local node is unable to store all the records itself. 

The default value for those are:
```rs
max_records: 1024,
max_provided_keys: 1024,
```

We could be using those instead of `1000`.

grumbach (approved):
Great one! ✅ 🚢

safenode/src/network/cmd.rs
joshuef: But it's only repubbed if needed. So here we have a lot of logic for free.

It's also worth noting the the standard provider flow doesn't necessarily get to the holders of the data I think? It could be cached or someone else could announce they are a provider and there you go... 

Which can be both good and bad

safenode/src/network/cmd.rs
joshuef: > The only difference between both of the approaches is that, the iterative get_record() call short circuits when it finds the first node that holds the data and the data is returned from that node.

I think this is key, and where we need to decide what behaviour we want. 

I was still aiming to get data from the supposed holders eg. It may be we dont need to... but may be we do?

safenode/src/network/mod.rs
joshuef: It's still not clear to me what a record vs provider record is here. For each record you want to be a provider of it no? 

(makes sense w/r/t 1k btw, yupyup)

safenode/src/network/mod.rs
```diff
@@ -85,8 +89,27 @@ impl SwarmDriver {
...
     pub fn new(addr: SocketAddr) -> Result<(Network, mpsc::Receiver<NetworkEvent>, SwarmDriver)> {
         let mut cfg = KademliaConfig::default();
+
+        // TODO: (make this dynamic?)
+        // repub every 5s as topology changes
+        let _ = cfg.set_replication_interval(Some(Duration::from_secs(5)));
+
+        // 1mb packet size
+        let _ = cfg.set_max_packet_size(1024 * 1024);
+        let _ = cfg.set_max_packet_size(1024 * 1024);
```
bochaco, line 99: it's a duplicated cmd/statement/line?
would this limit be a problem for our 1mb chunks?

PR #178, opened 2023-04-25
iit for wan testnet tool

joshuef:


wan_testnet_tool/up.sh
```diff
@@ -0,0 +1,165 @@
...
+  aws s3 cp \
+    "$network_contacts_path" \
+    "s3://sn-node/testnet_tool/$testnet_channel/network-contacts" \
+    --acl public-read
+}
+
+function pull_genesis_dbc_and_copy_to_s3() {
+  local genesis_ip=$(cat "./workspace/$testnet_channel/genesis-ip")
+  local genesis_dbc_path="./workspace/$testnet_channel/genesis-dbc"
+  echo "Pulling Genesis DBC from Genesis node"
```
oetyng, line 121: This shouldn't be necessary as we can have this from start with source code 👍

oetyng: (And same for the key)

jacderida:
Hi Josh, is the plan for this to replace the `sn_testnet_tool` repo? It's just that I've done a lot of work on the testnet tool for running on AWS. It's a bit different from how it works on DO, but my plan was to change the DO setup to use the same code I'm using for AWS.

I get why it can be useful to be in the same repo, but my suggestion would be to keep this infrastructure separate from the application code.

If we were going to bring it in, I would prefer to use the updated mechanism for AWS rather than the way we've done it for DO. Should I just commit my work to this repo?

Issue #46, opened 2023-04-05
Trigger replication flows when close group changes


joshuef:
We'll need to pull over the replication flows from the safe code (ie the random send of relevant data loop), and update nodes as/when our close nodes change (and so the holders change).

This could be on node join or leave

maqi:
As discussed on the https://github.com/maidsafe/safe_network/pull/172 and https://github.com/maidsafe/safe_network/pull/155
For replication, we could:
1, Use targeted replication, as shown in the comment of PR 155
2, Use kad provider model, which:
   a, store data to kad as record directly
   b, make data holder as Provider, and store ProviderRecord to kad

It seems libp2p can carry out replication using built-in configs via the approach 2, which save lot of our work
And the PR https://github.com/maidsafe/safe_network/pull/179 implemented the approach 2.a .

The approach 2.b requires some extra work to query the data based on fetched ProviderRecord.
Which we might be needed with the grown of data size to be uploaded to the network.

joshuef:
Closing this as using a `recordStore` sorts this

PR #179, opened 2023-04-25
feat: store chunk to network as kad record directly

maqi:
This PR contains the work of store chunk to network as kad record directly.
This allows the data replication to be covered by ligp2p directly.

Closes https://github.com/maidsafe/safe_network/pull/172
Closes https://github.com/maidsafe/safe_network/pull/155

Related to Issue https://github.com/maidsafe/safe_network/issues/46


safenode/src/network/mod.rs
```diff
@@ -296,6 +335,24 @@ impl Network {
...
+        let (sender, receiver) = oneshot::channel();
+        self.send_swarm_cmd(SwarmCmd::GetProvidedData { key, sender })
+            .await?;
+
+        receiver
+            .await
+            .map_err(|_e| Error::InternalMsgChannelDropped)
+    }
+
+    /// Register self as Provider for Data
```
RolandSherwin: Just a small one, the doc needs to be updated here.

safenode/src/node/api.rs
```diff
@@ -200,12 +202,19 @@ impl Node {
         match query {
             Query::Register(query) => self.registers.read(&query, User::Anyone).await,
             Query::GetChunk(address) => {
-                let resp = self
-                    .chunks
-                    .get(&address)
+                match self
+                    .network
+                    .get_provided_data(RecordKey::new(address.name()))
```
RolandSherwin, line 207: So, if I'm not wrong, Client sends a `GetChunk` and nodes fetch the data for it? 
If so, maybe we can directly call `network.swarm.behaviour_mut().kademlia.get_record(key);` from the client. This has the advantage of short-circuiting when the data is found. Might speed up the queries by a bit.

Maybe we can do this in the next PR?

safenode/src/network/cmd.rs
```diff
@@ -74,43 +71,16 @@ impl SwarmDriver {
...
-                            .send(crate::protocol::messages::QueryResponse::GetChunk(Err(
-                                ProtocolError::ProvideRecordNotFound,
-                            )))
-                            .map_err(|_| Error::InternalMsgChannelDropped)?;
-
-                        warn!("No data found in kademlia store for key: {:?}", key);
-                    }
-                }
+                let query_id = self.swarm.behaviour_mut().kademlia.get_record(key);
+                let _ = self.pending_query.insert(query_id, sender);
```
joshuef, line 76: So does this not actually the fire the request away into the network again here? 

I'm not sure we're actually getting data from _this node_?

safenode/src/network/mod.rs
```diff
@@ -85,8 +90,27 @@ impl SwarmDriver {
...
+        let _ = cfg.set_replication_factor(
+            NonZeroUsize::new(CLOSE_GROUP_SIZE).ok_or_else(|| Error::InvalidCloseGroupSize)?,
+        );
         let _ = cfg.set_query_timeout(Duration::from_secs(5 * 60));
 
+        // Require iterative queries to use disjoint paths for increased resiliency in the presence of potentially adversarial nodes.
+        let _ = cfg.disjoint_query_paths(true);
+
+        // Provider records never expire
+        let _ = cfg.set_provider_record_ttl(None);
```
RolandSherwin: I think we might have to set, `cfg.set_record_ttl(None)` to make the normal data `Records` not expire.

safenode/src/network/cmd.rs
maqi: `get_record` will fire the query to network, and a response will be given back later.

safenode/src/node/api.rs
maqi: > we can directly call network.swarm.behaviour_mut().kademlia.get_record(key); from the client

yes, I think we shall let client do this directly

maqi: pushed change that makes client do the `get_record` directly.

RolandSherwin: Hey, sorry, do we still need `Query::GetChunk` and `Cmd::StoreChunk` now that the clients talk with the swarm directly and not the peers?

maqi: we may still need them for the later on usage, say some data exchanged among nodes.
so, I think we can keep them

RolandSherwin (approved):

Issue #173, opened 2023-04-24
Come to a decision how to store the genesis key in source.


oetyng:
The DER/PEM is a normal way to keep keys in a filesystem, but in code I suppose the string works too?

Maybe it shall be easy to share and pass it around? So better in a file then, or?

Selecting the hex string from code, and copy paste is very simple.
But maybe less accessible to entering a file, finding the place where it is defined, than just pick up the key file.

The actual file could perhaps be a release artifact? With our checksum etc.
The validity of it shouldn't be depending on github though... eventually.

oetyng:
Also, if stored in file, should there be a line break at end of it? (Github warns when it doesn't have that.)

PR #180, opened 2023-04-25
feat: use faucet for e2e tests distribution

### LINKED ISSUE #154: Local network using a single faucet client dole out tokens for the concurrent test runs to use.
With #161, #162, #164 and #165, local testnets can be spun up efficiently, with a single faucet client who would dole out tokens for the concurrent test runs to use. If we want to increase throughput and performance, we could spin up (virtually) any number of clients who has a gotten a chunk each of the genesis.

This will allow E2E of the real transfer flows, with an actual genesis, and multiple different tests running concurrently on the same network without issues of having dbcs to use.

oetyng:
Resolves #154.

Implements a (currently ignored) transfer E2E test which successfully carries out transfer from genesis wallet to a faucet wallet.
Subsequent steps of transferring from faucet still hits network issues, to be solved.

This also fixes various bugs around genesis, transfer handling at nodes and clients.

Additionally, there is continued development and improvement of the faucet client logic.


safenode/src/domain/dbc_genesis.rs
```diff
@@ -6,32 +6,43 @@
...
 
+/// Total supply of tokens that will eventually exist in the network: 4,294,967,295 * 10^9 = 4,294,967,295,000,000,000.
+#[cfg(test)]
+const TOTAL_SUPPLY: u64 = u32::MAX as u64 * u64::pow(10, 9);
+
+/// The secret key for the genesis DBC.
+///
+/// This key is public for auditing purposes. Hard coding its value means all nodes will be able to
+/// validate it.
+const GENESIS_DBC_SK: &str = "0c5152498fc5b2f9ed691ef875f2c16f1f950910391f7ba1df63e9f0ce4b2780";
```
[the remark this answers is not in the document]
oetyng: This is a feature, not a bug.
The genesis key shall be public for audit purposes.

safenode/src/client/api.rs
```diff
@@ -210,7 +211,12 @@ impl Client {
             match select_all(list_of_futures).await {
                 (Ok(res), _, remaining_futures) => {
                     let res = res.map_err(Error::Network);
-                    info!("Got response for the req: {req:?}, res: {res:?}");
+                    let res_string = match &res {
```
oetyng, line 346: Bit ugly, but the value must be extracted from the result it seems, for it to not override the implemented display, which is meant to cut down on very large and verbose details.

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,122 @@ impl SendClient for Client {
...
+                })
+                .collect();
+
+            let ok_responses = spends.len();
+            // As to not have a single rogue node deliver a bogus spend,
+            // and thereby have us fail the check here
+            // (we would have more than 1 spend in the BTreeSet), we must
+            // look for a majority of the same responses, and ignore any other responses.
+            if ok_responses >= 1 {
+                //close_group_majority() {
```
oetyng: oops, remaining from bug hunt, will fix

oetyng: fixed

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,122 @@ impl SendClient for Client {
...
+                // Majority of nodes in the close group returned an Ok response.
+                use itertools::*;
+                if let Some(spend) = spends
```
oetyng: This logic exists in other place as well. Would be good to reuse it eventually.

safenode/src/domain/dbc_genesis.rs
```diff
@@ -43,25 +54,156 @@ pub enum Error {
...
+    let faucet_balance = faucet_wallet.balance();
+    if faucet_balance.as_nano() > 0 {
+        println!("Faucet wallet balance: {faucet_balance}");
+        return faucet_wallet;
+    }
+
+    // Transfer to faucet. We will transfer almost all of the genesis wallet's
+    // balance to the faucet, only leaving enough to pay for transfer fees.
+
+    let initial_fee_margin = 500_000;
```
oetyng, line 123: This will be fixed later, so that we actually transfer everything from genesis.

oetyng: (And there will be solution to the live transfers from first faucet so that there is no issue with transfer fees.)

safenode/src/domain/dbc_genesis.rs
```diff
@@ -146,3 +288,34 @@ pub(super) fn split(
...
+        .expect("Genesis test path to be successfully created.");
+    home_dirs
+}
+
+// We need deterministic and fix path for the faucet wallet.
+// Otherwise the test instances will not be able to find the same faucet instance.
+async fn get_faucet_dir() -> PathBuf {
+    let mut home_dirs = dirs_next::home_dir().expect("A homedir to exist.");
+    home_dirs.push(".safe");
+    home_dirs.push("test_faucet");
```
oetyng, line 316: This structure will be improved upon if/when we want multiple parallel faucets.

safenode/src/domain/dbc_genesis.rs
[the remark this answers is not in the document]
joshuef: I think we can ignore these devskim warnings here in the code

safenode/src/client/api.rs
```diff
@@ -165,11 +165,12 @@ impl Client {
     }
 
     pub(crate) async fn send_to_closest(&self, request: Request) -> Result<Vec<Result<Response>>> {
-        info!("Sending {:?} to the closest peers.", request.dst());
+        println!("Getting the closest peers to {:?}.", request.dst());
```
joshuef, line 183: Is this change from logging to println intentional to keep ? (same as below).

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,121 @@ impl SendClient for Client {
...
     }
 }
+
+#[async_trait::async_trait]
+impl VerifyingClient for Client {
+    async fn verify(&self, dbc: &Dbc) -> Result<()> {
+        let mut received_spends = std::collections::BTreeSet::new();
+
+        for spend in &dbc.signed_spends {
+            let address = dbc_address(spend.dbc_id());
```
joshuef: This fn seems quite big. Can we split apart the logic into some smaller fns for readability please?

Perhaps just a `verify_spend` function and `verify_response`?

safenode/src/client/api.rs
oetyng: Yes. The logging was previously removed from client so none of this was visible.

safenode/src/client/wallet.rs
oetyng: It's not that simple actually, it get's more complicated to try split this up. And makes it harder to reason about IMO.

Look closer at the logic (or try it) and you will see the same.

Is this absolutely necessary now?

oetyng: I'm moving the VerifyingClient to it's own file, and splitting out the block in the loop to own `fn verify_spend`.
This should help with focus when reading that.

resources/GENESIS.key
```diff
@@ -0,0 +1 @@
+0c5152498fc5b2f9ed691ef875f2c16f1f950910391f7ba1df63e9f0ce4b2780
```
maqi, line 1: may not be critical, but seems git complains the file with `lacking a line break at the end of file` with a red stop sign
same to the other file of GENESIS.dbc

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,121 @@ impl SendClient for Client {
...
+                .flatten()
+                .flat_map(|resp| {
+                    if let Response::Cmd(CmdResponse::Spend(Ok(()))) = resp {
+                        Some(())
+                    } else {
+                        println!("Spend error {resp:?}.");
+                        None
+                    }
+                })
+                .collect();
```
grumbach: Can't we just .count() here?

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,121 @@ impl SendClient for Client {
...
+        let mut received_spends = std::collections::BTreeSet::new();
+
+        for spend in &dbc.signed_spends {
+            let address = dbc_address(spend.dbc_id());
+            let query = Query::Spend(SpendQuery::GetDbcSpend(address));
+
+            let responses = self
+                .send_to_closest(Request::Query(query))
+                .await
+                .map_err(|err| Error::CouldNotVerifyTransfer(err.to_string()))?;
```
maqi: for each signed_spend, we will have to wait for ALL the responses returned then carry to the next
this may explains why the process is slow?
can they be carried out in parallel?

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,121 @@ impl SendClient for Client {
...
+
+            // Get all Ok results of the expected response type `GetDbcSpend`.
```
maqi: this seems a code block (including the majority check part) with above?
also, seems for the `send_to_closest` , it doesn't need to wait_till_get_all_responses, but can return_early_when_get_majority_ok_responses, which can avoid a single no-response node hold back the whole process till time_out?

safenode/src/client/wallet.rs
```diff
@@ -73,14 +75,121 @@ impl SendClient for Client {
...
+                    .max_by_key(|(_, v)| v.len())
+                    .map(|(k, _)| k)
+                {
+                    // Majority of nodes in the close group returned the same spend.
+                    let _ = received_spends.insert(spend);
+                    continue;
+                }
+            }
+
+            // The parent is not recognised by all peers in its close group.
```
maqi: `all` or `majority` ?

safenode/src/client/wallet.rs
oetyng: And clarified what we are actually getting from the network.

safenode/src/domain/dbc_genesis.rs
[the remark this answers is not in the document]
maqi: This is now stored in a file. can be removed?

safenode/src/domain/dbc_genesis.rs
```diff
@@ -6,32 +6,43 @@
...
...deda5a7cb56abb1b68c073ffc58d623d5667ae8ea981642fb64ea4cd1d4f64b5f1a00df9c162ac9a5dabee87f1c5103f83af3e04bbeb371222a7a472397387e76e72944f645a0859c79004e1d904da04b6760252dc710168a67e8c43ce15cfd71bbc460f67bb15e43fb3da670ca3bb01cb133d1e49da4c3daf12b9b6692d3d776ca0d251f064a83d215b2c60c893e800e9775a92497e2b8ae6a4bb94da8a1c7cc3acaa492ed0c5d40a0011df334b4139f9dc4deb937e90ad0d535bdd7b334e02756de49890a4c4c80429965c9e3ac8344e6f84f1f326dedf91f1cf90a722cb16f4a81f05ca4515fb27d4eaf76fe5d82f175f9ab6733a3ede04cc0f9913afe836b006b4490e030bbbe4afb72e6078104a4624831937a32fe291a894d4d6a372d655c5a82116a38df751906789adc68000000000000002a0dd319adfeeb6cb46acdb6ee72aa7d0f82f3a0bab3f0409279f7289d4d35c431c2c8fa15fa24b703135308ecf72a71396000000000000000177a2315fc70b8072892cc6b32002606b9313f0955d2479e67ed79ee14cd8e6f3cb3b77f3d07f1028945eedfedea47204239096d8af7064a07fd1a98c6fb29e0ddf966a63837d37be71a3651ad385fb40ef8ac8d74ed524f0bf2367008062a79318a465ddf34b54fa622d570c68df841d26d321c55313342ea9aeadffb01bd170dd319adfeeb6cb46acdb6ee72aa7d0f82f3a0bab3f0409279f7289d4d35c431c2c8fa15fa24b703135308ecf72a713960000000000000001dd319adfeeb6cb46acdb6ee72aa7d0f82f3a0bab3f0409279f7289d4d35c431c2c8fa15fa24b703135308ecf72a71396";
```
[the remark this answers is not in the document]
maqi: This is now stored in a file. can be removed?

safenode/src/client/wallet.rs
oetyng: fixed

safenode/src/client/wallet.rs
oetyng: That would be great yes.
Thanks for pointing it out. Will look at that right away!

safenode/src/client/wallet.rs
oetyng: will fix

oetyng: fixed

safenode/src/client/wallet.rs
oetyng: Looking to see what can be done there.

safenode/src/client/wallet.rs
oetyng: use of VerifyingClient was really fast though. 56 ms.
But will be different when not local network..

safenode/src/client/wallet.rs
oetyng: Yes, this would need another low level api for client though, instead of `send_to_closest` (`send_and_get_responses` don't have the correct beahviour either).

I will implement that then.

safenode/src/client/wallet.rs
oetyng: fixed

safenode/src/domain/dbc_genesis.rs
```diff
@@ -43,25 +54,156 @@ pub enum Error {
...
+    /// a source transaction, as there was nothing before it.
+    pub static ref GENESIS_DBC: Dbc = match sn_dbc::Dbc::from_hex(GENESIS_DBC_HEX) {
+        Ok(dbc) => dbc,
+        Err(err) => panic!("Failed to read genesis DBC: {err:?}"),
+    };
+}
+
+/// Return if provided DbcTransaction is genesis parent tx.
+pub(crate) fn is_genesis_parent_tx(parent_tx: &DbcTransaction) -> bool {
+    parent_tx == &GENESIS_DBC.src_tx
```
grumbach, line 69: I find it weird to have a genesis tx, with no inputs right?

grumbach (commented):
Generally looks good to me, will probably have to think about how we want to represent genesis checks, having a genesis tx with no inputs seems forced uppon us by the current dbc design, is there any other way?

safenode/src/client/wallet.rs
oetyng: > seems for the send_to_closest , it doesn't need to wait_till_get_all_responses, but can return_early_when_get_majority_ok_responses,

Actually, we can't return early with majority of Ok responses, because we need majority of _same item_ in those Ok responses.

Seems it will be a bit of logic leak if going into client code and have it wait for majority of signed_spend being same..?

safenode/src/domain/dbc_genesis.rs
oetyng: It actually has one input. But an input that doesn't exist in the network.

oetyng:
> Generally looks good to me, will probably have to think about how we want to represent genesis checks, having a genesis tx with no inputs seems forced uppon us by the current dbc design, is there any other way?

The genesis actually has an input, but it's an inpout that doesn't exist in the network, and it has the same id as the genesis id, and that is what I find a bit weird. 
We could give the pre-genesis "dbc" (it never existed as a dbc, only a key and an amount) a fix key (thus id) which is not the genesis id. I feel that would make a bit more sense at least.

Getting around that the pre-genesis is invalid seems not doable.

safenode/src/client/wallet.rs
oetyng: fixed

safenode/src/client/wallet.rs
oetyng: We need the actual signed spend returned, so we can compare it to the one(s) in the dbcs to validate.

oetyng: This has been moved to client api.

resources/GENESIS.key
oetyng: I'm not sure it should have that?

maqi (approved):

safenode/src/domain/dbc_genesis.rs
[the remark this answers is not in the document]
oetyng: We have this as an issue #173 in wait for discussions and a decision.

resources/GENESIS.key
oetyng: This is also added as part of issue #173.

PR #182, opened 2023-04-26
remove timeout from `send_to_closest` + send to `self`

### LINKED ISSUE #193: Fix bug: nodes deadlocking when self among close group recipients of msg
This is because the msg is sent over the network to itself, which causes `deadlock` that jamms all comms for the node.

Instead use some sort of backchannel at the path of queries and cmds, so that the node doesn't send the msg to self over the network, but handles it locally at once, and then includes that response with the responses of the others.

Like following pseudo code:
```
if close_group.contains(self) {
     close_group.remove(self);
     let self_response = local_code.send(msg);
     let group_response = network.send(msg, close_group);
     return group_response.concat(self_response);
}
```

RolandSherwin:
fixes #193 
- fix(network): get our `PeerId` from Network
    - use `network.peer_id` instead of storing it around in `PendingGetClosest`
- fix(network): remove timeout from `send_to_closest`
    - The request_response Behaviour contains an inbuilt timeout. Hence, remove our custom timeout implementation.
- feat(network): set timeout through the `RequestResponse` behaviour
- fix(network): route `Request` and `Response` to self 
	- While using the `RequestResponse` behaviour, we get a `OutboundFailure::DialFailure`
if we try to send a request to `self`
	- So if `self` is the recipient of the `Request`, then route the request
  directly to `self` without using the `RequestResponse` behaviour.
	- This request, then follows the normal flow without having any custom
  branch on the upper layers. The produced `Response` is also routed
  back to `self`
- fix(test): initialize logging once for unit tests
	- else the test panics as we initialize multiple `Subscriber` instances, at multiple sub-processes when we call `cargo test`

### DIFF
--- a/safenode/src/network/mod.rs
+++ b/safenode/src/network/mod.rs
@@ -30,7 +30,9 @@ use libp2p::{
     },
     mdns,
     multiaddr::Protocol,
-    request_response::{self, ProtocolSupport, RequestId, ResponseChannel},
+    request_response::{
+        self, Config as RequestResponseConfig, ProtocolSupport, RequestId, ResponseChannel,
+    },
     swarm::{Swarm, SwarmBuilder},
     Multiaddr, PeerId, Transport,
 };
@@ -49,6 +51,11 @@ use xor_name::XorName;
 /// This is the group size used in safe network protocol to be responsible for
 /// an item in the network.
 pub(crate) const CLOSE_GROUP_SIZE: usize = 8;
+
+// Timeout for requests sent/received through the request_response behaviour.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+// Sets the keep-alive timeout of idle connections.
+const CONNECTION_KEEP_ALIVE_TIMEOUT: Duration = Duration::from_secs(10);
 
 /// Majority of a given group (i.e. > 1/2).
 #[inline]
@@ -75,7 +82,7 @@ impl SwarmDriver {
     /// Creates a new `SwarmDriver` instance, along with a `Network` handle
     /// for sending commands and an `mpsc::Receiver<NetworkEvent>` for receiving
     /// network events. It initializes the swarm, sets up the transport, and
-    /// configures the Kademlia and mDNS behaviors for peer discovery.
+    /// configures the Kademlia and mDNS behaviours for peer discovery.
     ///
     /// # Returns
     ///
@@ -84,10 +91,10 @@ impl SwarmDriver {
     ///
     /// # Errors
     ///
-    /// Returns an error if there is a problem initializing the mDNS behavior.
+    /// Returns an error if there is a problem initializing the mDNS behaviour.
     pub fn new(addr: SocketAddr) -> Result<(Network, mpsc::Receiver<NetworkEvent>, SwarmDriver)> {
-        let mut cfg = KademliaConfig::default();
-        let _ = cfg
+        let mut kad_cfg = KademliaConfig::default();
+        let _ = kad_cfg
             // how often a node will replicate records that it has stored, aka copying the key-value pair to other nodes
             // this is a heavier operation than publication, so it is done less frequently
             .set_replication_interval(Some(Duration::from_secs(20)))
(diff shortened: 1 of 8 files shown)


safenode/src/network/mod.rs
```diff
@@ -145,26 +144,31 @@ impl SwarmDriver {
...
         let peer_id = PeerId::from(keypair.public());
 
         info!("Node (PID: {}) with PeerId: {peer_id}", std::process::id());
 
+        // RequestResponse configuration
+        let mut req_res_config = RequestResponseConfig::default();
+        let _ = req_res_config.set_request_timeout(REQUEST_TIMEOUT);
+        let _ = req_res_config.set_connection_keep_alive(CONNECTION_KEEP_ALIVE_TIMEOUT);
+
+        let request_response = request_response::Behaviour::new(
```
oetyng, line 170: Shouldn't this variable be called `behaviour`?
`request_response` is a very fuzzy name. It is a `Behaviour` instance, right?

safenode/src/network/mod.rs
```diff
@@ -296,50 +300,31 @@ impl Network {
...
-        if CLOSE_GROUP_SIZE > closest_peers.len() {
-            warn!("Not enough peers in the k-bucket to satisfy the request");
-            return Err(Error::NotEnoughPeers);
-        }
-
-        trace!(
-            "Got the {} closest_peers to the given XorName-{xor_name}, nodes: {closest_peers:?}",
-            closest_peers.len()
+    /// Send `Request` to the closest peers. `Self` might be present among the recipients.
+    pub async fn node_send_to_closest(&self, request: &Request) -> Result<Vec<Result<Response>>> {
```
oetyng, line 311: It would be nice to have traits for shared objects such as Network, that don't expose apis that are not intended for certain layers of the application, such as client or node.

oetyng: So when as a client using the network it won't list all apis that shouldn't be used by a client, only those that are supposed to be used by it.

safenode/src/network/mod.rs
```diff
@@ -296,50 +300,31 @@ impl Network {
...
+        info!(
+            "Sending {request:?} with dst {:?} to the closest peers.",
+            request.dst().name()
         );
+        // todo: if `self` is present among the closest peers, the request should be routed to self?
+        let closest_peers = self.node_get_closest_peers(*request.dst().name()).await?;
```
oetyng, line 316: I wonder if we quite know what happens with messages sent to ourselves?

Can we setup a test for it?

Just to confirm that we can send any number of msgs to ourselves and it will just flow through without any deadlock or funny loops or such.

safenode/src/network/mod.rs
RolandSherwin: I've renamed it to `req_res_behaviour` as we have multiple structs implementing the `Behaviour` trait.

RolandSherwin: I think we have a couple of other structs that implement the `Behaviour` trait, and they're just called by their crate/struct name, e.g. `kademlia`, `mdns`. So I think `request_response` fits well in this context? 

The variable names are self-explanatory inside this function, but adding `_behaviour` to them is also an option.

safenode/src/network/mod.rs
RolandSherwin: Yeah, that would be great and would lead to some concise function names. Do you mind if I create an issue for it and solve it on a follow-up PR?

safenode/src/network/mod.rs
RolandSherwin: > Just to confirm that we can send any number of msgs to ourselves and it will just flow through without any deadlock or funny loops or such.

The msg to `self` indeed causes an issue, as confirmed by a test. I have updated the PR with a fix for it.

safenode/src/network/mod.rs
oetyng: Yeah, will mull this over, later.. :)

safenode/src/network/mod.rs
oetyng: Please do! Thanks!

oetyng (approved):
Nice one!!

safenode/src/network/mod.rs
RolandSherwin: Raise issue https://github.com/maidsafe/safe_network/issues/203!

maqi (dismissed):

oetyng (changes requested):
This PR is unfortunately not solving the node many to many msg issue. Or perhaps it is but causing other issues instead. Hard to say, but network is even less stable than before, unfortunately :(

oetyng (approved):
Ok, test changes have been made, and the test passes. Unless any other tests fail, let's get this in.

And I'll deal with the effects for transfer when we cross that bridge (right up after this actually..).

PR #185, opened 2023-04-26
fix: make client cfg consistent with node

maqi:
Make client cfg consistent with node
The previous client cfg using default packet size of 16k,
which actually blocks any large chunk (hence large files) to be uploaded to the network.


oetyng (approved):
LGTM

PR #196, opened 2023-04-27
Draft: temp disable double spend detection

oetyng:
(On top of #195.)

This will prevent killing a dbc on accidental resend.
This will be disabled as soon as client wallet is refactored to mark dbcs as spend before publishing it, and store the actual spend for later re-publish at _any_ time in the future.


oetyng:
Superseded.

PR #197, opened 2023-04-27
feat: have nodes broadcast valid spends

### LINKED ISSUE #198: Node to broadcast valid spend upon receival.
This significantly decreases the chances of incomplete transmission of the spend to the close group.

oetyng:
Resolves #198.

As nodes broadcast a valid spend, it is very unlikely that it will only sit at one or a few nodes, due to client connection issues. It will most likely either go through in full, or not at all.

Also there is a temp disabling of _punishment_ on double spent detection.
This will prevent killing a dbc on accidental resend.
The detection is still there and hinders double spends, but the _punishment_ is disabled.

This can/will be enabled again as soon as client wallet is refactored to mark dbcs as spent before publishing it, and store the actual spend for later re-publish at any time in the future.

### DIFF
--- a/safenode/src/domain/storage/spends.rs
+++ b/safenode/src/domain/storage/spends.rs
@@ -347,6 +347,7 @@ mod tests {
             .expect("The exact same spend should be added.");
     }
 
+    #[ignore = "We have temporarily disabled the punishment of double spends. NB it is still detected and hindered, just not punished."]
     #[tokio::test]
     async fn double_spend_attempt_is_detected() {
         let mut storage = init_file_store();
--- a/safenode/src/domain/storage/spends.rs
+++ b/safenode/src/domain/storage/spends.rs
@@ -101,20 +101,22 @@ impl SpendStorage {
         if let Ok(existing) = self.get(&address).await {
             let tamper_attempted = signed_spend.spend.hash() != existing.spend.hash();
             if tamper_attempted {
-                // We don't error if the double spend failed, as we rather want to
-                // announce the double spend attempt to close group. TODO: how to handle the error then?
-                self.try_store_double_spend(&existing, signed_spend).await?;
+                trace!("Tamper attempt detected, jsut rejecting the request.");
+                return Err(Error::AlreadyExists(address));
+                // // We don't error if the double spend failed, as we rather want to
+                // // announce the double spend attempt to close group. TODO: how to handle the error then?
+                // self.try_store_double_spend(&existing, signed_spend).await?;
 
-                // The spend is now permanently removed from the valid spends.
-                // We don't error if the remove failed, as we rather want to
-                // announce the double spend attempt to close group.
-                // The double spend will still be detected by querying for the spend.
-                self.remove(&address, &self.valid_spends_path).await?;
+                // // The spend is now permanently removed from the valid spends.
+                // // We don't error if the remove failed, as we rather want to
+                // // announce the double spend attempt to close group.
+                // // The double spend will still be detected by querying for the spend.
+                // self.remove(&address, &self.valid_spends_path).await?;
 
-                return Err(Error::DoubleSpendAttempt {
-                    new: Box::new(signed_spend.clone()),
-                    existing: Box::new(existing),
-                });
+                // return Err(Error::DoubleSpendAttempt {
+                //     new: Box::new(signed_spend.clone()),
+                //     existing: Box::new(existing),
+                // });
             }
         }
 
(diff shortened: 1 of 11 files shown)


safenode/src/domain/storage/spends.rs
```diff
@@ -101,20 +101,22 @@ impl SpendStorage {
...
-                // We don't error if the remove failed, as we rather want to
-                // announce the double spend attempt to close group.
-                // The double spend will still be detected by querying for the spend.
-                self.remove(&address, &self.valid_spends_path).await?;
-
-                return Err(Error::DoubleSpendAttempt {
-                    new: Box::new(signed_spend.clone()),
-                    existing: Box::new(existing),
-                });
+                trace!("Tamper attempt detected, jsut rejecting the request.");
```
maqi: just

safenode/src/domain/storage/spends.rs
```diff
@@ -101,20 +101,22 @@ impl SpendStorage {
...
+                return Err(Error::AlreadyExists(address));
+                // // We don't error if the double spend failed, as we rather want to
```
maqi, line 106: hmm... so the following block is then OK to be removed?

safenode/src/domain/storage/spends.rs
```diff
@@ -101,20 +101,22 @@ impl SpendStorage {
...
-                // announce the double spend attempt to close group.
-                // The double spend will still be detected by querying for the spend.
-                self.remove(&address, &self.valid_spends_path).await?;
-
-                return Err(Error::DoubleSpendAttempt {
-                    new: Box::new(signed_spend.clone()),
-                    existing: Box::new(existing),
-                });
+                trace!("Tamper attempt detected, jsut rejecting the request.");
+                return Err(Error::AlreadyExists(address));
```
maqi, line 105: no longer `DoubleSpendAttempt` error to be returned, 
which will trigger a further handling within the caller?
https://github.com/maidsafe/safe_network/pull/197/files#diff-ae93be614203f23a4b33e38e2ac1aafcf961ae7d4e6ac9485fae33eb475e0938R326

safenode/src/node/api.rs
```diff
@@ -285,9 +297,32 @@ impl Node {
...
                 {
+                    Ok(()) => {
+                        warn!("Broadcasting valid spend: {:?}", signed_spend.dbc_id());
+                        let event = Event::ValidSpendReceived {
+                            spend: signed_spend,
+                            parent_tx,
+                            fee_ciphers,
+                            parent_spends,
+                        };
+                        match self.send_to_closest(&Request::Event(event)).await {
```
maqi: First, `send_to_closest` actually wait on response by default (and even wait on ALL responses)
Second, request/response model itself also indicated a response to be expected on sending out a request

Both, as I understand, not required for a `broadcasting` here?
i.e. here actually we can just spawn a thread to send the stuff out to avoid potential slow_down or even blocking ?

safenode/src/domain/storage/spends.rs
oetyng: Yes.
There can still be no double spend. The spend is rejected.

safenode/src/domain/storage/spends.rs
oetyng: Yes, it is disabled for now.

If it exists it exists. The client can tryt double spend but it will always be rejected, because we have the state of the first spend.

The double spend state is a way to _punish_ the double spending client, not to _hinder_ it. It is already hindered.

safenode/src/node/api.rs
oetyng: Yeah, absolutely. Is there an api for that?

safenode/src/domain/storage/spends.rs
oetyng: fixed

safenode/src/node/api.rs
oetyng: If we don't have an api for that, I would prefer that we raise an issue and a PR for extending the client with that, and then refactor this method here to use it.

safenode/src/domain/storage/spends.rs
maqi: then shall the block got `removed` instead of just `commented out` ?

safenode/src/node/api.rs
maqi: > Is there an api for that?

I don't know there is an existing one.
But, you can have a `spawned thread` , like here https://github.com/maidsafe/safe_network/blob/main/safenode/src/node/api.rs#L142-L146
to have a quick one to let it not being a blocker/slower

maqi (approved):

safenode/src/domain/storage/spends.rs
oetyng: Hm. I think we want the code still, just disabled for now. In such cases we usually keep the code commented out.

safenode/src/node/api.rs
oetyng: Issue raised for that improvement :+1:

PR #201, opened 2023-04-27
fix(network): add peers who dialed to us to the RoutingTable

RolandSherwin:


safenode/src/network/event.rs
```diff
@@ -199,11 +199,24 @@ impl SwarmDriver {
             SwarmEvent::ConnectionEstablished {
                 peer_id, endpoint, ..
             } => {
+                debug!(
+                    "Connection established with {peer_id:?}. We are the dialer {:?}",
```
maqi: Endpoint could be Listener as well?

safenode/src/network/event.rs
```diff
@@ -199,11 +199,24 @@ impl SwarmDriver {
...
                         let _ = sender.send(Ok(()));
                     }
+                } else {
+                    // If the connection was opened by someone else / relayed to us,
+                    // then add the peer to the routing table.
+                    let _routing_update = self
+                        .swarm
+                        .behaviour_mut()
+                        .kademlia
+                        .add_address(&peer_id, endpoint.get_remote_address().clone());
```
maqi, line 215: In theory, if this is a relayed connection, then we shall able to dial back using the `endpoint.get_remote_address()` ?

So, here we shall try a dial back, and if the dial succeeded, we then can add peer into RT.
instead of adding it to RT directly ?

safenode/src/network/event.rs
RolandSherwin: Aye, it can be a listener too. Will modify the log there.

safenode/src/network/event.rs
RolandSherwin: Do we have to manually dial back in case of a relayed connection, or is it automatically connected? I'm not really sure, I've added a todo here.

oetyng (approved):

oetyng (changes requested):
Something's not right:

```
======== Verifying testnet nodes ========
Checking nodes log files to verify all (25) nodes have joined. Logs path: /home/runner/.safe/node/local-test-network
Number of nodes found in logs: 26
The application panicked (crashed).
Message:  assertion failed: `(left == right)`
  left: `25`,
 right: `26`: Unexpected number of joined nodes. Expected 25, we have 26
Location: sn_testnet/src/check_testnet.rs:77
```

safenode/src/network/event.rs
```diff
@@ -200,10 +200,24 @@ impl SwarmDriver {
...
+                    // If the connection was opened by someone else, add them
+                    // to the routing table.
+                    let _routing_update = self
+                        .swarm
+                        .behaviour_mut()
+                        .kademlia
+                        .add_address(&peer_id, endpoint.get_remote_address().clone());
+
+                    self.event_sender.send(NetworkEvent::PeerAdded).await?;
+                } else if endpoint.is_relayed() {
```
b-zee, line 218: This will never be reached, as I assume either `is_listener` or `is_dialer` is true.

oetyng:
> Something's not right:
> 
> ```
> ======== Verifying testnet nodes ========
> Checking nodes log files to verify all (25) nodes have joined. Logs path: /home/runner/.safe/node/local-test-network
> Number of nodes found in logs: 26
> The application panicked (crashed).
> Message:  assertion failed: `(left == right)`
>   left: `25`,
>  right: `26`: Unexpected number of joined nodes. Expected 25, we have 26
> Location: sn_testnet/src/check_testnet.rs:77
> ```

I found that this issue is in other branches as well, so not caused by this PR.
There seems to be a ghost node of pid 0 added.
A simple fix is to not add a pid 0 node to the list we are checking len of.

joshuef:
Closing as superseded by #224

PR #202, opened 2023-04-27
refactor: wallet logic to client state first

### LINKED ISSUE #199: Refactor wallet to client-first state and spend idempotency.
When the client spends a dbc, ie signs the tx, the dbc must be marked locally as spent (ie pending). Only then should the client broadcast it.

The client stores the tx as pending until either a) all nodes respond with spent so the client locally changes it from pending to spent or b) no nodes respond with spent so the client locally changes it to unspent.

The best heuristic here is clients are in charge of their state, and the network is the source of truth for the state. If there’s ever a conflict in those states, the client can update their local state. Clients create events (are in charge), nodes store events (are source of truth).

The bitcoin flow here is very useful: unspent, unconfirmed (in mempool), confirmed. These three states are held by both the client and the node, and is easy for the client to check and resolve.

(The most difficult situation for a bitcoin client to resolve is a low-fee tx in mempool for a long time, which eventually clears from the mempool and becomes spendable again.

That mempool exists for us as well, as the `priority queue`. Fortunately though, we our clients can actually update it, so a spend is never stuck for a client! 

Another level of equivalency for us could be a tx with too few nodes storing it, and eventually is not stored anywhere, which should be extremely rare, we can expect it never happens.
But with our idempotent republish, it's even less of a risk.).

***

The client had the wrong order. It should 
1. Sign the tx.
2. Locally mark the dbc as spent. 
3. Broadcast the signed tx.

The client won’t try to spend it again, and will just retry until the tx is confirmed.
The client replays their pending txs, as to not get stuck with change dbcs not working due to parents not yet registered.
(Later it can readd non-spent dbcs to available again, if it turns out to be completely non-existent on the network, buit that's an advanced case).

It should be fine for a client to rebroadcast the same tx as many times as they want (with as long a gap as they want, maybe years later). 
Clients must manage the state, if they see it’s stuck then just replay the tx same as if it’s the first time.

oetyng:
Resolves #199.

The client had the wrong order. It should

1. Sign the tx.
2. Locally mark the dbc as spent.
3. Broadcast the signed tx.

The client won’t try to spend the dbc(s) again, sonce they are already marked as spent. It will however retry them if they are in the `pending_txs` list until the tx is confirmed.

The client replays their pending_txs on any send of tokens, as to not block those sends, by getting stuck with change dbcs not working due to parents not yet registered.
(Later it can readd non-spent dbcs to available again, if it turns out to be completely non-existent on the network, buit that's an advanced case).

It should be fine for a client to rebroadcast the same tx as many times as they want (with as long a gap as they want, maybe years later).
Clients must manage the state, if they see it’s stuck then just replay the tx same as if it’s the first time.

### DIFF
--- a/safenode/src/domain/wallet/local_store.rs
+++ b/safenode/src/domain/wallet/local_store.rs
@@ -101,6 +101,7 @@ impl KeyLessWallet {
             spent_dbcs: BTreeMap::new(),
             available_dbcs: BTreeMap::new(),
             dbcs_created_for_others: vec![],
+            unconfirmed_spends: BTreeMap::new(),
         }
     }
 
@@ -202,12 +203,17 @@ impl SendWallet for LocalWallet {
             }
         }
 
+        let transfer = client
+            .create_transfer(available_dbcs, to, self.address())
+            .await?;
+
         let TransferDetails {
             change_dbc,
             created_dbcs,
             ..
-        } = client.send(available_dbcs, to, self.address()).await?;
+        } = transfer.clone();
 
+        // First of all, update client local state.
         let spent_dbc_ids: BTreeSet<_> = created_dbcs
             .iter()
             .flat_map(|created| &created.dbc.signed_spends)
@@ -224,6 +230,9 @@ impl SendWallet for LocalWallet {
         self.wallet
             .dbcs_created_for_others
             .extend(created_dbcs.clone());
+
+        // Last of all, register the spend in the network.
+        let _result = client.send(transfer).await;
 
         Ok(created_dbcs)
     }
@@ -635,7 +644,7 @@ mod tests {
 
     #[async_trait::async_trait]
     impl SendClient for MockSendClient {
-        async fn send(
+        async fn create_transfer(
             &self,
             dbcs: Vec<(Dbc, DerivedKey)>,
             to: Vec<(Token, DbcIdSource)>,
@@ -647,6 +656,11 @@ mod tests {
                 .expect("There should be no issues creating this transfer.");
 
             Ok(transfer)
+        }
+        async fn send(&self, _transfer: TransferDetails) -> super::Result<()> {
+            // Here we just return Ok(()), without network calls,
+            // and without sending it to the network.
+            Ok(())
         }
     }
 
--- a/safenode/src/domain/wallet/local_store.rs
+++ b/safenode/src/domain/wallet/local_store.rs
@@ -245,10 +245,7 @@ impl SendWallet for LocalWallet {
         // Last of all, register the spend in the network.
         if let Err(error) = client.send(transfer.clone()).await {
             trace!("The transfer was not successfully registered in the network: {error:?}. It will be retried later.");
-            let _ = self
-                .wallet
-                .unconfirmed_txs
-                .push((transfer.tx_hash, transfer));
+            let _ = self.wallet.unconfirmed_txs.push(transfer);
         }
 
         Ok(created_dbcs)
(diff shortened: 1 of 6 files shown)


safenode/src/domain/wallet/local_store.rs
```diff
@@ -177,6 +178,22 @@ impl SendWallet for LocalWallet {
         to: Vec<(Token, PublicAddress)>,
         client: &C,
     ) -> Result<Vec<CreatedDbc>> {
+        // First resend any pending txs.
```
maqi: the line width is bit too narrow?
do we still follow the 80 or 100 width giude line?

safenode/src/domain/wallet/local_store.rs
```diff
@@ -225,10 +247,36 @@ impl SendWallet for LocalWallet {
             .dbcs_created_for_others
             .extend(created_dbcs.clone());
 
+        // Last of all, register the spend in the network.
+        let result = client.send(transfer.clone()).await;
```
maqi: make the blow code block to
```rust
if let Err(err) = client.send(transfer.clone()).await {
    ...
}
```
will be cleaner?

safenode/src/domain/wallet/local_store.rs
```diff
@@ -225,10 +247,36 @@ impl SendWallet for LocalWallet {
...
+
+        if result.is_err() {
+            println!("The transfer was not successfully registered in the network. It will be retried later.");
```
maqi: use `error`,
or `error` followed by a line of `println` if do want the user always see this

safenode/src/domain/wallet/local_store.rs
```diff
@@ -225,10 +247,36 @@ impl SendWallet for LocalWallet {
...
+        }
+
         Ok(created_dbcs)
     }
 }
 
+async fn resend_pending_txs<C: SendClient>(local: &mut LocalWallet, client: &C) {
+    for (index, (tx_hash, transfer)) in local.wallet.unconfirmed_txs.clone().into_iter().enumerate()
+    {
+        println!("Trying to republish pending tx: {tx_hash:?}..");
```
maqi: use trace instead ?

safenode/src/domain/wallet/local_store.rs
```diff
@@ -225,10 +247,36 @@ impl SendWallet for LocalWallet {
...
+        if client.send(transfer.clone()).await.is_ok() {
+            println!("Tx {tx_hash:?} was successfully republished!");
```
maqi: trace ?

safenode/src/domain/wallet/local_store.rs
oetyng: Good question.. I saw it as a line break at that particular place though, I think..

oetyng: But very narrow generally there. Will improve this right away :+1:

oetyng: Fixed.

safenode/src/domain/wallet/local_store.rs
oetyng: Being fixed with the above.

safenode/src/domain/wallet/local_store.rs
oetyng: Ok.

I've noticxed that the client output is extremely noise now that logs are added for everything. 

It's hard to follow simple flows now.

safenode/src/domain/wallet/local_store.rs
oetyng: fixed

safenode/src/domain/wallet/local_store.rs
oetyng: fixed

safenode/src/domain/wallet/local_store.rs
oetyng: fixed

safenode/src/domain/wallet/local_store.rs
oetyng: fixed

maqi (approved):

PR #204, opened 2023-04-27
feat(client): allow clients to dial specific network peers

joshuef:


bochaco (approved):

safenode/src/bin/safe/mod.rs
```diff
@@ -55,3 +57,28 @@ async fn get_client_dir() -> Result<PathBuf> {
     tokio::fs::create_dir_all(home_dirs.as_path()).await?;
     Ok(home_dirs)
 }
+
+// TODO: dedupe
+/// Parse multiaddresses containing the P2p protocol (`/p2p/<PeerId>`).
+/// Returns an error for the first invalid multiaddress.
+fn parse_peer_multiaddreses(multiaddrs: &[Multiaddr]) -> Result<Vec<(PeerId, Multiaddr)>> {
```
oetyng, line 64: parse_peer_multiaddreses

*addresses

safenode/src/bin/safe/cli/mod.rs
```diff
@@ -10,12 +10,17 @@ mod register;
...
 
 use clap::{Parser, Subcommand};
+use libp2p::Multiaddr;
 
 pub(super) use self::{files::files_cmds, register::register_cmds, wallet::wallet_cmds};
 
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
 pub(super) struct Opt {
+    /// Nodes we dial at start to help us get connected to the network. Can be specified multiple times.
```
jacderida, line 20: I would suggest a couple of little changes for this.

If we're going to allow the argument to be used multiple times, perhaps it should be renamed `peer` singular. Otherwise, if it's `peers`, we should probably tell the user how each of the multiaddr values are separated, i.e., by spaces, colons or whatever.

For the UI text, I would suggest:
```
Provide a peer to connect to a public network.

The argument can be used multiple times and the peer should be specified in the 'MultiAddr' format.

If not provided, a connection will be attempted to a local network.
```

I realise it's just been copied from the node, but I don't think the node is necessarily right either.

oetyng (approved):

safenode/src/bin/safe/cli/mod.rs
oetyng: @jacderida, I tried understand what change you requested exactly, so that I could push it. But didn't quite.
So I will merge this without the change, because I think we might need this for a bug we're seeing. But if you raise PR for it, we'll get it right in!

PR #212, opened 2023-04-29
fix: restore the use of file system storage

### LINKED ISSUE #211: Restore use of file system chunk store.
It has better performance (as shown [here](https://safenetforum.org/t/user-experience-for-the-clis/38101/334))   and less complexity than the kad storage.

The file store basically won't change anymore, and the msg handling is much simpler than the kad store version.

Also, it reduces the black box reliance on the kad store system.

oetyng:
Resolves #211.


joshuef:
I don't see any tests asserting that files are republished here. We'd need that if we're not relying on libp2p

Issue #207, opened 2023-04-28
Client-to-many-nodes-to-many-nodes messing fails.


oetyng:
This bug is reproduced in PR #206.

The security and validation of DBCs in the network depend on the below described flow.
The referenced PR demonstrates the bug using an E2E test named `node_many_to_many_ping_pong_succeeds`.

***

A client sends a `ClientPing` to close group nodes of a given name.
Those nodes send a `NodePing` to close group nodes of the hash of the original name.
A node that receives a `NodePing`, will return `NodePong(Ok(()))` if it considers itself to be among the close group nodes of the given name.

The nodes receiving a `ClientPing`, will return a `NodePong(Ok(()))` if a majority of the nodes they queried, returned a `NodePong(Ok(()))`.

The test will succeed if the client receives a majority of `NodePong(Ok(()))`.

oetyng:
Fixes [#192: nodes cannot query spend parent nodes](https://github.com/maidsafe/safe_network/issues/192).

maqi:
Issue resolved.
The PingPong test got passed within the PR https://github.com/maidsafe/safe_network/pull/218.

PR #213, opened 2023-04-29
refactor: impl get local peers

oetyng:
This is a suggestion for how we can perhaps help a bit with the msg issuing that has been seen (described [here](https://github.com/maidsafe/safe_network/issues/207)).

*** 
### Background
It struck me that at least part of it could come from msg overload..

 (especially if there is some issue with contention/deadlock, that we are investigating)

..when I was looking at the logs and saw manymany queries back to back to close group peers, to know who close group peers are.

We probably don't need to query the network so often to get the close peers, it could be an issue with repeated calls. For some requests, especially the repeated ones, it should be fine to just get them from locally.

***
### Clients
For client requests, there is no option to only get local closest, since they will always talk to random places across the network.

***
### Further work
NB: It is not fully identified in this PR, all places where we should use one or the other. So some more work with that is needed.
For now only the broadcast of `ValidSpendReceived` and `DoubleSpendAttempted` is using local closest only.
The `GetDbcSpend` query is asking network for closest though, since that is used for parents that reside on random places across the network.

### DIFF
--- a/safenode/src/network/mod.rs
+++ b/safenode/src/network/mod.rs
@@ -371,9 +371,9 @@ impl Network {
             .await
     }
 
-    /// Send `Request` to the the given `PeerId` and await for the response. If `self` is the recepiant,
-    /// then the `Request` is forwarded to itself and handled. Then a corresponding `Response` is created
-    /// and forwarded to iself. Hence the flow remains the same and there is no branching at the upper
+    /// Send `Request` to the the given `PeerId` and await for the response. If `self` is the recipient,
+    /// then the `Request` is forwarded to itself and handled, and a corresponding `Response` is created
+    /// and returned to itself. Hence the flow remains the same and there is no branching at the upper
     /// layers.
     pub async fn send_request(&self, req: Request, peer: PeerId) -> Result<Response> {
         let (sender, receiver) = oneshot::channel();


safenode/src/network/mod.rs
```diff
@@ -295,25 +295,57 @@ impl Network {
     /// Returns the closest peers to the given `XorName`, sorted by their distance to the xor_name.
     /// Excludes the client's `PeerId` while calculating the closest peers.
     pub async fn client_get_closest_peers(&self, xor_name: XorName) -> Result<Vec<PeerId>> {
```
b-zee: Does it make sense to add in `query_for` here too?

safenode/src/node/api.rs
```diff
@@ -403,7 +403,8 @@ impl Node {
     /// Retrieve a `Spend` from the closest peers
     async fn get_spend(&self, address: DbcAddress) -> Result<SignedSpend> {
         let request = Request::Query(Query::Spend(SpendQuery::GetDbcSpend(address)));
-        let responses = self.network.node_send_to_closest(&request).await?;
+        // NB: for discovery, might need query for the closest instead of local.
+        let responses = self.network.node_send_to_local_closest(&request).await?;
```
grumbach: Agreed, I think that for DBC related stuff we need to query to prevent attacks.
Here using `node_send_to_queried_closest` sounds safer.

grumbach (commented):
LGTM, one comment about DBC spends

safenode/src/network/mod.rs
oetyng: Perhaps so, I had it first actually. But then I realised that clients should never call only local stored closest (or maybe only in rare optimisation cases?) because all (or very close to all) their interaction with network is to random places across it.

So I though that it should just be default, and no need to distinguish?

safenode/src/node/api.rs
oetyng: I saw this one! And changed it just now.

It actually _has_ to have query because getting parents is from across the whole network.

oetyng: Broadcast of valid spend and double spend attempted are using locally known closest though.

b-zee (approved):

oetyng:
I saw so much repeated msgsing between nodes for exactly closest peers, so it seemed very overkill. 
Like over and over in the span of a few seconds or less.
But in the end now in this PR the get from local is not used anywhere (the last remaining, broadcast double spend attempt, which should be fairly rare compared to other msgs, will still have it just for ultimate security).

So the question is if this should be merged at all if not even used?

***

The case where I see it can definitely be used is when repeating calls for missing data on every peer died event. That's another PR though. (Because that's using pageing with batches of 25 data items. When nodes hold data that will be quite a number of repeated calls, which stops when returned collection is empty.)

I'm letting this brew for a while until we have a good picture of exactly where we should use locally known closest.

PR #214, opened 2023-04-29
chore: remove node msg handling mut access

oetyng:
This should decrease the cases of sequential access or deadlocks, which may have contributed to [this issue](https://github.com/maidsafe/safe_network/issues/207).

Although tests on this PR shows that the issues persist.

Nevertheless, it should be better to push down any locking further down, to decrease the amount of calls/logic that get serialized.

### DIFF
--- a/safenode/src/domain/storage/spends.rs
+++ b/safenode/src/domain/storage/spends.rs
@@ -310,7 +310,7 @@ mod tests {
     use sn_dbc::MainKey;
 
     #[tokio::test]
-    async fn write_and_read_300_spends() {
+    async fn write_and_read_100_spends() {
         // Test that a range of different spends can be stored and read as expected.
         let number_of_spends = 100;
         let storage = init_file_store();
@@ -330,6 +330,7 @@ mod tests {
                 .try_add(&spend)
                 .await
                 .expect("Failed to write spend.");
+
             let read_spend = storage
                 .get(&dbc_address(spend.dbc_id()))
                 .await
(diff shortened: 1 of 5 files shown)


safenode/src/domain/storage/spends.rs
```diff
@@ -333,9 +339,45 @@ mod tests {
...
+                    .get(&dbc_address(spend.dbc_id()))
+                    .await
+                    .expect("Failed to read spend.");
+
+                assert_eq!(spend.to_bytes(), read_spend.to_bytes());
+            })
+        });
+        let res = futures::future::join_all(tasks).await;
+
+        assert!(res.into_iter().all(|res| res.is_ok()))
```
maqi: shall there be an `overall` result check, to confirm the final result is expected?

oetyng: I'm not sure what would be the overall result check? Not following.

The add and read in the spawned tasks, also verify that the subsequently read spend is what was previously added.

maqi: that is with all those spends , is the final sum-up result is epxected.

i.e. given the base balance is O, then there is two followup spends A & B

maqi (approved):

safenode/src/domain/storage/spends.rs
oetyng: We just check the bytes are the same of stored and gotten spend. That is even more exact.

joshuef:
This apparently doesn't solve the issue and adds a whole lot of potential problems with the heavy use of RwLock. Closing as I _really_ don't think this is the way forwards here. 

(We've seen a lot of issues around RwLocks in the prior codebase and should strive to avoid them unless absolutely necessary).

PR #215, opened 2023-04-30
refactor: impl canonical network key

oetyng:
This was originally part of PR #212, but can be broken out to its own PR.

We've been keeping our `XorName` closeness logic in parallel with the kad closeness, i.e. mapping `PeerId` to our concept of id. We did some hashing that seems it would lead to us having one view of what is close peer, while kad had another view of it. 
Maybe it wasn't actually happening, but the two logics are there side by side, and there is not so much clarity there.

As I was looking at data replication, I got more clarity there, and I found what I believe to be the way to use only the kad closeness logic, and introducing some simplification and type safety.

This PR is an intermediate step and further clarity can be reached.

***

The closeness logic exists in `libp2p::kad::KadBucketKey`. It does all that work that we had our custom code for. We already use it for close calc, but there was not so much type safety to ensure we use only that, and a bit confusing where `XorName` is used, why and how - and how closeness of `PeerId` fits into closeness of `XorName`.

`KadBucketKey` (in this PR aliased as `Key`) works on a bit vector - which is also what `PeerId` is a wrapping type for, and it has a `preimage` which is what we pass in, and it has a key which is the `SHA256 digest` of the `preimage`.

So in this PR a canonical `NetworkKey` is implemented for cohesion. It holds the bit vector of either `PeerId` or XorName, without any modifications to it.
So it has two enum variants, `PeerId` and `XorName`. And it means we leave all bit vectors untouched before they get mapped to `KBucketKey`. 

The `NetworkKey` is now what is passed around in the paths where we use dst/distance/closeness metrics.

This should keep the distance metrics of all our items the same, and also clear up when to use what, how they relate etc.

### DIFF
--- a/safenode/src/node/api.rs
+++ b/safenode/src/node/api.rs
@@ -23,6 +23,7 @@ use crate::{
             Request, Response, SpendQuery,
         },
         storage::{dbc_address, registers::User, DbcAddress},
+        NetworkKey,
     },
 };
 
@@ -38,7 +39,6 @@ use std::{
     path::Path,
 };
 use tokio::{sync::mpsc, task::spawn};
-use xor_name::XorName;
 
 #[derive(Debug)]
 pub(super) struct TransferAction {
@@ -154,16 +154,12 @@ impl Node {
             }
             NetworkEvent::PeerAdded(peer) => {
                 self.events_channel.broadcast(NodeEvent::ConnectedToNetwork);
-                let target = {
-                    let mut rng = rand::thread_rng();
-                    XorName::random(&mut rng)
-                };
-
+                let key = NetworkKey::from_peer(peer);
                 let network = self.network.clone();
                 let _handle = spawn(async move {
-                    trace!("On PeerAdded({peer:?}) Getting closest peers for target {target:?}");
-                    let result = network.node_get_closest_peers(target).await;
-                    trace!("For target {target:?}, get closest peers {result:?}");
+                    trace!("On PeerAdded({peer:?}) Getting closest peers for target {key:?}...");
+                    let result = network.node_get_closest_peers(&key).await;
+                    trace!("Closest peers to {key:?} got: {result:?}.");
                 });
             }
             NetworkEvent::NewListenAddr(_) => {
--- a/safenode/src/client/api.rs
+++ b/safenode/src/client/api.rs
@@ -17,6 +17,7 @@ use crate::{
     protocol::{
         messages::{Cmd, CmdResponse, Query, QueryResponse, Request, Response, SpendQuery},
         storage::{dbc_address, dbc_name, Chunk, ChunkAddress},
+        NetworkKey,
     },
 };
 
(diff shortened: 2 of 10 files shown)


safenode/src/domain/storage/address/mod.rs
```diff
@@ -22,23 +24,26 @@ use serde::{Deserialize, Serialize};
...
-#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
 pub enum DataAddress {
     ///
     Chunk(ChunkAddress),
     ///
     Register(RegisterAddress),
     ///
     Spend(DbcAddress),
+    /// The new general identifier for a items in the network.
```
maqi: for an item in the network

However I feel this is bit tricky (as `item` covers Chunk/Register/Spend) which may cause some misuse that:
on one side we use address of `Chunk`, `Register` , `Spend`, but on other side use `Network` key.

Shall we `force` to use just one via type safe?

safenode/src/network/event.rs
```diff
@@ -85,7 +85,9 @@ pub enum NetworkEvent {
         channel: MsgResponder,
     },
     /// Emitted when the DHT is updated
-    PeerAdded,
+    PeerAdded(PeerId),
+    /// Emitted when the DHT is updated
+    PeersAdded(Vec<PeerId>),
```
maqi: PeerAdded can be PeersAdded with just one entry?

safenode/src/node/api.rs
```diff
@@ -129,20 +129,28 @@ impl Node {
...
-                    let result = network.node_get_closest_peers(target).await;
-                    trace!("For target {target:?}, get closest peers {result:?}");
+                    trace!("Getting closest peers for target {key:?}...");
+                    let result = network.node_get_closest_peers(&key).await;
+                    trace!("Closest peers to {key:?} got: {result:?}.");
                 });
             }
+            NetworkEvent::PeersAdded(peers) => {
+                self.events_channel.broadcast(NodeEvent::ConnectedToNetwork);
+                for peer in peers {
```
maqi: the purpose of this get_closest block is just to get the node try to discover other nodes.
So, no need to `get_closest for each peer` , just pick one (or a buset, say number_nodes/8 ) random peer to get_closest shall be enough.

safenode/src/node/api.rs
```diff
@@ -129,20 +129,28 @@ impl Node {
             NetworkEvent::RequestReceived { req, channel } => {
                 self.handle_request(req, channel).await?
             }
-            NetworkEvent::PeerAdded => {
+            NetworkEvent::PeerAdded(peer) => {
```
maqi, line 155: I think merge `PeerAdded` and `PeersAdded` will save a lot of traffic on `get_closest`

safenode/src/network/event.rs
oetyng: Could be, absolutely. Fixing.

safenode/src/node/api.rs
oetyng: Will fix!

safenode/src/domain/storage/address/mod.rs
oetyng: Yeah, I wasn't quite sure, so felt this is an intermediate step.

Thinking about it now... if good from get go or can wait..

safenode/src/node/api.rs
oetyng: But how many could actually be gotten in such a vec of peers? Is it many many?

safenode/src/node/api.rs
oetyng: Fixed!

safenode/src/domain/storage/address/mod.rs
oetyng: The differentiation in DataAddress is maybe not necessary since we anyway always know what the action (query/cmd) is by the name of it..

Maybe on the benefit side is that it's a bit clearer in code. But I don't think there are many places where we could be confused about it.

safenode/src/node/api.rs
oetyng: Fixed.

safenode/src/network/event.rs
oetyng: Fixed.

safenode/src/domain/storage/address/mod.rs
oetyng: I will defer a potential DataAddress update to a follow up PR.

maqi (approved):

PR #223, opened 2023-05-02
fix: avoid deadlock during spend dbc

maqi:
This PR contains the work trying to resolve the deadlock issue during the dbc spend process.
Which shall get the `multiple_sequential_transfers_succeed` test starts passing

Closes https://github.com/maidsafe/safe_network/pull/218

### DIFF
--- a/safenode/src/network/event.rs
+++ b/safenode/src/network/event.rs
@@ -255,23 +255,12 @@ impl SwarmDriver {
                 }
             }
             SwarmEvent::ConnectionClosed {
-                peer_id, endpoint, ..
+                peer_id,
+                endpoint,
+                cause,
+                ..
             } => {
-                // The connection closed due to no msg flow show as `endpoint::receiver`
-                // The connection closed due to peer dead show as `endpoint::dialer`
-                info!("Connection closed to Peer {peer_id} - {endpoint:?}");
-                // if endpoint.is_dialer() {
-                //     info!("Dead Peer {peer_id:?}");
-                //     if self
-                //         .swarm
-                //         .behaviour_mut()
-                //         .kademlia
-                //         .remove_address(&peer_id, endpoint.get_remote_address())
-                //         .is_some()
-                //     {
-                //         info!("Removed dead peer {peer_id:?} from RT");
-                //     }
-                // }
+                info!("Connection closed to Peer {peer_id} - {endpoint:?} - {cause:?}");
             }
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
                 if let Some(peer_id) = peer_id {
--- a/safenode/src/network/event.rs
+++ b/safenode/src/network/event.rs
@@ -91,7 +91,7 @@ pub enum NetworkEvent {
         channel: MsgResponder,
     },
     /// Emitted when the DHT is updated
-    PeerAdded,
+    PeerAdded(PeerId),
     /// Started listening on a new address
     NewListenAddr(Multiaddr),
 }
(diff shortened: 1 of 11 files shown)


oetyng:
Hey @maqi, I wanted to give some more info that maybe can be useful.

The fees used to have no problems being returned on query. Something happened, I am not sure what, I couldn't see much change in the messaging/network code though.

Maybe you can find it working a bit back in history, and see what changed.

oetyng:
For example, this commit returns the fees successfully: [a60ad2338190b4ca6d1341ea41bc1f266aea0810](https://github.com/maidsafe/safe_network/commit/a60ad2338190b4ca6d1341ea41bc1f266aea0810)

safenode/src/network/event.rs
```diff
@@ -217,18 +220,18 @@ impl SwarmDriver {
...
-                        .swarm
-                        .behaviour_mut()
-                        .kademlia
-                        .remove_address(&peer_id, endpoint.get_remote_address())
-                        .is_some()
-                    {
-                        info!("Removed dead peer {peer_id:?} from RT");
-                    }
-                }
+                // if endpoint.is_dialer() {
```
bochaco: If all this commented code is not needed we can remove it

safenode/src/node/api.rs
```diff
@@ -280,187 +307,258 @@ impl Node {
...
+    transfer_actor: mpsc::Sender<TransferAction>,
+    response_channel: MsgResponder,
+    signed_spend: Box<SignedSpend>,
+    parent_tx: Box<DbcTransaction>,
+    fee_ciphers: BTreeMap<NodeId, FeeCiphers>,
+) {
+    // First we fetch all parent spends from the network.
+    // They shall naturally all exist as valid spends for this current
+    // spend attempt to be valid.
+    trace!("handle_spend_dbc");
```
bochaco: Don't we need some context info so this log entry is useful, e.g. dbc-id, tx-id, or something else, something which can uniquely link this log entry to some flow/transaction ?

oetyng:
> For example, this commit returns the fees successfully: [a60ad2338190b4ca6d1341ea41bc1f266aea0810](https://github.com/maidsafe/safe_network/commit/a60ad2338190b4ca6d1341ea41bc1f266aea0810)

Oh, and also, the querying for spends from the network, i.e. validating a transfer, also works in that commit.


And, two commits later, at `feat: impl early return for majority get spend` (https://github.com/maidsafe/safe_network/commit/1513ef5f33993cc417e969d36ca50055884f10ea), the fee query is OK, but the subsequent spend request fails, with outbounderror timeout.

maqi:
@oetyng seems the `no response` is related to `swarm.next()` inside the `tokio::select`.
once changed to `swarm.select_next_some()`, the issue then gone.

However I checked the commit history, that `select` block remain untouched for a long time.
The one that I find might be related could be 5cd9f4af674a1e19ea64b1092959477afdeb4040, which was commited on 6 Apr
and was changing from using `futures::select` to `tokio::select`, and from `cmd_receiver.next()` to `cmd_receiver.recv()`

oetyng:
Okay interesting. I can try verify if you are not already?

oetyng:
Hey @maqi, I see tests are passing! Good job!

If you run the same test a couple of times back to back. Do they all still succeed?

Those could be in two ways: 
- Multiple ci runs (that will be new network every time), 
- Multiple times on the same network.

It is quite mixed with `success` / `fail`, when back to back on same network.

It is expected that this test would pass as many times as required until the Faucet is empty of tokens (about 1.2 billion times).

But if it succeeds on first, then that is good enough as a first step!! Very good!
So, next step would be to get it consistently passing back to back.

***

One note from running it locally: The spend cmds after genesis spend, seem to often return `CouldNotSendTokens("Failed to verify transfer validity in the network Not enough close group nodes accepted the spend. Got 0, required: 5.")`.

The transfer can however be validated right after, by querying for the spend.

A minor thing, considering the great achievement of this PR. Just a note for UX, that something can be fixed there in the coming.

maqi:
Hi @oetyng , the passing is now quite consistent (locally and CI), expecially now CI runs the dbc spend test across all three OS.

And thx again for the detailed history trace back. :)

=================================

Not sure, seems there is failing come back after rebase due to other PR merged.
Or could be just because of low performance of CI, as I even see failed UnitTest during the `compliation` step

=================================

Ok, seems the failing is due to client got added into RT, which prolongs the get_closest
With a fix, CI is not turned back to ALL green.

joshuef (approved):

PR #224, opened 2023-05-02
feat(node): add identify behaviour

b-zee:
Using the identify behaviour to exchange node information and adding addresses to the routing table based on that.

### DIFF
--- a/safenode/src/network/mod.rs
+++ b/safenode/src/network/mod.rs
@@ -293,6 +293,18 @@ impl Network {
     }
 
     /// Dial the given peer at the given address.
+    pub async fn add_to_routing_table(&self, peer_id: PeerId, peer_addr: Multiaddr) -> Result<()> {
+        let (sender, receiver) = oneshot::channel();
+        self.send_swarm_cmd(SwarmCmd::AddToRoutingTable {
+            peer_id,
+            peer_addr,
+            sender,
+        })
+        .await?;
+        receiver.await?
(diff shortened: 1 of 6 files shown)


safenode/src/network/mod.rs
```diff
@@ -203,12 +203,18 @@ impl SwarmDriver {
             query_interval: Duration::from_secs(5),
             ..Default::default()
         };
-
         let mdns = mdns::tokio::Behaviour::new(mdns_config, peer_id)?;
+
+        let identify_cfg =
+            libp2p::identify::Config::new("safe/0.1.0".to_string(), keypair.public())
```
joshuef: can we pull this from a const? (is this semvar? I wonder how this might interact with the protocol folder and changes there... something for the huddle)

b-zee: Good one, I think for now we should mostly ignore versions. As long as we don't go to major 1. But putting it in a const is probably a good idea.

joshuef (approved):

safenode/src/network/mod.rs
```diff
@@ -280,6 +292,18 @@ impl Network {
         receiver.await?
     }
 
+    /// Dial the given peer at the given address.
```
maqi, line 295: the code block of the function is nothing related to `dial`?
though maybe the later on swarm will carry out `dial` before add to RT ?

PR #235, opened 2023-05-04
Remove fees per input

grumbach:
This PR removes the fees per input design, which is an economically unviable design. 

Example scenario:
- each input has a fee, so for my 5$ input I will need a fee, let's say the smallest unit is 1$. 
- then I spend my 5$ and 1$ is fee, so far so good.
- now farmer nodes accumulate lots of small fee DBCs, in order for them to spend them, they need a fee
- to spend a 1$ DBC they will need a 1$ fee

In this system, eventually, all money on the network will turn into unuseable, value-less penny-DBCs

A viable alternative would be fee per Tx like bitcoin does.

### DIFF
--- a/safenode/src/domain/client_transfers/mod.rs
+++ b/safenode/src/domain/client_transfers/mod.rs
@@ -39,16 +39,13 @@ pub(crate) use self::{
     online::create_transfer as create_online_transfer,
 };
 
-use crate::protocol::messages::{FeeCiphers, NodeId, RequiredFee};
+use crate::protocol::messages::{FeeCiphers, NodeId};
 
 use sn_dbc::{
-    Dbc, DbcId, DbcIdSource, DbcTransaction, DerivedKey, PublicAddress, RevealedAmount,
-    SignedSpend, Token,
+    Dbc, DbcIdSource, DbcTransaction, DerivedKey, PublicAddress, RevealedAmount, SignedSpend, Token,
 };
 
 use std::collections::BTreeMap;
-
-type NodeFeesPerInput = BTreeMap<DbcId, BTreeMap<NodeId, (RequiredFee, DbcIdSource)>>;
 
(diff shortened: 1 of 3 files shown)


maqi:
Suggest this PR shall be merged after the PR 223 as this may affect the existing `multiple_sequential_transfers_succeed` test,
which is currently not being checked in CI

oetyng:
> This PR removes the fees per input design, which is an economically unviable design. 
> 

It isn't really if you consider fluctuating price. What is earned in high traffic (and cost) can be redeemed in low traffic (and cost).

Nonetheless, I thought that was a bit too high risk. It is much better to use the `reason` field, to distinguish fees. That allows `free merge of fees`.
The design is described in discourse. I wrote it a while ago.

I think that design is necessary with `fee per input`. It removes `reversed stake` characteristics which some considered an appealing idea. But in that format it was dubious anyway if it could be done at sufficiently low risk.

***

`Fee per tx` is quite complicated in comparison.

You can see that if you start drafting up those two solutions side by side.

***

If you however still want to try tx fees instead, then the code you need to remove is as seen in [this PR](https://github.com/maidsafe/safe_network/pull/239).

Most if not all is designed around `fee-per-input`, so you would have to redesign all that anyway, and simpler to just remove.

oetyng:
> Suggest this PR shall be merged after the PR 223 as this may affect the existing `multiple_sequential_transfers_succeed` test, which is currently not being checked in CI

@maqi, there's no problem for that test, if the whole `fee-per-input` design is removed. The necessary PR for that is this one: https://github.com/maidsafe/safe_network/pull/239

It removes all fee-logic, since it's all based on `fee-per-input` design. So if another design is attempted - such as `fee-per-tx` - then better re-write it from scratch. Too much work to try to adapt/refactor.

grumbach:
Replaced by #239

PR #239, opened 2023-05-04
feat: remove transfer fees

oetyng:
This is the complete removal of the `fees-per-input` design, and by that all of the transfer fee logic.
This is because if a `fees-per-tx` design is attempted, it is much easier to write it from scratch, since all of the current fee logic is designed around `fees-per-input`.

(However, worth to note is that `fee-per-input`, as far as seen yet at least, is a simpler design than `fee-per-tx`. And when using `free-merge-of-fees`, it is just as much economically viable.)

***

As found in PR #235:

> This PR removes the fees per input design, which is an economically unviable design.

It isn't really if considering fluctuating price. What is earned in high traffic (and cost) can be redeemed in low traffic (and cost).

Nonetheless, it is even better to use the `reason` field, to _distinguish fees_. That allows `free-merge-of-fees`.
The design is described in discourse.

That design is probably even necessary with `fee-per-input`, even though it removes the `reverse-stake` concept that has been talked about.

***

`Fee-per-tx` is quite complicated in comparison.

It can be seen when drafting up those two solutions side by side (`fee-per-tx`, and `fee-per-input` with `free-merge-of-fees`).

If however that solution is still attempted, then this PR is most practical, as most if not all is designed around `fee-per-input`, so you would have to redesign all that anyway, and simpler to just remove.


grumbach (approved):
LGTM ✅ 

Agreed, let's remove fees for now while we come up with a design that everybody agrees on. 
I would agree with you that fees-per-tx is probably more complex than fees-per-input on our network. 

On the other hand, although free-merge-of-fees would fix the serious dusting shortcoming of fees-per-input, I feel like this is a technical hard fix to fix an unreliable economy design. I think we should avoid dusting altogether. 

Let's try to keep things simple for now with your PR. The solution might be something we haven't thought of yet. 

Thanks for your insights and this PR @oetyng 🙇

oetyng:
> Thanks for your insights and this PR @oetyng 🙇 

Happy to help 😊
Go get that stuff working in the best way possible 🥳

oetyng:
> Thanks for your insights and this PR @oetyng 🙇 

Happy to help 😊
Go get that stuff working in the best way possible @grumbach 🥳

PR #246, opened 2023-05-08
feat: client complete upload with majority Ok resps

### LINKED ISSUE #243: Lower the client check during upload
The current client requires ALL 8 OK resp to consider a chunk got succeed in uploading.
AND, a single incorrect resp will fail not only this chunk but the whole file uploading.

This is OK (or even required for a quick problem spotting) for our local test, but seems an overkill to the public testnet.

I personally experienced three times on uploading a 20MB file to public testnet, and the log shows all three failure is on just 1 out of 8 incorrect resp (that is received 7 OK resps but got 1 incorrect one after the time out).

It's suggest to only log the error and continue with the uploading when received majority OK resps, 
instead of a termination of upload flow.

maqi:
This PR lower the client check on upload chunks, from requiring ALL (8) OK resps to Majority (5) .
This avoid the termination of a whole file's upload flow just because of one single delayed OK response within the public testnet,
as stated in the issue https://github.com/maidsafe/safe_network/issues/243

Closes https://github.com/maidsafe/safe_network/issues/243

### DIFF
--- a/safenode/src/client/api.rs
+++ b/safenode/src/client/api.rs
@@ -169,14 +169,15 @@ impl Client {
         let request = Request::Cmd(Cmd::StoreChunk(chunk));
         let responses = self.send_to_closest(request).await?;
 
-        let all_ok = responses
+        let all_oks = responses
             .iter()
-            .all(|resp| matches!(resp, Ok(Response::Cmd(CmdResponse::StoreChunk(Ok(()))))));
-        if all_ok {
+            .filter(|resp| matches!(resp, Ok(Response::Cmd(CmdResponse::StoreChunk(Ok(()))))))
+            .count();
+        if all_oks >= close_group_majority() {
             return Ok(());
         }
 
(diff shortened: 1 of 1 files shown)


safenode/src/client/api.rs
```diff
@@ -167,16 +167,19 @@ impl Client {
     pub(super) async fn store_chunk(&self, chunk: Chunk) -> Result<()> {
         info!("Store chunk: {:?}", chunk.address());
         let request = Request::Cmd(Cmd::StoreChunk(chunk));
+        // TODO: consider using `expect_closest_majority_ok` as within the dbc spend flow
```
dirvine: Can we put this in an issue please. Otherwise this is good to go.

dirvine (approved):

safenode/src/client/api.rs
maqi: issue https://github.com/maidsafe/safe_network/issues/247  crated for this

PR #248, opened 2023-05-08
refactor: impl canonical network key

maqi:
This PR contains the work from https://github.com/maidsafe/safe_network/pull/215 , to clarify the concept of `network address` 

Closes https://github.com/maidsafe/safe_network/pull/215


safenode/src/protocol/mod.rs
```diff
@@ -10,5 +10,102 @@
...
+/// This is the mapping from the XOR name used
+/// by for example self encryption, or the libp2p `PeerId`,
+/// to the key used in the Kademlia DHT.
+/// All our xorname calculations shall be replaced with the `KBucketKey` calculations,
+/// for getting proximity/distance to other items (wether nodes or data).
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+pub enum NetworkKey {
+    /// The NetworkKey is representing a PeerId.
+    PeerId(Vec<u8>),
+    /// The NetworkKey is representing an XorName of some data.
```
dirvine: I wonder if here we should include the types of data instead of just the generic XorAddr?

safenode/src/protocol/storage/address/mod.rs
```diff
@@ -10,19 +10,17 @@ mod chunk;
...
 
 use sn_dbc::DbcId;
 
 use serde::{Deserialize, Serialize};
 use xor_name::XorName;
 
 /// An address of data on the network.
-#[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize, Debug)]
 pub enum DataAddress {
```
dirvine: I.e. we could remove this enum and include these types in the `NetworkAdder`

safenode/src/protocol/error/mod.rs
```diff
@@ -29,4 +29,10 @@ pub enum Error {
     /// Errors in node transfer handling.
     #[error("Transfer error: {0:?}")]
     Transfers(#[from] TransferError),
+    /// Error when converting to a PeerId.
+    #[error("Could not convert to a PeerId: {0:?}")]
+    NotAPeerId(String),
+    /// Error when converting to an XorName.
+    #[error("Could not convert to an XorName: {0:?}")]
+    NotAnXorName(String),
```
bochaco: Neither `NotAPeerId` nor `NotAnXorName` sounds like they should be part of the protocol but just part of our own implementation, I don't see any scenario where an error going from a node to a peer or client could contain any of these two types of errors. These shouldn't be part of the protocol.

bochaco (changes requested):

safenode/src/protocol/error/mod.rs
maqi: I think they got put here just because they could happen in node OR client (yeah agreed may only related to implementation)
Never means they will got `transmitted` out .

In that case, do you think they shall got moved into `safenode/srd/node/error.rs` or somewhere else?

dirvine: Probably just `InvalidNetworkAddress` would work?

bochaco: If we are throwing that error when trying to convert something in our impl they shouldn't be part of the protocol, but in whatever place we are doing that and at most returned by such impl's API. I don't think this is a case of an invalid net addr but just impl internal errors.

safenode/src/protocol/messages/node_id.rs
```diff
@@ -6,28 +6,33 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::network::Error as NetworkError;
```
bochaco: The `protocol` cannot and should not depend on any safe crate, it has no dependency on any of our crates that implement node/client.

bochaco (changes requested):

safenode/src/protocol/mod.rs
```diff
@@ -12,3 +12,100 @@ pub mod error;
...
+    pub fn as_kbucket_key(&self) -> Key<Vec<u8>> {
+        Key::new(self.as_bytes())
+    }
+
+    /// Compute the distance of the keys according to the XOR metric.
+    pub fn distance(&self, other: &NetworkAddress) -> Distance {
+        self.as_kbucket_key().distance(&other.as_kbucket_key())
+    }
+
+    // NB: Leaving this here as to demonstrate what we can do with this.
```
bochaco, line 90: I don't tthink we need this, specially because it's not using any of the `protocol` mechanisms but some kad fns.

safenode/src/protocol/mod.rs
```diff
@@ -12,3 +12,100 @@ pub mod error;
 pub mod messages;
 /// Storage types for spends, chunks and registers.
 pub mod storage;
+
+use self::storage::{ChunkAddress, DbcAddress, RegisterAddress};
+
+use libp2p::{
```
bochaco, line 18: We shouldn't have dependencies to crates which are implementation of the `protocol`, this mod/crate doesn't and shouldn't depend on any SAFE crate or crates needed for implementation of the protocol.

bochaco (changes requested):

bochaco (approved):

PR #252, opened 2023-05-09
feat(node): Make mDNS use a feat

joshuef:
No to run a testnet I'm doing: 

```
killall safenode || true && rm ~/.safe/client && RUST_LOG=safenode,safe cargo run --bin testnet -- --flame -b --interval 1 && export SAFE_PEERS=$(rg "listening on \".+\"" ~/.safe -u | rg '/ip4.*$' -m1 -o | rg '"' -r '')   
```


safenode/src/peers_acquisition/error.rs
```diff
@@ -0,0 +1,23 @@
...
+#[derive(Debug, Error)]
+#[allow(missing_docs)]
+pub enum Error {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[cfg(feature = "otlp")]
+    #[error("OpenTelemetry Tracing error: {0}")]
+    OpenTelemetryTracing(#[from] opentelemetry::trace::TraceError),
+    #[error("Could not configure OTLP logging: {0}")]
+    OtlpConfigurationError(String),
```
RolandSherwin: I assume the `OpenTelemetry` errors don't actually belong here? 

Looking further, I don't see the `Error` enum used anywhere. The `peer_acquisition/mod.rs` makes use of `eyre` and so we don't need this file?

b-zee:
Could you add in the README.md that the testnet locally has to be run with e.g. `--features local-discovery`? I am assuming that is needed?

safenode/src/client/api.rs
```diff
@@ -113,7 +113,13 @@ impl Client {
                 self.events_channel
                     .broadcast(ClientEvent::ConnectedToNetwork);
 
-                let key = NetworkAddress::from_peer(peer_id);
+                let target = {
```
dirvine: To me this seems like we are trying to populate our RT in some fakish way. 
If we connect to any node then get closest to us then we can connect to the results of that. 
Our RT will be populated when we do spend parent checking or any other network search. 
IMO all we need to do initially is connect to the peers close to our address.

PR #253, opened 2023-05-09
feat(storage): implement DiskBackedRecordStore

### LINKED ISSUE #242: Disk backed storage
Implement a `RecordStore` offering disk backed storage.

It should:

- [ ] write all records to disk
- [ ] load records from disk

----

After the initial disk impl, we can look to provide a limited LRU cache for all result types (configurable), which is used first if available, if not we read from disk (but always write to disk, after LRU write)

RolandSherwin:
- This allows KAD to store the `Records` to fs instead of memory.
- Only data `Records` are stored to disk, `ProviderRecords` are still stored in memory

closes #242

### DIFF
--- a/safenode/src/domain/storage/disk_backed_record_store.rs
+++ b/safenode/src/domain/storage/disk_backed_record_store.rs
@@ -1,0 +1,318 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+// This was forked from the
+// use super::*;
+
+use libp2p::identity::PeerId;
+use libp2p::kad::{kbucket, K_VALUE};
+use smallvec::SmallVec;
+
+use libp2p::kad::{
+    record::{Key, ProviderRecord, Record},
+    store::{Error, RecordStore, Result},
+};
+use std::borrow::Cow;
+use std::collections::{hash_map, hash_set, HashMap, HashSet};
+use std::iter;
+
+/// In-memory implementation of a `RecordStore`.
+pub(crate) struct DiskBackedRecordStore {
+    /// The identity of the peer owning the store.
+    local_key: kbucket::Key<PeerId>,
+    /// The configuration of the store.
+    config: DiskBackedRecordStoreConfig,
+    /// The stored (regular) records.
+    records: HashMap<Key, Record>,
+    /// The stored provider records.
+    providers: HashMap<Key, SmallVec<[ProviderRecord; K_VALUE.get()]>>,
+    /// The set of all provider records for the node identified by `local_key`.
+    ///
+    /// Must be kept in sync with `providers`.
+    provided: HashSet<ProviderRecord>,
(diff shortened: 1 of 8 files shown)


safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -0,0 +1,483 @@
...
+                "Record with key {:?} already exists, not overwriting.",
+                r.key
+            );
+            return Ok(());
+        }
+
+        let num_records = self.disk_backed_records.len();
+        if num_records >= self.config.max_records {
+            warn!("Record not stored. Maximum number of records reached. Current num_records: {num_records}");
+            return Err(Error::MaxRecords);
```
maqi, line 161: `Error::RecordsTooMany` to be consistent with `Error::ValueTooLarge`

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -0,0 +1,483 @@
...
+        }
+    }
+
+    fn remove(&mut self, k: &Key) {
+        let filename = Self::key_to_hex(k);
+        let file_path = self.config.storage_dir.join(&filename);
+        match fs::remove_file(file_path) {
+            Ok(_) => {
+                trace!("Removed record from disk! filename: {filename}");
+                let _ = self.disk_backed_records.remove(k);
```
maqi: better move `let _ = self.disk_backed_records.remove(k);` before the match block
to free up the entry even the disk file is broken

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -0,0 +1,483 @@
...
+        let mut records = Vec::new();
+        for key in self.disk_backed_records.iter() {
+            if let Some(record) = self.get(key) {
+                records.push(record);
+            }
+        }
+        records.into_iter()
+    }
+
+    fn add_provider(&mut self, record: ProviderRecord) -> Result<()> {
```
maqi: this shall only get called when `PoviderRecord` is enabled ?
Which won't be used in the current cfg?
Maybe just leave it as `un-implemented` ?

Same to other ProviderRecord realted impls

safenode/src/domain/storage/disk_backed_record_store.rs
RolandSherwin: The methods of the above trait expects the `libp2p::store::Error`. I'm not really sure on how to workaround this and add/modify the error variants.

safenode/src/domain/storage/disk_backed_record_store.rs
RolandSherwin: Aye! Marking the methods as `todo!()` and removing the provider related fields. Since this is copied from `MemoryStore`, we can always add it back if we want a quick and working provider store.

RolandSherwin: Currently debugging an issue that raised after marking the methods as todo!()

RolandSherwin: The `kad_cfg.set_provider_publication_interval()` caused the node to trigger the provider APIs. Removing the cfg fixed the issue.

safenode/src/domain/storage/disk_backed_record_store.rs
maqi: Ah, I thought these are errors defined by us.
If they are the `libp2p` ones, then just stay with them.

joshuef: I don't think we can/should here. It should be converted at the upper layers. But to remain within the `RecordStore` trait we'll have to leave this I think

joshuef (approved):

PR #254, opened 2023-05-09
test(register): verify Registers data integrity during nodes churn

### LINKED ISSUE #56: Check register integrity over time
With 100% churn over X time, put registers should exist

bochaco:
Resolves #56

### DIFF
--- a/safenode/src/tests_e2e/registers_with_churn.rs
+++ b/safenode/src/tests_e2e/registers_with_churn.rs
@@ -1,0 +1,253 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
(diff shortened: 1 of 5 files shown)


joshuef (approved):

safenode/src/tests_e2e/registers_with_churn.rs
```diff
@@ -0,0 +1,253 @@
...
+                    } else {
+                        let _ = regs_erred.write().await.insert(addr, reg_error);
+                    }
+                }
+            }
+        }
+    });
+}
+
+async fn node_restart(addr: SocketAddr, delay_millis: u64) -> Result<()> {
```
maqi: Registers are read/write from disk on request.
Which actually a `disk-based-storage` requiring a targeted replication.

Meanwhile, a simple node restart won't clean out the previous disk files, which leaves a `back hole` for register to be read back during the churning test.
To resolve this:
1, Ensure during node restart, previous disk files will be cleaned up.
OR
2, instead of `node_restart` , have to `stop a node via rpc` then `add back a new node`.

bochaco: Pushed a commit implementing #1.

PR #262, opened 2023-05-11
Wallet cli ux

iancoleman:
This is a tidy up for wallet ux in the cli

* Remove `safe wallet deposit` since it has no way to be used, there's no way to export a dbc for later being imported. My feeling is this functionality will be enabled later anyhow with something like wallet backups such as `safe wallet import|export`. The `deposit` functionality is still retained in the code since it's used for genesis dbc.

* `safe wallet address` is a local-only operation, but was trying to connect to the network. This is fixed by instantiating the client object later in the cli process, rather than immediately at the start. This allows commands that run locally to avoid creating the client and contacting the network.

* Rename `safe wallet address` to `safe wallet receive` to be a pair with `send|receive`.

In the future it would be nice to also have some extra features, but are beyond the scope of this PR:

* wallet backup via `safe wallet export|import` which packages the whole wallet into a zip file (tx history, dbcs, name metadata etc)

*  use EIP2333 derivation with BIP39 mnemonics exposed via `safe wallet mnemonic show|load` (is this ergonomic?)

* encrypt all secrets on disk with a password. Enter the password before signing txs or any other use of secret keys

* cleaner separation of tx creation and signing, as a path to offline / hardware wallet transaction signing

### DIFF
--- a/safenode/src/bin/safe/mod.rs
+++ b/safenode/src/bin/safe/mod.rs
@@ -14,6 +14,7 @@ mod cli;
 use self::cli::{files_cmds, register_cmds, wallet_cmds, Opt, SubCmd};
 use clap::Parser;
 use eyre::Result;
+use libp2p::Multiaddr;
 
 use safenode::client::Client;
 use safenode::git_hash;
@@ -29,23 +30,26 @@ async fn main() -> Result<()> {
     let log_appender_guard = init_node_logging(&Some(tmp_dir.join("safe-client")))?;
 
     info!("Full client logs will be written to {:?}", tmp_dir);
-    println!("Instantiating a SAFE client...");
     println!("Current build's git commit hash: {}", git_hash::git_hash());
 
-    let secret_key = bls::SecretKey::random();
-    let peers = peers_from_opts_or_env(&opt.peers)?;
+    let peers = opt.peers;
     let root_dir = get_client_dir().await?;
 
-    let client = Client::new(secret_key, Some(peers)).await?;
-
     match opt.cmd {
-        SubCmd::Wallet(cmds) => wallet_cmds(cmds, &client, &root_dir).await?,
-        SubCmd::Files(cmds) => files_cmds(cmds, client.clone(), &root_dir).await?,
-        SubCmd::Register(cmds) => register_cmds(cmds, &client).await?,
+        SubCmd::Wallet(cmds) => wallet_cmds(cmds, peers, &root_dir).await?,
+        SubCmd::Files(cmds) => files_cmds(cmds, peers, &root_dir).await?,
+        SubCmd::Register(cmds) => register_cmds(cmds, peers).await?,
     };
 
     drop(log_appender_guard);
     Ok(())
+}
+
+async fn get_client(peers: Vec<Multiaddr>) -> Result<Client> {
+    println!("Connecting to network...");
+    let peers = peers_from_opts_or_env(&peers)?;
+    let secret_key = bls::SecretKey::random();
+    Ok(Client::new(secret_key, Some(peers)).await?)
 }
 
 async fn get_client_dir() -> Result<PathBuf> {
--- a/safenode/src/bin/safe/cli/files.rs
+++ b/safenode/src/bin/safe/cli/files.rs
@@ -6,10 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use safenode::{
-    client::{Client, Files},
-    protocol::storage::ChunkAddress,
-};
+use crate::{get_client, Multiaddr};
+use safenode::{client::Files, protocol::storage::ChunkAddress};
 
 use bytes::Bytes;
 use clap::Parser;
@@ -38,7 +36,12 @@ pub enum FilesCmds {
     },
 }
 
-pub(crate) async fn files_cmds(cmds: FilesCmds, client: Client, root_dir: &Path) -> Result<()> {
+pub(crate) async fn files_cmds(
+    cmds: FilesCmds,
+    peers: Vec<Multiaddr>,
+    root_dir: &Path,
+) -> Result<()> {
+    let client = get_client(peers).await?;
     let file_api: Files = Files::new(client);
     match cmds {
         FilesCmds::Upload { path } => upload_files(path, &file_api, root_dir).await?,
--- a/safenode/src/bin/safe/cli/register.rs
+++ b/safenode/src/bin/safe/cli/register.rs
@@ -6,6 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::{get_client, Multiaddr};
 use safenode::client::{Client, Error as ClientError};
 
 use clap::Subcommand;
@@ -35,11 +36,12 @@ pub enum RegisterCmds {
     },
 }
 
-pub(crate) async fn register_cmds(cmds: RegisterCmds, client: &Client) -> Result<()> {
+pub(crate) async fn register_cmds(cmds: RegisterCmds, peers: Vec<Multiaddr>) -> Result<()> {
+    let client = get_client(peers).await?;
     match cmds {
-        RegisterCmds::Create { name } => create_register(name, client).await?,
-        RegisterCmds::Edit { name, entry } => edit_register(name, entry, client).await?,
-        RegisterCmds::Get { names } => get_registers(names, client).await?,
+        RegisterCmds::Create { name } => create_register(name, &client).await?,
+        RegisterCmds::Edit { name, entry } => edit_register(name, entry, &client).await?,
+        RegisterCmds::Get { names } => get_registers(names, &client).await?,
     }
     Ok(())
 }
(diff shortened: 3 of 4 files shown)


oetyng:
Hm, I think there is some misunderstanding here. Deposit is absolutely possible to use.

Genesis is reissued to a faucet. A DBC is produced for every transfer from the faucet. It is printed in hex format to a file.
That file is sent to recipient out of band, who places it in the required path, and is then deposited to recipient wallet with the `deposit` cmd.

Any user who deposited a DBC from a faucet, can subsequently transfer in the same way, and send a DBC out of band to be deposited by another recipient.

iancoleman:
> A DBC is produced for every transfer from the faucet. It is printed in hex format to a file.

Ah thanks for clarifying! I wasn't aware the faucet worked this way. Is there a reason the faucet doesn't do a tx directly (ie why is it out of band?)

> Any user who deposited a DBC from a faucet, can subsequently transfer in the same way, and send a DBC out of band to be deposited by another recipient.

Is there a way for the wallet to transfer out of band? I couldn't work out how to create a dbc file using the wallet cli.

oetyng:
> Ah thanks for clarifying! I wasn't aware the faucet worked this way. Is there a reason the faucet doesn't do a tx directly (ie why is it out of band?)
> Is there a way for the wallet to transfer out of band? I couldn't work out how to create a dbc file using the wallet cli.

There is no comms or any features outside of the wallet yet that can facilitate a send of anything to anyone. No messaging nothing of the sorts.

The faucet cmds are described briefly in [README.md](https://github.com/maidsafe/safe_network/blob/main/README.md).

So the flow is the following:

1. The entity starting the network claims the genesis into a faucet, and thereby assumes the role of a faucet (see [faucet.rs](https://github.com/maidsafe/safe_network/blob/main/safenode/src/bin/faucet.rs)).
2. A user calls `safe wallet address` to get their wallet address.
3. User sends this address via any sort of interface to the faucet. (Chris was about to develop a web app and aws lambda for providing this service on testnets.)
4. `faucet send [amount] [address]` produces a hex string that is printed to stdout (see [faucet.rs](https://github.com/maidsafe/safe_network/blob/main/safenode/src/bin/faucet.rs)).
5. The hex string is sent in any way to user (in the web app it would be in a text box or downloaded as file).
6. User places the file in `received_dbcs` dir path (see [wallet_file.rs](https://github.com/maidsafe/safe_network/blob/main/safenode/src/domain/wallet/wallet_file.rs)) and calls `safe wallet deposit`.

Now the user can..

1.  ..call `safe wallet send [amount] [address]`
2. .. go to `created_dbcs` dir, find `public_address_[address]` dir, and find the created dbc file there (see [wallet_file.rs](https://github.com/maidsafe/safe_network/blob/main/safenode/src/domain/wallet/wallet_file.rs)).
3. Send the file to recipient via chat, email, pidgeon etc.
4. Recipient user repeats from step 6. above.

***

So, sending tokens means consuming at least one dbc you have, and producing a new dbc. It is outputted to `created_dbcs` dir.
Since `safe_network` has no messaging feature yet, a user has to manually take that dbc file and send it over to recipient in any preferred way.

happybeing:
What's the difference between the faucet and a wallet?

oetyng:
> What's the difference between the faucet and a wallet? 

The faucet has a wallet, it is a wallet you could say.
The cli faucet is a convenience api to guide the usage basically, wrt claiming the genesis and then simply just send out tokens.

It's sort of documentation by code to show what the steps are to start handling tokens.

The faucet that claims genesis in a test network, could also dole out larger chunks to other (community run) wallets, who can then act as faucets themselves.

It's just that the first thing necessary - since genesis dbc is publicly known - is for the entity who starts a network to claim the genesis, so it isn't taken by any random user, who would then possess the full supply :)

For the live network the first faucet claiming genesis is run by maidsafe. All initial distribution is made from that wallet.

iancoleman:
OK this makes it clear that 'deposit' is a crucial function. I'll revert that commit.

It feels more natural to call that function 'sync' since that would lend itself to naturally flowing into the remote side when it becomes implemented. For now though I'll just leave it as 'deposit', this can be changed when needed.

Thanks for the explanation.

iancoleman:
Still needs a bit of work because `peers` is required even for local operations, I'll keep working on this to make `safe wallet receive` run locally and not need any network info.

iancoleman:
The client can show the current wallet address without attempting to contact the network using `safe wallet receive`

`safe wallet address` has been renamed `safe wallet receive` to be a more intuitive pair with `safe network send`.

safenode/src/bin/safe/mod.rs
```diff
@@ -29,25 +30,28 @@ async fn main() -> Result<()> {
...
-        SubCmd::Files(cmds) => files_cmds(cmds, client.clone(), &root_dir).await?,
-        SubCmd::Register(cmds) => register_cmds(cmds, &client).await?,
+        SubCmd::Wallet(cmds) => wallet_cmds(cmds, peers, &root_dir).await?,
+        SubCmd::Files(cmds) => files_cmds(cmds, peers, &root_dir).await?,
+        SubCmd::Register(cmds) => register_cmds(cmds, peers).await?,
     };
 
     drop(log_appender_guard);
     Ok(())
 }
```
grumbach, line 49: Not introduced in this PR but is this drop necessary?

safenode/src/bin/safe/mod.rs
```diff
@@ -29,25 +30,28 @@ async fn main() -> Result<()> {
...
 
+async fn get_client(peers: Vec<Multiaddr>) -> Result<Client> {
+    println!("Connecting to network...");
+    let peers = peers_from_opts_or_env(&peers)?;
+    let secret_key = bls::SecretKey::random();
+    Ok(Client::new(secret_key, Some(peers)).await?)
```
grumbach, line 52: Does this mean we use a new bls key for each command we type?

grumbach (commented):
Agree on the renamings, nice one!

oetyng:
> safe wallet address has been renamed safe wallet receive to be a more intuitive pair with safe network send.

I would think that `receive` - just like `send` -  is a quite active verb, more related to actually _receiving_ a dbc. E.g. connecting to network and getting dbc through a (not yet implemented) "comms channel".

To me, `deposit`, `send`, `sync` and `receive` are in the same logical group, where `sync` is very general and could mean both `send` + `receive` or either. Too general even.
In the end, `deposit` and `sync` are the two that don't fit in the group, and we are left with the natural antonym pair of `send` and `receive`. And it is by that obvious that `receive` does not mean "print address".

The action of printing the wallet address to `stdout` I think is best represented by a cmd name like `address`, because all you do is to show the address. You don't _receive_ anything by that, and it's not even necessarily _to receive_ that you print it. It's basically just to _show_ the address. You might want to write it down somewhere, share to someone's contact book or what ever.

Then `receive` and `send` can be used for the two sides of syncing dbcs via the network (where optionally `sync` could do both..).

***

There is some shuffling inherent in this.
In a higher view flow, you'd imagine a use case like receving tokens, that'd consist of multiple wallet steps.

One example:

1. Show a QR code on your mobile to sender mobile.
2. Scan a QR from sender mobile into your mobile.

In step 1. we as recipient call `address` and then `display` it as a QR.
The sender scans it and `creates` a tx (signed spend) which is then `displayed` as a QR.
In step 2. we scan the QR and `sync` the signed spend to the network and `insert` the dbc to our wallet.
Note what happens in which step..
In step one we didn't receive anything. We displayed an address. In step 2 however, we _did_ rerceive something. Additionally we synced something.

So, what of all this is "send", " receive", "address", "display", "insert", "desposit" (etc..)?
Preferrably the interfacing with the wallet is sufficiently granular to not bunch too much of it into a single cmd.

joshuef:
> In step one we didn't receive anything. We displayed an address. In step 2 however, we did rerceive something. > Additionally we synced something.

To me, it's step 1.5 where we receive something. As it's _sent_ there, and committed to the spendbook etc. So that DBC value is now belonging to that address. 

Syncing to the wallet is perhaps _another_ step, thereafter?

---

I see what you're saying w/r/t `address/receive` @oetyng . `address` or `generate receive address` are perhaps more literally correct.

But, `receive` _feels_ pretty easy to understand to me (though perhaps not wholly clear with how/why/what you do with it; but more understandable than just `address` for me).

I think w/r/t `receive/sync` it kind of depends what the wallet is actually doing under the hood. If `receive` awaits for some info to be transmitted (via QR code, via writing DBC to inbox), then that's valid. And that's probably where we should be aiming down the line? 

-----
`receive` feels like an improvement to me over `address`, and it's not something we have to wed ourselves too. There's going to be more UX work on this at some juncture I expect (cc @JimCollinson ).

oetyng:
MetaMask has `Share Public Address` menu bar option, which opens the public address in the phone native share pane, where it can be copied, or inserted to any app of choice, where a bunch of default apps are listed first (different messenger apps etc).

That is almost the exact same functionality as the one discussed here - and it focuses on `share` (or the logically close `show`). Not `receive`.

MetaMask is a very popular wallet.

oetyng:
Yes, good point that there's no wedding with specific steps.

I don't want to block this by the discussions, just chip in with the views. Everything can be (and is) improved on iteratively.

There are some interesting correlations in `MetaMask` wallet, that I want to look into.

So, preferrably we keep as close as possible to established good patterns, without mistakenly go down into proliferated bad patterns (those existing in current cryptocurrency sphere).

There's also an actual difference in flow between SN dbc and btc/eth/monero (say), that we'll have to take into account down the line (no stress, it's developing).

I'll circle back in whatever place (PR/issue) is most suitable at the time with further findings.

iancoleman:
This PR is too small and the main part of it (delaying client creation until needed) is easy to put in later. I'm going to close this. Mostly I wanted a reason to get my hands on the code and I've achieved that goal without this needing to be merged.

Issue #260, opened 2023-05-11
Improve republish


joshuef:
Can we improve data republish _cost_ to us?

It'd be ideal to know if/when a node has gone offline to _flag_ certain data that they held for republish. And _only_ republish that. 

This would likely involve a filter in the `RecordStore::records` impl (as we have the random data there now). 

Are there other options here?

joshuef:
#271 goes some way here, #279 Would finish this off

joshuef:
Closing this as #279 is underway

PR #263, opened 2023-05-11
feat: only trigger replication when peer churning in close range

maqi:
This PR contains the work of only trigger replication when close range peer is churning.
This is aim to reduce the replication workload.

There are two timing spots that replication shall be triggered: 
a newly added peer falls into our close range; one of our close range peer dropped out

Depends on PR https://github.com/maidsafe/safe_network/pull/264

Related to issue https://github.com/maidsafe/safe_network/issues/260

### DIFF
--- a/safenode/src/domain/storage/disk_backed_record_store.rs
+++ b/safenode/src/domain/storage/disk_backed_record_store.rs
@@ -16,10 +16,13 @@ use libp2p::{
         store::{Error, RecordStore, Result},
     },
 };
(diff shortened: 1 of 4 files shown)


joshuef (approved):

joshuef:
The next step once this is in, @maqi is probably to remove the randomness and detect what data the lost peer should have been involved in and be sure to replicate that. What do you think? (I don't mean for this PR per se)

joshuef (approved):

PR #264, opened 2023-05-11
feat: dead peer detection

maqi:
This PR contains the work of detection dead peers and remove it from the kad.

Note: this relies on `flying requests among nodes`. i.e. the detection only works when there is data/requests being exchanged among the network.
A quiet network, say startup a network of 25 nodes locally, then kill one of it, won't trigger the detection.

### DIFF
--- a/safenode/src/network/event.rs
+++ b/safenode/src/network/event.rs
@@ -282,9 +282,22 @@ impl SwarmDriver {
                 info!("Connection closed to Peer {peer_id} - {endpoint:?} - {cause:?}");
             }
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
+                info!("Having OutgoingConnectionError {peer_id:?} - {error:?}");
                 if let Some(peer_id) = peer_id {
                     if let Some(sender) = self.pending_dial.remove(&peer_id) {
+                        info!("OutgoingConnectionError is due to a pending_dial to {peer_id}");
                         let _ = sender.send(Err(error.into()));
+                    }
+                    // A dead peer will cause a bunch of `OutgoingConnectionError`s
+                    // to be received within a short period.
+                    if let Some(value) = self.potential_dead_peers.get_mut(&peer_id) {
+                        *value += 1;
+                        if *value > 3 {
+                            trace!("Detected dead peer {peer_id:?}");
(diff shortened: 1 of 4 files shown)


safenode/src/network/event.rs
```diff
@@ -273,10 +273,23 @@ impl SwarmDriver {
                 info!("Connection closed to Peer {peer_id} - {endpoint:?} - {cause:?}");
             }
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
+                info!("Having OutgoingConnectionError {peer_id:?} - {error:?}");
```
joshuef, line 285: probably `warn` is better here?

safenode/src/network/event.rs
```diff
@@ -273,10 +273,23 @@ impl SwarmDriver {
...
                 if let Some(peer_id) = peer_id {
                     if let Some(sender) = self.pending_dial.remove(&peer_id) {
+                        info!("OutgoingConnectionError is due to a pending_dial to {peer_id}");
```
joshuef, line 288: Perhaps we can only log the error if we know it's _not_ due to another pending dial?

safenode/src/network/event.rs
```diff
@@ -273,10 +273,23 @@ impl SwarmDriver {
...
                         let _ = sender.send(Err(error.into()));
                     }
+                    // A dead peer will cause a bunch of `OutgoingConnectionError`s
+                    // to be received within a short period.
+                    if let Some(value) = self.potential_dead_peers.get_mut(&peer_id) {
+                        *value += 1;
```
joshuef, line 294: this `1` and `3`, can we `const` them up with some explanation of their significance please 🙇

safenode/src/network/mod.rs
```diff
@@ -264,6 +266,10 @@ impl SwarmDriver {
             pending_get_closest_peers: Default::default(),
             pending_requests: Default::default(),
             pending_query: Default::default(),
+            potential_dead_peers: LruCache::with_expiry_duration_and_capacity(
+                Duration::from_secs(10),
```
joshuef, line 289: can we name and document these values too please 🙇

joshuef (approved):

PR #265, opened 2023-05-11
fix(client): Register retrieval API was expecting incorrect error type

bochaco:
- Register retrieval API was not handling error responses correctly, thus when no success response was received it was reporting always the same `Error::UnexpectedResponses` variant.
- Also adding context info to `Network::Error::NotEnoughPeers` variant.

### DIFF
--- a/safenode/src/network/error.rs
+++ b/safenode/src/network/error.rs
@@ -68,8 +68,8 @@ pub enum Error {
     #[error("The oneshot::sender has been dropped")]
     SenderDropped(#[from] oneshot::error::RecvError),
 
-    #[error("Could not get CLOSE_GROUP_SIZE number of peers.")]
-    NotEnoughPeers,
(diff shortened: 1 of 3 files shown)


safenode/src/network/error.rs
```diff
@@ -68,8 +68,8 @@ pub enum Error {
     #[error("The oneshot::sender has been dropped")]
     SenderDropped(#[from] oneshot::error::RecvError),
 
-    #[error("Could not get CLOSE_GROUP_SIZE number of peers.")]
-    NotEnoughPeers,
+    #[error("Could no get enough peers ({required}) to satisfy the request, found {found}")]
```
joshuef:
```suggestion
    #[error("Could not get enough peers ({required}) to satisfy the request, found {found}")]
```

joshuef (approved):

PR #266, opened 2023-05-12
ci: add Ci job to check the memory usage

joshuef:


.github/workflows/memcheck.yml
```diff
@@ -0,0 +1,150 @@
...
+        if: failure()
+        continue-on-error: true
+
+
+      - name: Upload Heaptrack
+        uses: actions/upload-artifact@main
+        with:
+          name: heaptrack_safenode
+          path: heaptrack.safenode.*
+        continue-on-error: true
```
maqi: need an extra line break at the end of the file to get rid of the `red stop icon`

.github/workflows/memcheck.yml
```diff
@@ -0,0 +1,150 @@
...
+      # Start a heaptracked node instance to compare memory usage
+      - name: Start safenode with heaptrack
+        run: |
+          mkdir -p ~/.safe/heapnode
+          heaptrack ./target/release/safenode --root-dir ~/.safe/heapnode --log-dir ~/.safe/heapnode &
+          sleep 10
+
+      
+      - name: Start a client to upload files
+        run: cargo run --bin safe --release -- files upload -- "./resources"
```
maqi, line 71: "./resources" may change with the content under it.
for example at the time of I used it first time for e2e upload test, there is only one script file.
But now it has 4 script files and 2 key files.
This will affect the mem usage in this test later on.

Better put a comment here to explain the relationship and may restrict it to a subfolder or just a particular file.

.github/workflows/memcheck.yml
```diff
@@ -0,0 +1,150 @@
...
+          pkill safenode
+          echo "$(pgrep safenode | wc -l) nodes still running"
+
+      - name: Check for heaptrack file
+        run: ls -la
+
+      - name: Check memory usage
+        shell: bash
+        env:
+          MEM_LIMIT_MB: 10
```
maqi: As mentioned perviously, "./resources" may change with the content under it.

Better put a comment here to explain the relationship and may restrict the upload to a subfolder or just a particular file,
to make the threshold more stable.

maqi (changes requested):
Seems the test only got run as a task of merge queue ?
It has almost the same steps of e2e test, but just only on Ubuntu.
Wondering if shall make it as a CI run for PR as well, then remove the e2e run on Ubuntu.
Which will have the roughly same overall execution time, but can have the memory alert earlier ?

maqi (approved):

PR #269, opened 2023-05-16
chore: only wait for one OK response when client store chunk

### LINKED ISSUE #247: Improve Client store chunk performance
Current, when a client storing a chunk, it first get closest peers, send send out request to each and wait on all ALL responses (Ok or non-Ok) till complete and return.
This will make the flow slow when connection to one of the nodes is in problem, i.e. a response only got returned after TimeOut.

The performance can be improved by using the similar function as to `expect_closest_majority_ok` with the dbc spend flow, which will erturn earlier whenever received majority OK responses. 
OR even being more agressive to return on the first OK response.

maqi:
This PR improves the client chunk store performance by only wait for one OK response.

Closes https://github.com/maidsafe/safe_network/issues/247

### DIFF
--- a/safenode/src/client/api.rs
+++ b/safenode/src/client/api.rs
@@ -153,28 +153,15 @@ impl Client {
     pub(super) async fn store_chunk(&self, chunk: Chunk) -> Result<()> {
         info!("Store chunk: {:?}", chunk.address());
         let request = Request::Cmd(Cmd::StoreChunk(chunk));
-        let responses = self.send_to_closest(request).await?;
+        let response = self.send_and_wait_till_first_rsp(request).await?;
 
-        let all_oks = responses
-            .iter()
-            .filter(|resp| matches!(resp, Ok(Response::Cmd(CmdResponse::StoreChunk(Ok(()))))))
-            .count();
-        if all_oks >= close_group_majority() {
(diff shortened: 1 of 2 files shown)


safenode/src/client/api.rs
```diff
@@ -204,7 +203,21 @@ impl Client {
...
+            .into_iter()
+            .map(|res| res.map_err(Error::Network))
+            .collect_vec();
+        Ok(responses)
+    }
+
+    pub(crate) async fn send_and_wait_till_first_rsp(
+        &self,
+        request: Request,
+    ) -> Result<Vec<Result<Response>>> {
```
RolandSherwin: Just a simple one, but can we return a single response instead of a vec here?

maqi: there is chance some peer may response with error.
hence will be better `retain the capability of return with multiple responses`

RolandSherwin: If I'm not wrong, the underlying `Network::send_and_get_responses()` returns the **first successful response** if `expect_all_responses=false` or will return `vec<error>` if every node returned an error.

If we don't care about **all the errors**, then maybe we can dump the `Vec` here?

joshuef:
I'm not sure we want this. Or so little? It maybe puts too much work on the receiving node, rather than the client at least hitting majority?

RolandSherwin (approved):

PR #270, opened 2023-05-16
fix(node): perform random get_closest query on join

RolandSherwin:


safenode/src/node/api.rs
```diff
@@ -100,8 +101,10 @@ impl Node {
             transfer_actor: transfer_action_sender,
         };
 
+        let network_clone = network.clone();
         let _handle = spawn(swarm_driver.run());
         let _handle = spawn(async move {
+            let mut random_query_count = 0;
```
maqi: this shall be inside the loop

maqi (changes requested):

safenode/src/node/api.rs
```diff
@@ -116,6 +119,18 @@ impl Node {
                             None => error!("The `TransferAction` channel is closed")
                         }
                     }
+                    _ = tokio::time::sleep(INACTIVITY_TIMEOUT) => {
+                        if random_query_count < MAX_RANDOM_CLOSEST_QUERY_ATTEMPTS {
+                            let random_target: Vec<u8> = (0..64).map(|_| rand::random::<u8>()).collect();
```
maqi: you can use `PeerId::random()` directly

maqi (approved):

joshuef:
Let's try using `our_id` on connect instead of a random one. @RolandSherwin @maqi , and see how that goes. (Marking as DO NOT MERGE until we know more there)

maqi:
to make the nodes keep a healthy knowledge of the network during the quiet time, I'd suggest following change:
remove the MAX_ATTEMPTS (currently 5) cap, and fire a get_closest when there is a random inactivity period (say 10-30 seconds, make it random to avoid all nodes fire at the same time) experienced.

safenode/src/node/api.rs
```diff
@@ -116,10 +117,23 @@ impl Node {
                             None => error!("The `TransferAction` channel is closed")
                         }
                     }
+                    _ = tokio::time::sleep(INACTIVITY_TIMEOUT) => {
```
maqi: better make this bit random, say in a range of 20-40 seconds.
to avoid nodes firing get_closest at the same time due to `event handling syncing`

safenode/src/node/api.rs
```diff
@@ -116,10 +123,23 @@ impl Node {
...
+                        if let Ok(closest) = network_clone.node_get_closest_peers(&random_target).await {
+                            debug!("Network inactivity: get_closest returned {closest:?}");
+                        }
+                    }
                 }
             }
         });
 
+        // perform a get_closest query to self on node join. This should help populate the node's RT
+        let _ = network
```
maqi: At this point, `network` instance may not discover any peer yet?
I think this query shall only got fired on the first time of received `PeerAdded`/`ConnectedToNetwork` ?

RolandSherwin: Sorry, you're right. Got `Not enough peers in the k-bucket to satisfy the request`. Updating it.

maqi (approved):

joshuef (approved):

PR #271, opened 2023-05-16
fix(node): implementation of RecordStore::records() libp2p trait fn was not returning all records

bochaco:


joshuef:
I think, while you're rigth, our impl is off trait. It's probably what we want _just now_. And we should try and figure out how we want to push replication going forwards.

Do we want to stay w/ that logic in libp2p (ideally yes). And if so, what do we need to do to libp2p to make it work for us (I'm thinking PRing upstream). 

For clarity: As I understand things, libp2p pulls in all records _every_ replication interval and sends them to _all_ relevant nodes. Regardless if they have them or other ndoes might do this. For us, with larger `Records`, this is a massive memory hog.

So one option is to just dial back node size and record count. Spread things _thinner_ so to speak.

If we don't want to do that, something like the `random record selection` we have reduces message overlap between nodes and does not load as many `records` into mem each iteration... (while potentially still covering a large `Record` surface between all nodes)...

joshuef (approved):

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -257,6 +273,28 @@ impl RecordStore for DiskBackedRecordStore {
...
+pub(crate) struct RecordsIterator<'a> {
+    keys: hash_set::Iter<'a, Key>,
+    storage_dir: PathBuf,
+}
+
+impl<'a> Iterator for RecordsIterator<'a> {
+    type Item = Cow<'a, Record>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        for key in self.keys.by_ref() {
```
maqi, line 287: will this always return the `first` one?

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -228,13 +247,10 @@ impl RecordStore for DiskBackedRecordStore {
     }
 
     fn records(&self) -> Self::RecordsIter<'_> {
-        let mut records = vec![];
-        for key in self.replication_records.iter() {
-            if let Some(record) = self.get(key) {
-                records.push(record);
-            }
+        RecordsIterator {
+            keys: self.records.iter(),
```
maqi, line 251: if decided no longer using `replication_records` , then it shall be removed totally?

joshuef: > replication_records

I thiiink this could be useful in a follow up PR? Although, that's a PR you're looking at @maqi so let me know what you think

bochaco:
Superseded by #282

PR #274, opened 2023-05-17
fix(storage): removing panics from disk backed RecordStore

bochaco:


safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -246,19 +246,22 @@ impl RecordStore for DiskBackedRecordStore {
...
 
     fn providers(&self, _key: &Key) -> Vec<ProviderRecord> {
-        todo!("ProviderRecords are not used currently.")
+        // ProviderRecords are not used currently
+        vec![]
     }
 
     fn provided(&self) -> Self::ProvidedIter<'_> {
-        todo!("ProviderRecords are not used currently.")
+        // ProviderRecords are not used currently
```
maqi, line 259: better change it to `warn` log, so that we can trace the kad behaviour if interested?
same suggestion to other `todo` places

bochaco: i suspect this will flood the logs as these methods are called periodically by kad...not sure we need these log msgs as we don't implement this feature of RecordStore

maqi: it shall not. 
if does, the current `todo` shall already cause trouble.

bochaco: well, if we are not disabling it, then they are supposed to be called, they are part of a Trait which libp2p does callbacks on, so they will be called and unless there is a way to disable it, it will be called periodically and we should just ignore since by design we are not implementing that feature

We should never use `todo!`, `panic!`, `expect`, or `unimplemented!` in prod code, they all panic

maqi (approved):

PR #278, opened 2023-05-18
feat: log initial connection to the network, printing if client

joshuef:


RolandSherwin (approved):
Printing the commit hash would prevent some confusions, atleast for me!

safenode/src/client/api.rs
```diff
@@ -91,23 +92,31 @@ impl Client {
             match event {
                 ClientEvent::ConnectedToNetwork => {
                     info!("Client connected to the Network.");
+                    println!("Client successfully connected to the Network.");
```
maqi, line 94: user will only see this `println` , so actually only need to check the `logged_connected` flag here?
then don't need to pass the flag down and do other changes?

safenode/build.rs
```diff
@@ -5,9 +5,26 @@
...
+        .args(["rev-parse", "HEAD"])
+        .output()
+        .expect("Failed to execute git command");
+
+    let git_hash = String::from_utf8(output.stdout).unwrap().trim().to_string();
+
+    let mut file = File::create("src/git_hash.rs").expect("Failed to create git_hash.rs");
+    writeln!(
+        file,
+        "/// The git commit at build time\npub const GIT_HASH: &str = \"{}\";",
```
maqi: shall the detailed `build time` (i.e. `Instant::now()` in readable format) got written as well ?

joshuef: I guess that could be added if you think it's useful?

safenode/src/client/api.rs
joshuef: Ah yeh, clients already did it only once

safenode/build.rs
joshuef: I'm not sure it is? Does it matter if we have the commit? Gonna leave that off, we can always add it in later if desired 👍

PR #282, opened 2023-05-18
feat: Targeted replication

### LINKED ISSUE #279: Add in churn based, targetted replication
Atop the work in https://github.com/maidsafe/safe_network/pull/271 which resetablishes the libp2p standard, but shifts to use it as a replication backstop (staggered over all nodes to avoid mem issues).

We should aim to replicate data as/when our routing table changes.

We have the fn `pub(crate) fn trigger_replication(&mut self) {` which is called to track this, but it was set up to rely on libp2p republishing intervals, and also publishes only a random subset of data.

We should be able to improve this to:

- [ ] Find out what data has _new_ holders on every churn. 
- [ ] Send a message to that node _only_ with data we have for them.

Can we do this using existing messages? (Perhaps it has to be out of band w/r/t libp2p?)

maqi:
This PR trying to carry out targeted replication in churn, and using a periodic record replication as a backstop.

Closes https://github.com/maidsafe/safe_network/pull/271
Closes https://github.com/maidsafe/safe_network/issues/279

### DIFF
--- a/safenode/src/domain/storage/disk_backed_record_store.rs
+++ b/safenode/src/domain/storage/disk_backed_record_store.rs
@@ -238,11 +238,15 @@ impl RecordStore for DiskBackedRecordStore {
 
     // A backstop replication shall only trigger within pre-defined interval
     fn records(&self) -> Self::RecordsIter<'_> {
+        trace!(
+            "Periodic replication shall be triggered: {:?}",
+            self.replication_start + self.config.replication_interval < Instant::now()
+        );
         RecordsIterator {
             keys: self.records.iter(),
             storage_dir: self.config.storage_dir.clone(),
-            is_triggered: self.replication_start + self.config.replication_interval
-                < Instant::now(),
+            // Set to false to ensure periodic replication disabled.
+            is_triggered: false,
         }
     }
 
--- a/safenode/src/domain/storage/disk_backed_record_store.rs
+++ b/safenode/src/domain/storage/disk_backed_record_store.rs
@@ -6,11 +6,10 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::network::CLOSE_GROUP_SIZE;
 use libp2p::{
     identity::PeerId,
     kad::{
-        kbucket::Key as KBucketKey,
+        kbucket::{Distance, Key as KBucketKey},
         record::{Key, ProviderRecord, Record},
         store::{Error, RecordStore, Result},
     },
@@ -25,13 +24,10 @@ use std::{
     vec,
 };
 
-// Control the random replication factor, which means `one in x` copies got replicated each time.
-const RANDOM_REPLICATION_FACTOR: usize = CLOSE_GROUP_SIZE / 2;
-
 // Each node will have a replication interval between these bounds
 // This should serve to stagger the intense replication activity across the network
-pub(crate) const REPLICATION_INTERVAL_UPPER_BOUND: Duration = Duration::from_secs(180);
-pub(crate) const REPLICATION_INTERVAL_LOWER_BOUND: Duration = Duration::from_secs(10);
+pub(crate) const REPLICATION_INTERVAL_UPPER_BOUND: Duration = Duration::from_secs(540);
+pub(crate) const REPLICATION_INTERVAL_LOWER_BOUND: Duration = Duration::from_secs(180);
 
 /// A `RecordStore` that stores records on disk.
 pub(crate) struct DiskBackedRecordStore {
@@ -42,8 +38,6 @@ pub(crate) struct DiskBackedRecordStore {
     config: DiskBackedRecordStoreConfig,
     /// A set of keys, each corresponding to a data `Record` stored on disk.
     records: HashSet<Key>,
-    /// Records for the next replication.
-    replication_records: Vec<Key>,
     /// Time that replication triggered.
     replication_start: Instant,
 }
@@ -90,7 +84,6 @@ impl DiskBackedRecordStore {
             local_key: KBucketKey::from(local_id),
             config,
             records: Default::default(),
-            replication_records: Default::default(),
             replication_start: Instant::now(),
         }
     }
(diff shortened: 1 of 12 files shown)


safenode/src/node/api.rs
```diff
@@ -252,6 +252,9 @@ impl Node {
                 self.send_response(Response::Cmd(resp), response_channel)
                     .await;
             }
+            Cmd::Replicate(_) => {
+                warn!("Replication cmd shall not be received here");
```
bochaco: Does the msg shall actually be that the "Cmd is not processed here", "ignored"? because the Cmd effectively will be always received there as your are monitoring all the Cmds. Plus, because of the same reason, this shall not be a `warn`, at most it can be a `debug`, or just `trace`, there is nothing to warn about here as far as I can see?

bochaco (changes requested):

safenode/src/network/msg/mod.rs
```diff
@@ -33,12 +33,17 @@ impl SwarmDriver {
...
                 } => {
                     trace!("Received request with id: {request_id:?}, req: {request:?}");
-                    self.event_sender
-                        .send(NetworkEvent::RequestReceived {
-                            req: request,
-                            channel: MsgResponder::FromPeer(channel),
-                        })
-                        .await?
+
+                    if let Request::Cmd(Cmd::Replicate(ReplicatedData::Chunk(chunk))) = request {
```
bochaco: Why are we short-circuiting the req handling flow here? I think this introduces a design issue: some requests are handled here, some are handled somewhere else, others somewhere else....that's not a good design and will be error prone for sure.

bochaco (changes requested):

safenode/src/node/api.rs
maqi: `Cmd::Replicate` got handled within SwarmDriver already, which won't be sent to `Node` level.
Hence put it as a warning as it may indicates a leak of handling.

bochaco: yeah, I saw that, that is a huge problem, changing the flow of where things are handled, you'll need to handle the Cmd here since this is the place where Cmds are handled, otherwise you are breaking the design of the flow completely.

safenode/src/network/msg/mod.rs
maqi: Well, if using the normal flow on receiving a Request::Cmd, it will be:
   swarm_driver -> node -> handling -> swarm_driver
Where swarm_driver is calling the `disk_storage` to write the record to disk.
As the record is a chunk, which is bit mem consuming, hence made it short-cut here, to avoid an internal un-necessary trip.

It is just `Request` is a facility for peering messaging to be used, over which we can define which are for the Node-Node communication and which are for SwarmDriver-SwarmDriver communication. 

In case `Request::Cmd` is confusing, we can have something like `Request::PeerDirect` to communicate

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -246,10 +240,19 @@ impl RecordStore for DiskBackedRecordStore {
         }
     }
 
+    // A backstop replication shall only trigger within pre-defined interval
     fn records(&self) -> Self::RecordsIter<'_> {
+        // Only need to load portion of the records.
```
joshuef: Hmm, this is not the targetted replication. 

So if we're maintaining parity of `RecordStore`, this has to return all records.

So our periodic libp2p backed flow will do just that, and send out everything. (And so will likely be over a larger period).

Then above, we should probably avoid calling `records` to avoid pulling _everything_ in. 

If we can't track only the files/keys we want at a higher level, we may want to try upstreaming some helper fns to libp2p?

safenode/src/domain/storage/disk_backed_record_store.rs
```diff
@@ -227,14 +236,14 @@ impl RecordStore for DiskBackedRecordStore {
...
     fn records(&self) -> Self::RecordsIter<'_> {
-        let mut records = vec![];
-        for key in self.replication_records.iter() {
-            if let Some(record) = self.get(key) {
-                records.push(record);
-            }
+        RecordsIterator {
+            keys: self.records.iter(),
+            storage_dir: self.config.storage_dir.clone(),
+            is_triggered: self.replication_start + self.config.replication_interval
```
bochaco: I don't think we need to return an `RecordsIterator` when we already know here up front weather they will all be returned or not when iterated, isn't it simpler and cheaper to just return an empty iterator if this is `false` here? and otherwise return the `RecordsIterator` ?

joshuef:
Let's increase the backstop intervals. 10s is very low if we have event driven replication on the go. I'd say something like every 3 mins to 9 mins perhaps? (as a starting point), and we can evaluate how useful that is as we go forwards?

safenode/src/domain/storage/disk_backed_record_store.rs
maqi: has to return a `RecordsIterator` holding an `iter` to a non-temp hashset<key>.
will be tricky if not using `keys: self.records.iter()` to calm the clippy erros regarding lifetime.
Hence, prefer to stay with the current.

bochaco: true, then just return `RecordsIterator` with an empty `keys`, that'll be fine.

safenode/src/tests_e2e/data_with_churn.rs
```diff
@@ -135,14 +137,30 @@ async fn data_availability_during_churn() -> Result<()> {
...
     }
 
+    // The churning of storing_chunk/querying_chunk are all random,
+    // which will have a high chance that newly stored chunk got queried BEFORE
+    // the original holders churned out.
+    // i.e. the test may pass even without any replication
+    // Hence, we carry out a final round of query all data to confirm storage.
+    for net_addr in content.read().await.iter() {
+        println!("Final querying content at {net_addr:?}");
+        assert!(
```
bochaco, line 196: I'm not sure that last check we are adding to the churn test is valid...the test retries (set to 5 times now) when querying any piece of content failed, this is because a content is expected to fail at some point in time if one of the nodes holding a replica has been churned, so that's why the test retries trying to make sure a piece of content is effectively gone.
Now this last check we are adding is querying all content created but we don't retry, I think it's expected to fail since, even if replication is working among nodes, it could have been that 1 sec. earlier one of the content's replica was churned and the replication didn't take place just yet, but that doesn't mean it was not gonna take place a few secs later.

Perhaps for this final check we want to also retry each failure (`MAX_NUM_OF_QUERY_ATTEMPTS` times), with proper delays in between (`CONTENT_QUERY_RATIO_TO_CHURN`), till we can say it's lost...?....

Plus, maybe we want to stop the churning and content creation tasks before doing this last check.

bochaco (changes requested):

safenode/src/tests_e2e/data_with_churn.rs
maqi: This round of final query check is AFTER those data got queried and experienced 5 times query already.
The churning that happens in-bettwen or simutaneous shall not affect so much (fails all 8 copies) to causing a failure, if no-retry happens.
Actually, I think adding a retry may `cover` a potential issue ?
Same as to stop the con-current tasks.

bochaco: > This round of final query check is AFTER those data got queried and experienced 5 times query already.

no, the queries are done on random content, the querying task doesn't query all content but pick one at random each time

> Actually, I think adding a retry may cover a potential issue ?

no, we may be giving false negatives since replication may be taking place still, just that it needs more time to be triggered or completed

> The churning that happens in-between or simutaneous shall not affect so much (fails all 8 copies)

this is adding randomness to the result, you need to stop the churning as it could be they churn those nodes holding the content you are querying, just to make sure and remove the randomness where you don't want it
..in fact, keeping the churning task may affect replication in some fashion you don't want to affect at that last check...

joshuef (approved):

safenode/src/tests_e2e/data_with_churn.rs
joshuef: This should be sorted by my commits updating the churn test 👍

PR #284, opened 2023-05-19
Record based dbc spend

grumbach:
## Description

- #251 

### Summary generated by Reviewpad on 19 May 23 01:49 UTC
This pull request contains various changes to the codebase including updates to dependencies in the `Cargo.lock` file, changes to error handling in the SAFE protocol crate, modifications to network events and protocols, variable renames, and updates to structs, enums, and methods across multiple files. These changes also include removal of unused modules and code, deleting multiple files, and simplification of code.


safenode/src/protocol/error.rs
```diff
@@ -70,18 +72,26 @@ pub enum StorageError {
...
     #[error("Spend not found: {0:?}")]
     SpendNotFound(DbcAddress),
-    /// We failed to store spend
-    #[error("Spend was not stored: {0:?}")]
-    SpendNotStored(DbcAddress),
-    /// A double spend attempt was detected.
-    #[error("A double spend attempt was detected. Incoming and existing spend are not the same: {new:?}. Existing: {existing:?}")]
+    /// Insuficient valid spends found to make it valid, less that majority of closest peers
+    #[error("Insuficient valid spends found: {0:?}")]
+    InsuficientValidSpendsFound(DbcAddress),
```
joshuef: `Insufficient` ;)

safenode/src/node/spendbook.rs
```diff
@@ -0,0 +1,352 @@
...
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use libp2p::kad::{Record, RecordKey};
+use sn_dbc::{DbcTransaction, SignedSpend};
+use tokio::sync::RwLock;
+
+/// The entitiy managing spends in a Node
+#[derive(Default)]
+pub(crate) struct SpendBook {
```
oetyng, line 23: I think `SpendBook` is not the right level of abstraction here.

I was talking with Dirvine about this recently, and we agreed that the `SpendBook` is _the entire network_.

A `spend` is like a single entry (word) on one page of that book.
A single node holds a page out of the book. 

But a node doesn't hold the book.

grumbach: Interesting idea, what should we rename it to then?

grumbach:
Rebased into #309

PR #287, opened 2023-05-19
fix: ci churn tests.

joshuef:
Reduce use of local-discovery in CI.
Ensures node-1 does not churn so we can use that specifically for our SAFE_PEER and it's vaild over churn calls.

### Summary generated by Reviewpad on 19 May 23 21:25 UTC
This pull request fixes issue with CI churn tests, reducing the use of local-discovery and ensuring node-1 does not churn so it can be used specifically for SAFE_PEER. The changes include modifications to scripts and workflows.


joshuef:
Closing as #281 sorts this out using a better testing method all round

PR #289, opened 2023-05-21
feat(client): add a network redial feature; to redial if there's no activity.

joshuef:
Clients coudl hang if for some reason the initial dial did not go through.

Here we redial if there has been no network activity

### Summary generated by Reviewpad on 22 May 23 02:17 UTC
This pull request adds a network redial feature if there is no activity from the network. It also error outs if the `inactive-client-redials` feature is not enabled and there is inactivity from the network.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -87,35 +87,36 @@ impl Client {
                 }
 
                 info!("Client waiting for a network event");
-
-                tokio::select! {
-                    event = network_event_receiver.recv() => {
+                match tokio::time::timeout(INACTIVITY_TIMEOUT, network_event_receiver.recv()).await
+                {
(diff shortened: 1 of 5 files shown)


safenode/src/client/api.rs
```diff
@@ -76,16 +80,36 @@ impl Client {
                 }
 
                 info!("Client waiting for a network event");
-                let event = match network_event_receiver.recv().await {
-                    Some(event) => event,
-                    None => {
-                        error!("The `NetworkEvent` channel has been closed");
-                        continue;
+
+                tokio::select! {
```
bochaco: It's perhaps cleaner with `match tokio::time::timeout(INACTIVITY_TIMEOUT, network_event_receiver.recv()).await { ... }` ?

PR #298, opened 2023-05-24
feat: paying for Chunks to upload them

bochaco:


sn_cli/src/subcommands/files.rs
```diff
@@ -87,7 +92,25 @@ async fn upload_files(files_path: PathBuf, file_api: &Files, root_dir: &Path) ->
...
+            let chunks_addrs: Vec<_> = chunks
+                .iter()
+                .map(|c| sn_protocol::NetworkAddress::ChunkAddress(ChunkAddress::new(*c.name())))
+                .collect();
+
+            let (dbc, proofs) = wallet_client.pay_for_storage(chunks_addrs.iter()).await?;
+
+            // FIXME: provide the correct lemma and path for the current chunk being uploaded
+            let payment = PaymentProof {
+                dbc_id: bls::PublicKey::from_bytes(dbc.public_address().to_bytes())?,
```
oetyng, line 108: Small comment. Not sure if you're trying to achieve something else, and I'm misunderstanding, but the dbc id, if that's what you want, is gotten from `dbc.id()`. The public address is the recipient of the dbc.

(After having gone through all changes now, I'm guessing there's incomplete placeholder logic still?)

sn_cli/src/subcommands/wallet.rs
```diff
@@ -119,3 +132,54 @@ async fn send(amount: String, to: String, client: &Client, root_dir: &Path) -> R
...
+            let (_, chunks) = file_api.chunk_bytes(bytes)?;
+            chunks.iter().for_each(|c| {
+                chunks_addrs.push(NetworkAddress::ChunkAddress(ChunkAddress::new(*c.name())))
+            });
+        }
+    }
+
+    println!(">> PAYING FOR {} chunks", chunks_addrs.len());
+    match wallet_client.pay_for_storage(chunks_addrs.iter()).await {
+        Ok((new_dbc, proofs)) => {
```
oetyng, line 162: Is there just one dbc created as payment?
Shouldn't there be one per node in a close group or some such?

sn_domain/src/wallet/error.rs
```diff
@@ -17,6 +17,9 @@ pub enum Error {
     /// Failed to create transfer.
     #[error("Transfer error {0}")]
     CreateTransfer(#[from] crate::client_transfers::Error),
+    /// Could not build storage payment reason for content
+    #[error("Could not build storage payment reason for content: {0}")]
+    StoragePaymentReason(String),
```
oetyng, line 22: Think perhaps the error names would be better if it describes the error explicitly. Like `CouldNotBuildStoragePaymentReason` or anything like that? (Although that particular one is a rather long name..)

sn_domain/src/wallet/local_store.rs
```diff
@@ -168,6 +168,7 @@ impl SendWallet for LocalWallet {
         &mut self,
         to: Vec<(Token, PublicAddress)>,
         client: &C,
+        reason_hash: Option<Hash>,
```
oetyng, line 171: Theere should be a `DbcReason` type available (if it wasn't removed), which can be used all out through sn code for more type safety.

sn_node/src/api.rs
```diff
@@ -214,10 +214,73 @@ impl Node {
...
                 debug!("That's a store chunk in for :{:?}", addr.name());
 
+                // NOTE: there is a bug in proof.validate_with_data api: https://github.com/filecoin-project/merkletree/issues/95
+                info!(
+                    ">> Verifying payment proof for chunk store {:?} ...",
+                    addr.name()
+                );
+
+                // We need to fetch the PaymentProof's DBC to obtain the reason-hash
+                let dbc_id = payment.dbc_id;
```
oetyng, line 228: The `dbc_id` field could be of `DbcId` type, for type safety.

oetyng (commented):
Nice one to have this one finally coming along!

Would be great if you could write up the systematization/design of this feature in the PR description.

I'm guessing this is still very much wip,  but would be great to be able to follow how what you are striving for in each step connects to the end design goal! It would make it easier to chip in with ideas and tips!

PR #299, opened 2023-05-25
WIP - chore: extract smaller crates from `safenode` crate 

jacderida:
- e8e6241 **chore: introduce sn_cli crate**

- b82505b **chore: extract new sn_protocol crate**

- 72bb51b **chore: extract sn_domain crate**

  Some of the functionality in `dbc_genesis` had to be moved from `sn_domain` to the `client` module
  in `safenode` because it uses the client for creating and working with the faucet.

  The scope of various functions and modules were also changed from `pub(crate)` to `pub` so they
  can be used by other crates.

- 4c53c83 **chore: extract logging and networking crates**

- feed355 **chore: extract sn_client crate**

- 9658de4 **chore: move shared modules to sn_domain**

  This may not be the perfect place for them to be, but this code is shared between both client and
  node and it just seemed a little bit excessive for them to be in their own crate, since they are
  pretty much just two functions that are not really related to each other.

- 7b57509 **chore: extract sn_node crate**


joshuef (approved):

joshuef:
I'm going to force merge this, with its fix for churn + dead peers detection. 

[`62d8c5e` (#299)](https://github.com/maidsafe/safe_network/pull/299/commits/62d8c5e8d09462f44986dd7308dbb85e897ddb5f)

It'll get main to the new repo setup. We'll need to keep debugging this `double run` issue for churn

PR #307, opened 2023-05-29
feat: refactor replication flow to using pull model

### LINKED ISSUE #294: Swtich to `pull` model to further reduce communication traffic during replication
Even with the targeted replication, it's still 1/16 - 7/16 of data copies need to be sent to the peer, and that peer will need to accept 2xtotal copies from network within a shorten period.
That's a lot of simultanious traffic and potential a still high peak mem usage, expecial in case of periodic stopback replication, that a node will re-publish all its copies.

If do really need to resolve the mem usage issue in a perfect accurate way we may have to had a more complicated flow, for example:

1. Using the current targeted replication logic to find out which peer & copies shall the replication sent to
2. Instead of currently sent replication copy directly, just sent the list of key to the peer
3. When the peer received the list, if don't have the copy, fetch it from the holder
   To avoid fetching from multiple, the ongoing work need to be recorded. May need a cache like following:
```
   on_going_replications: BtreeMap<key, BTreeMap<PeerId, HolderStatus>>;
   enum HolderStatus {
       // A claimed holder, not queried yet
       pending,
       // Query has been sent, waiting for the response
       ongoing,
       // Failed to fetch a copy from the holder
       failed,
   }
```
4. In case of periodic replication, we do the same just sent to closest peers, with the list of keys that we think that peer shall hold.
    And leave the received peer to decide whether need to fetch the missing copies if detected one.

However, this seems like we did all the job of libp2p::record, and we can then discard it totally ?

maqi:
## Description

This PR contains the work to refactor replication flow to using pull model as described within th e issue https://github.com/maidsafe/safe_network/issues/294.
This will reduce the communication traffic a lot, with a much safer coverage, and flatten the peak traffic/mem usage via controlling the parallel fetching threads.

It also contains a work of refactor dead peer detection, which now is more deterministic.

### Summary generated by Reviewpad on 29 May 23 13:23 UTC
This pull request includes various changes across different files. The changes include updates to enums, new modules, additions to dependencies, and several code cleanups. The changes also introduce a `ReplicationFetcher` module for fetching data from holders on the network. Additionally, a new variant called `ReplicatedDataNotFound` was added to the `StorageError` enum. Finally, a new variant called `GetReplicatedData` was added to the `Query` and `QueryResponse` enums to allow clients to retrieve replicated data from the network.
 

Closes https://github.com/maidsafe/safe_network/issues/294

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -22,23 +22,29 @@ use libp2p::mdns;
 
 use libp2p::{
     autonat::{self, NatStatus},
-    kad::{kbucket::Key as KBucketKey, GetRecordOk, Kademlia, KademliaEvent, QueryResult, K_VALUE},
+    kad::{
+        kbucket::Key as KBucketKey, record::Key as RecordKey, GetRecordOk, Kademlia, KademliaEvent,
+        QueryResult, K_VALUE,
+    },
     multiaddr::Protocol,
     request_response::{self, ResponseChannel as PeerResponseChannel},
     swarm::{behaviour::toggle::Toggle, DialError, NetworkBehaviour, SwarmEvent},
     Multiaddr, PeerId,
 };
 use sn_protocol::{
-    messages::{Cmd, ReplicatedData, Request, Response},
-    storage::Chunk,
+    error::Error as ProtocolError,
+    messages::{Cmd, QueryResponse, ReplicatedData, Request, Response},
     NetworkAddress,
 };
 use sn_record_store::DiskBackedRecordStore;
 #[cfg(feature = "local-discovery")]
 use std::collections::hash_map;
-use std::collections::HashSet;
+use std::collections::{BTreeMap, HashSet};
 use tokio::sync::oneshot;
 use tracing::{info, warn};
+
+// To reduce the number of messages exchanged, patch max 500 replication keys into one request.
+const MAX_PRELICATION_KEYS_PER_REQUEST: usize = 500;
 
 // Defines how close that a node will trigger repliation.
 // That is, the node has to be among the REPLICATION_RANGE closest to data,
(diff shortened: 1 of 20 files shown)


sn_networking/src/event.rs
```diff
@@ -268,39 +270,45 @@ impl SwarmDriver {
...
-                            let _prior_none_value = self.potential_dead_peers.insert(peer_id, 1);
+                    // Related errors are: WrongPeerId, ConnectionRefused
+                    let err_string = format!("{error:?}");
+                    let is_wrong_id = err_string.contains("WrongPeerId");
+                    let is_all_connection_refused = if let DialError::Transport(ref errors) = error
+                    {
+                        errors.iter().all(|(_, error)| {
+                            let err_string = format!("{error:?}");
+                            err_string.contains("ConnectionRefused")
+                            // if let TransportError::Other(err) = error {
```
joshuef: lets remove this commented code

sn_networking/src/event.rs
```diff
@@ -268,39 +270,45 @@ impl SwarmDriver {
...
+                    if is_wrong_id || is_all_connection_refused {
+                        self.try_trigger_replication(&peer_id, true);
+                        if self
+                            .swarm
+                            .behaviour_mut()
+                            .kademlia
+                            .remove_peer(&peer_id)
+                            .is_some()
+                        {
+                            trace!("First time detected dead peer {peer_id:?}");
```
joshuef: Is this the first time? Could it not happen again?

joshuef (commented):
I'm not realllly sure how to evaluate this. What are our benchmarks vs the current implementation? 

There's 500 lines added here more or less... But I'm not sure how necessary it is.

The current impl is pretty easy to reason about, eg. This feels as you say, probably less mem, but is it significant and worth the added complexity?

To be clear: I'm not against this, but I would like a concrete way to be measuring success on the replication front. (Passing a specific churn test bar, which we increase? While keeping mem lower eg?)

sn_networking/src/event.rs
maqi: yes, it shall be the first time.
It will only happen again when the peer join using the same id and dropped out again.
which, autually indicates some issue of the dead peer detection algorithm, in case we don't allow such `re-join with same peer_id` happens.

joshuef (approved):

PR #309, opened 2023-05-30
feat: Record based DBC Spends

grumbach:
## Description

- implements #251 
- refactor spend logic and checks
- add extra double spend checks
- use Kad Records instead of in-memory store for Spends
- stop broadcasting Spends/DoubleSpends to neighbours every time we receive them, instead ask neighbours only when we need to check

### Summary generated by Reviewpad on 30 May 23 06:20 UTC
This pull request updates the sn_dbc dependency from version 17.0.X to 18.0.0 with the serdes feature enabled in multiple files across the codebase. It also includes changes to the `sn_protocol`, `sn_client`, `sn_storage`, and `sn_node` modules. Specifically, it includes changes to the various enum and struct implementations, modifications to import statements, changes in error handling, deletion of code related to node transfers, reordering of struct methods, and addition of new methods and constants.


sn_client/src/api.rs
```diff
@@ -312,21 +313,39 @@ impl Client {
...
+                        match received_spend.verify(received_spend.spent_tx_hash()) {
+                            Ok(_) => {
+                                trace!("Verified signed spend got from network while getting Spend for {dbc_id:?}");
+                                ok_responses.push(received_spend);
+                            }
+                            Err(err) => {
+                                warn!("Invalid signed spend got from network while getting Spend for {dbc_id:?}: {err:?}.");
+                            }
+                        }
                     }
```
grumbach, line 325: Trust but verify what other nodes send us

sn_node/src/spendbook.rs
```diff
@@ -0,0 +1,352 @@
...
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use libp2p::kad::{Record, RecordKey};
+use sn_dbc::{DbcTransaction, SignedSpend};
+use tokio::sync::RwLock;
+
+/// The entitiy managing spends in a Node
+#[derive(Default)]
+pub(crate) struct SpendBook {
```
grumbach, line 23: @oetyng raised the legitimate question wether we should rename this to something else:

```
I think `SpendBook` is not the right level of abstraction here.

I was talking with Dirvine about this recently, and we agreed that the `SpendBook` is _the entire network_.

A `spend` is like a single entry (word) on one page of that book.
A single node holds a page out of the book. 

But a node doesn't hold the book.

_Originally posted by @oetyng in https://github.com/maidsafe/safe_network/pull/284#discussion_r1209093170_
```

sn_node/src/spendbook.rs
```diff
@@ -0,0 +1,352 @@
...
+                        warn!("Ignoring invalid double spend reported by dirty liar peer");
+                    }
+                    None
+                }
+                Response::Query(QueryResponse::GetDbcSpend(Err(e))) => {
+                    warn!("Peer sent us an error while getting spend from network: {e:?}");
+                    None
+                }
+                _ => {
+                    // TODO check what it means if we get a different response type
```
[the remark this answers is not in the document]
grumbach: Here the `Response::Query(QueryResponse::GetDbcSpend(Err(e)))` type should be simplified. 
The possible errors trimmed down to what is possible in this specific case. 
This type is too vast.

bochaco (approved):

sn_node/src/spendbook.rs
oetyng: Nothing complicated I think. It's just a store of spends, right?

So, like chunk store, it's spend store? But there are of course a possible number of variations of that.

joshuef: Given that _here_ it's just a lock, i'd call it `SpendLock` myself. 

I wouldn't sweat it too much here though, as discussed yesterday we should be bale to remove this lock entirely with #315

joshuef (approved):

PR #313, opened 2023-05-30
fix: small fixes to the CI script

maqi:
## Description

This PR contains small fixes to the Ci script:
1, Have an OS denpendent process killer
2, No longer tar client file, as the folder is OS dependent and the log file is not much useful anymore.


joshuef:
I actually find the client log quite useful so I'd like to keep that if we can please!

joshuef (approved):

PR #314, opened 2023-05-30
fix(node): correct dead peer detection

maqi:
The previous DialError was obsoleted, hence the dead peer detection needs to be refactored for a more solid error.

### Summary generated by Reviewpad on 30 May 23 10:53 UTC
This pull request fixes the dead peer detection in the node, which was using an obsolete DialError and required refactoring. The patch removes unused dependencies and decreases the replication range to improve the performance and ensures a corrected partially targeted replication.


joshuef:
@maqi can we add another regex check to churn to ensure that we're seeing dead peer detection happening. With at least one whole round of node churning?

joshuef (approved):

Issue #315, opened 2023-05-30
In Kad, store spends values in a Vec instead of a single SignedSpend


grumbach:
from

```rs
        let kademlia_record = Record {
            key: RecordKey::new(dbc_addr.name()),
            value: signed_spend_bytes,
            publisher: None,
            expires: None,
        };
```

to

```rs
        let kademlia_record = Record {
            key: RecordKey::new(dbc_addr.name()),
            value: vec![signed_spend_bytes],
            publisher: None,
            expires: None,
        };
```

Limit it to two. Maybe not a vec?

RolandSherwin:
I think https://github.com/maidsafe/safe_network/issues/318 can make it easier to call `PUT` more than once for the same key, as that would be the case here?

PR #319, opened 2023-05-31
feat: store double spends when we detect them

grumbach:
## Description

#315 

### Summary generated by Reviewpad on 31 May 23 10:33 UTC
This pull request includes changes to the SpendBook struct and several related functions. The changes mainly address the handling of double spend attempts, including the storage of both attempted spends in the book, and improvements to error messaging for invalid spends or spend parents.


joshuef:
I'm wondering if we need a test (or does it already exist??) to check: after doubel spend attempt, when we try and spend a child, do we get an error with both spends listed and see that this is dead money?

joshuef (approved):

Issue #310, opened 2023-05-30
Simplify Wallet code, get rid of Traits


grumbach:
Currently we have 6 Traits for wallet:

```rust
pub trait SendClient: Send + Sync + Clone {
pub trait VerifyingClient: Send + Sync + Clone {
pub trait Wallet {

pub trait SigningWallet {
pub trait SendWallet: DepositWallet {
pub trait DepositWallet: Wallet {
```

With only 7 uses:

```rust
impl SendClient for Client {
impl VerifyingClient for Client {
impl Wallet for LocalWallet {

impl SigningWallet for LocalWallet {
impl DepositWallet for LocalWallet {
impl SendWallet for LocalWallet {

impl SendClient for MockSendClient {
```

The intention was to enable user to implement their own wallet. 
This makes reading through wallet code very hard as we have to guess which implementation of each we are using when a function is called. As this code is not stable yet, we should simplify it to make progress/debugging easier on this code. 

At this stage, we should KISS and get rid of ALL those traits.

PR #327, opened 2023-06-02
refactor: get rid of wallet traits

grumbach:
## Description

- addresses #310 

### Summary generated by Reviewpad on 02 Jun 23 07:06 UTC
This pull request includes various changes made to different files in the project. Some notable changes include the removal of several traits and their implementations from one file, a modification to the faucet module that may impact its functionality, changes in the implementation of the `LocalWallet` struct and associated methods in another file, and the deletion of a file containing an implementation of the `SendClient` trait. Other changes include modifications to function signatures and restructuring of code. It is recommended that all changes are carefully reviewed.


sn_client/src/wallet/mod.rs
```diff
@@ -6,30 +6,56 @@
...
+pub struct WalletClient {
     client: Client,
-    wallet: W,
+    wallet: LocalWallet,
+    /// These have not yet been successfully confirmed in
+    /// the network and need to be republished, to reach network validity.
+    /// We maintain the order they were added in, as to republish
+    /// them in the correct order, in case any later spend was
+    /// dependent on an earlier spend.
+    unconfirmed_txs: Vec<TransactionOutputs>,
```
grumbach: This logic was in the `LocalWallet`, moved it to client code since it requires Network activity aka Client.

sn_client/src/wallet/mod.rs
```diff
@@ -6,30 +6,56 @@
...
+        println!("Sending transfer to the network: {transfer:#?}");
+        if let Err(error) = self.client.send(transfer.clone()).await {
+            println!("The transfer was not successfully registered in the network: {error:?}. It will be retried later.");
+            self.unconfirmed_txs.push(transfer);
+        }
+
+        // return created DBCs even if network part failed???
         match &dbcs[..] {
             [info, ..] => Ok(info.dbc.clone()),
             [] => Err(Error::CouldNotSendTokens(
```
grumbach: Left that behaviour as is but must admit I'm not too sure about this so left a comment, anyone know why we do this? (The wallet is already mutated at this point but the network might not know anything as of yet)

sn_client/src/wallet/mod.rs
```diff
@@ -38,12 +64,68 @@ impl<W: SendWallet> WalletClient<W> {
...
+            if self.client.send(transfer.clone()).await.is_ok() {
+                println!("Tx {:?} was successfully republished!", transfer.tx_hash);
+                let _ = self.unconfirmed_txs.remove(index);
+                // We might want to be _really_ sure and do the below
+                // as well, but it's not necessary.
+                // use crate::domain::wallet::VerifyingClient;
+                // client.verify(tx_hash).await.ok();
+            }
+        }
+    }
```
grumbach: Again moved from Wallet code

sn_client/src/wallet.rs
```diff
@@ -0,0 +1,152 @@
...
+        let dbcs = transfer.created_dbcs.clone();
+
+        // send to network
+        println!("Sending transfer to the network: {transfer:#?}");
+        if let Err(error) = self.client.send(transfer.clone()).await {
+            println!("The transfer was not successfully registered in the network: {error:?}. It will be retried later.");
+            self.unconfirmed_txs.push(transfer);
+        }
+
+        // return created DBCs even if network part failed???
```
oetyng, line 58: Hi Anselme!

As you say, this is not totally clear.

Above we don't yet have an evaluation of the error returned. It could be transient errors, or something that means the transfer can never be registered in the network for some reason (something invalid, or double spend, but the actual real paths I haven't looked at now).

All errors there lead to the transfer being added to pending list.
There are two options there, or maybe three, and they are mostly about UX.

`A.` Ensure it's only adding to pending list on transient/recoverable errors.
`B.` Allow removing transfers from the pending list (if by inspection it can be determined that it can't be registered, or... also even if user doesn't want to send it anymore).
`C.` Combine the two above.

So. Let's say it's a transient error.
The user still wants the created dbcs (and store in wallet, so they are not lost). User can decide if they want to resend the pending tx and confirm it before handing over the created dbcs. Or user hands over the dbcs and tries confirming later. (Depends on what the recipient/sender agrees on of course).

If it's not a transient error, then I think it's appropriate to handle that above at `ln 54`, and not add the tx to the pending list.
(However the UX might be improved by keeping the failed tx somewhere anyway, just as a history of actions - together with the error description/reason why it failed).
In this case, I also think the created dbcs should be dropped without ever being exposed to user, as they will never become valid.

oetyng, line 58: Hi @grumbach!

As you say, this is not totally clear.

Above we don't yet have an evaluation of the error returned. It could be transient errors, or something that means the transfer can never be registered in the network for some reason (something invalid, or double spend, but the actual real paths I haven't looked at now).

All errors there currently lead to the transfer being added to pending list (which will later be retried before any other send is executed).

There are two-three options there, for updating the logic, and they are mostly about UX.

`A.` Ensure it's only adding to pending list on transient/recoverable errors.
`B.` Allow removing transfers from the pending list (if by inspection it can be determined that it can't be registered, or... also even if user doesn't want to send it anymore).
`C.` Combine the two above.
(I have noted somewhere in a comment, that I think managing the pending list is probably a good UX.)

So. Let's say it's a transient error.
The user still wants the created dbcs (and store in wallet, so they are not lost). User can decide if they want to resend the pending tx and confirm it before handing over the created dbcs. Or user hands over the dbcs and tries confirming later. (Depends on what the recipient/sender agrees on of course).

If it's not a transient error, then I think it's appropriate to handle that above at `ln 54`, and not add the tx to the pending list.
(However the UX might be improved by keeping the failed tx somewhere anyway, just as a history of actions - together with the error description/reason why it failed).
In this case, I also think the created dbcs should be dropped without ever being exposed to user, as they will never become valid.

sn_transfers/src/client_transfers/mod.rs
```diff
@@ -55,7 +55,7 @@ pub struct Inputs {
 /// The created dbcs and change dbc from a transfer
 /// of tokens from one or more dbcs, into one or more new dbcs.
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-pub struct Outputs {
+pub struct TransactionOutputs {
```
oetyng: This here is not a trivial one, it comes back and bites us all the time.

I think though.... that perhaps `TransferOutputs` would be better for this one, considering the names in other parts, for consistency.

I've tried defining the difference, in some comment somewhere.
So far, the simplest convention I think is to only call the type `DbcTransaction` a `transaction`, and everything else, which is more general referring to the operations and steps, `transfer`.

What do you think?

sn_client/src/wallet.rs
grumbach: Thanks @oetyng your help is always appreciated!

What about rejecting it, and letting the user retry later on his own? 

> keeping the failed tx somewhere anyway, just as a history of actions

This could be logged straight away to the user instead of kept in the system?

sn_transfers/src/client_transfers/mod.rs
grumbach: Agreed, good one!

sn_client/src/wallet.rs
oetyng: Happy to help!

> What about rejecting it, and letting the user retry later on his own?

Absolutely, if the attempted `tx` is kept somewhere, and the included dbcs are kept aside so they are not used in any other `tx` in the meantime.

So, that'd be a sort of `pending_tx` list.

What to be aware of is to keep the exact same `tx` if an attempt to send it to network was made, because it could have succeeded partially (on some nodes), and then you are committed to that exact `tx`, as any other (using any of the included dbcs) would be a double spend.

I guess if we really want to be able to drop a `tx` that we tried and failed to send to network, we would need to make sure no node saw it as valid (and thereby will be propagating it).

(There is a post on discourse where mav describes the flow).

> This could be logged straight away to the user instead of kept in the system?

Sure. It's a UX design question I guess. If we're talking of a tx that is invalid and of no use to us, then - for simplicity - I guess not much of it even needs to be shown to the user? And instead logged if for debug purposes?
(This cli wallet probably won’t be super advanced, so a log of failed txs probably is not high on the feature list.)

bochaco (approved):

PR #328, opened 2023-06-02
fix(client): carry out client random query when inactivity

maqi:
### Summary generated by Reviewpad on 02 Jun 23 10:40 UTC
This pull request adds a client feature that carries out random queries when the client has been inactive for too long. It uses the `rand` crate and performs a random get_chunk query to target: {random_target}. The patch also includes updates to the Cargo and sn_client files.
 

Close https://github.com/maidsafe/safe_network/pull/322

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -37,6 +37,9 @@ use xor_name::XorName;
 /// The timeout duration for the client to receive any response from the network.
 const INACTIVITY_TIMEOUT: std::time::Duration = tokio::time::Duration::from_secs(10);
 
+/// The initial rounds of `get_random` allowing client to fill up the RT.
+const INITIAL_GET_RANDOM_ROUNDS: usize = 5;
+
 impl Client {
     /// Instantiate a new client.
     pub async fn new(signer: SecretKey, peers: Option<Vec<(PeerId, Multiaddr)>>) -> Result<Self> {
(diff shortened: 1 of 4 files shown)


oetyng:
I'm wondering if it’s worth specifying a specific `ping/pong` protocol?

It can seem like unnecessary (and perhaps is?), but the gains is that it's very easy to distinguish the requests when perusing logs, for example.

It can lower the cognitive load in such cases, and it’s a very small addition of types.

Just a thought.

maqi:
> I'm wondering if it’s worth specifying a specific `ping/pong` protocol?
> It can seem like unnecessary (and perhaps is?), but the gains is that it's very easy to distinguish the requests when perusing logs, for example.

Though the `ping/pong` protocol indicates a point-to-point communication more for `keep-alive-style`,
a similiar `query-response with network knowledge` type msg could be easier for track, and quicker to fill the kad RT.

However, this might go back to the previous join flow little bit, and have overlapped work with libp2p ?

sn_client/src/error.rs
```diff
@@ -46,6 +46,10 @@ pub enum Error {
     #[error("Client received no network activity for {0:?}")]
     InactiveClient(Duration),
 
+    /// Client received an unexpected event.
+    #[error("Client received an unexpected event")]
```
joshuef: Can we not handle this event? Why not? This seems a bit too catch all..

joshuef:
I think something like `RoutingTableUpdate<Address>` is a fair enough msg to have. It's what we're doing here and as @oetyng says it'll at least be clearer when we see that what's going on there. On receipt it could well just be exactly the same `get_closest` funcitonality, but at least we'll see that coming in and can see how often that happens (and perhaps more easily dig into why a client gets into that state?)

sn_client/src/api.rs
```diff
@@ -127,15 +127,35 @@ impl Client {
...
+                        }
+                    }
+                    break;
                 }
-                ClientEvent::InactiveClient(timeout) => {
-                    error!("Inactive client for {timeout:?}, inactive-client-redial is not set, and so shutting down");
-                    return Err(Error::InactiveClient(timeout));
+                _ = tokio::time::sleep(INACTIVITY_TIMEOUT) => {
+                    let mut rng = rand::thread_rng();
+                    let random_target = ChunkAddress::new(XorName::random(&mut rng));
```
joshuef, line 140: Why do we have this logic twice? Perhaps we can dedupe for clarity?

joshuef (changes requested):

sn_client/src/api.rs
maqi: the `ClientEvent::InactiveClient(timeout)` was from the redial, which I was not sure what it is about.
Meanwhile, this `` is the place I think shall be the best related to put the `random get` in.
Hence there are two places having the same logic put in.

Once the `redial` stuff got clarified (removal or retain), one of the `get random` places shall be removed.

sn_client/src/error.rs
maqi: That's just being too `safe`, can be removed actually

joshuef (approved):
Okay, let's get this in. 

But we may want to clarify this messaging in the protocol itself down the line

PR #334, opened 2023-06-05
Feat: chunk payment proof using a binary merkle tree

bochaco:
The payment is optional for now, i.e. the nodes verify the payment proofs only if they were provided, and they only verify if it was issued for the Chunk being stored, but **they don't yet** check if the node itself was payed with or if the input/parents were spent, that'd be for subsequent PRs. 
A new `wallet pay` CLI cmd is introduced, as well as a new flag `--pay` for files upload CLI cmd (by default files upload won't try to make the payment yet).

### Summary generated by Reviewpad on 05 Jun 23 19:23 UTC
This pull request includes multiple changes and updates across various files. These changes include adding new error types and variants related to payment proof verification, dependencies updates, modifications to functions to include payment proof arguments, and the addition of new modules and structs for generating and validating payment proofs. Overall, these changes aim to improve the functionality of the Safe Network API by enhancing payment proof verification and storage.


sn_transfers/src/payment_proof/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                path: proof.path().to_vec(),
+            },
+        );
+    }
+
+    Ok((reason_hash, payment_proofs))
+}
+
+/// Verify if the payment proof is valid and contains a valid audit trail for the given xorname
+pub fn validate_payment_proof(addr_name: XorName, payment: &PaymentProof) -> Result<()> {
```
joshuef, line 140: I'm wondering if there's good literature explaining the merkle tree and chunk addressing we can link to here. Would be good to include in the comments if you've come across anything decent @bochaco ?

sn_client/src/file_apis.rs
```diff
@@ -91,37 +95,64 @@ impl Files {
...
     /// without storing them onto the network.
     #[instrument(skip_all, level = "debug")]
     pub fn calculate_address(&self, bytes: Bytes) -> Result<XorName> {
-        chunk_bytes(bytes).map(|(name, _)| name)
+        self.chunk_bytes(bytes).map(|(name, _)| name)
+    }
+
+    /// Tries to chunk the bytes, returning the data-map address and chunks, without storing anything to network.
+    #[instrument(skip_all, level = "trace")]
+    pub fn chunk_bytes(&self, bytes: Bytes) -> Result<(XorName, Vec<Chunk>)> {
```
maqi, line 126: `self` is not required for the processing?

sn_transfers/src/payment_proof/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+    use super::*;
+    use eyre::{eyre, Result};
+    use sn_protocol::storage::ChunkAddress;
+    use tiny_keccak::{Hasher, Sha3};
+    use xor_name::XorName;
+
+    // Helper to generate the sha3-256 hash of the provided bytes, with the provided prefix.
+    // We use a prefix since that's what the 'merkletree' crate we use adds to each tree node,
+    // e.g. prefix '0' to leaves and prefix '1' to root node.
+    fn hash(prefix: u8, bytes_l: &[u8], bytes_r: &[u8]) -> [u8; 32] {
```
maqi: better using 32 as a const.
I assume it is `XOR_NAME_LENGTH` actually ?
same to the other places of using `32`, including in comments

sn_client/src/wallet.rs
```diff
@@ -64,6 +69,31 @@ impl WalletClient {
...
+        content_addrs: impl Iterator<Item = &NetworkAddress>,
+    ) -> Result<(Dbc, PaymentProofsMap)> {
+        // TODO: calculate the amount to pay to each node, perhaps just 1 nano to begin with.
+        let amount = Token::from_nano(1);
+
+        // FIXME: calculate closest nodes to pay for storage, we are now just burning the output amount.
+        let to = vec![(amount, PublicAddress::new(SecretKey::random().public_key()))];
+
+        // Let's build the payment proofs for list of content addresses
+        let (reason_hash, payment_proofs) = build_payment_proofs(content_addrs)
```
maqi, line 84: hmm...   I am bit lost here.
The function is `pay_for_storage` , which, as I understand, the `payment_proofs` shall be:
1, granted by the network nodes
OR
2, deduced from sender's balance

However, it seems the `build_payment_proofs` and `local_send` are all `local operations` which has no network actions involved.
Meanwhile, when the network nodes received the `payment_proofs` together with the chunk to be stored, it can only validate `the proof is related to that chunk` , but, how they can validate the `payment proof` is a valid one, so that they can accept the store?

sn_transfers/src/payment_proof/mod.rs
bochaco: I'll add a paragraph with some info and a link.

sn_client/src/file_apis.rs
bochaco: it's just to make the API nicer, in this way every api has the struct name or instance as namespace rather than isolated/scattered functions.

sn_client/src/wallet.rs
bochaco: I've added a comment in the description, this is just a first step, subsequent PRs will bring in additional payment verification/validations. One of these PRs will include the output DBC as part of the payment proof, nodes will be able to check input/parent DBCs were spent.

sn_transfers/src/payment_proof/mod.rs
joshuef: Lovely stuff 💪

joshuef (approved):

PR #335, opened 2023-06-06
fix(build): ensure to pull GIT_HASH via `env` call in apps directly

joshuef:
abstracting this call to the build_info lib meant that it was trying to fetch an actual env var at runtime, as opposed to the one supplied at build time

### Summary generated by Reviewpad on 06 Jun 23 00:04 UTC
This pull request fixes a build issue where the build_info library was fetching an actual env var at runtime instead of the one supplied at build time. The solution is to abstract the call to the build_info lib. The GIT_HASH is now pulled via `env` call in apps directly.

### DIFF
--- a/sn_build_info/src/lib.rs
+++ b/sn_build_info/src/lib.rs
@@ -5,12 +5,19 @@
 // under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
+use std::process::Command;
 
-/// The git commit pulled from the env var GIT_HASH (which is set during build at build.rs)
-pub fn git_hash() -> String {
-    if let Ok(git_hash) = std::env::var("GIT_HASH") {
-        git_hash
-    } else {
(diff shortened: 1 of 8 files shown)


sn_build_info/src/lib.rs
```diff
@@ -5,12 +5,19 @@
...
 
-/// The git commit pulled from the env var GIT_HASH (which is set during build at build.rs)
-pub fn git_hash() -> String {
-    if let Ok(git_hash) = std::env::var("GIT_HASH") {
-        git_hash
-    } else {
-        "---- No git commit hash found ----".to_string()
-    }
+/// set GIT_HASH env var to current commit
+pub fn pre_build_set_git_commit_env() -> Result<(), Box<dyn std::error::Error>> {
```
b-zee, line 11: This function never returns an `Result::Err`.

b-zee:
An alternative solution would be to include the `build.rs` with the `sn_build_info` crate. Then have a single function in that `lib.rs`:

```
pub fn git_hash() -> &'static str {
    env!("GIT_HASH")
}
```

That would mean `sn_cli` doesn't need a `build.rs`. `sn_node` already has a `build.rs` file.

joshuef:
You run into issues with cached builds then, sadly!

b-zee (approved):

PR #351, opened 2023-06-07
test: more replication flow statistics during mem_check test

maqi:
### Summary generated by Reviewpad on 07 Jun 23 08:34 UTC
This pull request includes changes to the 'mem_check' test that add more replication flow statistics.


sn_client/src/file_apis.rs
```diff
@@ -315,7 +315,7 @@ impl Files {
             let missing_chunks: Vec<XorName> = chunks_info
                 .iter()
                 .filter_map(|expected_info| {
-                    if !retrieved_chunks.iter().any(|retrieved_chunk| {
```
joshuef, line 318: What's this change about? Is it supposed to be in this PR?

maqi: that's a fix to a bug when throwing an error during the data_with_churn tests.
The error supposed to contain a list of `missing chunks` , but the previous log acutally put the `fetched chunks` in.

joshuef (approved):

PR #353, opened 2023-06-07
feat: bail out if empty list of addreses is provided for payment proof generation

bochaco:
### Summary generated by Reviewpad on 07 Jun 23 17:58 UTC
This pull request adds a feature that makes the code bail out if an empty list of addresses is provided for payment proof generation. The patch contains changes in multiple files to implement this feature.

### DIFF
--- a/sn_cli/src/cli.rs
+++ b/sn_cli/src/cli.rs
@@ -14,7 +14,19 @@ use crate::subcommands::SubCmd;
 #[derive(Parser)]
 #[command(author, version, about, long_about = None)]
 pub(crate) struct Opt {
-    /// Nodes we dial at start to help us get connected to the network. Can be specified multiple times.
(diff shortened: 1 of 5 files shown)


sn_cli/src/cli.rs
```diff
@@ -15,7 +15,7 @@ use crate::subcommands::SubCmd;
 #[command(author, version, about, long_about = None)]
 pub(crate) struct Opt {
     /// Nodes we dial at start to help us get connected to the network. Can be specified multiple times.
-    #[clap(long = "peer")]
+    #[clap(long = "peer", global = true)]
```
jacderida, line 30: Can we consider documenting this argument in the same way it is for the node?
https://github.com/maidsafe/safe_network/blob/0be2ef056215680b02ca8ec8be4388728bd0ce7c/sn_node/src/bin/safenode/main.rs#L70

jacderida (approved):

PR #358, opened 2023-06-08
feat: include output DBC within payment proof for Chunks storage

### LINKED ISSUE #342: Attach the output DBC when uploading a set of paid Chunks
Pass the output DBC when uploading Chunks. We now pass the reason-hash, audit trail, and path, which allow to only verify if it was issued for the Chunks being uploaded, but nodes are not able to check if they have been payed for that Chunk. This would allow them to also check parent/input DBCs.

bochaco:
Resolves #342 

### Summary generated by Reviewpad on 08 Jun 23 14:35 UTC
This pull request includes various changes across multiple files related to the `PaymentProofsMap` type. It adds new fields and modifies the return types of some functions. It also removes some unused imports and changes imports from one module to another. Additionally, it updates the `CmdResponse::StoreChunk` function in the `sn_protocol` module to accept payment proofs. Finally, it modifies the `PaymentProof` struct in the `sn_protocol` module to add a new field `dbc` and update the `audit_trail` field.

### DIFF
--- a/sn_protocol/src/messages/cmd.rs
+++ b/sn_protocol/src/messages/cmd.rs
@@ -14,7 +14,7 @@ use crate::{
 use super::RegisterCmd;
 
 // TODO: remove this dependency and define these types herein.
-pub use sn_dbc::{Hash, SignedSpend};
+pub use sn_dbc::{Dbc, Hash, SignedSpend};
 
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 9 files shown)


joshuef (approved):

sn_protocol/src/messages/cmd.rs
```diff
@@ -103,11 +103,14 @@ pub type MerkleTreeNodesType = [u8; 32];
 
 #[derive(Clone, Eq, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
 pub struct PaymentProof {
+    // Output DBC for nodes to check the Chunk payment is valid and inputs have
+    // been effectivelly spent on the network.
+    pub dbc: Dbc,
```
bochaco, line 108: ...I believe we'll need a `Vec<Dbc>` here actually, I'll take care of that on later PRs since this is likely to change anyways.

Issue #323, opened 2023-06-02
Enable `record_filtering` and perform validations


RolandSherwin:
- Enable `kad_config.set_record_filtering(KademliaStoreInserts::FilterBoth)` which would emit a `KademliaEvent::InboundRequest::PutRecord` while trying to store records into the data_store. 
- This `KademliaEvent` can be propagated to the node level, where we can deserialize the `Record` (#318) to get back the actual data. This can now be validated before storing it into the record store.

PR #361, opened 2023-06-08
impl `RecordHeader` and perform PUT validations

### LINKED ISSUE #318: impl `RecordHeader` to store the `RecordKind`
Currently, one can overwrite a stored DBC with a `PUT Chunk` call with `Record::key = DBC's key`. This would trigger the `PUT chunk` flow without any checks to stop it and would overwrite the key's value.

Thus, we can prepend a `RecordHeader` to `Record::value`.  Now, whenever we get a `PUT` call, we perform a `GET` for that key, partially deserialize the value to get the `RecordHeader::kind` and then decide on the next steps.

```rust
struct RecordHeader {
	pub kind: RecordKind,
}
enum RecordKind {
	Chunk,
	DBC,
	Register
}
```

RolandSherwin:
Closes #318 and #323 

- Implement `RecordHeader` to make sure we are not overwriting any `Record`
- Perform validations on PUT when we get them, comparing them with the local copy.
- Move `replication logic` to the node level. This allows us to perform validation on the `ReplicationData` as well.
- Trim `RecordHeader` and return only the requested item during `GET`
- Sending `Request` without awaiting a `Response` has been flushed out.


sn_protocol/src/storage/header.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl Display for RecordKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "RecordKind({self:?})")
+    }
+}
+
+impl RecordHeader {
+    // Bincode serializes enums with unit variants as a u32, not a u8, so it would take up 4 bytes.
```
joshuef: I don't think we can use bincode here. The protocol should be language agnostic and `bincode` is rust specific

joshuef (changes requested):

sn_protocol/src/storage/header.rs
RolandSherwin: Updated the PR to use `rmp_serde` instead.

sn_networking/src/lib.rs
```diff
@@ -520,51 +548,127 @@ impl Network {
...
             .map_err(|_e| Error::InternalMsgChannelDropped)
     }
 
-    /// Get `ReplicatedData` from local Storage
-    pub async fn get_replicated_data(
-        &self,
-        address: NetworkAddress,
-    ) -> Result<Result<QueryResponse>> {
+    /// Get the Record from the network
+    pub async fn get_record_from_network(&self, key: RecordKey) -> Result<Record> {
```
grumbach, line 589: Excellent renaming! Was always confused with the previous name!

sn_node/src/api.rs
```diff
@@ -188,131 +203,151 @@ impl Node {
...
+
+                        let success = self.validate_and_store_chunk(chunk_with_payment).await?;
+                        trace!("ReplicatedData::Chunk with {chunk_addr:?} has been validated and stored. {success:?}");
+                        addr
+                    }
+                    ReplicatedData::DbcSpend(spends_with_parent) => {
+                        // Put validations make sure that we have >= 1 spends and with the same
+                        // dbc_id
+                        let dbc_addr =
+                            DbcAddress::from_dbc_id(spends_with_parent[0].signed_spend.dbc_id());
```
grumbach: `[0]` is unsafe, might panic if the vec is empty

sn_node/src/put_validation.rs
```diff
@@ -0,0 +1,433 @@
...
+            expires: None,
+        };
+        self.network.put_local_record(record).await.map_err(|_| {
+            let err = ProtocolError::SpendNotStored(format!("Cannot PUT Spend with {dbc_addr:?}"));
+            error!("Cannot put spend {err:?}");
+            err
+        })?;
+
+        // Notify the sender of any double spend
+        if spends_with_parent.len() > 1 {
```
grumbach: Here we don't check if it is the same DBC, are we sure that it's not, aka are we sure it is a double spend and not twice the same dbc?

sn_node/src/put_validation.rs
```diff
@@ -0,0 +1,433 @@
...
+
+                }).collect::<Vec<_>>();
+
+                if spends_with_parent.is_empty() {
+                    warn!("No spend with valid Record key found. Ignoring DbcSpend PUT request");
+                    return Err(ProtocolError::RecordKeyMismatch.into());
+                }
+                // Since spends_with_parent is not empty and they contain the same DbcId (RecordKeys are
+                // the same), get the DbcId
+                let dbc_id = *spends_with_parent[0].signed_spend.dbc_id();
```
grumbach: Here I would rather use a match instead of relying on `[0]`which is a dangerous pattern

sn_node/src/put_validation.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Perform validations on the provided `Vec<SpendWithParent>`. Returns `Some<Vec<SpendWithParent>>` if
+    /// the spends has to be stored to the `RecordStore` where the spends are aggregated and can
+    /// have a max of only 2 elements. Any double spend error has to be thrown by the caller.
+    ///
+    /// The Vec<SpendWithParent> must all have the same dbc_id.
+    ///
+    /// - If the SpendWithParent for the provided DbcId is present locally, check for new spends by
+    /// comparing it with the local copy.
+    /// - If incoming spends_with_parent.len() > 1, aggregate store them directly as they are a double spent.
+    /// - If incoming spends_with_parent.len() == 1, then check for parent_inputs and the closet(dbc_id)
```
grumbach: `closet` typo

sn_node/src/put_validation.rs
```diff
@@ -0,0 +1,433 @@
...
+        present_locally: bool,
+    ) -> Result<Option<Vec<SpendWithParent>>, ProtocolError> {
+        // get the DbcId; used for validation
+        let dbc_addr = DbcAddress::from_dbc_id(&dbc_id);
+        let record_key = RecordKey::new(dbc_addr.name());
+
+        if present_locally {
+            debug!("DbcSpend with DbcId {dbc_id:?} already exists, checking if it's the same spend/double spend",);
+            // fetch the locally stored record; should be present
+            //
```
grumbach: Typo?

sn_node/src/put_validation.rs
```diff
@@ -0,0 +1,433 @@
...
+            }
+        }
+
+        // Check the parent spends and check the closest(dbc_id) for any double spend
+        // if so aggregate the spends and return just 2 spends.
+        let spends_with_parent = if spends_with_parent.len() == 1 {
+            debug!(
+                "Received a single SpendWithParent, verifying the parent and checking for double spend"
+            );
+            let spends_with_parent = spends_with_parent.remove(0);
```
grumbach: Here a match would be less dangerous than check then remove

sn_node/src/replication.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Setup the record storage distance range.
+        let sorted_peers: Vec<PeerId> = if let Ok(sorted_peers) =
+            sort_peers_by_address(all_peers.clone(), &our_address, CLOSE_GROUP_SIZE + 1)
+        {
+            sorted_peers
+        } else {
+            return Ok(());
+        };
+        let distance_bar =
+            NetworkAddress::from_peer(sorted_peers[CLOSE_GROUP_SIZE]).distance(&our_address);
```
grumbach: `sorted_peers[CLOSE_GROUP_SIZE]` from a `Vec<PeerId>` is unsafe

sn_node/src/replication.rs
```diff
@@ -0,0 +1,207 @@
...
+        // Plus 1 because the result contains self.
+        let sorted_peers: Vec<PeerId> = if let Ok(sorted_peers) =
+            sort_peers_by_address(all_peers, &churned_peer_address, 2 * CLOSE_GROUP_SIZE + 1)
+        {
+            sorted_peers
+        } else {
+            return Ok(());
+        };
+
+        let distance_bar = NetworkAddress::from_peer(sorted_peers[CLOSE_GROUP_SIZE])
```
grumbach: Same as above, I suggest we use another way like:
- pattern matching
- get() -> Option and check it

sn_node/src/replication.rs
```diff
@@ -0,0 +1,207 @@
...
+            let closest_peers: Vec<_> = if let Ok(sorted_peers) =
+                sort_peers_by_key(sorted_peers.clone(), &record_key, CLOSE_GROUP_SIZE + 1)
+            {
+                sorted_peers
+            } else {
+                continue;
+            };
+
+            // Only carry out replication when self within REPLICATION_RANGE
+            let replicate_range = NetworkAddress::from_peer(closest_peers[REPLICATION_RANGE]);
```
grumbach: same here, since we use this one a lot, maybe we could get it out in a variable and reuse it everywhere we need?

sn_node/src/spends.rs
```diff
@@ -0,0 +1,229 @@
...
+        .iter()
+        .any(|o| o.dbc_id() == spend_with_parent.signed_spend.dbc_id())
+    {
+        return Err(Error::InvalidParentTx(format!(
+            "The DBC we're trying to spend: {:?} is not an output of the parent tx: {:?}",
+            spend_with_parent.signed_spend, spend_with_parent.parent_tx
+        )));
+    }
+
+    // get the parent spends form the network
```
grumbach:
```suggestion
    // get the parent spends from the network
```

sn_protocol/src/storage/spend.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use serde::{Deserialize, Serialize};
+use sn_dbc::{DbcTransaction, SignedSpend};
+use std::hash::{Hash, Hasher};
+
+type ParentTx = DbcTransaction;
+
+/// The `SignedSpend` along with its corresponding `ParentTx`
+#[derive(custom_debug::Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SpendWithParent {
```
grumbach, line 17: With https://github.com/maidsafe/sn_dbc/pull/278 in, we should be able to get rid of this

grumbach: Maybe in a follow up PR?

grumbach (commented):
Big one, and a good one! Left a couple comments mostly about code safety, but generally looking good!

grumbach (approved):
LGTM 🎉

joshuef (approved):

PR #381, opened 2023-06-13
feat(client): expose req/resp timeout to client cli

joshuef:
### Summary generated by Reviewpad on 13 Jun 23 03:22 UTC
This pull request adds the ability for the client to set the timeout duration for requests and responses from the network. Additionally, it increases the default timeout duration from 10 seconds to 30 seconds.

### DIFF
--- a/sn_cli/src/cli.rs
+++ b/sn_cli/src/cli.rs
@@ -6,11 +6,11 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use clap::Parser;
-use libp2p::Multiaddr;
-
 use crate::subcommands::SubCmd;
-
+use clap::Parser;
+use color_eyre::Result;
+use libp2p::Multiaddr;
+use std::time::Duration;
 // Please do not remove the blank lines in these doc comments.
 // They are used for inserting line breaks when the help menu is rendered in the UI.
 #[derive(Parser)]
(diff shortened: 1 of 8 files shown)


sn_cli/src/cli.rs
```diff
@@ -35,4 +35,8 @@ pub(crate) struct Opt {
     /// Available sub commands.
     #[clap(subcommand)]
     pub cmd: SubCmd,
+
+    /// Timeout in seconds for the CLI to wait for a data response from the network.
+    #[clap(long = "timeout", global = true)]
```
b-zee: I personally prefer keeping the parsing with clap. Like this:

```suggestion
    #[clap(long = "timeout", global = true, value_parser = |t: &str| -> Result<Duration> { Ok(t.parse().map(Duration::from_secs)?) })]
```

But then a `fn` instead of inlined.

sn_networking/src/lib.rs
```diff
@@ -72,7 +72,7 @@ use tracing::warn;
 pub(crate) const CLOSE_GROUP_SIZE: usize = 8;
 
 // Timeout for requests sent/received through the request_response behaviour.
-const REQUEST_TIMEOUT: Duration = Duration::from_secs(10);
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
```
b-zee: Perhaps a rename to `REQUEST_TIMEOUT_DEFAULT` to highlight this is a fallback value?

I do prefer defaults to be more higher up though, so it could also be an idea to not make `new_client` accept an `Option` and specify this default somewhere in `main`.

b-zee: What is the reasoning behind increasing it from `10` to `30`?

sn_cli/src/cli.rs
joshuef: had no idea you could do that 👍

sn_networking/src/lib.rs
joshuef: Folk were complaining things were timing out, so I thought default maybs should go up.

b-zee (approved):

PR #382, opened 2023-06-13
fix(sn_node): resolve OOM issue

maqi:
### Summary generated by Reviewpad on 13 Jun 23 10:33 UTC
This pull request fixes an issue where returning an error could cause an OOM (Out Of Memory) issue under certain network circumstances by returning  `Ok(())` instead of `Error::MaxRecords`.

### DIFF
--- a/sn_record_store/src/lib.rs
+++ b/sn_record_store/src/lib.rs
@@ -216,7 +216,11 @@ impl RecordStore for DiskBackedRecordStore {
         let num_records = self.records.len();
         if num_records >= self.config.max_records {
             warn!("Record not stored. Maximum number of records reached. Current num_records: {num_records}");
-            return Err(Error::MaxRecords);
+            // To avoid returning error causing libp2p holding un-necessary copy in cache,
+            // to achieve specified `quorumn::ALL` successes,
+            // which may result in OOM(Out Of Memory) issue under certain network circumstance.
+            // Here returning `OK` instead of `Error::MaxRecords`.
+            return Ok(());
         }
 
         let filename = Self::key_to_hex(&r.key);


sn_record_store/src/lib.rs
```diff
@@ -216,7 +216,11 @@ impl RecordStore for DiskBackedRecordStore {
         let num_records = self.records.len();
         if num_records >= self.config.max_records {
             warn!("Record not stored. Maximum number of records reached. Current num_records: {num_records}");
-            return Err(Error::MaxRecords);
+            // To avoid returning error causing libp2p holding un-necessary copy in cache,
+            // to achieve specified `quorumn::ALL` successes,
+            // which may result in OOM(Out Of Memory) issue under certain network circumstance.
+            // Here returning `OK` instead of `Error::MaxRecords`.
+            return Ok(());
```
joshuef, line 223: Hmm, this feels like it breaks the underlying kad... we have not stored the data...

The copies in cache are held because of this no? Until we have a node with sufficient capacity? So really we'd want that extra capcity, or to prune... (And in the end put pressure to pay nodes more to get that space). 

I think for now this may be an expected issue if we're not growing the network

joshuef (changes requested):
I'm not sure this is desireable

sn_record_store/src/lib.rs
dirvine: I tend to agree. We are in error here and have reached capacity and cannot store the data. 
The fix would seem more to be libp2p holding the data in cache. If the node errors we should disconnect from it. 
We could try another (but I would not).
That means holding data on not only the closest nodes, which again is not optimal. 
Therefore the key issue is grow the network but in meantime on this error disconnect and save the cache from being held, perhaps?

joshuef: I'd even say stay, keep cache (which I assume should be disbersed upon new members), and make sure we're updating store cost... If we can assess this cache size, we miiight have a real nice level data velocity too perhaps? (If it's cached until written... higher cache w/o being full will mean we have a data bottleneck and so should also increase storecost?

joshuef:
Closing this, as while it does pinpoint what's going on, it's not the answer here 👍

PR #390, opened 2023-06-14
fix: improve client upload speed

maqi:
### Summary generated by Reviewpad on 14 Jun 23 16:22 UTC
This pull request contains client improvements including reduced memory usage during file uploading and improved upload speed. Changes were made in `sn_client/src/api.rs`, `sn_client/src/file_apis.rs`, `sn_networking/src/cmd.rs`, and `sn_networking/src/lib.rs`. In `sn_client/src/file_apis.rs`, the maximum batch size for chunk uploads was increased to 64, and uploading progress is now logged. In `sn_networking/src/lib.rs`, a new function was added to send requests to half of the closest peers.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -261,15 +261,31 @@ impl Client {
         RegisterOffline::create(self.clone(), xorname, tag)
     }
 
-    /// Store `Chunk` to its close group.
+    /// Store `Chunk` to spcified target.
     pub(super) async fn store_chunk(
         &self,
         chunk: Chunk,
         payment: Option<PaymentProof>,
+        closest_peers: Vec<PeerId>,
     ) -> Result<()> {
-        info!("Store chunk: {:?}", chunk.address());
+        let address = *chunk.name();
+        info!("Store chunk: {:?}", address);
         let request = Request::Cmd(Cmd::StoreChunk { chunk, payment });
-        let response = self.send_and_wait_till_first_rsp(request).await?;
+
+        // Result will be: just one with `StoreChunk(Ok(_))` response;
+        // or a vector of error responses, which only take the first into account.
+        let mut responses = self
+            .network
+            .send_and_get_responses(closest_peers, &request, false)
+            .await
+            .into_iter()
+            .map(|res| res.map_err(Error::Network))
+            .collect_vec();
+        let response = if let Some(response) = responses.pop() {
+            response?
+        } else {
+            return Err(Error::UnexpectedResponses);
+        };
 
         if matches!(response, Response::Cmd(CmdResponse::StoreChunk(Ok(_)))) {
             return Ok(());
@@ -281,6 +297,14 @@ impl Client {
 
         // If there were no store chunk errors, then we had unexpected responses.
         Err(Error::UnexpectedResponses)
+    }
+
+    /// Return the closest_peers to self, fetched from local.
+    pub(super) async fn get_closest_local_peers(&self) -> Result<Vec<PeerId>> {
+        Ok(self
+            .network
+            .get_closest_local_peers(&NetworkAddress::from_peer(self.network.peer_id))
+            .await?)
     }
 
     /// Retrieve a `Chunk` from the kad network.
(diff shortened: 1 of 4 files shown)


sn_client/src/api.rs
```diff
@@ -302,10 +314,18 @@ impl Client {
         Ok(responses)
     }
 
-    pub(crate) async fn send_and_wait_till_first_rsp(&self, request: Request) -> Result<Response> {
+    pub(crate) async fn send_half_range_and_wait_till_first_rsp(
```
joshuef: Can we get some docs on this and why we'd use it?

sn_client/src/file_apis.rs
```diff
@@ -25,7 +25,7 @@ use tracing::trace;
 use xor_name::XorName;
 
 // Maximum number of concurrent chunks to be uploaded/retrieved for a file
-const CHUNKS_BATCH_MAX_SIZE: usize = 5;
+const CHUNKS_BATCH_MAX_SIZE: usize = 16;
```
joshuef: Can we get a bit of data on why this was chosen?

sn_client/src/file_apis.rs
```diff
@@ -170,8 +170,10 @@ impl Files {
         let payment = payment_proofs.get(&address.name().0).cloned();
         // TODO: re-enable requirement to always provide payment proof
         //.ok_or(super::Error::MissingPaymentProof(address))?;
-
-        self.client.store_chunk(chunk, payment).await?;
+        let closest_peers = self.client.client_get_closest_peers_from_local().await?;
```
joshuef: If we do this all the time, when/how do we populate enough info from the network? 

What happens if this data is stale?

sn_networking/src/lib.rs
```diff
@@ -520,6 +520,21 @@ impl Network {
             .await)
     }
 
+    /// Send `Request` to half of the closest peers. `Self` is not present among the recipients.
```
joshuef: Can we get docs on when/why

sn_networking/src/lib.rs
```diff
@@ -627,6 +642,27 @@ impl Network {
         sort_peers_by_address(closest_peers, key, CLOSE_GROUP_SIZE)
     }
 
+    /// Based on the local knowledge, returns half closest peers to the given `XorName`,
+    /// sorted by their distance to the xor_name.
+    /// If `client` is false, then include `self` among the `closest_peers`
+    pub async fn client_get_closest_peers_from_local(
```
joshuef: this naming should probably reflect the `half`

joshuef (changes requested):

sn_client/src/file_apis.rs
maqi: > If we do this all the time, when/how do we populate enough info from the network?

Here only the result from the `libp2p::get_closest_peers` got trimmed for the half-range result.
And it is only `libp2p::get_closest_peers` will populate network knowledge.
So no affect to the network info pupulation.

> What happens if this data is stale?

The nodes received the `Cmd::StoreChunk` shall carry out a `libp2p::put_record`, which will broadcast the copy to all 8 closest.
So, the data won't stale? or, have the simiar chance of stale as current?

sn_client/src/file_apis.rs
maqi: following is some comparision result from benchmark:
```
                            max_batch_size                 1MB throughput                  10MB throughput
main branch                     5                                0.57MB/s                         3.4MB/s
    PR 390                      5                                0.88MB/s                         5.8MB/s
    PR 390                     16                               0.66MB/s                         6.3MB/s
    PR 390                     32                               0.7MB/s                           6.8MB/s
```
hence set this value to 32 currently.

maqi:
The increased `CLIENT_AVG_MEM_LIMIT_MB` is just because of the improved speed, there is less mem_usage check points to spread out the peak mem usage at the beginning due to the self_encryption

joshuef (approved):

PR #393, opened 2023-06-15
Add in HardRestart RPC command

joshuef:
### Summary generated by Reviewpad on 15 Jun 23 04:15 UTC
This pull request contains changes to multiple files related to the SafeNode project. Updates include adding a new RPC function to restart the node, implementing file rotation and naming, changing the logging behavior, adding a Clone trait implementation to the RunningNode struct, and modifying request and response type messages. Additionally, a new enum variant and function were added to initiate a full restart of the binary in a new process.


b-zee:
This makes me a bit uneasy. I think normally such things should be left to the OS/system manager (e.g. systemd).

Also, just from reading the code, does this actually work? If `cmd.status()` is executed, the parent process won't exit until the child process exits. If the parent process exits early, then the child process will be shutdown with it, as far as I know.

I'm actually against restart functionality at all, again, as it's better left to the OS. What I've seen in software is a 'reload' function, where e.g. systemd allows you to do `systemctl reload <service>`, which will execute a command that should reload the configuration of the node.

joshuef:
Fair @b-zee , if we do that though, to maintain any keys running a node (and so it's peerId) we'll have to pass _out_ keys. Orrrr, we'll be passing _in_ keys to encrypt keys to pass out keys to restart. 

All of which could be fine and doable. This is just intended as a first step towards such things. Potentially avoiding some of that. (If we can not send keys anywhere, that'd be the ideal).

If you think it's worth holding off till there's a more robust solution we can do that 👍 

-------

It does work though, new process is created, old one shut down. PeerID is not yet passed across, that was my next step in the plan.

----

I think it would not work for updates as is on linux though, or at least ubuntu seems to lock the bin while it's running, so we'd have to do some alt-bin shenanigans to get that working. (or passing out keys in some manner... or... something else?)

b-zee:
After `hard-restart` I see the parent process is still alive:

```
$ ps -ef --forest | grep safenode
zamaan    133461   70058  0 15:08 pts/2    00:00:00  |           \_ target/release/safenode --rpc=127.0.0.1:10001
zamaan    134565  133461  0 15:08 pts/2    00:00:00  |               \_ /home/b-zee/src/safe_network/target/release/safenode --rpc=127.0.0.1:10001
```

Where the parent ID (`PPID`) of the second shows it's a child process. This adds another child (leaf node) with every hard restart.

One avenue I guess is to daemonize with https://docs.rs/daemonize/latest/daemonize/. But I'd like to caution against it: https://users.rust-lang.org/t/creating-daemon-processes/86832/6
> I would recommend against using that crate on modern Linux systems, that style of creating forking daemons is obsolete. See the "New-Style Daemons" section in [man 7 daemon 42](https://www.man7.org/linux/man-pages/man7/daemon.7.html) for a description of how things are done now. You can just write the program normally and log to stderr, any special requirements would go in a systemd unit file.

This critique also supports my stance on leaving these things to e.g. systemd.

joshuef:
It does feel like we can hand this off to some other process. It will complicate the process of key-handover a fair bit though. But is probably more correct.

(That leftover process though would just be a missing `exit` call, I think after I refactored things up a layer... But I'm fine with taking a different approach to this :+1: )

Issue #311, opened 2023-05-30
Fix e2e test: `multiple_sequential_transfers_succeed`


grumbach:
Fix e2e test: `multiple_sequential_transfers_succeed`

```rs
    let tokens =
        get_tokens_from_faucet(first_wallet_balance, first_wallet.address(), &client).await;
    std::thread::sleep(std::time::Duration::from_secs(5));
    println!("Verifying the transfer from faucet...");
    client.verify(&tokens).await?;                                            // <- this here fails
```

In this test the tokens issued by the faucet are deemed invalid. 
I found out that those token's parent's `SignedSpends` are not known to the network. 

In other words: `token.signed_spends` are not stored on the Network

The Faucet should make sure to upload (aka. spend the money `spend_put`) those before returning the funds.

To make this issue easier to fix, #310 should be fixed first.

PR #395, opened 2023-06-15
fix: DBC transfers, fix e2e test, add double spend test

grumbach:
## Description

- fix issue in parent spends check
- add more `parent_tx` checks
- fix and reactivate `multiple_sequential_transfers_succeed`e2e test #311
- add a double spend e2e test

This PR makes DBC transfers work as well as 2 e2e tests to prove it. 

### Summary generated by Reviewpad on 15 Jun 23 05:38 UTC
This pull request involves multiple changes across various projects. Changes include updating dependencies, modifying error messages, adding new functions, and modifying logging statements. Specific changes include updating the sn_dbc version from 19.0.0 to 19.0.1 with the "serdes" feature still enabled, changing function visibility, modifying function parameters, adding new imports, and changing import statements. Additionally, modifications to logging statements to use proper logging macros can be found, along with changes to the default keep-alive timeout value.

### DIFF
--- a/sn_protocol/src/error.rs
+++ b/sn_protocol/src/error.rs
@@ -110,16 +110,9 @@ pub enum Error {
     #[error("Spend parents are invalid: {0}")]
     InvalidSpendParents(String),
 
-    /// One or more parent spends of a requested spend had a different dst tx hash than the signed spend src tx hash.
-    #[error(
-        "The signed spend src tx ({signed_src_tx_hash:?}) did not match a valid parent's dst tx hash: {parent_dst_tx_hash:?}. The trail is invalid."
-    )]
-    TxTrailMismatch {
-        /// The signed spend src tx hash.
-        signed_src_tx_hash: Hash,
-        /// The dst hash of a parent signed spend.
-        parent_dst_tx_hash: Hash,
-    },
+    /// One or more parent spends of a requested spend has an invalid hash
+    #[error("Invalid parent spend hash: {0}")]
+    BadParentSpendHash(String),
     /// The provided source tx did not check out when verified with all supposed inputs to it (i.e. our spends parents).
     #[error(
         "The provided source tx (with hash {provided_src_tx_hash:?}) when verified with all supposed inputs to it (i.e. our spends parents).."
--- a/sn_protocol/src/error.rs
+++ b/sn_protocol/src/error.rs
@@ -95,8 +95,8 @@ pub enum Error {
     #[error("Insufficient valid spends found: {0:?}")]
     InsufficientValidSpendsFound(DbcAddress),
     /// Node failed to store spend
-    #[error("Failed to store spend: {0:?}")]
-    FailedToStoreSpend(DbcAddress),
+    #[error("Failed to store spend: {0}")]
+    FailedToStoreSpend(String),
     /// Node failed to get spend
     #[error("Failed to get spend: {0:?}")]
     FailedToGetSpend(DbcAddress),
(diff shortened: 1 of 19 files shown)


sn_protocol/src/error.rs
```diff
@@ -95,8 +95,8 @@ pub enum Error {
     #[error("Insufficient valid spends found: {0:?}")]
     InsufficientValidSpendsFound(DbcAddress),
     /// Node failed to store spend
-    #[error("Failed to store spend: {0:?}")]
-    FailedToStoreSpend(DbcAddress),
+    #[error("Failed to store spend: {0}")]
+    FailedToStoreSpend(String),
```
bochaco, line 99: I think in general we should avoid having strings as the context for Error types. 

Any application (implementing the protocol) trying to act upon these errors should be able to use the context information from them. In this case, the context information is the address of the DBC Spend that couldn't be stored (there could be more information attached to it).

It sounds to me here we are trying to avoid having defined more specific error types by having this String? if so, we should add those additional Error types so applications can correctly detect the different scenarios and in some cases auto correct and retry the operation, having a string won't allow them to do such things.

sn_protocol/src/error.rs
```diff
@@ -110,24 +110,13 @@ pub enum Error {
...
-    )]
-    TxTrailMismatch {
-        /// The signed spend src tx hash.
-        signed_src_tx_hash: Hash,
-        /// The dst hash of a parent signed spend.
-        parent_dst_tx_hash: Hash,
-    },
+    /// The DBC we're trying to Spend came with an invalid parent Tx
+    #[error("Invalid Parent Tx: {0}")]
+    InvalidParentTx(String),
```
bochaco, line 115: Same comment as previous change to error type.

sn_protocol/src/error.rs
```diff
@@ -110,24 +110,13 @@ pub enum Error {
...
+    /// One or more parent spends of a requested spend has an invalid hash
+    #[error("Invalid parent spend hash: {0}")]
+    BadParentSpendHash(String),
```
bochaco, line 118: Same comment as previous ones about error types, it sounds like here the context is the parent spend hash? if there are different cases of how a parent spend hash is invalid then we may want different error types for them?

sn_protocol/src/error.rs
```diff
@@ -110,24 +110,13 @@ pub enum Error {
...
-        "The provided source tx (with hash {provided_src_tx_hash:?}) when verified with all supposed inputs to it (i.e. our spends parents).."
-    )]
-    InvalidSourceTxProvided {
-        /// The signed spend src tx hash.
-        signed_src_tx_hash: Hash,
-        /// The hash of the provided source tx.
-        provided_src_tx_hash: Hash,
-    },
+    #[error("The provided source tx is invalid: {0}")]
+    InvalidSourceTxProvided(String),
```
bochaco, line 121: Same comment as previous changes to error types.

bochaco (commented):
> This PR makes DBC transfers work

Are the instructions from the main `README.md` to claim genesis DBC and transfer funds from it to a wallet supposed to work with this PR? I gave it a quick try but it doesn't seem so...?

EDIT: ahh...it's just that the instructions are not mentioning the step of depositing the dbc on the wallet which needs to be donde manually, after ding that I can see the balance, so seems to be working.

sn_protocol/src/error.rs
grumbach: Agreed, there is a bad error management and API overall in this area, I was thinking of reworking it in a follow up PR.

sn_protocol/src/error.rs
grumbach: Yeah maybe not indeed, duely noted, will fix all that in a follow up.

bochaco (approved):

PR #403, opened 2023-06-15
fix(safe): check if upload path contains a file

RolandSherwin:


sn_cli/src/subcommands/files.rs
```diff
@@ -94,8 +94,10 @@ async fn upload_files(
         PaymentProofsMap::default()
     };
 
+    let mut atleast_one_file_exists = false;
```
joshuef: `at_least_one_file_exists`

joshuef (changes requested):

PR #409, opened 2023-06-16
tryout for CI: RecordHeader

RolandSherwin:
Tryout #361


sn_networking/src/error.rs
```diff
@@ -78,4 +78,7 @@ pub enum Error {
 
     #[error("Record was not found locally")]
     RecordNotFound,
+
+    #[error("Bincode error:: {0}")]
+    Bincode(#[from] bincode::Error),
```
joshuef, line 83: I think we'll want to reframe this error for the lib. What's relevant, that bincode errored or that what we put into bindcode was wrong?

PR #411, opened 2023-06-18
ci: use `local-discovery` for released binaries

jacderida:
It was pointed out by @willief that these features were not on the released binaries, so they couldn't be used to establish a connection.

### Summary generated by Reviewpad on 18 Jun 23 20:52 UTC
This pull request adds the `local-discovery` feature to the released binaries for establishing connections.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -43,12 +43,12 @@ build-release-artifacts arch:
   cargo clean
   if [[ $arch == arm* || $arch == armv7* || $arch == aarch64* ]]; then
     cargo install cross
-    cross build --release --target $arch --bin safe
-    cross build --release --target $arch --bin safenode
+    cross build --release --target $arch --bin safe --features local-discovery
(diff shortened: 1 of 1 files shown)


joshuef:
This is intentionally disabled as we shouldn't need local discovery per se. Testnets all require some contact peers. (And local-discovery does increase mem usage of nodes)

joshuef:
Just for clarity for folk reading this down the line.

All `local-discovery` does is enable `MDNS` for nodes on the same local network, such that they might discover each other without the need to have `peers` provided. 

For a local network that may be desireable + simpler. And simplify some client interactions too. But with public testnets it can be bothersome and confusing (someone starts another network on their local machine, but it actually also connects to an exissting public node eg; or a client locally w/ discovery set connects to an old local testnet and not the public testnet you'd desired eg).

So we leave it off by default in general 👍

PR #425, opened 2023-06-20
feat(node): trigger replication when inactivity

maqi:
### Summary generated by Reviewpad on 20 Jun 23 15:19 UTC
This pull request adds a feature to trigger replication to the closest peers when there is inactivity by adding a new `SwarmCmd` called `RequestReplication`. The change allows an inactivity peer to request for a replication flow to be triggered, considering the requester as a newly added peer.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -550,6 +550,17 @@ impl Network {
         Ok(())
     }
 
+    /// Send `Request` to the closest peers to self
+    pub async fn send_req_no_reply_to_self_closest(&self, request: &Request) -> Result<()> {
+        info!("Sending {request:?} to self closest peers.");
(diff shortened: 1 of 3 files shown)


sn_networking/src/event.rs
```diff
@@ -418,7 +418,7 @@ impl SwarmDriver {
     }
 
     /// Replication is triggered when the newly added peer or the dead peer was among our closest.
```
joshuef:
```suggestion
    /// Replication can triggered when a newly added peer or a dead peer was among our closest nodes. 
    /// Nodes may also request replication if they are not seeing much activity (as they may have been dropped
    /// from neighbouring routing tables for some reason, eg.
```

joshuef (approved):

PR #426, opened 2023-06-20
feat: send and verify storage cost payment made to special network-owned output

bochaco:
### Summary generated by Reviewpad on 20 Jun 23 20:52 UTC
This pull request adds the ability to send storage cost payments to a special network-owned output. It introduces a new function called `local_send_storage_payment` to support this functionality, which creates a new network-owned DBC and associated outputs with a `Hash` derived from the `reason_hash` and the set of input dbcs to spend using SHA-3.

### DIFF
--- a/sn_transfers/src/client_transfers/mod.rs
+++ b/sn_transfers/src/client_transfers/mod.rs
@@ -58,7 +58,7 @@ pub struct TransferOutputs {
     /// The dbcs that were created containing
     /// the tokens sent to respective recipient.
     #[debug(skip)]
-    pub created_dbcs: Vec<CreatedDbc>,
+    pub created_dbcs: Vec<Dbc>,
     /// The dbc holding surplus tokens after
     /// spending the necessary input dbcs.
     #[debug(skip)]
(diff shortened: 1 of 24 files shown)


sn_transfers/src/client_transfers/mod.rs
```diff
@@ -87,5 +84,5 @@ pub struct CreatedDbc {
     pub dbc: Dbc,
     /// This is useful for the sender to know how much they sent to each recipient.
     /// They can't know this from the dbc itself, as the amount is encrypted.
-    pub amount: RevealedAmount,
+    pub amount: Token,
```
RolandSherwin: I think we can remove this field now that the `Token` can be obtained directly from the `Dbc`. Maybe this entire struct can be removed as a result?

sn_transfers/src/client_transfers/transfer.rs
```diff
@@ -55,13 +42,66 @@ fn select_inputs(
...
 
     for (dbc, derived_key) in available_dbcs {
         let input_key = dbc.id();
 
-        let dbc_balance = match dbc.revealed_amount(&derived_key) {
-            Ok(revealed_amount) => Token::from_nano(revealed_amount.value()),
+        let dbc_balance = match dbc.token() {
+            Ok(token) => token,
             Err(err) => {
                 warn!("Ignoring input Dbc (id: {input_key:?}) due to not having correct derived key: {err:?}");
```
RolandSherwin: Should we update the warning here, as we might not need to use the derived_key to reveal amounts anymore.

sn_transfers/src/client_transfers/mod.rs
bochaco: sounds good, I'll give it a try

RolandSherwin (approved):

PR #428, opened 2023-06-21
feat(node): expose log markers in public api

joshuef:
Start using them for some more crucial log messages, which we'll treat as an API and have to version accordingly.

### Summary generated by Reviewpad on 21 Jun 23 01:58 UTC
This pull request adds log markers to the public API, which are used for more crucial log messages. These log markers will be treated as an API and versioned accordingly.

### DIFF
--- a/sn_node/src/log_markers.rs
+++ b/sn_node/src/log_markers.rs
@@ -1,0 +1,43 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use libp2p::PeerId;
+use sn_protocol::messages::{Cmd, CmdResponse};
+use std::time::Duration;
+// this gets us to_string easily enough
+use strum::Display;
+
+/// Public Markers for generating log output,
+/// These generate apprioriate log level output and consistent strings.
+/// Changing these log markers is a breaking change.
+#[derive(Debug, Clone, Display)]
+pub enum Marker<'a> {
+    /// The node has started
+    NodeConnectedToNetwork,
+
+    /// No network activity in some time
+    NoNetworkActivity(Duration),
+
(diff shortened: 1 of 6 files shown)


sn_node/src/log_markers.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// NodeCmd::StoreChunk was received
+    StoreChunkReceived {
+        /// The chunk being stored
+        address: &'a ChunkAddress,
+        /// payment
+        payment: &'a Option<PaymentProof>,
+    },
+    /// NodeCmd::SpendDbc Received
+    // SpendDbcReceived(SpendWithParent),
```
RolandSherwin: This log line can be removed, I think it is a leftover from a refactor.

sn_node/src/log_markers.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// No network activity in some time
+    NoNetworkActivity(Duration),
+
+    /// Put request received from + for
+    KademliaPutRequest {
+        /// The peer that sent the request
+        peer: PeerId,
+        /// The key of the record being put
+        key: &'a RecordKey,
+    },
```
RolandSherwin: This will not be triggered with our current setup, as we don't have kad's PUT/replication enabled. 
Maybe we can add this later when it is required?

sn_node/src/log_markers.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Peer was added to the routing table
+    PeerAddedToRoutingTable(PeerId),
+
+    /// NodeCmd::StoreChunk was received
+    StoreChunkReceived {
+        /// The chunk being stored
+        address: &'a ChunkAddress,
+        /// payment
+        payment: &'a Option<PaymentProof>,
+    },
```
RolandSherwin: Just a small question, should we also have a corresponding `StoreChunkResponded` marker that denotes that the cmd has been completed? Might help people to track the time it took for each cmd to process.

Also, maybe we can have a match statement inside `SwarmCmd::SendResponse` and put all the `*Responded` markers there.

RolandSherwin: Similarly, we can put all the `*Received` markers under a match statement?

sn_node/src/log_markers.rs
joshuef: updated for cmd in/out

sn_node/src/log_markers.rs
joshuef: 👍

RolandSherwin (approved):

PR #430, opened 2023-06-21
Obtain `parent_tx` from `SignedSpend`

RolandSherwin:
- Removes `SpendWithParent` and instead uses the updated `sn_dbc` lib which stores the `parent_tx` inside `SignedSpend`


sn_node/src/put_validation.rs
```diff
@@ -470,37 +467,23 @@ impl Node {
...
-                if spends_with_parent.parent_tx.hash()
-                    != spends_with_parent.signed_spend.dbc_creation_tx_hash()
-                {
-                    return Err(ProtocolError::InvalidSpendParents(format!(
-                    "parent tx hash does not match the parent tx in the DBC we're trying to spend: {:?} != {:?}",
-                    spends_with_parent.signed_spend.dbc_creation_tx_hash(),
-                    spends_with_parent.parent_tx.hash(),
-                )));
-                }
-
```
RolandSherwin, line 493: This check can be safely removed, as `dbc_creation_tx_hash()` is now derived from the `signed_spend.spend.dbc_creation_tx.hash()`

sn_node/src/spends.rs
```diff
@@ -68,26 +68,22 @@ pub(crate) async fn get_aggregated_spends_from_peers(
...
                 // make sure the spent_tx hash matches
-                && spend_with_parent
-                    .signed_spend
-                    .verify(spend_with_parent.signed_spend.spent_tx_hash())
+                && signed_spend
+                    .verify(signed_spend.spent_tx_hash())
                     .is_ok()
-                // make sure the parent_tx hash matches
-                && spend_with_parent.parent_tx.hash()
-                    == spend_with_parent.signed_spend.dbc_creation_tx_hash()
```
RolandSherwin, line 90: Can be safely removed

grumbach (approved):
🚢

PR #434, opened 2023-06-22
Send `Cmd::Store` to network closest nodes

### LINKED ISSUE #433: Check if node among the closest before storing data
Make sure that we store data that actually belongs to us. Return error if we're not the closest.

RolandSherwin:


RolandSherwin:
Superseded by #443

PR #436, opened 2023-06-22
Dialing without peer ID

b-zee:
This allows us to pass `--peer=1.2.3.4:1234` instead of a multiaddr containing the peer ID.

This changes the bootstrapping somewhat and the API surrounding how we dial to accomodate for passing a multiaddr without peer ID.


joshuef (approved):

joshuef:
Tested on a wee testnet, no problems with the new format 💪

PR #438, opened 2023-06-22
feat(node): write secret key to disk and re-use

b-zee:
If no `secret-key` file exists, a key is generated and written to that file. If it exists, we read the key from the file and use that. An option is added to the command-line to force a new key to be generated, even if the file exists.

### DIFF
--- a/sn_node/src/bin/safenode/main.rs
+++ b/sn_node/src/bin/safenode/main.rs
@@ -19,10 +19,10 @@ use sn_peers_acquisition::PeersArgs;
 
 use clap::Parser;
 use eyre::{eyre, Error, Result};
-use libp2p::Multiaddr;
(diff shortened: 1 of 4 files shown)


sn_node/src/bin/safenode/main.rs
```diff
@@ -86,6 +86,11 @@ struct Opt {
     /// Only applies when --log-dir is also set to output logs to file.
     #[clap(long)]
     json_log_output: bool,
+
+    /// Renew secret key (file) that is used for our peer ID. If the `secret-key` file exists
```
jacderida: Minor style point. I would prefer to keep the use of the backtick notation out of the UI text because it doesn't get parsed in the terminal.

jacderida (approved):

PR #443, opened 2023-06-23
feat: replicate to lost record peer; client upload chunk to chunk's closest; forwarding chunk

maqi:
### Summary generated by Reviewpad on 23 Jun 23 08:23 UTC
This pull request contains two patches. The first one introduces a feature that allows tracking of lost records from peers. The second patch fixes the issue where client upload to peers was not always being sent to the closest peers to the chunk and instead defaulted to sending it to the closest peers to the node.

### DIFF
--- a/sn_client/src/file_apis.rs
+++ b/sn_client/src/file_apis.rs
@@ -169,9 +169,9 @@ impl Files {
         let payment = payment_proofs.get(&address.name().0).cloned();
         // TODO: re-enable requirement to always provide payment proof
         //.ok_or(super::Error::MissingPaymentProof(address))?;
-        let all_peers = self.client.get_all_local_peers().await?;
+
         let dst = NetworkAddress::from_chunk_address(address);
-        let closest_peers = sort_peers_by_address(all_peers, &dst, CLOSE_GROUP_SIZE)?;
+        let closest_peers = self.client.get_closest_local_peers(&dst).await?;
         self.client
             .store_chunk(chunk, payment, closest_peers)
             .await?;
--- a/sn_client/src/file_apis.rs
+++ b/sn_client/src/file_apis.rs
@@ -17,7 +17,11 @@ use bytes::Bytes;
 use futures::future::join_all;
 use itertools::Itertools;
 use self_encryption::{self, ChunkInfo, DataMap, EncryptedChunk, MIN_ENCRYPTABLE_BYTES};
-use sn_protocol::storage::{Chunk, ChunkAddress};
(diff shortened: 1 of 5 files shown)


sn_client/src/file_apis.rs
```diff
@@ -165,7 +169,9 @@ impl Files {
         let payment = payment_proofs.get(&address.name().0).cloned();
         // TODO: re-enable requirement to always provide payment proof
         //.ok_or(super::Error::MissingPaymentProof(address))?;
-        let closest_peers = self.client.get_closest_local_peers().await?;
+        let all_peers = self.client.get_all_local_peers().await?;
+        let dst = NetworkAddress::from_chunk_address(address);
+        let closest_peers = sort_peers_by_address(all_peers, &dst, CLOSE_GROUP_SIZE)?;
```
RolandSherwin: Should we call `get_closest_local_peers(chunk_address)` here? It should provide the sorted, closest peers to the chunk directly?

sn_client/src/file_apis.rs
```diff
@@ -196,16 +202,17 @@ impl Files {
...
                 //.ok_or(super::Error::MissingPaymentProof(chunk_addr))?;
 
                 tasks.push(task::spawn(async move {
-                    client
-                        .store_chunk(chunk, payment, closest_peers_clone)
-                        .await?;
+                    let dst = NetworkAddress::from_chunk_address(chunk_addr);
+                    let closest_peers =
+                        sort_peers_by_address(all_peers_clone, &dst, CLOSE_GROUP_SIZE)?;
+                    client.store_chunk(chunk, payment, closest_peers).await?;
```
RolandSherwin, line 215: Same as the above one? Should we call the `local_closet_peers(chunk)` directly?

sn_client/src/file_apis.rs
RolandSherwin: And hence a smaller `Vec<peers>`?

sn_client/src/file_apis.rs
maqi: No. by using `sort_peers_by_address`, there is then only one call from node to swarm_driver.
using `local_closet_peers(chunk)`, result in one call from node to swarm_driver for each chunk.
Though it is via `local event channel`, prefer to avoid such multiple calls.

RolandSherwin: That is right! Looks good to me!

RolandSherwin (approved):

PR #446, opened 2023-06-26
feat(node): add handle for Cmd::Response(Ok)

joshuef:
This should stop a lot of pointless logging around the response there

### Summary generated by Reviewpad on 26 Jun 23 03:14 UTC
This pull request adds a handle for Cmd::Response(Ok), which should stop pointless logging around the response.


grumbach (approved):
LGTM

PR #448, opened 2023-06-26
fix: replicate data only to nodes in the close group

joshuef:
### Summary generated by Reviewpad on 26 Jun 23 06:26 UTC
This pull request fixes the issue where data was being unnecessarily replicated to all nodes instead of just those in the close group. The replication now only happens to the nodes in the close group.

### DIFF
--- a/sn_node/src/replication.rs
+++ b/sn_node/src/replication.rs
@@ -55,7 +55,14 @@ impl Node {
             None => return,
         };
         trace!("Being out of closest range, replicating {addr:?} to {close_peers:?}");
+        let mut peers_updated = 0;
         for peer in close_peers.iter() {
+            peers_updated += 1;
+
+            if peers_updated > CLOSE_GROUP_SIZE {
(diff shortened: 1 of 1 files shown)


sn_node/src/replication.rs
```diff
@@ -55,7 +55,14 @@ impl Node {
             None => return,
         };
         trace!("Being out of closest range, replicating {addr:?} to {close_peers:?}");
+        let mut peers_updated = 0;
         for peer in close_peers.iter() {
+            peers_updated += 1;
```
maqi, line 60: I think the root reason is that libp2p's `get_closest_local_peers` function is little bit misleading that it actually get ALL peers just sort by distance. 
The name is very misleading as for us closest only means `CLOSE_GROUP_SIZE` peers.

To fix it more properly, shall we do a sized chop at the place of calling libp2p's `get_closest_local_peers`
i.e. at the place of handling `SwarmCmd::GetClosestLocalPeers`

PR #447, opened 2023-06-26
feat(networking): updgrade to libp2p 0.52.0

joshuef:
### Summary generated by Reviewpad on 26 Jun 23 04:46 UTC
This pull request updates the libp2p dependency across several files to version 0.52. The source of the dependency has also changed to a specific branch in some cases. There are various changes to the behavior of the Swarm type in `sn_networking`, including the addition of a new protocol and modifications to the local-discovery feature. Additionally, there are changes to struct implementations and import statements in several files.

PR #452, opened 2023-06-27
Error on dialing local addresses on global network

b-zee:
Depending on #447

The `global_only` transport prevents dialing local addresses. This appears to happen sometimes as the Identify protocol broadcasts local addresses that seem to end up in the routing table.

Dialing a local address will yield a `TransportError::MultiaddrNotSupported`, which is a bit ambiguous, but we could potentially capture it/ignore it.


joshuef:
this is in main

Issue #250, opened 2023-05-08
Make sure registers use Record storage like chunks


joshuef:
- Registers ops should be checked and merged into the record store on receipt.
- If there has been an update, the new record value should be saved
- If there has been an update, we send an update to all peers to replicate this op.
- More concerns?

PR #453, opened 2023-06-27
feat: Registers refactor + Records

grumbach:
## Description

- refactors registers into something simpler, removing duplicates and unnecessary types
- move all register logic to `sn_register` make it an independent crate with no dependency to the monorepo. 
- store registers in Records #250 
- registers merge during replication if there's a conflict

reviewpad:summary


sn_registers/src/register.rs
```diff
@@ -150,14 +128,15 @@ impl RegisterReplica {
...
+        self.crdt.write(entry, children, self.owner)
     }
 
     /// Apply a signed data CRDT operation.
-    pub fn apply_op(&mut self, op: RegisterOp<Entry>) -> Result<()> {
+    pub fn apply_op(&mut self, op: RegisterOp) -> Result<()> {
         self.check_entry_and_reg_sizes(&op.crdt_op.value)?;
+        self.check_register_op(&op)?;
         self.crdt.apply_op(op)
     }
```
grumbach, line 140: Adds the register op verification/signature at the `sn_register` level

sn_registers/src/register.rs
```diff
@@ -166,16 +145,25 @@ impl RegisterReplica {
...
     /// `Ok(())` if the permissions are valid,
     /// `Err::AccessDenied` if the action is not allowed.
-    pub fn check_permissions(&self, action: Action, requester: Option<User>) -> Result<()> {
-        let requester = requester.unwrap_or(self.authority);
+    pub fn check_user_rights(&self, action: Action, requester: User) -> Result<()> {
         // First checks if the requester is the owner.
-        if action == Action::Read || requester == self.policy.owner {
+        if action == Action::Read || requester == self.owner {
             Ok(())
         } else {
```
grumbach: I'm wondering if we shouldn't give all rights (read+write) to the owner?

sn_registers/src/register.rs
```diff
@@ -27,58 +23,42 @@
...
-            },
-        )
+    pub fn new_owned(owner: User, name: XorName, tag: u64) -> Self {
+        let permissions = Default::default();
+        Self::new(owner, name, tag, permissions)
+    }
+
+    /// Verifies the integrity of the Register.
+    pub fn verify(&self) -> Result<()> {
+        // NB TODO, add op log to reg and verify capability
```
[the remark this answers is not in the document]
grumbach: For now, to keep it simple, registers are replicated as is without checks. 
On the other hand, one needs to have the proper rights to write to a register. 
This means if nodes are honest, registers can't be corrupted by a bad client.

sn_client/src/register.rs
```diff
@@ -6,37 +6,29 @@
...
+use sn_registers::{
+    Action, Entry, EntryHash, Permissions, Register, RegisterAddress, User, UserRights,
 };
+
+use std::collections::{BTreeSet, LinkedList};
 use xor_name::XorName;
 
 /// Ops made to an offline Register instance are applied locally only,
 /// and accumulated till the user explicitly calls 'sync'. The user can
 /// switch back to sync with the network for every op by invoking `online` API.
```
RolandSherwin, line 23: I think the `online` API is gone now? Should this doc be updated here?

sn_node/src/put_validation.rs
```diff
@@ -254,6 +298,55 @@ impl Node {
...
+            .await
+            .map_err(|err| {
+                warn!("Error while fetching local record {err}");
+                ProtocolError::RegisterNotStored(*reg_addr.name())
+            })?;
+        let record = match maybe_record {
+            Some(r) => r,
+            None => {
+                error!("Register with addr {reg_addr:?} already exists locally, but not found in local storage");
+                return Err(ProtocolError::RecordKeyMismatch);
```
RolandSherwin: I think this error was a leftover from a previous change. Maybe we don't need it anymore and we can just return back `RegisterNotStored(*reg_addr.name())`?

sn_protocol/src/error.rs
```diff
@@ -88,25 +60,19 @@ pub enum Error {
     /// Received a `Vec<SignedSpend>` with more than two spends
     #[error("Incoming SpendDbc PUT with incorrect number of SignedSpend")]
     MaxNumberOfSpendsExceeded,
-    /// Spend not found.
```
joshuef, line 91: Why have we removed _some_ comments here?

sn_registers/src/permissions.rs
```diff
@@ -11,17 +11,17 @@ use super::Action;
...
-    /// `Some(false)` explicitly denies this permission (even if `Anyone` has permissions).
-    /// Use permissions for `Anyone` if `None`.
+    /// `Some(false)` explicitly denies writes (even if `Anyone` has the right to write).
+    /// `None` use default: the rights for `Anyone`
     write: Option<bool>,
 }
 
-impl Permissions {
-    /// Constructs a new public permission set.
+impl UserRights {
```
joshuef: This gave me pause... `UserPermissions`? 

I guess you were trying to clarify the overall permission map vs per user perms? 

If so it feels like  `UserPermissions` would address this without adding another term into the mix?

sn_registers/src/register.rs
joshuef: As they can effectively update to give themselves this, it does seem like they may as well just have perms there already, yeh 👍

sn_client/src/register.rs
grumbach: It's not gone, the `*_online` named functions do this job now.

sn_protocol/src/error.rs
grumbach: The comments where identical to the error name, should we put them back?

sn_registers/src/permissions.rs
grumbach: Make sense yeah

RolandSherwin (approved):

PR #455, opened 2023-06-27
feat: append peer id to node's default root dir

jacderida:
BREAKING CHANGE: The default root directory for the node will now have the node's peer ID appended to its path.

Using the peer ID facilitates running multiple node processes on a single machine, without the need to explicitly use the `--root-dir` argument for each node to specify its own root directory. Previously, if they didn't do that, all the nodes would share the same record store.

### Summary generated by Reviewpad on 27 Jun 23 13:28 UTC
This pull request introduces a new dependency `dirs-next` and changes the `root_dir` parameter to be an optional `PathBuf` in multiple places. It also modifies the implementation of `SwarmDriver::get_root_dir_path()` and `Node` and `RunningNode` struct to accommodate this change. Additionally, it adds an `error` variant to the `Error` enum in `error.rs` related to root directory configuration.

### DIFF
--- a/sn_node/src/api.rs
+++ b/sn_node/src/api.rs
@@ -21,7 +21,7 @@ use sn_protocol::{
     NetworkAddress,
 };
 use sn_registers::RegisterStorage;
-use std::{net::SocketAddr, path::Path, time::Duration};
+use std::{net::SocketAddr, path::PathBuf, time::Duration};
 use tokio::task::spawn;
 
 /// Once a node is started and running, the user obtains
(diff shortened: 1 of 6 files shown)


sn_node/src/api.rs
```diff
@@ -37,6 +37,17 @@ impl RunningNode {
         self.network.peer_id
     }
 
+    /// Returns the root directory path for the node.
+    ///
+    /// This will either be a value defined by the user, or a default location, plus the peer ID
+    /// appended. The default location is platform specific:
+    ///  - Linux: $HOME/.local/share/safe/node/peer-id
```
b-zee: Could you write `<peer-id>` to signal it's not the literal `peer-id`?

b-zee (approved):

PR #457, opened 2023-06-28
ci: ensure we use rust-cache v2 everywhere

joshuef:
### Summary generated by Reviewpad on 28 Jun 23 00:06 UTC
This pull request updates the CI pipelines to use rust-cache v2 everywhere. The previous version of rust-cache is replaced with the latest version. The latest version allows the CI to continue even when caching fails.


grumbach (approved):
👍

PR #460, opened 2023-06-28
networking: no blocking send events +  increase node event channel size

joshuef:
### Summary generated by Reviewpad on 28 Jun 23 05:19 UTC
This pull request modifies several files in the codebase.

In `sn_networking/src/msg/mod.rs`, changes were made to handle received requests. The code now uses the `send_event` method to send a `RequestReceived` event to the `event_sender`. Additionally, the error handling for failed notifications seems to have been removed.

In `sn_networking/src/event.rs`, the diff includes changes to the `send_event` method. The use of `event_sender.send` has been replaced with `self.send_event` for the `NewListenAddr`, `NatStatusChanged`, and `PeerAdded` events. The `await` keyword after sending these events has also been removed. Additionally, a call to the `log_kbuckets` function has been added before sending the `PeerAdded` event.

The `impl SwarmDriver` block in the `sn_networking/src/cmd.rs` file has been modified. Specifically, lines 273-280 were changed to replace the `self.event_sender.send(...)` call with a call to `self.send_event(...)`.

In the `event.rs` file, a constant named `NODE_EVENT_CHANNEL_SIZE` with a value of `10_000` has been added. This constant is used to define the size of the event channel in the `NodeEventsChannel` struct. The default implementation of `NodeEventsChannel` has been modified to use the new constant value. This change increases the channel size to accommodate more events.

In the diff, the following changes were made:

- Added a new field `event_sender` in the `SwarmDriver` struct.
- Added a new method `send_event` to send events asynchronously.
- Replaced the usage of `await` with non-blocking calls for certain commands in the `Network` struct.
- Updated the `send_swarm_cmd` method to handle full capacity of the `SwarmCmd` channel and provide error handling.

In the `sn_networking/src/error.rs` file, a new variant `NoSwarmCmdChannelCapacity` has been added to the `Error` enum. This variant is used when there is no SwarmCmd channel capacity available.

This pull request includes various changes to improve event handling, error handling, and code organization in the codebase.

### DIFF
--- a/sn_networking/src/msg/mod.rs
+++ b/sn_networking/src/msg/mod.rs
@@ -53,9 +53,7 @@ impl SwarmDriver {
                             None => {
                                 // responses that are not awaited at the call site must be handled
                                 // separately
-                                self.event_sender
-                                    .send(NetworkEvent::ResponseReceived { res: response })
-                                    .await?;
+                                self.send_event(NetworkEvent::ResponseReceived { res: response });
                             }
                         }
                     } else {
--- a/sn_networking/src/msg/mod.rs
+++ b/sn_networking/src/msg/mod.rs
@@ -31,16 +31,10 @@ impl SwarmDriver {
                     ..
                 } => {
                     trace!("Received request with id: {request_id:?}, req: {request:?}");
-                    if let Err(error) = self
-                        .event_sender
-                        .send(NetworkEvent::RequestReceived {
-                            req: request,
-                            channel: MsgResponder::FromPeer(channel),
-                        })
-                        .await
-                    {
-                        error!("Failed in notify {request_id:?} NetworkEvent: {error:?}");
-                    }
+                    self.send_event(NetworkEvent::RequestReceived {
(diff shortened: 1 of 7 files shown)


RolandSherwin (approved):

sn_networking/src/msg/mod.rs
```diff
@@ -31,16 +31,10 @@ impl SwarmDriver {
...
-                        .send(NetworkEvent::RequestReceived {
-                            req: request,
-                            channel: MsgResponder::FromPeer(channel),
-                        })
-                        .await
-                    {
-                        error!("Failed in notify {request_id:?} NetworkEvent: {error:?}");
-                    }
+                    self.send_event(NetworkEvent::RequestReceived {
+                        req: request,
```
RolandSherwin, line 35: There are a couple more places where we still use the `self.event_sender` directly and await on the result. Could not tag them here for some reason.
I assume they should be made sync as well?

RolandSherwin (approved):

PR #461, opened 2023-06-28
feat: client do not wait for response when upload

maqi:
### Summary generated by Reviewpad on 28 Jun 23 10:11 UTC
This pull request introduces a new feature where the client no longer waits for a response when uploading. The `send_and_complete` function is added to the `sn_networking` module to send a request to a set of peers and collect all the responses. This eliminates the need to wait for a single response and handles multiple error responses at once. Overall, this patch improves the efficiency of the upload process in the client.

### DIFF
--- a/.github/workflows/benchmark-prs.yml
+++ b/.github/workflows/benchmark-prs.yml
@@ -39,7 +39,7 @@ jobs:
 
      
       - name: Build sn bins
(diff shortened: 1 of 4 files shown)


maqi:
No longer relevant as now client using put_record to upload

PR #463, opened 2023-06-28
fixes: mem dead issue; failing mem_check test

maqi:
### Summary generated by Reviewpad on 28 Jun 23 11:32 UTC

The first patch is a chore to tidy up the `try_trigger_replication` function in the `replication.rs` file. It simplifies the function logic and improves code readability. The changes include:
- Initializing variables more efficiently.
- Reducing the computation work by only considering nearby peers.
- Avoiding unnecessary calculations and checks.
- Setting up the record storage distance range correctly.

The second patch is a fix for avoiding duplicated replications. It addresses two scenarios where duplicated replications could occur:
1. When a dead peer is detected multiple times, resulting in duplicate replication events.
2. When the `kad::remove_peer` function removes a peer from the `kBucketsTable` with a delay, causing duplicate replications. 

The fix includes:
- Adding a set to track dead peers to prevent duplicate detection.
- Checking if a peer is already marked as dead before triggering the removal event.
- Removing the peer from the `kBucketsTable` to avoid duplicate replications.


joshuef (approved):

.github/workflows/benchmark-prs.yml
```diff
@@ -127,6 +128,20 @@ jobs:
...
           SN_LOG: "all"
 
+      - name: Confirm the heapnode connected to the network
+        shell: bash
+        timeout-minutes: 1
+        run : |
+          echo "Directory of log_dir is $log_dir"
+          ls $log_dir
+          discovered_count=$(rg "PeerAddedToRoutingTable" ~/.safe/heapnode -c --stats | rg "(\d+) matches" | rg "\d+" -o)
+          echo "Heapnode discovered $restart_count nodes"
```
RolandSherwin: I think you meant to echo the `$discovered_count` variable here?

sn_client/src/api.rs
```diff
@@ -143,13 +146,19 @@ impl Client {
             }
             match client_events_rx.recv().await {
                 Ok(ClientEvent::ConnectedToNetwork) => {
-                    info!("Client connected to the Network.");
+                    is_connected = true;
+                    info!("Client connected to the Network {is_connected:?}.");
+                    println!("Client connected to the Network");
```
RolandSherwin: I think we already have the progress bar that prints to the console when the client is connected?

PR #466, opened 2023-06-29
refactor: demystify permissions

grumbach:
### Summary generated by Reviewpad on 29 Jun 23 10:31 UTC

- Register-related functionality for retrieving user permissions has been removed.
- Module visibility, imports, and method signatures have been optimized and simplified.
- Update and removal of variants, structs, and methods related to permissions handling.

These changes aim to improve code organization, reduce unnecessary code, and introduce a simplified permissions model for register access.

Let me know if you need more information or further assistance!

### DIFF
--- a/sn_client/src/register.rs
+++ b/sn_client/src/register.rs
@@ -11,9 +11,7 @@ use crate::{Client, Error, Result};
 use sn_protocol::messages::{
     Cmd, CmdResponse, Query, QueryResponse, RegisterCmd, RegisterQuery, Request, Response,
 };
-use sn_registers::{
-    Action, Entry, EntryHash, Permissions, Register, RegisterAddress, User, UserPermissions,
-};
+use sn_registers::{Entry, EntryHash, Permissions, Register, RegisterAddress, User};
 
 use std::collections::{BTreeSet, LinkedList};
 use xor_name::XorName;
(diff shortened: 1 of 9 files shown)


happybeing:
Is there any info on the previous and new permissions models? Even just a high level would be interesting.

grumbach:
> Is there any info on the previous and new permissions models? Even just a high level would be interesting.

Good question. 
So the previous permission model is kind of clunky:
- the owner can always read
- everyone can read
- it lists the users who can write as well as the users who cannot write (like a ban list)
- for those not listed, we referred to the default behaviour
- if the default behaviour was not specified then we deny access
- if `Anyone` can write but you are in the ban list you cannot write (although nothing prevents you from trying with another key...)

All this was a bit complex and clunky, both from the user standpoint and the dev standpoint so something had to be done. Now what we have is:
- the owner can always read and write
- everyone can read
- we have a list of those the the owner allowed to write, those not listed cannot write

Hope this makes things clearer @happybeing?

RolandSherwin (approved):
Great simplification! :rocket:

happybeing:
Yes, very helpful thanks. So to prevent read, a writer would use an additional layer of encryption?

Is that how private data will be implemented, and if so is there still an intention to allow for private data to be deletable, with/without partial refund?

Any obvious implications for Jim's app layer examples where he showed his ideas for helping people manage their data and permissions?

Don't worry trying to answer everything. I'll be content of you need to just get on but if it helps then I'm curious to understand. I always find the data model interesting and have long wondered how different app level functions might need to be implemented, and what might be possible, and what might be very difficult in terms of decentralised web 2 things.

I remember @joshuef put thought into ways to implement a capability based permissions model, so also wonder if that's still under consideration as it seemed advantageous.

Great to see this. 😄

grumbach:
> So to prevent read, a writer would use an additional layer of encryption?

Indeed to prevent read, a user would just have to encrypt its data client side before even uploading anything, just like chunks!

> Is that how private data will be implemented, and if so is there still an intention to allow for private data to be deletable, with/without partial refund?

We don't have deletion implemented as of now, due to the CRDT nature of Registers, history is kept always for now. This could be a nice to have indeed, we'll need to think about a viable design. 

> Any obvious implications for Jim's app layer examples where he showed his ideas for helping people manage their data and permissions?

This refactor is mostly simplification, this shouldn't change what was possible before.

PR #467, opened 2023-06-29
feat: carry out validation for record_store::put

maqi:
### Summary generated by Reviewpad on 29 Jun 23 16:04 UTC

1. The file `api.rs` has a new condition added to the `match` statement in the `impl Client` block.
2. The file `cmd.rs` has a change in the implementation of the `SwarmDriver` struct, specifically in line 194 where the `put` method is now `put_verified`.
3. The file `lib.rs` has changes related to introducing a new module called `record_store`, importing and using the `DiskBackedRecordStore` and related configurations, and reordering the creation and usage of channels.
4. The file `sn_record_store/src/lib.rs` was renamed to `sn_networking/src/record_store.rs` and has changes to imports, struct fields, and method implementations.
5. The `Cargo.lock` file has changes related to dependencies, adding `quickcheck` and removing `sn_record_store`.
6. The file `put_validation.rs` has changes in the `Node` implementation, including adding new functions and handling different record types.
7. The `Cargo.toml` file has changes related to updating dependencies and removing/readding packages.
8. The file `lib.rs` in the `sn_protocol` module has changes related to import statements.

Please review these changes in the diff and let me know if you need any further information.


RolandSherwin:
Do you think this old commit might be useful here? https://github.com/maidsafe/safe_network/commit/e6e6b77a75b133787fe1e253d7e4d5b570fb2e7a

sn_networking/src/record_store.rs
```diff
@@ -208,40 +241,29 @@ impl RecordStore for DiskBackedRecordStore {
...
-        if r.value.len() >= self.config.max_value_bytes {
-            warn!(
-                "Record not stored. Value too large: {} bytes",
-                r.value.len()
-            );
-            return Err(Error::ValueTooLarge);
+    fn put(&mut self, record: Record) -> Result<()> {
+        if self.records.contains(&record.key) {
+            trace!("Unverified Record {:?} already exists.", record.key);
+            return Ok(());
```
RolandSherwin: We might want to send out a `NetworkEvent::UnverifiedRecord` even if we already have the record with us? E.g. a double spend attempt would trigger a PUT with the same record key.

maqi: considering there is `double spend ` case, then it will be better just throw it to the validation and leave it to be detected and handled there.

sn_node/src/put_validation.rs
```diff
@@ -25,6 +25,28 @@ use sn_transfers::payment_proof::validate_payment_proof;
...
+                let chunk_with_payment = try_deserialize_record::<ChunkWithPayment>(&record)?;
+                self.validate_and_store_chunk(chunk_with_payment).await
+            }
+            RecordKind::DbcSpend => {
+                let register = try_deserialize_record::<Register>(&record)?;
+                self.validate_and_store_register(register).await
+            }
+            RecordKind::Register => {
+                let signed_spends = try_deserialize_record::<Vec<SignedSpend>>(&record)?;
+                self.validate_and_store_spends(signed_spends).await
```
RolandSherwin, line 46: I think we might also have to check if `Record::key == key obtained from data`, i.e., if it is a chunk, we might have to make sure that `Key(Chunk.address()) == Record::key`, if it's dbc, check for `Key(DBC.id()) == Record::key`.

This will prevent someone from crafting a valid `Record::value` but then store it in a random place in the network using a different `Record::value`. A malicious client can overwrite a valid Record using the above method.

maqi: this sounds a existing bug within the `validate_and_store_spends` function to the main branch as well.
shall this be addressed in a separate PR to main?

RolandSherwin (approved):

PR #469, opened 2023-06-29
New Logging Setup

jacderida:
- 27369a9 **feat: provide `--log-output-dest` arg for `safenode`**

  BREAKING CHANGE: the `--log-dir` argument has been removed in favour of `--log-output-dest`.

  This argument uses a 'special' value, which can be "stdout", "default", or a custom directory path.
  The "default" is the default root directory, plus "logs" appended to that path.

  The `sn_logging` crate now provides a `LogOutputDest` enum which makes for a nicer interface on the
  logging initialisation, removing the need for an `Option<PathBuf>` to specify output to a file.

- 0f20a34 **feat: provide `--log-output-dest` arg for `safe`**

  The behaviour for this is mostly the same as `safenode`, except if the argument is not
  specified, no logging will take place (this does not affect println output).

- 9b6ccc9 **feat: introduce `--log-format` arguments**

  BREAKING CHANGE: for both `safe` and `safenode` the `--json-log-output` argument is removed in
  favour of this new `--log-format` argument.

  The argument functions in a similar fashion to `--log-output-dest`, in that it uses a 'special'
  value, which for the time being can either be "default" or "json". It has the potential to be
  expanded.
  
 - 7b64b22 **feat: remove option from `--log-output-dest` arg**

   Benno submitted a patch that removes the need for `log_output_dest` to be expressed as an Option,
   which makes the argument easier to work with.
   
- 9e0e049 **chore: incorporate various feedback items**

  * Keep the logging guard alive for the scope of the whole program in `safe`
  * Remove duplicate code in favour of function call
  * Use the term "root-dir" rather than "default" to avoid confusion with the fact that "stdout" is
    actually the default value   

- f01ad07 **ci: update node logging paths**

  In light of the changes to the logging setup, the testnet tool is modified such that it will use the
  default root and logging directories.

  The merge workflow is also updated to look for logs in the new directories. For the correct
  usage of paths, in some cases this meant using different mechanisms on Windows vs. Unix.

  The memcheck workflow was also updated to the same effect, but it only runs on Linux, so
  cross-platform considerations weren't required.

- 0acdeb4 **chore: use data-dir rather than root-dir**

  Completely separate the `--log-output-dest` argument from the `--root-dir` argument by removing the
  use of the term "root-dir" from the logging output destinations, which could have been confusing.
  
- 2efdb94 **ci: update benchmark workflows for new directories**

  This was missed in the previous commit, but is now updated in the same fashion as the merge and
  memcheck workflows.
  
### Summary generated by Reviewpad on 29 Jun 23 20:48 UTC

1. Changes related to the `root_dir_path` function in the `impl RunningNode` block, updating the code comment to provide the correct Windows path format for `root_dir_path`.
2. Changes related to logging configuration, including importing additional modules, modifying function calls, and updating variable initialization.
3. Various changes to the `faucet.rs` file, such as removing import statements, modifying the `main` function, and adding a log message.
4. Changes to the `sn_node/tests/sequential_transfers.rs` file, including importing statements, updating the use of `init_logging` function, modifying a variable, and adding test cases.
5. Changes related to logging configuration and formatting, introducing new enums and modifying existing code.
6. Changes to the `data_with_churn.rs` file, including modifying import statements, updating the `init_logging` function, and adding a comment.
7. Changes to the `main.rs` file, adding import statements, modifying the main function, updating logging initialization, and performing code clean-up.
8. Changes to the `cli.rs` file, including importing functions and enums, adding new functions and fields, and updating the documentation.

Please review these changes thoroughly.

### DIFF
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -15,34 +15,21 @@ mod subcommands;
 use crate::cli::Opt;
 use crate::subcommands::{files::files_cmds, register::register_cmds, wallet::wallet_cmds, SubCmd};
 use sn_client::Client;
-use sn_logging::init_logging;
 #[cfg(feature = "metrics")]
 use sn_logging::metrics::init_metrics;
 
 use clap::Parser;
 use color_eyre::Result;
 use std::path::PathBuf;
-use tracing_core::Level;
 
 #[tokio::main]
 async fn main() -> Result<()> {
     let opt = Opt::parse();
-    // For client, default to log to std::out
-    // This is ruining the log output for the CLI. Needs to be fixed.
-    let tmp_dir = std::env::temp_dir();
-    let logging_targets = vec![
-        ("safe".to_string(), Level::INFO),
-        ("sn_client".to_string(), Level::INFO),
-        ("sn_networking".to_string(), Level::INFO),
-    ];
-    let log_appender_guard =
-        init_logging(logging_targets, &Some(tmp_dir.join("safe-client")), false)?;
     #[cfg(feature = "metrics")]
     tokio::spawn(init_metrics(std::process::id()));
 
     debug!("Built with git version: {}", sn_build_info::git_info());
     println!("Built with git version: {}", sn_build_info::git_info());
-    info!("Full client logs will be written to {:?}", tmp_dir);
     println!("Instantiating a SAFE client...");
 
     let secret_key = bls::SecretKey::random();
@@ -66,7 +53,6 @@ async fn main() -> Result<()> {
         SubCmd::Register(cmds) => register_cmds(cmds, &client).await?,
     };
 
-    drop(log_appender_guard);
     Ok(())
 }
 
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -15,9 +15,9 @@ mod subcommands;
 use crate::cli::Opt;
 use crate::subcommands::{files::files_cmds, register::register_cmds, wallet::wallet_cmds, SubCmd};
 use sn_client::Client;
-use sn_logging::init_logging;
 #[cfg(feature = "metrics")]
 use sn_logging::metrics::init_metrics;
+use sn_logging::{init_logging, LogFormat};
 
 use clap::Parser;
 use color_eyre::Result;
@@ -33,7 +33,11 @@ async fn main() -> Result<()> {
             ("sn_client".to_string(), Level::INFO),
             ("sn_networking".to_string(), Level::INFO),
         ];
-        let _log_appender_guard = init_logging(logging_targets, log_output_dest, false)?;
+        let _log_appender_guard = init_logging(
+            logging_targets,
+            log_output_dest,
+            opt.log_format.unwrap_or(LogFormat::Default),
+        )?;
     }
     #[cfg(feature = "metrics")]
     tokio::spawn(init_metrics(std::process::id()));
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -27,18 +27,20 @@ use tracing::Level;
 #[tokio::main]
 async fn main() -> Result<()> {
     let opt = Opt::parse();
-    if let Some(log_output_dest) = opt.log_output_dest {
+    let _log_appender_guard = if let Some(log_output_dest) = opt.log_output_dest {
         let logging_targets = vec![
             ("safe".to_string(), Level::INFO),
             ("sn_client".to_string(), Level::INFO),
             ("sn_networking".to_string(), Level::INFO),
         ];
-        let _log_appender_guard = init_logging(
+        init_logging(
             logging_targets,
             log_output_dest,
             opt.log_format.unwrap_or(LogFormat::Default),
-        )?;
-    }
+        )?
+    } else {
+        None
+    };
     #[cfg(feature = "metrics")]
     tokio::spawn(init_metrics(std::process::id()));
 
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -15,16 +15,26 @@ mod subcommands;
 use crate::cli::Opt;
 use crate::subcommands::{files::files_cmds, register::register_cmds, wallet::wallet_cmds, SubCmd};
 use sn_client::Client;
+use sn_logging::init_logging;
 #[cfg(feature = "metrics")]
 use sn_logging::metrics::init_metrics;
 
 use clap::Parser;
 use color_eyre::Result;
 use std::path::PathBuf;
+use tracing::Level;
 
 #[tokio::main]
 async fn main() -> Result<()> {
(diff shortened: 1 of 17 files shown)


sn_node/src/bin/safenode/main.rs
```diff
@@ -274,18 +290,73 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
     });
 }
 
-async fn create_secret_key_file(path: impl AsRef<Path>) -> Result<tokio::fs::File, std::io::Error> {
-    let mut opt = tokio::fs::OpenOptions::new();
+fn init_logging(
+    log_output_dest: Option<LogOutputDestArg>,
```
RolandSherwin: Just a small Q, why do we have to parse the `opt.log_output_dest` into `LogOutputDestArg` here and then pass it around?
Instead, could we directly send `opt.log_output_dest` to  `fn init_logging` to obtain the `LogOutputDest`?

jacderida: Hi Roland,

Thanks for the question. I'm sorry but I don't quite understand what you're asking here. Would you mind posting a code change suggestion?

Cheers,

Chris

sn_node/src/bin/safenode/main.rs
```diff
@@ -274,18 +290,73 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
...
+                    .join("safe")
+                    .join("node")
+                    .join(peer_id.to_string())
+                    .join("logs");
+                LogOutputDest::Path(path)
+            }
+            LogOutputDestArg::Path(path) => LogOutputDest::Path(path),
+        }
+    } else {
+        LogOutputDest::Stdout
```
b-zee: It's confusing that the default is not "default", but "stdout".

sn_node/src/bin/safenode/main.rs
```diff
@@ -274,18 +290,73 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
...
+    let output_dest = if let Some(log_output_dest) = log_output_dest {
+        match log_output_dest {
+            LogOutputDestArg::Stdout => LogOutputDest::Stdout,
+            LogOutputDestArg::Default => {
+                let path = dirs_next::data_dir()
+                    .ok_or_else(|| eyre!("could not obtain data directory path".to_string()))?
+                    .join("safe")
+                    .join("node")
+                    .join(peer_id.to_string())
+                    .join("logs");
```
b-zee:
```suggestion
                let path = get_root_dir(peer_id)?.join("logs");
```

sn_cli/src/main.rs
```diff
@@ -15,34 +15,35 @@ mod subcommands;
...
-    ];
-    let log_appender_guard =
-        init_logging(logging_targets, &Some(tmp_dir.join("safe-client")), false)?;
+    if let Some(log_output_dest) = opt.log_output_dest {
+        let logging_targets = vec![
+            ("safe".to_string(), Level::INFO),
+            ("sn_client".to_string(), Level::INFO),
+            ("sn_networking".to_string(), Level::INFO),
+        ];
+        let _log_appender_guard = init_logging(
```
b-zee: Shouldn't we keep this guard until the end of `main` scope?

sn_node/src/bin/safenode/main.rs
```diff
@@ -32,25 +33,57 @@ use tokio::{
...
-    log_dir: Option<PathBuf>,
+    /// Valid values are "stdout", "default", or a custom path.
+    ///
+    /// The default location is platform specific:
+    ///  - Linux: $HOME/.local/share/safe/node/<peer-id>/logs
+    ///  - macOS: $HOME/Library/Application Support/safe/node/<peer-id>/logs
+    ///  - Windows: C:\Users\<username>\AppData\Roaming\safe\node\<peer-id>\logs
+    #[allow(rustdoc::invalid_html_tags)]
+    #[clap(long, value_parser = parse_log_output, verbatim_doc_comment)]
+    log_output_dest: Option<LogOutputDestArg>,
```
b-zee, line 71: What's the rationale behind renaming to this from `log_dir`?

sn_node/src/bin/safenode/main.rs
RolandSherwin: Sure Chris, I was suggesting to move the `parsing` logic to the `init_logging` function
This would allow us to bypass the intermediate `LogOutputDestArg` state.

```suggestion
fn init_logging(
    log_output_dest: Option<String>,
    peer_id: PeerId,
    format: Option<LogFormat>,
) -> Result<(String, Option<WorkerGuard>)> {
    let logging_targets = vec![
        ("safenode".to_string(), Level::INFO),
        ("sn_transfers".to_string(), Level::INFO),
        ("sn_networking".to_string(), Level::INFO),
        ("sn_node".to_string(), Level::INFO),
    ];

    let output_dest = if let Some(log_output_dest) = log_output_dest {
        match log_output_dest.as_str() {
            "stdout" => LogOutputDest::Stdout,
            "default" => {
                let path = dirs_next::data_dir()
                    .ok_or_else(|| eyre!("could not obtain data directory path".to_string()))?
                    .join("safe")
                    .join("node")
                    .join(peer_id.to_string())
                    .join("logs");
                LogOutputDest::Path(path)
            }
            // The path should be a directory, but we can't use something like `is_dir` to check
            // because the path doesn't need to exist. We can create it for the user.
            value => LogOutputDest::Path(PathBuf::from(value)),
        }
    } else {
        LogOutputDest::Stdout
    };

```

jacderida: Hi Roland,

Thanks. If you look at the definition for the `log_output_dest` argument, the parsing function is actually assigned as part of the definition. So what you would actually be suggesting would be to turn that argument into just a `String` rather than a 'special' type. I think we probably want to retain the use of the special type.

sn_node/src/bin/safenode/main.rs
```diff
@@ -329,13 +400,13 @@ fn get_root_dir(peer_id: PeerId) -> Result<PathBuf> {
 
 /// The keypair is located inside the root directory. At the same time, when no dir is specified,
 /// the dir name is derived from the keypair used in the application: the peer ID is used as the directory name.
-async fn get_root_dir_and_keypair(root_dir: Option<PathBuf>) -> Result<(PathBuf, Keypair)> {
+fn get_root_dir_and_keypair(root_dir: Option<PathBuf>) -> Result<(PathBuf, Keypair)> {
     match root_dir {
         Some(dir) => {
-            tokio::fs::create_dir_all(&dir).await?;
+            std::fs::create_dir_all(&dir)?;
```
b-zee, line 420: What is the reason behind changing this?

b-zee (approved):

PR #472, opened 2023-07-03
WIP test(node): impl TestSwarm + verify chunk PUT

RolandSherwin:
WIP

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3695,6 +3695,7 @@ dependencies = [
  "rand 0.8.5",
  "rmp-serde",
  "serde",
+ "sn_logging",
  "sn_protocol",
  "thiserror",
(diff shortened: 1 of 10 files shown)


joshuef:
It's not clear to me what we're getting here... Why do we need this?

RolandSherwin:
> It's not clear to me what we're getting here... Why do we need this?

I was wondering if we might need unit tests to isolate and test certain parts of our codebase. We already have the larger tests that might as well cover these functions, but abstracting away the network might allow us to test node features in isolation. 

The mock network would be kept as simple as possible, just a simple key value pair for the data without reimplementing kademlia.

RolandSherwin:
Closing because it is outdated, will work on this again.

PR #479, opened 2023-07-05
WIP feat: client get_closest from network for each entry

maqi:
### Summary generated by Reviewpad on 05 Jul 23 10:20 UTC
This pull request introduces a new feature where the client can get the closest peers from the network for each entry. The code changes made in this pull request include modifications to the `api.rs` and `file_apis.rs` files. In `api.rs`, the `closest_peers` argument has been removed from the `store_chunk` method and the method now uses the `client_send_to_closest` method from the network to send and get responses. Additionally, the `get_all_local_peers` and `get_closest_local_peers` methods have been removed from the `Client` struct. In `file_apis.rs`, the `store_chunk` method no longer retrieves the closest local peers and instead directly calls the `store_chunk` method from the `Client` struct. The `get_all_local_peers` method has also been removed from the `Files` struct.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -19,7 +19,7 @@ use indicatif::ProgressBar;
 use itertools::Itertools;
 use libp2p::{
     kad::{RecordKey, K_VALUE},
-    Multiaddr, PeerId,
+    Multiaddr,
 };
 use sn_dbc::{DbcId, SignedSpend};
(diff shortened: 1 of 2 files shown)


maqi:
Closed as in favour of the `using put_record for uploading` approach,
as implemented in the PR https://github.com/maidsafe/safe_network/pull/482

PR #480, opened 2023-07-05
fix(node): verify incoming `Record::key`

RolandSherwin:
- The `Record::key` should be the same one obtained from the data inside `Record::value`
- This prevents someone from storing data to the wrong part of the network or prevents overwriting of valid data.

### Summary generated by Reviewpad on 05 Jul 23 13:01 UTC
This pull request fixes the issue with verifying the incoming `Record::key`. The patch ensures that the `Record::key` is the same one obtained from the data inside `Record::value`, preventing data storage to the wrong part of the network or overwriting of valid data. The patch includes changes to the `sn_node/src/put_validation.rs` file, with 44 insertions and 9 deletions.

### DIFF
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -25,6 +25,9 @@ use sn_transfers::payment_proof::validate_payment_proof;
 use std::collections::HashSet;
 use xor_name::XorName;
 
+// The max length of `Vec<SignedSpend>` that is permitted
+const MAX_SIGNED_SPENDS_LENGTH: usize = 2;
+
 impl Node {
     /// Validate and store a record to the RecordStore
     pub(crate) async fn validate_and_store_record(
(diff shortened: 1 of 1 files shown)


sn_node/src/put_validation.rs
```diff
@@ -144,6 +171,12 @@ impl Node {
         &mut self,
         signed_spends: Vec<SignedSpend>,
     ) -> Result<CmdOk, ProtocolError> {
+        // Prevents someone from crafting large Vec and slowing down nodes.
+        if signed_spends.len() > 2 {
```
maqi: better make this as an const

sn_node/src/put_validation.rs
```diff
@@ -144,6 +172,12 @@ impl Node {
...
+            warn!("Discarding incoming DbcSpend PUT as it contains more than 2 SignedSpends");
```
maqi: same here make it taking variable from the const

maqi (approved):

PR #497, opened 2023-07-10
feat: client upload Register via put_record

maqi:
### Summary generated by Reviewpad on 10 Jul 23 10:39 UTC

- In the `sn_protocol/src/messages/mod.rs` file, an `allow(clippy::large_enum_variant)` attribute was added to suppress the `clippy::large_enum_variant` lint warning. A new variant was defined in the `NetworkEvent` enum.
- In the `sn_protocol/src/response.rs` file, the `CreateRegister` and `EditRegister` responses were removed from the `CmdResponse` enum.
- In the `sn_protocol/src/cmd.rs` file, an `allow(clippy::large_enum_variant)` attribute was added to the `SwarmCmd` enum. A new variant `StartListening` was also added to the `SwarmCmd` enum.
- In the `sn_api/src/cmd.rs` file, the `RegisterCmd` variant was removed from the `Cmd` enum and its associated code was also removed. The usage of `RegisterCmd` in the `Cmd::dst` and `Cmd::fmt` implementations was also eliminated.
- In the `sn_api/src/api.rs` file, a new function `get_register_from_network` was added to the `Client` struct. This function retrieves a register from the network based on a provided address.

These changes aim to improve code organization, suppress warnings, and simplify the logic for handling register creation and edit commands in the project.

### DIFF
--- a/sn_protocol/src/messages/register.rs
+++ b/sn_protocol/src/messages/register.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use sn_registers::{EntryHash, Permissions, RegisterAddress, RegisterOp, User};
+use sn_registers::{Permissions, RegisterAddress, RegisterOp, User};
 
 #[allow(unused_imports)] // needed by rustdocs links
 use crate::messages::QueryResponse;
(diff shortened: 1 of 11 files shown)


sn_protocol/src/messages/register.rs
```diff
@@ -16,47 +16,6 @@ use sn_registers::Register;
...
-    ///
-    /// [`GetRegister`]: QueryResponse::GetRegister
-    Get(RegisterAddress),
-    /// Retrieve the current entries from the [`Register`] at the given address.
-    ///
-    /// Multiple entries occur on concurrent writes. This should eventually lead to a
-    /// [`ReadRegister`] response.
-    ///
-    /// [`ReadRegister`]: QueryResponse::ReadRegister
-    Read(RegisterAddress),
```
joshuef, line 34: Are we not here removing a lot of our register API surface? 

Might we want to replicate these at the client if they're being removed at the protocol level?

maqi: not sure, but so far they are only used when transmission messages,
and as clippy complains shows, none of them is required any more after switched to `kad::record` approach.

joshuef (approved):

PR #498, opened 2023-07-10
chore: add more logging around replication

joshuef:
### Summary generated by Reviewpad on 10 Jul 23 13:24 UTC

Patch 1/2 is a chore that adds more logging around replication. It includes changes to the `api.rs`, `log_markers.rs`, and `replication.rs` files. The changes in `api.rs` add logging statements for when a peer is removed from the routing table and when a lost record is detected. The changes in `log_markers.rs` add new log markers for these events. The changes in `replication.rs` add logging statements for when keys are being fetched for replication.

Patch 2/2 is a feature that makes replication flows per peer. It includes changes to the `cmd.rs`, `lib.rs`, `replication_fetcher.rs`, and `api.rs` files. The changes in `cmd.rs` modify the `AddKeysToReplicationFetcher` command to use the new `add_keys_to_replicate_per_peer` method in the replication fetcher. The changes in `lib.rs` rename the `notify_fetch_result` method to `notify_replication_fetch_result` and update its usage in the `api.rs` file. The changes in `replication_fetcher.rs` modify the `add_keys_to_replicate_per_peer` method to add keys to replicate per peer and limit the number of parallel fetches per peer. The changes in `api.rs` update the usage of the `notify_replication_fetch_result` method.

Overall, these patches improve the logging and replication flow in the codebase.

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -158,7 +158,26 @@ impl SwarmDriver {
                 let _ = sender.send(peers);
             }
             SwarmCmd::AddKeysToReplicationFetcher { peer, keys, sender } => {
-                let keys_to_fetch = self.replication_fetcher.add_keys(peer, keys);
+                // check if we have any of the data before adding it.
+                let existing_keys: HashSet<NetworkAddress> = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .store_mut()
+                    .record_addresses();
+
+                // remove any keys that we already have from replication fetcher
+                self.replication_fetcher.remove_held_data(&existing_keys);
+
+                let non_existing_keys: Vec<NetworkAddress> = keys
+                    .iter()
+                    .filter(|key| !existing_keys.contains(key))
+                    .cloned()
+                    .collect();
+
+                let keys_to_fetch = self
+                    .replication_fetcher
+                    .add_keys_to_replicate_per_peer(peer, non_existing_keys);
                 let _ = sender.send(keys_to_fetch);
             }
             SwarmCmd::NotifyFetchResult {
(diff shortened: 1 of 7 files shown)


sn_networking/src/cmd.rs
```diff
@@ -158,7 +158,26 @@ impl SwarmDriver {
                 let _ = sender.send(peers);
             }
             SwarmCmd::AddKeysToReplicationFetcher { peer, keys, sender } => {
-                let keys_to_fetch = self.replication_fetcher.add_keys(peer, keys);
+                // check if we have any of the data before adding it.
+                let existing_keys: HashSet<NetworkAddress> = self
```
maqi, line 149: node level already carried out the filtering.

This block of filtering code was originally in SwarmDriver.
But got moved up to Node when moving the whole replication flow up to Node.

It maybe good to have it here, so the knowledge is more synced.
But in that case, the filtering in Node shall then be removed to avoid un-necessary work.

sn_networking/src/replication_fetcher.rs
```diff
@@ -31,51 +31,67 @@ pub(crate) enum HolderStatus {
 #[derive(Default)]
 pub(crate) struct ReplicationFetcher {
     to_be_fetched: BTreeMap<NetworkAddress, BTreeMap<PeerId, (Instant, HolderStatus)>>,
-    on_going_fetches: usize,
+    on_going_fetches: BTreeMap<PeerId, usize>,
```
maqi: this per `Peer` measure may have a risk of contacting duplicate peers for the same entry,
or too peers contacted concurrently causing self choked when receiving responses.

sn_networking/src/replication_fetcher.rs
```diff
@@ -118,9 +136,9 @@ impl ReplicationFetcher {
             }
         }
 
-        for failed_key in all_failed_entries {
-            let _ = self.to_be_fetched.remove(&failed_key);
-        }
+        // for failed_key in all_failed_entries {
```
maqi: Just a note: decide to retain or remove this block of code before merge

maqi (approved):

PR #499, opened 2023-07-10
only query one node at once

joshuef:
## Description

atop #498 
reviewpad:summary


joshuef:
As @maqi noted, this will just spread out amongst all nodes, so no real effect.

We'll need to randomised the peers to ensure that we're not accidentally ordering and overloading peers across the network in general

PR #502, opened 2023-07-10
test(node): verify data location

RolandSherwin:


sn_node/src/protocol/safenode_proto/req_resp_types.proto
```diff
@@ -23,11 +23,26 @@ message NodeInfoResponse {
...
+  repeated bytes connected_peers = 1;
+  repeated string listeners = 2;
+}
+
 // Stream of node events
 message NodeEventsRequest {}
 
 message NodeEvent {
-  string event = 1;
+  bytes event = 1;
```
RolandSherwin, line 38: An earlier version of the test relied on the `NodeEvent` to obtain the address of the data that was PUT into each node.
But this has been later changed as it was unreliable. 

Thus, the transmitted `NodeEvent as bytes` is not used anymore. But I do think that it can be useful down the line, hence keeping the changes here.

maqi (changes requested):
May this PR to be split into couple of smaller PRs .
Seems it contains certain part not related to a test itself, and be more suitable to be raised PR separately.

also the PR title is bit too narrowed, and there is no PR description ?

joshuef:
As this will be replaced by some smaller PRs i'm going to close this one to clean up the board 👍

PR #503, opened 2023-07-10
feat(node): shuffle data waiting for fetch

joshuef:
This should more evenly spread out replications across the network as oppsosed to forcing all ndoes to query the same space at the same time

### Summary generated by Reviewpad on 10 Jul 23 19:50 UTC
This pull request introduces a new feature to the node's replication fetch process. It implements shuffling of data waiting to be fetched in order to evenly spread out replications across the network. The patch modifies the `replication_fetcher.rs` file and increases the value of `MAX_PARALLEL_FETCH` to 8. Additionally, it adds code to randomize the order of keys to fetch, resulting in a more balanced distribution of replication requests. The pull request also includes some debug logs for tracking the number of records awaiting fetch.


grumbach (approved):
✅ 🚢

PR #507, opened 2023-07-11
feat: use libp2p patch for NotDialing default

b-zee:


joshuef:
Closed in favour of #511

PR #512, opened 2023-07-12
 chore(metrics): remove network stats 

RolandSherwin:
### Summary generated by Reviewpad on 12 Jul 23 08:54 UTC
This pull request fixes the network metrics by collecting metrics from all the network interfaces instead of just the default one. It modifies the `metrics.rs` file to update the `NetworkMetrics` struct to include statistics for all interfaces. The `init_metrics` function has been updated to loop through all the network interfaces and sum up the received and transmitted bytes and calculate the total MBytes received and transmitted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1298,23 +1298,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "default-net"
-version = "0.15.0"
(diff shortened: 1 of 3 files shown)


joshuef:
I don't think this is necessary, these stats per node are actually already not that helpful. I think we could perhaps just drop them?

joshuef (approved):

Issue #513, opened 2023-07-12
Stop using random keys


grumbach:
https://github.com/maidsafe/safe_network/blob/a30859dca3c5fbddb4e826c201d72bcfc4cc4e63/sn_cli/src/main.rs#L51

We use a different key at every command run. The direct implication I’m seeing is, now that registers require signatures, I get:
```
Error: Register error Access denied for user: Key(PublicKey(110f..2d1f)).
Error: Register error Access denied for user: Key(PublicKey(0fca..fb6c)).
...
```
Since we use a different key every time, if I create a register with: `safe register create myname`
I can’t write to it with a subsequent `safe register edit myname hello` because I have a new key, basically I loose access to it as soon as the command is over.

PR #514, opened 2023-07-12
feat: safer registers requiring signatures

grumbach:
### Summary generated by Reviewpad on 12 Jul 23 09:10 UTC
This pull request includes several file diffs with various changes related to imports, struct definitions, method implementations, and other code modifications. Here is a summary of the key changes:

1. The file `query.rs` in the root directory:
   - Changes to imports, the `Query` enum, and associated function implementations.
   - Addition of a new variant `GetRegister` in the `Query` enum.
   - Modifications to the `dst()` and `Display` implementations.

2. The file `sn_node/src/get_validation.rs`:
   - Replacement of an import statement from `sn_registers::Register` to `sn_registers::SignedRegister`.
   - Update to return type of the `get_register_from_network` function to `Result<SignedRegister>`.
   - Modifications to use `SignedRegister` instead of `Register` in the `impl_node` function.

3. The file `sn_client/src/register.rs`:
   - Addition of the import `bls::PublicKey`.
   - Changes related to the `sn_protocol` and `sn_registers` modules.
   - Modifications to use the `Permissions` struct for creating a new register with the `new_owner_only()` method.
   - Update in the `check_user_permissions` method to handle `PublicKey` instead of `User`.
   - Changes in the `write` and `sync` methods of `ClientRegister`.
   - Renaming of the `get_register` method to `get_register_from_peers`.
   - Modifications related to debug statements, register retrieval, merging, and validation.

4. The file `sn_registers/src/lib.rs`:
   - Changes in the `Register` struct to include the `SignedRegister` struct.
   - Updates in imports related to the new structs and modules.

5. The file `sn_protocol/src/messages/response.rs`:
   - Removal of an unused import and modifications in other import statements.
   - Replacement of `std::collections::BTreeSet` with `sn_registers::SignedRegister`.
   - Additions of the variant `GetRegister` with a result of `SignedRegister` in the `QueryResponse` enum.

6. The file `sn_client/src/cmd.rs`:
   - Updates in imports related to the `sn_protocol` module.
   - Changes in the type of `Register` to `SignedRegister`.
   - Removal of the `handle_register_query` function.
   - Modifications in the `handle_register_cmd` function, including parameter changes and other logic updates.

7. The file `sn_client/src/error.rs`:
   - Addition of a new variant `RegisterAlreadyClaimed` to the `Error` enum with a field of type `bls::PublicKey`.
   - Addition of an error message for when a register is already created by another owner.

8. The file `error.rs`:
   - Addition of two new error variants to the `Error` enum: `DifferentBaseRegister` and `InvalidSecretKey`.
   - These variants handle cases where signed register merging fails due to different base registers and invalid secret keys provided for signing.

9. The file `register_op.rs`:
   - Changes related to imports, traits, methods, and their implementations for the `RegisterOp` struct.
   - Improved functionality with the addition of a generic `sign_with` method and other modifications.

10. The file `api.rs` in the `sn_client` module:
    - Changes to the `sign` method to accept a generic type for signing data.
    - Updates to the `create_register` and `ClientRegister::create` methods, including log messages and asynchronous handling.

11. The file `sn_protocol/src/messages/mod.rs`:
    - Modifications in import statements to include or remove certain module types.

12. The file `sn_protocol/src/register.rs`:
    - Removal of an import for `User` and `Permissions` from the `sn_registers` module.
    - Updates related to the `RegisterQuery` enum, `RegisterCmd` variant fields, and `dst` methods.

13. The file `api.rs`:
    - Changes to imports, function parameter types, and method calls related to `Register` and `SignedRegister`.
    - Modifications in the `validate_and_store_register` and `register_validation` functions.
    - Usage of `merged_register.verified_merge` method instead of `merged_register.merge`.

14. The file `api.rs` in the `sn_node` module:
    - Introduction of a new variant `Query::GetRegister` in the `Query` enum.
    - Handling of the `Query::GetRegister` variant, including trace logging and calling the `get_register_from_network` method.

These changes include updates and improvements in various aspects of the codebase, such as register handling, error types, signatures, and query operations.


sn_client/src/register.rs
```diff
@@ -218,10 +228,11 @@ impl ClientRegister {
     // Create a new `ClientRegister` instance with the given name and tag.
     fn new(client: Client, name: XorName, tag: u64) -> Result<Self> {
         let public_key = client.signer_pk();
-        let owner = User::Key(public_key);
-        let perms = Permissions::new_anyone_can_write();
+        // NB TODO permissions should be configurable
+        // using owner only for now, as it is the most restrictive
+        let perms = Permissions::new_owner_only();
```
grumbach: This is not pressing matter but we should, once we get to client work, think about this

sn_client/src/register.rs
```diff
@@ -292,17 +303,58 @@ impl ClientRegister {
...
+                    }
+                }
+            })
+            .collect::<Vec<SignedRegister>>();
+
+        debug!(
+            "Got {}/{} valid Register responses for Register at {address:?}",
+            regs.len(),
+            responses.len()
+        );
```
grumbach: For now, even if only 1 node responded with a valid response we take it. Maybe we want to be stricter and require majority?

sn_protocol/src/messages/response.rs
```diff
@@ -44,18 +43,12 @@ pub enum QueryResponse {
...
-    /// Response to [`RegisterQuery::GetOwner`].
-    GetRegisterOwner(Result<User>),
-    /// Response to [`RegisterQuery::Read`].
-    ReadRegister(Result<BTreeSet<(EntryHash, Entry)>>),
-    /// Response to [`RegisterQuery::GetPermissions`].
-    GetRegisterPermissions(Result<Permissions>),
+    /// Response to [`GetRegister`]
+    ///
+    /// [`GetRegister`]: crate::messages::Query::GetRegister
+    GetRegister(Result<SignedRegister>),
```
grumbach: All this was un-secure AND unused by the client, replaced it with a simpler and secure API.

sn_registers/src/register.rs
```diff
@@ -27,7 +27,7 @@ const MAX_REG_NUM_ENTRIES: u16 = 1024;
 #[derive(Clone, Eq, PartialEq, PartialOrd, Hash, Serialize, Deserialize, Debug)]
 pub struct Register {
     /// Owner of the Register
-    owner: User,
+    owner: PublicKey,
```
grumbach, line 30: Changed it so the owner of a Register is a `PublicKey`, as it cannot be `User::Anyone`

sn_cli/src/subcommands/register.rs
```diff
@@ -48,7 +48,8 @@ async fn create_register(name: String, client: &Client) -> Result<()> {
     let xorname = XorName::from_content(name.as_bytes());
     println!("Creating Register with '{name}' at xorname: {xorname:x} and tag {tag}");
 
-    let _register = client.create_register(xorname, tag).await?;
+    // clients currently only support public registers as we create a new key at each run
+    let _register = ClientRegister::create_public_online(client.clone(), xorname, tag).await?;
```
grumbach, line 52: This shouldn't be necessary anymore after we fix #513

sn_client/src/register.rs
```diff
@@ -28,14 +32,49 @@ pub struct ClientRegister {
...
+            register,
+            ops: LinkedList::new(),
+        };
+
+        Ok(reg)
+    }
+
+    /// Create a new public Register (Anybody can write to it) and send it so the Network.
+    pub async fn create_public_online(client: Client, name: XorName, tag: u64) -> Result<Self> {
+        let public_key = client.signer_pk();
```
maqi: taking `create, create_public_online, create_online` all into account,
to avoid duplication and any mis-synced coding step among them,
I'd suggest to have a common function to be a common util func to all of these three, which takes perm in as well.
and make these three function names more explicit

sn_registers/src/register.rs
```diff
@@ -36,31 +36,144 @@ pub struct Register {
...
+    /// Returns a bytes version of the Register used for signing
+    /// Use this API when you want to sign a Register withtout providing a secret key to the Register API
+    pub fn bytes(&self) -> Result<Vec<u8>> {
+        bincode::serialize(self).map_err(|_| Error::SerialisationFailed)
+    }
+
+    /// Sign a Register into a SignedRegister
+    pub fn into_signed(self, secret_key: &SecretKey) -> Result<SignedRegister> {
+        let bytes = self.bytes()?;
+        let signature = secret_key.sign(bytes);
```
RolandSherwin: Would we want to call `let signature = self.sign(secret_key)` here? This would make sure that the signer is the owner
I assume, we would not want to sign the `Register` with a `random secret_key` as that would result in the failure of  `SignedRegister::verify`.

sn_protocol/src/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     InvalidRegister(RegisterAddress),
     #[error("Register is Invalid: {0}")]
     RegisterError(#[from] sn_registers::Error),
+    #[error("The Register was already created by another owner: {0:?}")]
+    RegisterAlreadyClaimed(bls::PublicKey),
```
maqi, line 41: better using Register::name or Register::address to be consistent with other logging fmt

sn_protocol/src/messages/query.rs
```diff
@@ -29,6 +29,13 @@ pub enum Query {
     /// [`Chunk`]:  crate::storage::Chunk
     /// [`GetChunk`]: super::QueryResponse::GetChunk
     GetChunk(ChunkAddress),
+    /// Retrieve a [`Register`] at the given address.
+    ///
+    /// This should eventually lead to a [`GetRegister`] response.
+    ///
+    /// [`Register`]: sn_registers::Register
```
RolandSherwin: The response would be a `SignedRegister` instead of a `Register` here.
```suggestion
    /// Retrieve a [`SignedRegister`] at the given address.
    ///
    /// This should eventually lead to a [`GetRegister`] response.
    ///
    /// [`SignedRegister`]: sn_registers::SignedRegister
```

sn_protocol/src/messages/mod.rs
```diff
@@ -28,7 +28,7 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 use sn_dbc::SignedSpend;
-use sn_registers::Register;
+use sn_registers::SignedRegister;
```
maqi, line 31: do we have the user scenario to store/operate un-signed register on network?
If don't, maybe just `assume register shall always be signed` ?
Then we can just stay with the existing name of `Register`, but they shall always be signed?

I just feel the current `Register` and `SignedRegister` are bit confusing and misleading

sn_protocol/src/messages/query.rs
```diff
@@ -29,6 +29,13 @@ pub enum Query {
...
+    /// [`GetRegister`]: super::QueryResponse::GetRegister
+    GetRegister(RegisterAddress),
```
maqi, line 38: It's for signed or un-signed?

sn_protocol/src/messages/register.rs
```diff
@@ -22,14 +17,10 @@ use xor_name::XorName;
 pub enum RegisterCmd {
     /// Create a new [`Register`] on the network.
```
RolandSherwin: Just a tiny suggestion, we might want to change the references from `Register` -> `SignedRegister`.
```suggestion
    /// Create a new [`SignedRegister`] on the network.
```

Also, the doc `/// A [Register] cmd that is stored in a log on Adults.` is outdated, I think.

sn_registers/src/register.rs
```diff
@@ -36,31 +36,144 @@ pub struct Register {
     permissions: Permissions,
 }
 
+/// A Signed Register on the SAFE Network
+/// This cryptographically secure version of the Register is used to make sure that the data cannot be tampered with
+#[derive(Clone, Debug, Serialize, Deserialize, PartialOrd, PartialEq, Eq, Hash)]
+pub struct SignedRegister {
```
maqi, line 42: for the naming structure, it seems we shall have:
SignedRegister, UnSignedRegister then both using BaseRegister as their base (or UnSignedRegister is the same to BaseRegister)

Or, from user perspective, we shall have the naming of:
PrivateRegister, PulicRegister, and both using BaseRegister as their base

sn_protocol/src/messages/query.rs
grumbach: It's signed, indeed confusing, Roland also pointed this out, adapted the comments. Thanks!

sn_protocol/src/messages/mod.rs
grumbach: `Register` is the local copy of your register, it's the one you're operating on, `SignedRegister` is the signed/verifiable version that we store on the Network. In theory, users should only deal with `Registers` and Network only with `SignedRegisters`.

sn_registers/src/register.rs
grumbach: > UnSignedRegister is the same to BaseRegister

Yes

> Or, from user perspective, we shall have the naming of:
> PrivateRegister, PulicRegister, and both using BaseRegister as their base

We already use the Private and Public nomenclature for permissions, using it also for Signed and Raw register might be a bit dangerous

sn_protocol/src/error.rs
grumbach: Agree that we need consistency, but in this context, I think it might be more interesting to know who already created the register rather than the address.

RolandSherwin (approved):

PR #520, opened 2023-07-13
Avoid get closest local calls into the swarm itself

joshuef:
Avoid get closest local calls into the swarm itself

### Summary generated by Reviewpad on 13 Jul 23 10:07 UTC

1. The file `Cargo.lock` has a new dependency on `custom_debug`.
2. The `Cargo.toml` file now includes a new dependency, `custom_debug` version 0.5.1.
3. The `SwarmDriver` struct now has a new field `all_local_peers` of type `Vec<PeerId>`, which is initialized with a single element in the constructor.
4. There are multiple changes related to the `SwarmCmd` enum and its implementations in the `SwarmDriver` struct.
5. Some changes have been made to handle swarm events in the `SwarmDriver` implementation.
6. Logging has been simplified and improved in certain areas.
7. Various modifications have been made to handle specific events and incorporate logging.

Please review these changes and provide any necessary feedback.


sn_networking/src/cmd.rs
```diff
@@ -348,6 +334,23 @@ impl SwarmDriver {
         Ok(())
     }
 
+    /// Update local peers stored on the SwarmDriver
+    /// Also logs the k-buckets information
+    ///
+    /// This should prevent having to dive into the swarm driver to get the local peers
+    pub(crate) fn update_local_peers(&mut self, changed_peer_id: PeerId) {
+        let mut all_peers: Vec<PeerId> = vec![];
+        for kbucket in self.swarm.behaviour_mut().kademlia.kbuckets() {
```
maqi, line 342: the dive into the kad::kbuckets as I understand shall not taking too much resources and causing noticable mem spike issue.
Actually, this extra container of `self.all_local_peers` could be more to be concerned.

If we do want to reduce the calling to swarm_driver, it will be better put this extra container into node level and send the all_peers info together with the event of PeerAdded or PeerRemoved

joshuef: Right now we do this for every `trigger_replication` call. Which there are many many many of. So I thiiink it can be affecting things. I'm still not sure.

This PR atm suffers from some wild CPU. So it feels liike it's unblocked _something_. Im trying to determine what that is...

Yeh I did wonder about pushing it up to Network. Seemed messier, but it would avoid the call into here and some channel work entirely yeh

PR #524, opened 2023-07-14
feat: randomly select unused port

jacderida:
If a port is not specified, an unused port will be selected at random. If you pass a port with the value 0 to the `libp2p` infrastructure, it will give you a port it says is provided by the OS, but it doesn't mention whether it guarantees that the port is not in use. I was not able to determine if this is the case.

We can add this simple mechanism to our own code to select a port that is unused. This will help for running multiple nodes on the same machine, particularly in the context of the testnet tool.

### Summary generated by Reviewpad on 14 Jul 23 21:11 UTC
This pull request adds the feature of randomly selecting an unused port if a port is not specified. This is implemented to help run multiple nodes on the same machine, particularly in the context of the testnet tool. The `get_unused_port()` function shuffles a range of ports and attempts to bind them until an unused port is found. If the provided port is 0, the function is called and the selected port is used. The implementation includes changes to the `sn_node/src/bin/safenode/main.rs` file, where the `get_unused_port()` function is defined and used to determine the node's socket address. Overall, this patch improves the flexibility and convenience of running multiple nodes on a single machine.

### DIFF
--- a/sn_node/src/bin/safenode/main.rs
+++ b/sn_node/src/bin/safenode/main.rs
@@ -14,6 +14,7 @@ mod rpc;
 use clap::Parser;
 use eyre::{eyre, Error, Result};
 use libp2p::{identity::Keypair, Multiaddr, PeerId};
+use rand::seq::SliceRandom;
 #[cfg(feature = "metrics")]
 use sn_logging::metrics::init_metrics;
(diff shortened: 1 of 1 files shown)


jacderida:
The `libp2p` infrastructure guarantees that the port that is selected is unused, so this PR is not necessary.

PR #525, opened 2023-07-17
feat(networking): remove LostRecordEvent

joshuef:
Kad docs on the event we use here:
``` 
FinishedWithNoAdditionalRecord {
/// If caching is enabled, these are the peers closest
        /// _to the record key_ (not the local node) that were queried but
        /// did not return the record, sorted by distance to the record key
        /// from closest to farthest. How many of these are tracked is configured
        /// by [`KademliaConfig::set_caching`]. If the lookup used a quorum of
        /// 1, these peers will be sent the record as a means of caching.
        /// If the lookup used a quorum > 1, you may wish to use these
        /// candidates with [`Kademlia::put_record_to`] after selecting
        /// one of the returned records.
        ```
        
This reads to me like it'll return for any/all peers on the route to the data (as they wont necessarily have that data), and so this might well not be how we want to trigger replication.

I'm leaning towards removing this for now, and see how we go and coming back with a more comprehensive approach for "healing" lost data (that wont trigger as frequently as this seems to).

### Summary generated by Reviewpad on 17 Jul 23 09:39 UTC
This pull request removes the `LostRecordEvent` from the networking feature. The event is currently causing high memory usage and the removal is done temporarily to see if it has any impact. The patch removes the event from multiple locations in the codebase.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -81,11 +81,6 @@ jobs:
       - name: Check the whole workspace can build
         run: cargo build --all-targets --all-features
 
-      - name: Check we could publish sn_node
-        run: cargo publish --dry-run -p sn_node
-      
-      - name: Check we could publish sn_cli
-        run: cargo publish --dry-run -p sn_cli
 
   unit:
     if: "!startsWith(github.event.head_commit.message, 'chore(release):')" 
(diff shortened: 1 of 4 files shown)


grumbach (approved):
Sounds good to me, less is more!

PR #534, opened 2023-07-18
ci: use custom action for testnet admin

jacderida:
- cca6589 **ci: use custom action for testnet management**

  The Powershell script for collecting log artifacts was incorrect, and instead of fixing the
  same script in multiple places, this prompted the creation of a custom action to start and stop the
  local network. After the network is stopped the logs are tar'd and uploaded as artifacts.

  There are three other workflows that launch a local testnet. They use the action to start the
  network, but their logging process involves also collecting logs for heaptrack, so their logging
  scripts remain directly in the workflows. We can make an extension to the custom action later for
  supporting a heaptrack node.

- b0ba039 **chore: remove all workflow step `id` attributes**

  You only need to assign these IDs if you're referring to the step later, like if it has some
  outputs.

  Also change the title of the Nightly workflow, which had an ugly full stop in the middle of it.

- ec755a0 **ci: apply custom action to nightly workflow**

  I missed applying the new local network action to the nightly workflow, which also starts and stops
  a local network.

### Summary generated by Reviewpad on 18 Jul 23 21:16 UTC

- Added a new feature to improve user authentication
- Fixed a bug in the login page
- Updated the README file with new instructions

Please let me know if you need further assistance or have any questions.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -81,7 +81,6 @@ jobs:
       - name: Check the whole workspace can build
         run: cargo build --all-targets --all-features
 
-
   unit:
     if: "!startsWith(github.event.head_commit.message, 'chore(release):')" 
     name: Unit Tests
@@ -135,13 +134,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        include:
-          - os: ubuntu-latest
-            node_data_path: /home/runner/.local/share/safe/node
-          - os: windows-latest
-            node_data_path: C:\\Users\\runneradmin\\AppData\\Roaming\\safe\\node
-          - os: macos-latest
-            node_data_path: /Users/runner/Library/Application Support/safe/node
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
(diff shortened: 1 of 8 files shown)


.github/workflows/merge.yml
```diff
@@ -229,84 +194,35 @@ jobs:
...
-        continue-on-error: true
-        run: |
-          Get-ChildItem -Path "${{ matrix.node_data_path }}" -Filter "*.log*" -Recurse -File | `
-            ForEach-Object {& 'tar' -czvf 'log_files.tar.gz' $_.FullName}
-        if: failure() && matrix.os == 'windows-latest'
-
-      - name: Upload Node Logs
-        uses: actions/upload-artifact@main
+      - name: Stop the local network and upload logs
+        if: always()
```
maqi, line 204: this will upload log zips for successfully runs as well.
could be bit waste of resources.
Shall it got split into two and make the `upload logs` one only execute in failure?

jacderida: Hi Qi,

Thanks for the feedback.

Yeah, I think I turned this on just for testing. I will switch it back to `failure()`. However, it might not do much harm to leave this on. The logs aren't that big, and workflow runs are cleared out every 90 days by the GHA infrastructure.

Cheers,

Chris

maqi (approved):

PR #537, opened 2023-07-19
feat(networking): make record store r/w non blocking

joshuef:
### Summary generated by Reviewpad on 19 Jul 23 10:40 UTC

The first patch makes the record store read/write operations non-blocking. It also adds a TODO comment about cleaning up records if an insert fails. Additionally, it spawns a tokio task to asynchronously write records to disk and handle errors.

The second patch removes an unused function called `add_to_routing_table` in the networking module of the codebase. The function is no longer needed and has been deleted.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -85,6 +85,11 @@ const REQ_RESPONSE_VERSION_STR: &str = concat!("/safe/node/", env!("CARGO_PKG_VE
 const IDENTIFY_CLIENT_VERSION_STR: &str = concat!("safe/client/", env!("CARGO_PKG_VERSION"));
 const IDENTIFY_PROTOCOL_STR: &str = concat!("safe/", env!("CARGO_PKG_VERSION"));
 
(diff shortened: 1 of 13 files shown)


sn_networking/src/lib.rs
```diff
@@ -487,17 +487,6 @@ impl Network {
...
-    pub async fn add_to_routing_table(&self, peer_id: PeerId, peer_addr: Multiaddr) -> Result<()> {
-        let (sender, receiver) = oneshot::channel();
-        self.send_swarm_cmd(SwarmCmd::AddToRoutingTable {
-            peer_id,
-            peer_addr,
-            sender,
-        })?;
-        receiver.await?
-    }
-
```
RolandSherwin: If we are not using the function, then I think we can get rid of the corresponding `SwarmCmd` as well?

maqi (approved):

joshuef:
atop https://github.com/maidsafe/safe_network/pull/543

PR #539, opened 2023-07-19
feat(node): add markers for initial connection attempt/timeout

aed900:
Added AttemptingNetworkConnection marker for attempt to dial initial peers and NetworkConnectionTimingOut marker if timing out.

### DIFF
--- a/.github/workflows/generate-benchmark-charts.yml
+++ b/.github/workflows/generate-benchmark-charts.yml
@@ -66,6 +66,21 @@ jobs:
           node-path: target/release/safenode
           platform: ubuntu-latest
 
+
+      - name: Start a faucet client to claim genesis
+        run: cargo run --bin faucet --release -- claim-genesis
+        env:
+          SN_LOG: "all"
+        timeout-minutes: 2
+
+      - name: Create and fund a wallet to pay for files storage
+        run: |
+          cargo run --bin faucet --release -- send 1000 $(cargo run --bin safe --release -- wallet address | tail -n 1) | tail -n 1 > dbc_hex
+          cat dbc_hex | cargo run --bin safe --release -- wallet deposit --stdin
+        env:
+          SN_LOG: "all"
+        timeout-minutes: 2
+
       - name: Start a heaptracked node instance to compare memory usage
         run: |
           mkdir -p $HEAPNODE_DATA_PATH


sn_node/src/bin/safenode/main.rs
```diff
@@ -296,6 +296,25 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
     let _handle = tokio::spawn(async move {
         loop {
             match node_events_rx.recv().await {
+                Ok(NodeEvent::AttemptingNetworkConnection) => {
+                    Marker::AttemptingNetworkConnection.log();
+                    trace!("Attempting to connect to the network: dialing initial peers");
+                    let inactivity_timeout = Instant::now();
+
+                    loop {
+                        if inactivity_timeout.elapsed() >= Duration::from_secs(60) {
```
b-zee, line 305: This will block the event receiver for 60 seconds. That means other events will not be handled for this period.

sn_node/src/bin/safenode/main.rs
```diff
@@ -296,6 +296,25 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
...
+                            Marker::NetworkConnectionTimingOut.log();
+                            trace!("Attempt to dial initial peers and connect to the network is timing out after {:?}", inactivity_timeout);
+                        }
+
+                        match node_events_rx.recv().await {
+                            Ok(NodeEvent::ConnectedToNetwork) => {
```
b-zee: If other events are coming in here, we might not be handling them down in the real event handlers. E.g. `NodeEvent::BehindNat`, etc.

In case of `ConnectedToNetwork` this will prevent `Marker::NodeConnectedToNetwork.log()` from being called by that handler.

sn_node/src/api.rs
```diff
@@ -205,13 +205,21 @@ impl Node {
...
+                                node_event_sender.broadcast(NodeEvent::AttemptingNetworkConnection);
+                                first_peer = false;
+                            }
                             if let Err(err) = network.dial(addr.clone()).await {
                                 tracing::error!("Failed to dial {addr}: {err:?}");
                             };
                         }
                     });
+                } else {
+                    self.events_channel.broadcast(NodeEvent::AttemptingNetworkConnection);
```
b-zee: What's the rationale behind broadcasting this event on `NewListenAddr`?

Listening on a new address isn't really equivalent to attempting to connect?

sn_node/src/bin/safenode/main.rs
```diff
@@ -296,6 +296,25 @@ fn monitor_node_events(mut node_events_rx: NodeEventsReceiver, ctrl_tx: mpsc::Se
     let _handle = tokio::spawn(async move {
         loop {
             match node_events_rx.recv().await {
+                Ok(NodeEvent::AttemptingNetworkConnection) => {
```
b-zee: Might it be possible to use the existing `PeerAdded` event for this functionality of checking we are connecting to something? Previously that event was actually called `ConnectedToNetwork` I believe.

sn_node/src/bin/safenode/main.rs
aed900: Added an is_connected boolean to handle this instead.

sn_node/src/bin/safenode/main.rs
aed900: Seems sequential to trigger this marker on the first dial of an initial peer before we get to the point of peers being added to routing table and the ConnectedToNetwork event.

PR #541, opened 2023-07-20
feat: persistent cli key and better get

### LINKED ISSUE #513: Stop using random keys

grumbach:
## Description

- fixes #513
- makes `safe register get [name]` print out the register value instead of just saying `Successfully retrieved Register`!

reviewpad:summary

### DIFF
--- a/sn_cli/src/subcommands/register.rs
+++ b/sn_cli/src/subcommands/register.rs
@@ -62,9 +62,8 @@ async fn edit_register(name: String, entry: String, client: &Client) -> Result<(
     match client.get_register(xorname, tag).await {
         Ok(mut register) => {
             println!(
-                "Successfully retrieved Register '{name}' from {}, {}!",
-                register.name(),
-                register.tag()
+                "Successfully retrieved Register '{name}' from {:?}!",
+                register.address(),
             );
             println!("Editing Register '{name}' with: {entry}");
             match register.write_online(entry.as_bytes()).await {
@@ -103,9 +102,8 @@ async fn get_registers(names: Vec<String>, client: &Client) -> Result<()> {
         match client.get_register(xorname, tag).await {
             Ok(register) => {
                 println!(
-                    "Successfully retrieved Register '{name}' from {}, {}!",
-                    register.name(),
-                    register.tag()
+                    "Successfully retrieved Register '{name}' from {:?}!",
+                    register.address(),
                 );
                 let entries = register.read();
                 println!("Register entries:");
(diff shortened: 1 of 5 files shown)


sn_cli/src/subcommands/register.rs
```diff
@@ -101,11 +101,22 @@ async fn get_registers(names: Vec<String>, client: &Client) -> Result<()> {
...
         match client.get_register(xorname, tag).await {
-            Ok(register) => println!(
-                "Successfully retrieved Register '{name}' from {}, {}!",
-                register.name(),
-                register.tag()
-            ),
+            Ok(register) => {
+                println!(
+                    "Successfully retrieved Register '{name}' from {}, {}!",
+                    register.name(),
```
maqi: using `register.address()` will give both.
actually, it shall be `address` to be used across to be consistent and avoid misleading

sn_cli/src/subcommands/register.rs
```diff
@@ -101,11 +101,22 @@ async fn get_registers(names: Vec<String>, client: &Client) -> Result<()> {
...
+                    register.tag()
+                );
+                let entries = register.read();
```
maqi, line 108: better get `Register` with its `Debug` fmt refactored to giving those info ? if they are interested.
also, printly out all bytes might be not necessory, just first couple of hex will be enough and more readable?
like what we do for XorName Debug?

sn_client/src/register.rs
```diff
@@ -58,8 +58,7 @@ impl ClientRegister {
 
     /// Create a new public Register (Anybody can write to it) and send it so the Network.
     pub async fn create_public_online(client: Client, name: XorName, tag: u64) -> Result<Self> {
```
maqi, line 60: better change `name: XorName, tag: u64` to `RegisterAddress` to avoid potential misusing

grumbach: `name: XorName, tag: u64` is used throughout the entire high level API, changing this here means changing our entire API. I don't have a strong opinion on this but we might want to think it through as it changes app devs interface. 
Probably out of scope for this PR.

sn_cli/src/subcommands/register.rs
grumbach: Added a `address()` method to `ClientRegister`, be carefull the `register` here is a `ClientRegister` (includes a `Client`), not a `Register`

joshuef (approved):

PR #542, opened 2023-07-20
chore(networking): simplify OutgoingConnErr handling

joshuef:
### Summary generated by Reviewpad on 20 Jul 23 12:30 UTC
This pull request simplifies the handling of OutgoingConnErr in the networking module. It includes changes to the event.rs file where several lines of code have been modified. The modifications involve the handling of OutgoingConnectionError events, specifically related to dead peers. Previously, the code checked for specific error conditions such as WrongPeerId, ConnectionRefused, and HandshakeTimedOut. Now, the code uses pattern matching to handle these errors in a more concise manner. Additionally, the code removes dead peers from the peer list and logs the changes to the KBuckets. Overall, these changes aim to improve the clarity and efficiency of the OutgoingConnErr handling process.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -21,13 +21,14 @@ use libp2p::{
     multiaddr::Protocol,
     request_response::{self, ResponseChannel as PeerResponseChannel},
     swarm::{behaviour::toggle::Toggle, DialError, NetworkBehaviour, SwarmEvent},
-    Multiaddr, PeerId,
+    Multiaddr, PeerId, TransportError,
 };
 use sn_protocol::{
     messages::{Request, Response},
     NetworkAddress,
 };
 use std::collections::HashSet;
+use std::io::ErrorKind;
 use tokio::sync::oneshot;
 use tracing::{info, warn};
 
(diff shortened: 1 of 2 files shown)


sn_networking/src/event.rs
```diff
@@ -255,32 +256,30 @@ impl SwarmDriver {
...
+
+                let peer_died = match error {
+                    DialError::WrongPeerId { .. } => true,
+                    DialError::Transport(errors) => errors.iter().any(|(_, error)| matches!(error, TransportError::Other(err) if err.kind() == ErrorKind::ConnectionRefused) ),
+                    _ => false,
+                };
+
+                if peer_died {
+                    info!("Detected dead peer {failed_peer_id:?}");
+                    if self.dead_peers.insert(failed_peer_id) {
```
b-zee: In theory, could it happen that a peer is detected as a dead peer twice? E.g. after a few hours.

In that case, inserting will return `false`, but perhaps the peer is still in the routing table?

joshuef: they can come back. but they'd be removed from routing on the first `death`, i think? (and removed from `dead_peers` on rejoin).

(This insert returns if it existed already... so should be functionally the same as before but with one less call?)

b-zee (approved):

joshuef:
superseded by https://github.com/maidsafe/safe_network/pull/548

PR #543, opened 2023-07-20
chore(networking): remove PutRecord kad query tracking.

joshuef:
This doesn't tell us anything useful and is more state to keep track of in the swarm driver

### Summary generated by Reviewpad on 20 Jul 23 12:35 UTC
This pull request removes the PutRecord kad query tracking in the networking module. This change eliminates unnecessary state and reduces complexity in the swarm driver.

### DIFF
--- a/sn_cli/src/subcommands/register.rs
+++ b/sn_cli/src/subcommands/register.rs
@@ -55,14 +55,18 @@ async fn create_register(name: String, client: &Client) -> Result<()> {
     println!("Creating Register with '{name}' at xorname: {xorname:x} and tag {tag}");
 
     // clients currently only support public registers as we create a new key at each run
-    let _register = ClientRegister::create_public_online(client.clone(), xorname, tag).await?;
+    let our_register = ClientRegister::create_public_online(client.clone(), xorname, tag).await?;
 
     let mut verification_attempts = 0;
 
     while verification_attempts < VERIFICATION_ATTEMPTS {
         match client.get_register(xorname, tag).await {
-            Ok(_) => {
-                println!("Successfully created register '{name}' at {xorname:?}, {tag}!");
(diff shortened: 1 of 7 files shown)


sn_cli/src/subcommands/register.rs
```diff
@@ -43,15 +47,40 @@ pub(crate) async fn register_cmds(cmds: RegisterCmds, client: &Client) -> Result
...
     let _register = ClientRegister::create_public_online(client.clone(), xorname, tag).await?;
-    println!("Successfully created register '{name}' at {xorname:?}, {tag}!");
-    Ok(())
+
+    let mut verification_attempts = 0;
+
+    while verification_attempts < VERIFICATION_ATTEMPTS {
+        match client.get_register(xorname, tag).await {
+            Ok(_) => {
+                println!("Successfully created register '{name}' at {xorname:?}, {tag}!");
```
grumbach: Being a bit pedantic here but `Ok(_)` here could well be another register created concurrently, we might want to check if it matches ours.

grumbach (commented):
LGTM

sn_cli/src/subcommands/register.rs
joshuef: Hmm, 
```
                if our_register.name() == returned_register.name();
```
is enough? comparing contents  might fail if we get _newer_ ?

joshuef: no, that'd be the same whatever for that get.... hmm. if we're doing this we need to tolerate potential _new_ entries

joshuef: Okay, i think I have a good enough way forwards here. Just checking the unferlying and if not matching erroring so the user can take action.

PR #544, opened 2023-07-20
feat: not to carry out existence check explicitly

maqi:
### Summary generated by Reviewpad on 20 Jul 23 14:20 UTC
This pull request includes a feature addition that avoids explicit existence checks for already existing entries in the record store. Instead of sending duplicated data for validation, the patch does nothing for already existing entries and handles the double spend case separately.


maqi:
not preferred

PR #552, opened 2023-07-22
feat: replicate when our close group changes

RolandSherwin:
- Instead of triggering replication on every churn, we check if the churn has affected our close group and only then trigger the replication flow. 
-  The `ReplicationFetcher` is now updated on every successful PUT to our record store. 
- The `Cmd:RequestReplication` has been removed as well

This PR is part of the "Close group based replication PRs". 2/3
Depends on https://github.com/maidsafe/safe_network/pull/551
### Summary generated by Reviewpad on 22 Jul 23 06:44 UTC

- In the `record_store.rs` file:
  - The `get_record_keys_closest_to_target` method has been removed.
  - The `set_distance_range` method has been removed.
  - The `record_addresses_ref` method has been added.
  - The `read_from_disk` method signature has been updated.

- In the `event.rs` file:
  - An `import` statement for `sort_peers_by_address` has been added.
  - Two new variants `CloseGroupUpdated` and `KeysForReplication` have been added to the `NetworkEvent` enum.
  - The `check_for_change_in_our_close_group` function has been added.
  - The `get_all_local_peers` function has been added.
  - Calls to `check_for_change_in_our_close_group` have been added in the `handle_kademlia_event` and `handle_swarm_event` functions.
  - The `log_kbuckets` function now takes a reference to `peer` instead of `&PeerId`.

- In the `cmd.rs` file:
  - The `RequestReplication` variant of the `Cmd` enum has been removed.
  - The corresponding pattern in the `impl Cmd` block has been removed.

- In a separate file:
  - Several messages have been added and modified.
  - Import statements and functions have been added and updated.

- In the `lib.rs` file:
  - Various changes have been made, including import statements, method implementations, and method renames.

- In the `rpc.rs` file:
  - A new function `record_addresses` has been added to the `SafeNode` trait implementation.

- In the `log_markers.rs` file:
  - Import statements have been modified.
  - Enum variants and struct fields have been modified.

- In the `cmd.rs` file:
  - Variants have been moved, added, and removed, and their implementations have been updated.

- In the `event.rs` file:
  - Import statements have been added.
  - Serialization and deserialization capabilities have been added to the `NodeEvent` enum.
  - Methods for converting `NodeEvent` to and from bytes have been added.

- In the `log_markers.rs` file:
  - Import statements and enum variants have been modified.
  - A new method `fetching_keys_for_replication` has been added to the `Marker` struct.

- In the `cmd.rs` file:
  - Variants have been moved, added, and removed, and their implementations have been updated.

Please let me know if you need further explanations or have any questions about these changes.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -29,7 +29,6 @@ use sn_protocol::{
 };
 use sn_registers::SignedRegister;
 use sn_transfers::client_transfers::SpendRequest;
-use std::num::NonZeroUsize;
 use std::time::Duration;
 use tokio::task::spawn;
 use tracing::trace;
@@ -40,6 +39,11 @@ const INACTIVITY_TIMEOUT: std::time::Duration = tokio::time::Duration::from_secs
 
 /// The initial rounds of `get_random` allowing client to fill up the RT.
 const INITIAL_GET_RANDOM_ROUNDS: usize = 5;
+
+// returns the primitive K_VALUE
+const fn k_value() -> usize {
+    K_VALUE.get()
+}
 
 impl Client {
     /// Instantiate a new client.
@@ -196,25 +200,23 @@ impl Client {
             // it may take some time to fill up the RT.
             // To avoid such delay may fail the query with RecordNotFound,
             // wait till certain amount of peers populated into RT
-            if let Some(peers_added) = NonZeroUsize::new(self.peers_added) {
-                if peers_added >= K_VALUE {
-                    if let Some(progress) = &self.progress {
-                        progress.finish_with_message("Connected to the Network");
-                        // Remove the progress bar
-                        self.progress = None;
-                    }
+            if self.peers_added >= k_value() {
+                if let Some(progress) = &self.progress {
+                    progress.finish_with_message("Connected to the Network");
+                    // Remove the progress bar
+                    self.progress = None;
+                }
 
-                    self.events_channel
-                        .broadcast(ClientEvent::ConnectedToNetwork)?;
-                } else {
-                    debug!("{}/{} initial peers found.", self.peers_added, K_VALUE);
+                self.events_channel
+                    .broadcast(ClientEvent::ConnectedToNetwork)?;
+            } else {
+                debug!("{}/{} initial peers found.", self.peers_added, K_VALUE);
 
-                    if let Some(progress) = &self.progress {
-                        progress.set_message(format!(
-                            "{}/{} initial peers found.",
-                            self.peers_added, K_VALUE
-                        ));
-                    }
+                if let Some(progress) = &self.progress {
+                    progress.set_message(format!(
+                        "{}/{} initial peers found.",
+                        self.peers_added, K_VALUE
+                    ));
                 }
             }
             self.peers_added += 1;
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -182,45 +182,34 @@ impl Client {
     }
 
     fn handle_network_event(&mut self, event: NetworkEvent) -> Result<()> {
-        match event {
-            // Clients do not handle requests.
-            NetworkEvent::RequestReceived { .. } => {}
-            // Clients do not handle responses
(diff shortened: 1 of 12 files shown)


sn_networking/src/replication_fetcher.rs
```diff
@@ -5,142 +5,195 @@
...
+        incoming_keys: Vec<NetworkAddress>,
+        locally_stored_keys: &HashSet<RecordKey>,
+    ) -> Vec<(RecordKey, Option<PeerId>)> {
+        self.retain_keys(locally_stored_keys);
+
+        // add non existing keys to the fetcher
+        incoming_keys
+            .into_iter()
+            .filter_map(|incoming| incoming.as_record_key())
+            .filter(|incoming| !locally_stored_keys.contains(incoming))
```
maqi, line 69: these two `filter_map` and `filter` shall be merged into just one `filter_map`

sn_client/src/api.rs
```diff
@@ -182,45 +182,34 @@ impl Client {
...
-                            // Remove the progress bar
-                            self.progress = None;
-                        }
+        if let NetworkEvent::PeerAdded(peer_id) = event {
+            debug!("PeerAdded: {peer_id}");
+            // In case client running in non-local-discovery mode,
+            // it may take some time to fill up the RT.
+            // To avoid such delay may fail the query with RecordNotFound,
+            // wait till certain amount of peers populated into RT
+            if let Some(peers_added) = NonZeroUsize::new(self.peers_added) {
```
grumbach: There's no else to this clause, was wondering if it can happen and if we still want to do `self.peers_added += 1;` at the end of the function if it's `else`?

sn_networking/src/lib.rs
```diff
@@ -387,11 +386,12 @@ impl SwarmDriver {
             pending_record_put: Default::default(),
             replication_fetcher: Default::default(),
             local,
-            // We use 63 here, as in practice the capactiy will be rounded to the nearest 2^(n-1).
+            // We use 63 here, as in practice the capacity will be rounded to the nearest 2^(n-1).
```
grumbach: Noticed there's another typo here: should be ` 2^n-1` instead of `2^(n-1)`
63 made no sense if it was `2^(n-1)`, checked with the link below, it should indeed be `2^n-1`

sn_networking/src/replication_fetcher.rs
```diff
@@ -5,142 +5,195 @@
 // under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
+#![allow(clippy::mutable_key_type)]
```
grumbach, line 8: why do we need this?

sn_networking/src/replication_fetcher.rs
grumbach: That's possibly more optimised indeed but not sure it will increase readability?

grumbach (commented):
Generally looking good!

sn_networking/src/replication_fetcher.rs
grumbach: That's possibly more optimised indeed but not sure it will increase readability?

sn_client/src/api.rs
RolandSherwin: Sorry, I'm not sure if I understand you, `self.peers_added` just keeps track of the num of peers we've added, I think we might want to do that no matter what.

sn_networking/src/replication_fetcher.rs
RolandSherwin: This is raised when the `RecordKey` is put inside `HashSet`. RecordKey is just a wrapper around `Bytes` and bytes has this field `data: AtomicPtr<()>` that can enable interior mutability. Think this is an open issue, https://github.com/rust-lang/rust-clippy/issues/5812

maqi (approved):

PR #556, opened 2023-07-24
Get record concensus

maqi:
### Summary generated by Reviewpad on 24 Jul 23 14:32 UTC
This pull request includes changes in the `cmd.rs`, `event.rs`, and `lib.rs` files. 

In the `cmd.rs` file:
- Line 185-194: If an existing `get_record` task is replaced, a warning message is logged.

In the `event.rs` file:
- Line 366-393: The logic for handling `KademliaEvent::OutboundQueryProgressed` events related to `get_record` requests has been modified. The code now accumulates the successful `get_record` responses from peers and returns the record once a majority is reached. If the record is not found, an error is returned when the limit of requests is reached.
- Line 465-489: The logic for handling `KademliaEvent::OutboundQueryProgressed` events related to failed `get_record` requests has been modified. If the request reaches the end of the query and the record is not found, the code avoids waiting indefinitely for the response by returning an error.

In the `lib.rs` file:
- Line 94: The `PendingGetClosest` type has been defined to store pending `get_closest_peers` requests.
- Line 97: The `PendingGetRecord` type has been defined to store pending `get_record` requests.
- Line 109-110: The `pending_get_closest_peers` and `pending_get_record` fields have been added to the `SwarmDriver` struct to store pending requests.
- Line 383-407: The logic for handling `get_record` requests and responses has been modified. The code now accumulates the successful `get_record` responses from peers and returns the record once a majority is reached. The code also handles split resolvement policy and updates self when needed.

Overall, these changes improve the handling of `get_record` requests and responses in the code.

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -93,7 +93,7 @@ jobs:
           cat dbc_hex | cargo run --bin safe --release -- wallet deposit --stdin
         env:
           SN_LOG: "all"
-        timeout-minutes: 2
+        timeout-minutes: 10
 
       # The resources file we upload may change, and with it mem consumption.
       # Be aware!
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -244,13 +244,25 @@ jobs:
           fi
         if: always()
 
-      - name: Tar log files
+      - name: Tar node log files
         shell: bash
         continue-on-error: true
         run: |
           find $HEAPNODE_DATA_PATH -iname '*.log*' | tar -zcvf heap_node_log_files.tar.gz --files-from -
           find $NODE_DATA_PATH -iname '*.log*' | tar -zcvf nodes_log_files.tar.gz --files-from -
+        if: failure()
+
+      - name: Tar client log files
+        shell: bash
+        continue-on-error: true
+        run: |
           find $CLIENT_LOG_PATH -iname '*.log*' | tar -zcvf client_log_files.tar.gz --files-from -
+        if: failure()
+
+      - name: Tar log files for upload
+        shell: bash
+        continue-on-error: true
+        run: |
           find . -iname '*log_files.tar.gz' | tar -zcvf log_files.tar.gz --files-from -
         if: failure()
 
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -27,7 +27,7 @@ use self::{
     circular_vec::CircularVec,
     cmd::SwarmCmd,
     error::Result,
-    event::NodeBehaviour,
+    event::{GetRecordResultMap, NodeBehaviour},
     record_store::{
         DiskBackedRecordStore, DiskBackedRecordStoreConfig, REPLICATION_INTERVAL_LOWER_BOUND,
         REPLICATION_INTERVAL_UPPER_BOUND,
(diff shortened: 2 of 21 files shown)


.github/workflows/memcheck.yml
```diff
@@ -241,13 +241,25 @@ jobs:
...
         continue-on-error: true
         run: |
           find $HEAPNODE_DATA_PATH -iname '*.log*' | tar -zcvf heap_node_log_files.tar.gz --files-from -
           find $NODE_DATA_PATH -iname '*.log*' | tar -zcvf nodes_log_files.tar.gz --files-from -
+        if: failure()
+
+      - name: Tar client log files
+        shell: bash
+        continue-on-error: true
+        run: |
```
RolandSherwin, line 258: I think the reason for us not getting the client logs is that the `$CLIENT_LOG_PATH` has changed and as well as the option to enable client logs. We need to pass the following arg to enable it `--log-output-dest data-dir`. I will work on this and raise a PR for it.

sn_networking/src/lib.rs
```diff
@@ -621,13 +622,78 @@ impl Network {
...
-        let (sender, receiver) = oneshot::channel();
-        self.send_swarm_cmd(SwarmCmd::GetNetworkRecord { key, sender })?;
+    /// Carry out re-attempts if required
+    /// In case a target_record is provided, only return when fetched target.
+    /// Otherwise count it as a failure when all attempts completed.
+    pub async fn get_record_from_network(
+        &self,
+        key: RecordKey,
+        target_record: Option<Record>,
+        re_attampt: bool,
```
RolandSherwin: `re_attampt` -> `re_attempt`

joshuef:
```suggestion
        re_attempt: bool,
```

sn_node/src/put_validation.rs
```diff
@@ -206,8 +206,11 @@ impl Node {
         };
         let dbc_addr = DbcAddress::from_dbc_id(&dbc_id);
 
-        debug!("validating and storing spends {:?}", dbc_addr.name());
         let key = NetworkAddress::from_dbc_address(dbc_addr).to_record_key();
+        debug!(
+            "validating and storing spends {:?} - {key:?}",
+            dbc_addr.name()
```
RolandSherwin: I think we use `PrettyPrintRecordKey` to log keys everywhere now.
```suggestion
            "validating and storing spends {:?} - {:?}",
            dbc_addr.name(),
            PrettyPrintRecordKey::from(key.clone())
```

sn_node/src/put_validation.rs
```diff
@@ -233,6 +236,18 @@ impl Node {
...
+            warn!("Got a double spend for the SpendDbc PUT with dbc_id {dbc_id:?}",);
+            let mut proof = validated_spends.iter();
+            if let (Some(spend_one), Some(spend_two)) = (proof.next(), proof.next()) {
+                return Err(ProtocolError::DoubleSpendAttempt(
+                    Box::new(spend_one.to_owned()),
+                    Box::new(spend_two.to_owned()),
+                ))?;
+            }
+        }
+
```
RolandSherwin: If we were to move this check to here, we would not be storing the double spend right?
Assuming, local copy had 1 spend and we got a PUT with another spend for the same `DbcId`, the above `signed_spend_validation` would aggregate those two and would produce a `Vec<SignedSpend>` with `len = 2`. We might want to store that to the network?

joshuef (approved):

PR #563, opened 2023-07-25
chore(testnet): always start the faucet

joshuef:
### Summary generated by Reviewpad on 25 Jul 23 19:17 UTC
This pull request updates the code related to the testnet and faucet in the project. It removes the option to launch the DBC faucet server and instead always starts it. It also includes some code refactoring and cleanup. After the changes, the testnet and faucet are launched successfully.

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -83,12 +83,6 @@ jobs:
         shell: bash
         run: echo "The SAFE_PEERS variable has been set to $SAFE_PEERS"
 
-      - name: Start a faucet client to claim genesis
-        run: cargo run --bin faucet --release -- claim-genesis
-        env:
-          SN_LOG: "all"
-        timeout-minutes: 2
-
       - name: Create and fund a wallet to pay for files storage
         run: |
           cargo run --bin faucet --release -- send 500 $(cargo run --bin safe --release -- wallet address | tail -n 1) | tail -n 1 > dbc_hex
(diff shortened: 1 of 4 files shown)


.github/workflows/memcheck.yml
```diff
@@ -83,16 +83,8 @@ jobs:
...
-        run: cargo run --bin faucet --release -- claim-genesis
-        env:
-          SN_LOG: "all"
-        timeout-minutes: 2
-
       - name: Create and fund a wallet to pay for files storage
-        run: |
-          cargo run --bin faucet --release -- send 500 $(cargo run --bin safe --release -- wallet address | tail -n 1) | tail -n 1 > dbc_hex
-          cat dbc_hex | cargo run --bin safe --release -- wallet deposit --stdin
+        run: cargo run --bin safe --release -- wallet get-faucet
```
grumbach: get faucet will only get 10DBC, is that enough?

grumbach (approved):
Wondering if we want a flag to disable the faucet if it's default now?

.github/workflows/memcheck.yml
joshuef: for now yeh, storage is 1 nano fixed price atm

joshuef:
> Wondering if we want a flag to disable the faucet if it's default now?

I thought about that, but you can't do anything without it at the mo, so I left that off. 

I'm not against that if folk find it useful? (maybe follow up)

PR #564, opened 2023-07-25
feat(registers): make Registers to be stored and addressable based on its own metadata

bochaco:
### Summary generated by Reviewpad on 25 Jul 23 19:22 UTC

1. In the `lib.rs` file:
- Added the module `Metadata` in the `register` module.
- Added import statement for `Metadata` from `sn_registers`.
- Changed the argument of the `create_register` function from `xorname` to `metadata`.
- Updated the log message in the `create_register` function to include the `metadata` parameter.

2. In the `registers.rs` file:
- Import the `Metadata` struct from the `sn_client` crate is added.
- The `xor_name` module import and usage is removed.
- The code now creates a `Metadata` instance using `reg_nickname` and retrieves the `XorName` from it.
- The print statement for log messages is updated to include the register nickname.
- The code to create a register is updated to use the `create_register` method with the `metadata` parameter instead of the `xorname` parameter.

3. In the `sn_cli/src/subcommands/register.rs` file:
- Import `Metadata` from `sn_client::{Client, ClientRegister, Error as ClientError, Metadata}` has been added.
- The `create_register` function now creates a register using `Metadata` instead of `XorName`.
- The `create_register` function now prints the register name after successfully creating it.
- The `edit_register` function now retrieves the register using `Metadata` instead of `XorName`.
- The `edit_register` function now prints the register name when trying to retrieve it.

4. In the `reg_crdt.rs` file:
- Added import statement for the `Metadata` type.
- Modified the `RegisterCrdt` struct to include a `metadata` field of type `Metadata`.
- Created a new method `new` in the `RegisterCrdt` struct that takes `metadata` and `tag` as inputs and sets the `address` field using the `metadata` and `tag`.
- Modified the `merge` method to check if the addresses match before merging the data.
- Added a new method `metadata` in the `RegisterCrdt` struct to retrieve the metadata of the register.
- Modified the creation of `RegisterCrdt` instances in the test module to use randomly generated `Metadata` instead of `RegisterAddress`.
- Added a `tag` variable for the `new` method in the `RegisterCrdt` struct.

5. In the `error.rs` file:
- Added a new variant `MetadataTooBig` to the `Error` enum.
- The `MetadataTooBig` variant contains two fields: `size` of type usize and `max` of type usize.
- Added a new error message for the `MetadataTooBig` variant.
- The `EntryTooBig` variant has not been changed.

6. In the `data_with_churn.rs` file:
- Imported `Metadata` module from `sn_client`.
- Removed unused import of `XorName` from `xor_name` module.
- Updated the implementation of creating registers by using `Metadata` instead of `XorName` and `RegisterAddress`.
- Replaced the address of the newly created register with the address obtained from the `Metadata` object.
- Updated the error handling when creating registers to include the tag value in the error message.

7. In the `metadata.rs` file:
- Added the `use super::{register::MAX_REG_ENTRY_SIZE, Error};` statement.
- Modified the `std` import to include `std::result::Result`.
- Added the `xor_name::XorName` import.
- Added a new struct `Metadata` with its implementation.
- Added a new function `new` in the `Metadata` implementation.
- Added a new function `xor_name` in the `Metadata` implementation.
- Changed the type of `Entry` from `Vec<u8>` to `pub type Entry = Vec<u8>`.

In summary, this pull request includes changes to various files to introduce the `Metadata` struct, update function arguments and implementations related to registers, and add new error variants.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3928,6 +3928,7 @@ dependencies = [
  "rand 0.8.5",
  "self_encryption",
  "serde",
+ "serde_json",
  "thiserror",
  "tiny-keccak",
  "xor_name",
--- a/sn_cli/src/subcommands/register.rs
+++ b/sn_cli/src/subcommands/register.rs
@@ -8,7 +8,7 @@
 
 use clap::Subcommand;
 use color_eyre::Result;
-use sn_client::{Client, ClientRegister, Error as ClientError};
+use sn_client::{Client, ClientRegister, Error as ClientError, Metadata};
 use xor_name::XorName;
 
 #[derive(Subcommand, Debug)]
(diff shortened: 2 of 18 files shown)


bochaco:
Closed till it's clearer for everyone.

PR #569, opened 2023-07-26
feat(networking): record store prunes more frequently.

### LINKED ISSUE #566: Prune stored data every X record writes
After a successful write, we should check if we should prune far away data. (we should keep the prune on reaching capacity, this is in addition).

We could track this, or just use probability. Less state the better perhaps?

I'd start of at pruning once every ~100 successful writes? We can tweak that number as we go.

joshuef:
We now prune every 1/100 PUTs to keep space available more clear, and remove unneeded data faster

fixes #566

### Summary generated by Reviewpad on 26 Jul 23 07:35 UTC
This pull request updates the `record_store.rs` file in the `sn_networking` module. It introduces a new constant `PRUNE_ONCE_EVERY_PUTS` which determines how frequently the records in the store are pruned. Previously, pruning was done when the number of records exceeded the maximum limit, but now it also prunes once every X PUTs. The goal is to keep space available and remove unneeded data faster. The patch also includes some code changes to implement the new pruning logic.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -34,6 +34,12 @@ pub const REPLICATION_INTERVAL_LOWER_BOUND: Duration = Duration::from_secs(180);
 /// Max number of records a node can store
 const MAX_RECORDS_COUNT: usize = 2048;
 
+/// How frequenly we prune the records in the store
+/// eg, if 100 here, every 100 valid PUTs we'll prune the store
+/// This should give us some margin about storing data for nodes that are no longer our responsibility
+/// And may eventually feed into price discovery / storage costs and give us a buffer there too
+const PRUNE_ONCE_EVERY_PUTS: usize = 100;
+
 /// A `RecordStore` that stores records on disk.
 pub struct DiskBackedRecordStore {
     /// The identity of the peer owning the store.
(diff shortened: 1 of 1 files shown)


sn_networking/src/record_store.rs
```diff
@@ -198,7 +195,36 @@ impl DiskBackedRecordStore {
...
+    /// for new records.
+    ///
+    /// This can happen if we're full, or once in every X PUTs
+    ///
+    /// An error is returned if we prune and we're _still_ at max_record count
+    fn prune_storage_if_needed(&mut self) -> Result<()> {
+        let num_records = self.records.len();
+        let mut should_prune = num_records >= self.config.max_records;
+        if !should_prune {
+            // Use os rng to call ensure we prune at least PRUNE_ONCE_EVERY_PUTS
```
b-zee: By generating a number in the `0..100` range, it's not 'ensured' that `should_prune` is `true` once in max. 100 times. Statistically it could very well happen `0` isn't generated for a while.

joshuef: Indeed, it could also happen one after another. It will average out though.

We could just modulo here though perhaps and get a decent check interval actually

joshuef: Aye that's cleaner

RolandSherwin (approved):

PR #576, opened 2023-07-26
Add in GetStoreCost messaging

### LINKED ISSUE #567: Increase storage price as space declines
We can measure space in terms of stored records (that's our current method anyway).

We want the price of storage to increase exponentially as the storage space decreases. (Storage space will be kept relevant to our position in the network via https://github.com/maidsafe/safe_network/issues/566 , and so availability there should increase as more nodes join the network and price decreases).

We can probably divide `CAPACITY` by `Total SAFECOINS * NANOs` to gauge the price at any given point. 

This will differ between nodes, but (in the end) we'd accept a majority / average price.

Perhaps as a first stop we accept within a certain bounds to get things moving?

joshuef:
Atop the store cost work.

This adds basic protocol support for retrieving StoreCost from nodes.

I'm imagining we'd first do `GetClosest<PaymentRoot>`, and then query each for their storecost (which will need to be signed).

I'm not sure if we can roll that easily into one command over kad, or it'll need to be such a combo of calls? (or indeed if this is a bit of a naiive approach altogether?)

TODO: 

- [x] Sign the token payload as well (I'm not sure how/where to sign at the moment 🤔 )

reviewpad:summary

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -630,7 +630,7 @@ impl Network {
             .map_err(|_e| Error::InternalMsgChannelDropped)?
     }
 
-    /// Get the the cost of storing the next record from the network
+    /// Get the cost of storing the next record from the network
     pub async fn get_local_storecost(&self) -> Result<Token> {
         let (sender, receiver) = oneshot::channel();
         self.send_swarm_cmd(SwarmCmd::GetLocalStoreCost { sender })?;
@@ -778,7 +778,7 @@ impl Network {
         self.send_swarm_cmd(SwarmCmd::SetRecordDistanceRange { distance })
     }
 
-    /// Send `Request` to the the given `PeerId` and await for the response. If `self` is the recipient,
+    /// Send `Request` to the given `PeerId` and await for the response. If `self` is the recipient,
     /// then the `Request` is forwarded to itself and handled, and a corresponding `Response` is created
     /// and returned to itself. Hence the flow remains the same and there is no branching at the upper
     /// layers.
@@ -792,7 +792,7 @@ impl Network {
         receiver.await?
     }
 
-    /// Send `Request` to the the given `PeerId` and do _not_ await a response here.
+    /// Send `Request` to the given `PeerId` and do _not_ await a response here.
     /// Instead the Response will be handled by the common `response_handler`
     pub fn send_req_ignore_reply(&self, req: Request, peer: PeerId) -> Result<()> {
         let swarm_cmd = SwarmCmd::SendRequest {
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -49,6 +49,7 @@ use libp2p::{
     Multiaddr, PeerId, Transport,
 };
 use rand::Rng;
+use sn_dbc::Token;
 use sn_protocol::{
     messages::{Request, Response},
     NetworkAddress,
@@ -410,6 +411,7 @@ impl SwarmDriver {
                 swarm_cmd_sender,
                 peer_id,
                 root_dir_path,
+                keypair,
             },
             network_event_receiver,
             swarm_driver,
@@ -485,9 +487,15 @@ pub struct Network {
     pub swarm_cmd_sender: mpsc::Sender<SwarmCmd>,
     pub peer_id: PeerId,
     pub root_dir_path: PathBuf,
+    keypair: Keypair,
 }
 
 impl Network {
+    /// Signs the given data with the node's keypair.
+    pub fn sign(&self, msg: &[u8]) -> Result<Vec<u8>> {
+        self.keypair.sign(msg).map_err(Error::from)
+    }
+
     ///  Listen for incoming connections on the given address.
     pub async fn start_listening(&self, addr: Multiaddr) -> Result<()> {
         let (sender, receiver) = oneshot::channel();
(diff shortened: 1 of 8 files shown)


sn_networking/src/lib.rs
```diff
@@ -615,6 +616,16 @@ impl Network {
             .map_err(|_e| Error::InternalMsgChannelDropped)?
     }
 
+    /// Get the the cost of storing the next record from the network
```
maqi: duplicated `the`

sn_networking/src/lib.rs
```diff
@@ -83,9 +84,9 @@ const IDENTIFY_CLIENT_VERSION_STR: &str = concat!("safe/client/", env!("CARGO_PK
 const IDENTIFY_PROTOCOL_STR: &str = concat!("safe/", env!("CARGO_PKG_VERSION"));
 
 /// Duration to wait for verification
-const VERIFICATION_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+const REVERIFICATION_WAIT_TIME_S: std::time::Duration = std::time::Duration::from_secs(3);
 /// Number of attempts to verify a record
-const VERIFICATION_ATTEMPTS: usize = 3;
+const VERIFICATION_ATTEMPTS: usize = 30;
```
maqi: 30? how comes requiring that high number?
given each iteration is 3s, it becomes 90s in total to complete?

sn_networking/src/record_store.rs
```diff
@@ -210,6 +211,33 @@ impl DiskBackedRecordStore {
...
         }
     }
+
+    #[allow(dead_code)]
+    /// Calculate the cost to store data for our current store state
+    pub fn store_cost(&self) -> Token {
+        let records_len = self.records().count();
+
+        let mut cost = Token::from_nano(1);
+        const NANOS_PER_SNT: u64 = 1_000_000_000;
```
maqi: move to the beginning of the file?

sn_networking/src/record_store.rs
```diff
@@ -210,6 +211,33 @@ impl DiskBackedRecordStore {
...
+        // So we get the total amount of nanos in the network,
+        // we divide that number by our MAX_RECORD_COUNT to know
+        // what to charge for each additional record
+        let step_per_record = ((2 ^ 32) * NANOS_PER_SNT) / MAX_RECORDS_COUNT as u64;
+        let step_amount = Token::from_nano(step_per_record);
+
+        // we want to spread the cost of storing data across the MAX_CAPACITY
+        // doubling the cost for each record store
+        for _ in 0..records_len {
+            if let Some(new_cost) = cost.checked_add(step_amount) {
```
maqi: I don't get this.

first, what's the diff to just using : `cost.saturating_add(records_len * step_amount)`

second, the `step_amount` is just a `cost_per_record`. 
However, as I understanding, the `stepping charge` shall be something like:
For 1-10 records, charge 1 per record,
For 11-20 records, charge 2 per record,
For 21-40 records, charge 3 per record,
......
But here, seems the cost will be change `for each record`, which seems like more aggressive?
Also, this will make all nodes (storing the same record) giving more various cost charge of that record, as they may holding different set of records due to different space address or lifespan in the network.
Which could be a headache to sync ?

sn_protocol/src/error.rs
```diff
@@ -55,6 +55,9 @@ pub enum Error {
     SpendIsEmpty,
 
     // ---------- payment errors
+    /// Failed to get the storecost from kademlia stoe
```
maqi: store

sn_networking/src/record_store.rs
```diff
@@ -210,6 +211,33 @@ impl DiskBackedRecordStore {
...
+        let records_len = self.records().count();
+
+        let mut cost = Token::from_nano(1);
+        const NANOS_PER_SNT: u64 = 1_000_000_000;
+
+        // if we imagine that at FULL capacity we want _all the money_ (ie, we never want to reach full capacity)
+        // So we get the total amount of nanos in the network,
+        // we divide that number by our MAX_RECORD_COUNT to know
+        // what to charge for each additional record
+        let step_per_record = ((2 ^ 32) * NANOS_PER_SNT) / MAX_RECORDS_COUNT as u64;
```
bochaco: Can we just use the `sn_transfers::TOTAL_SUPPLY` const, i.e. `TOTAL_SUPPLY / MAX_RECORDS_COUNT` ? I see it's not public, perhaps it shall be public in sn_protocol or some other public crate/API.

sn_networking/src/lib.rs
joshuef: This should probs not be in this PR. Buttt, i saw `Ci failure without, and anselme was having a real bad time trying to verify uploads, so it needs to be higher. 

Could be 30s, or so. I'll open that up as a separate pr.

sn_networking/src/record_store.rs
joshuef: Ah, yes! I was looking for something like that. I only saw in sn_dbc and wanted to get something up for discussion w/o more repo faff.

joshuef: I'll update main for this.

sn_networking/src/record_store.rs
joshuef: >  Also, this will make all nodes (storing the same record) giving more various cost charge of that record, as they may holding different set of records due to different space address or lifespan in the network.
We can expect this anyway. Records stored vary in the hundreds sometimes. (at least on once machine). So we'll need to deal with that. (Which the plan is to just average the majority or responses. and pay _that_).

>But here, seems the cost will be change for each record,


Ahhh, yeh here I'm missing any module calc like the pruning!

RolandSherwin (approved):

PR #577, opened 2023-07-26
feat(node): add marker for a network connection timeout

### LINKED ISSUE #518: Add Messaging around failure to connect
Nodesshould, after a certain amount of time (1min?) of being unable to connect show a message stating as much.

(similar to how clients note they don't have any network contacts yet).

Let's add additional `Marker` for Connecting to the network and if the node is bailing too

aed900:
Added a NetworkConnectionTimedOut marker/error message if connection is timed out.

Closes #518

### DIFF
--- a/sn_node/src/api.rs
+++ b/sn_node/src/api.rs
@@ -146,12 +146,15 @@ impl Node {
                         }
                     }
                     _ = tokio::time::sleep(inactivity_timeout) => {
-                        let random_target = NetworkAddress::from_peer(PeerId::random());
-
                         Marker::NoNetworkActivity( inactivity_timeout ).log();
+                        let random_target = NetworkAddress::from_peer(PeerId::random());
                         debug!("No network activity in the past {inactivity_timeout:?}, performing a random get_closest query to target: {random_target:?}");
-                        if let Ok(closest) = network_clone.node_get_closest_peers(&random_target).await {
-                            debug!("Network inactivity: get_closest returned {closest:?}");
+                        match network_clone.node_get_closest_peers(&random_target).await {
+                            Ok(closest) => debug!("Network inactivity: get_closest returned {closest:?}"),
+                            Err(e) => {
+                                warn!("get_closest query failed after network inactivity timeout - check your connection: {}", e);
+                                Marker::OperationFailedAfterNetworkInactivityTimeout.log();
+                            }
                         }
                     }
                 }
--- a/sn_node/src/log_markers.rs
+++ b/sn_node/src/log_markers.rs
@@ -17,6 +17,9 @@ use strum::Display;
 /// Changing these log markers is a breaking change.
 #[derive(Debug, Clone, Display)]
 pub enum Marker<'a> {
+    /// An operation failed after timing out due to a period of no network activity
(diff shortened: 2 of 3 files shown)


sn_node/src/api.rs
```diff
@@ -134,6 +135,10 @@ impl Node {
                     net_event = network_event_receiver.recv() => {
                         match net_event {
                             Some(event) => {
+                                if is_conn_active == false {
+                                    is_conn_active = true;
+                                    Marker::AttemptingNetworkConnection.log();
```
joshuef: I'm not sure we need this (nor how accurate it is).

sn_node/src/api.rs
```diff
@@ -146,12 +146,15 @@ impl Node {
...
                         Marker::NoNetworkActivity( inactivity_timeout ).log();
+                        let random_target = NetworkAddress::from_peer(PeerId::random());
                         debug!("No network activity in the past {inactivity_timeout:?}, performing a random get_closest query to target: {random_target:?}");
-                        if let Ok(closest) = network_clone.node_get_closest_peers(&random_target).await {
-                            debug!("Network inactivity: get_closest returned {closest:?}");
+                        match network_clone.node_get_closest_peers(&random_target).await {
+                            Ok(closest) => debug!("Network inactivity: get_closest returned {closest:?}"),
+                            Err(e) => {
+                                Marker::NetworkConnectionTimedOut.log();
+                                println!("Network connection has timed out: {}", e);
```
joshuef: Do we need both the log marker and this println? the info should log to stdout, I believe

sn_node/src/log_markers.rs
```diff
@@ -17,6 +17,9 @@ use strum::Display;
 /// Changing these log markers is a breaking change.
 #[derive(Debug, Clone, Display)]
 pub enum Marker<'a> {
+    /// Network connection has timed out after a period of no network activity
+    NetworkConnectionTimedOut,
```
joshuef: I'm not sure this is clear enough, we may want to add more here like "CheckYourConnection"?

sn_node/src/api.rs
```diff
@@ -146,12 +146,14 @@ impl Node {
...
-
                         Marker::NoNetworkActivity( inactivity_timeout ).log();
+                        let random_target = NetworkAddress::from_peer(PeerId::random());
                         debug!("No network activity in the past {inactivity_timeout:?}, performing a random get_closest query to target: {random_target:?}");
-                        if let Ok(closest) = network_clone.node_get_closest_peers(&random_target).await {
-                            debug!("Network inactivity: get_closest returned {closest:?}");
+                        match network_clone.node_get_closest_peers(&random_target).await {
+                            Ok(closest) => debug!("Network inactivity: get_closest returned {closest:?}"),
+                            Err(_) => {
+                                Marker::NetworkConnectionTimedOutCheckYourConnection.log();
```
joshuef: Reading this again, no ConnectionTImedOut, right?  (or perhaps it did, we dont match the error).

We've had a timeout, tried an op, and that's failed. So I think we just need the log message to reflect that.

joshuef (approved):

PR #578, opened 2023-07-27
Tweak store cost + one in one out for data

joshuef:
### Summary generated by Reviewpad on 27 Jul 23 05:18 UTC

- Added the "sn_transfers" dependency to the Cargo.lock and Cargo.toml files in the sn_networking module.
- Modified the VERIFICATION_ATTEMPTS constant from 30 to 10 in the lib.rs file of the sn_networking module.
- Made changes to the record_store.rs file in the sn_networking module:
  - Added the "sn_transfers" import.
  - Used the TOTAL_SUPPLY constant from the sn_transfers module.
  - Removed the PRUNE_ONCE_EVERY_PUTS constant.
  - Updated the prune_storage_if_needed_for_record method to prune records based on the distance to the local key, instead of using PRUNE_ONCE_EVERY_PUTS.
- Updated the TOTAL_SUPPLY constant to be pub in the dbc_genesis.rs file of the sn_transfers module.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -187,36 +187,51 @@ impl DiskBackedRecordStore {
     /// Prune the records in the store to ensure that we free up space
     /// for the incoming record.
     ///
-    /// An error is returned if we prune and we're _still_ at max_record count
+    /// An error is returned if we are full and the new record is not closer than
+    /// the furthest record
     fn prune_storage_if_needed_for_record(&mut self, r: &Key) -> Result<()> {
         let num_records = self.records.len();
 
+        // we're not full, so we don't need to prune
+        if num_records < self.config.max_records {
+            return Ok(());
+        }
+
         // sort records by distance to our local key
         let mut records = self.records.iter().cloned().collect::<Vec<_>>();
-        records.sort_by_key(|k| {
+        records.sort_unstable_by_key(|k| {
             let kbucket_key = KBucketKey::from(k.to_vec());
             self.local_key.distance(&kbucket_key)
         });
 
         // now check if the incoming record is closer than our furthest
         // if it is, we can prune
-        let furthest_record = records.last();
-        if let Some(furthest_record) = furthest_record {
+        if let Some(furthest_record) = records.last() {
             let furthest_record_key = KBucketKey::from(furthest_record.to_vec());
             let incoming_record_key = KBucketKey::from(r.to_vec());
+
             if incoming_record_key.distance(&self.local_key)
                 < furthest_record_key.distance(&self.local_key)
             {
+                trace!(
+                    "{:?} will be pruned to make space for new record: {:?}",
+                    PrettyPrintRecordKey::from(furthest_record.clone()),
+                    PrettyPrintRecordKey::from(r.clone())
+                );
                 // we should prune and make space
                 self.remove(furthest_record);
 
-                trace!("Data was pruned to make space for new record");
-            } else {
-                // we should not prune, but we should check if we're at max capacity
-                if num_records >= self.config.max_records {
-                    warn!("Record not stored. Maximum number of records reached. Current num_records: {num_records}");
-                    return Err(Error::MaxRecords);
+                // check if the furthest record is was within our distance range and if so
+                // warn
+                if let Some(distance_range) = self.distance_range {
+                    if furthest_record_key.distance(&self.local_key) < distance_range {
+                        warn!("Pruned record would also be within our distance range.");
+                    }
                 }
+            } else {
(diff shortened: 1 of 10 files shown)


sn_networking/src/record_store.rs
```diff
@@ -36,11 +37,8 @@ pub const REPLICATION_INTERVAL_LOWER_BOUND: Duration = Duration::from_secs(180);
 /// Max number of records a node can store
 const MAX_RECORDS_COUNT: usize = 2048;
 
-/// How frequenly we prune the records in the store
-/// eg, if 100 here, every 100 valid PUTs we'll prune the store
-/// This should give us some margin about storing data for nodes that are no longer our responsibility
-/// And may eventually feed into price discovery / storage costs and give us a buffer there too
-const PRUNE_ONCE_EVERY_PUTS: usize = 100;
+/// ~Number of puts per price step
```
b-zee, line 40: What does the tilde signify here?

sn_networking/src/record_store.rs
```diff
@@ -50,11 +48,11 @@ pub struct DiskBackedRecordStore {
...
     /// A set of keys, each corresponding to a data `Record` stored on disk.
     records: HashSet<Key>,
-    /// Distance range specify the acceptable range of record entry.
-    /// `None` means accept all.
-    distance_range: Option<Distance>,
     /// Currently only used to notify the record received via network put to be validated.
     event_sender: Option<mpsc::Sender<NetworkEvent>>,
+    /// Distance range specify the acceptable range of record entry.
+    /// None means accept all records.
+    distance_range: Option<Distance>,
```
b-zee, line 55: Any specific reason for this rearrangement?

sn_networking/src/record_store.rs
```diff
@@ -202,71 +185,86 @@ impl DiskBackedRecordStore {
...
-        if should_prune {
-            self.prune_storage();
-            let new_num_records: usize = self.records.len();
-            trace!("A pruning reduced number of record in hold from {num_records} to {new_num_records}");
-            if new_num_records >= self.config.max_records {
-                warn!("Record not stored. Maximum number of records reached. Current num_records: {num_records}");
-                return Err(Error::MaxRecords);
+        // sort records by distance to our local key
+        let mut records = self.records.iter().cloned().collect::<Vec<_>>();
+        records.sort_by_key(|k| {
```
b-zee: According to the Rust docs on `sort_by_key`, we can use the 'unstable' sort variant:

> When applicable, unstable sorting is preferred because it is generally faster than stable sorting and it doesn’t allocate auxiliary memory. See [sort_unstable_by_key](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable_by_key).

sn_networking/src/record_store.rs
```diff
@@ -202,71 +185,86 @@ impl DiskBackedRecordStore {
...
+            let kbucket_key = KBucketKey::from(k.to_vec());
+            self.local_key.distance(&kbucket_key)
+        });
+
+        // now check if the incoming record is closer than our furthest
+        // if it is, we can prune
+        let furthest_record = records.last();
+        if let Some(furthest_record) = furthest_record {
```
b-zee:
```suggestion
        if let Some(furthest_record) = records.last() {
```

sn_networking/src/record_store.rs
```diff
@@ -202,71 +185,86 @@ impl DiskBackedRecordStore {
     }
 
     /// Prune the records in the store to ensure that we free up space
-    /// for new records.
-    ///
-    /// This can happen if we're full, or once in every X PUTs
+    /// for the incoming record.
     ///
     /// An error is returned if we prune and we're _still_ at max_record count
```
b-zee: The 'if we prune' part is incorrect. The error is returned without pruning, if I am reading correctly.

sn_networking/src/record_store.rs
```diff
@@ -202,71 +185,86 @@ impl DiskBackedRecordStore {
...
+        // Calculate the starting cost
+        let mut cost = TOTAL_SUPPLY / factor;
+
+        trace!("Starting cost is {:?}", cost);
+
+        trace!("Record count is {:?}", self.records.len());
+        let relevant_records_len = if let Some(distance_range) = self.distance_range {
+            self.records
+                .iter()
+                .cloned()
```
b-zee:
```suggestion
```

sn_networking/src/record_store.rs
joshuef: Approximate (as there's not necessarily an exact number of puts in the final step eg)

b-zee (approved):

PR #579, opened 2023-07-27
fix(node): set distance range to prune records

RolandSherwin:
Sets the distance range as the distance to our 8th closest node.
### Summary generated by Reviewpad on 27 Jul 23 07:05 UTC

First patch: 
- In the `sn_networking/src/event.rs` file, the `get_all_local_peers` function has been modified to remove self from the peers obtained from the local routing table. 
- In the `sn_networking/src/event.rs` file, the `check_for_change_in_our_close_group` function has been modified to remove self from the closest peers if they are not empty.

Second patch: 
- In the `sn_networking/src/lib.rs` file, the `SwarmCmd` enum has been modified to include a new command `SetRecordDistanceRange` that sets the acceptable range of `Record` entry. 
- In the `sn_networking/src/lib.rs` file, the `set_record_distance_range` function has been added to the `Network` struct to set the acceptable range of record entry. 
- In the `sn_networking/src/cmd.rs` file, the `SetRecordDistanceRange` command has been implemented to set the distance range in the `DiskBackedRecordStore` struct. 
- In the `sn_networking/src/record_store.rs` file, the `set_distance_range` function has been added to the `DiskBackedRecordStore` struct to set the distance range used for pruning. 
- In the `sn_node/src/replication.rs` file, the `replicate` function has been modified to calculate the distance range based on the last element of the close group and set it using the `set_record_distance_range` function.

Overall, these patches fix issues related to the inclusion of self in the close group and set the distance range for record entry pruning.

### DIFF
--- a/sn_node/src/replication.rs
+++ b/sn_node/src/replication.rs
@@ -54,6 +54,16 @@ impl Node {
             }
             self.send_replicate_cmd_without_wait(&our_address, &peer_id, remaining_keys.to_vec())?;
         }
+
+        // set the distance range used for pruning
+        let distance_range = match our_close_group.get(CLOSE_GROUP_SIZE) {
+            Some(peer) => NetworkAddress::from_peer(*peer).distance(&our_address),
+            None => {
+                debug!("Could not obtain distance_range");
+                return Ok(());
+            }
+        };
+        self.network.set_record_distance_range(distance_range)?;
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


sn_node/src/replication.rs
```diff
@@ -54,6 +54,16 @@ impl Node {
...
+
+        // set the distance range used for pruning
+        let distance_range = match our_close_group.get(CLOSE_GROUP_SIZE) {
+            Some(peer) => NetworkAddress::from_peer(*peer).distance(&our_address),
+            None => {
+                debug!("Could not obtain distance_range");
+                return Ok(());
+            }
+        };
+        self.network.set_record_distance_range(distance_range)?;
```
joshuef, line 66: Do we want to set this only if the range has changed? Or if our close group changed?

RolandSherwin: A change in close group is gonna lead to a change in the distance range. So we can overwrite the distance_range value?

joshuef (approved):

PR #583, opened 2023-07-27
Put record error improvements

grumbach:
## Description

- make errors actionable by printing `XorName` formatted `kad::RecordKeys` so client can copy paste them and get the failing record manually
- records put retries up to 3 times when `verify` is activated and fails

---

### Summary generated by Reviewpad on 27 Jul 23 09:45 UTC
This pull request introduces two new features. 

In the first patch, the `sn_networking` module now provides actionable error messages for record key errors. Previously, the error messages only displayed the `kad::RecordKey` type, which was not very useful for clients. Now, the `Error` enum includes a new variant `FailedToVerifyRecordWasStored` that wraps a `PrettyPrintRecordKey` struct. This struct implements `std::fmt::Display` and `std::fmt::Debug` to generate a hex-string representation of the record key, making it more actionable for clients. The patch also includes a new test that ensures the hex-string representation of `kad::RecordKey` matches the representation of the corresponding `xor_name::XorName`.

In the second patch, the `Network` struct now supports retries when putting records to the network. There is a new constant `PUT_RECORD_RETRIES` that determines the number of retries. The `put_record` method has been split into two private methods: `put_record_with_retries` and `put_record_once`. The former performs the actual putting of the record and verifies its storage, while the latter is used for a single attempt without verification. If verification fails, the method will retry according to the specified number of retries. If verification still fails after all retries, an error is returned.

### DIFF
--- a/sn_networking/src/error.rs
+++ b/sn_networking/src/error.rs
@@ -8,6 +8,7 @@
 
 use super::{cmd::SwarmCmd, NetworkEvent};
 
+use bytes::Bytes;
 use libp2p::{
     kad,
     request_response::{OutboundFailure, RequestId},
@@ -37,13 +38,11 @@ pub enum Error {
     #[error("Outgoing response has been dropped due to a conn being closed or timeout: {0}")]
     OutgoingResponseDropped(Response),
 
-    #[error("Could not retrieve the record after storing it: {0:?}")]
-    FailedToVerifyRecordWasStored(kad::RecordKey),
+    #[error("Could not retrieve the record after storing it: {0:}")]
+    FailedToVerifyRecordWasStored(PrettyPrintRecordKey),
 
-    #[error(
-        "Record retrieved from the network does not match the one we attempted to store {0:?}"
-    )]
-    ReturnedRecordDoesNotMatch(kad::RecordKey),
+    #[error("Record retrieved from the network does not match the one we attempted to store {0:}")]
+    ReturnedRecordDoesNotMatch(PrettyPrintRecordKey),
 
     #[error("Could not create storage dir: {path:?}, error: {source}")]
     FailedToCreateRecordStoreDir {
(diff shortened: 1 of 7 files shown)


sn_networking/src/error.rs
```diff
@@ -93,3 +92,53 @@ pub enum Error {
     #[error("No SwarmCmd channel capacity")]
     NoSwarmCmdChannelCapacity,
 }
+
+/// Pretty print a `kad::RecordKey` as a hex string.
+/// So clients can use the hex string for xorname and record keys interchangeably.
+/// This makes errors actionable for clients.
+/// The only cost is converting ked::RecordKey into it before sending it in errors: `record_key.into()`
+#[derive(Clone)]
+pub struct PrettyPrintRecordKey(kad::RecordKey);
```
maqi, line 101: this better to be used across all the logs when logging `record_key`, otherwise it won't be tracable along the flow.

ideally, I'd prefer to `overload` the `Debug` trait of the `RecordKey`, however, not sure how to do it in our own crate without raising PR into libp2p crate.

sn_networking/src/lib.rs
```diff
@@ -631,9 +636,33 @@ impl Network {
...
+        }
+    }
+
+    /// Put `Record` to network
+    /// Verify the record is stored after putting it to network
+    /// Retry up to `PUT_RECORD_RETRIES` times if we can't verify the record is stored
+    async fn put_record_with_retries(&self, record: Record) -> Result<()> {
+        let mut retries = 0;
+        while retries < PUT_RECORD_RETRIES {
+            let res = self.put_record_once(record.clone(), true).await;
```
maqi, line 654: the flag true here will trigger a multiple attempts of `get_record` internally?
which means the total completion time could be: ` PUT_RECORD_RETRIES  * GET_ATTEMPTS * interval` ?

sn_networking/src/error.rs
grumbach: Agreed, but as you say, we can't overload it since the type is not defined in our crate, will adapt the logs where I find them!

sn_networking/src/lib.rs
```diff
@@ -631,9 +636,33 @@ impl Network {
...
+            if !matches!(res, Err(Error::FailedToVerifyRecordWasStored(_))) {
+                return res;
+            }
+            retries += 1;
```
maqi, line 658: shall we wait a little bit before a re-try ?

grumbach: The verification already waited for a long time here, so there's actually all the verification waits times the verification retries here already.

sn_networking/src/lib.rs
grumbach: Worse case scenario yes. But only if the error matches `FailedToVerifyRecordWasStored` all other errors or success will be immediately returned.

joshuef (approved):

PR #584, opened 2023-07-28
fix(network): fix infinite loop during verification

RolandSherwin:
### Summary generated by Reviewpad on 28 Jul 23 09:05 UTC
This pull request fixes an issue with infinite loops during verification in the network module. The patch modifies the `impl Network` to handle verification attempts and check if the returned record matches the PUT record. If a different record is found, it informs the user and continues the verification. If the verification fails after multiple attempts, an error is returned.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -677,6 +677,10 @@ impl Network {
         }
 
         // Verify the record is stored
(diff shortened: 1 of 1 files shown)


RolandSherwin:
Will be replaced by #556

PR #586, opened 2023-07-28
feat(node): add relay/hole-punching functionality using multiple transport options (tcp/quic)

aed900:
Add relay/hole-punching functionality using multiple transport options (tcp/quic).

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -361,21 +361,16 @@ impl SwarmDriver {
     pub(crate) fn dial(&mut self, mut addr: Multiaddr) -> Result<(), DialError> {
         debug!(%addr, "Dialing manually");
 
-        if addr.to_string().contains("/p2p-circuit/") {
-            info!("Dialing manually relay address: {}", addr.clone());
-            self.swarm.dial(addr.clone())
-        } else {
-            let peer_id = multiaddr_pop_p2p(&mut addr);
-            let opts = match peer_id {
-                Some(peer_id) => DialOpts::peer_id(peer_id)
-                    // If we have a peer ID, we can prevent simultaneous dials.
-                    .condition(PeerCondition::NotDialing)
-                    .addresses(vec![addr])
-                    .build(),
-                None => DialOpts::unknown_peer_id().address(addr).build(),
-            };
-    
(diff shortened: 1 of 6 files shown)


sn_networking/src/cmd.rs
```diff
@@ -361,16 +361,21 @@ impl SwarmDriver {
...
-                .condition(PeerCondition::NotDialing)
-                .addresses(vec![addr])
-                .build(),
-            None => DialOpts::unknown_peer_id().address(addr).build(),
-        };
-
-        self.swarm.dial(opts)
+        if addr.to_string().contains("p2p-circuit") {
+            info!("Dialing manually relay address: {}", addr.clone());
+            self.swarm.dial(addr.clone())
```
joshuef: do we want to avoid all the conditions below? Why?

sn_networking/src/event.rs
```diff
@@ -234,6 +302,16 @@ impl SwarmDriver {
                 }
             },
             SwarmEvent::NewListenAddr { address, .. } => {
+                if address.iter().any(|p| p == libp2p::multiaddr::Protocol::P2pCircuit) {
+                    info!("Relayed listen address added: {}", address.clone());
```
joshuef, line 331: lets make this a Marker

joshuef:
Why do we have relay and hole punching together?

Lets get these split out into separate PRs so we can verify them independently. 

Same for TCP/quic. I want to know we're using a given protocol for messaging or not. Lets not mix all these changes together.

sn_networking/src/event.rs
aed900: Which is more helpful here?
RelayedListenAddressAdded
RelayReservationAccepted

And should I add another for RelayedConnectionEstablished in the ConnectionEstablished event?

sn_networking/src/cmd.rs
aed900: This was for debugging some dial issues. Removed now.

aed900:
Split into separate PRs:
TCP/QUIC feat flag: https://github.com/maidsafe/safe_network/pull/630
Relay: https://github.com/maidsafe/safe_network/pull/666
Hole-punching: https://github.com/maidsafe/safe_network/pull/667

PR #587, opened 2023-07-28
replication: fix bugs + add unit tests

RolandSherwin:
- fixes a bug where the state is not progressed when `MAX_PARALLEL_FETCHES` was reached. 
- also adds unit tests to replication fetcher
- increase sleep time to fix the record store unit test failure that happens due to reading a half written file. 
### Summary generated by Reviewpad on 28 Jul 23 17:19 UTC

In the first patch, the `next_keys_to_fetch` function in the `ReplicationFetcher` implementation is modified. The change ensures that the state progresses even if the `MAX_PARALLEL_FETCHES` limit is reached. Previously, if the number of ongoing fetches reached the limit, no further fetches would be performed. With this change, fetches can still be performed if there are fetches in progress but the `FETCH_TIMEOUT` is not hit yet.

The second patch adds unit tests for the `ReplicationFetcher` implementation. These tests cover scenarios such as fetching from the network if fetching from a peer fails, trying with multiple peers before fetching from the network, and verifying the behavior when the maximum number of parallel fetches is reached.

Overall, these patches improve the behavior and test coverage of the `ReplicationFetcher` implementation.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -505,7 +505,13 @@ mod tests {
 
         // loop over store.get 10 times to make sure async disk write has had time to complete
         for _ in 0..10 {
-            if store.get(&r.key).is_some() {
+            // try to check if it is equal to the actual record. This is needed because, the file
+            // might not be fully written to the fs and would cause intermittent failures.
+            // If there is actually a problem with the PUT, the assert statement below would catch it.
+            if store
+                .get(&r.key)
+                .is_some_and(|record| Cow::Borrowed(&r) == record)
+            {
                 break;
             }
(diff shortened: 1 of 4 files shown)


sn_networking/src/record_store.rs
```diff
@@ -498,13 +498,8 @@ mod tests {
 
         assert!(store.put_verified(returned_record).is_ok());
 
-        // loop over store.get 10 times to make sure async disk write has had time to complete
```
joshuef, line 506: I don't think this is the right way to tackle this. 

We can compare the record returned with the original and only `break` then

maqi (approved):

PR #596, opened 2023-08-01
fix(ci): provide correct platform value

RolandSherwin:
### Summary generated by Reviewpad on 01 Aug 23 17:48 UTC
This pull request fixes the incorrect platform value in the GitHub workflows for benchmarking and memcheck. It changes the platform value from the matrix.os variable to "ubuntu-latest" in order to provide the correct platform value for the respective workflows.

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -75,7 +75,7 @@ jobs:
           interval: 2000
           node-path: target/release/safenode
           faucet-path: target/release/faucet
-          platform: "ubuntu-latest"
+          platform: ubuntu-latest
           set-safe-peers: false
 
       # In this case we did *not* want SAFE_PEERS to be set to another value by starting the testnet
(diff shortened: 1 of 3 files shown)


.github/workflows/memcheck.yml
```diff
@@ -248,4 +248,4 @@ jobs:
           name: heaptrack_safenode
           path: heaptrack.safenode.*
         continue-on-error: true
-        if: always()
\ No newline at end of file
+        if: always()
```
maqi, line 251: prefer retain the last empty line.
i.e. the file diff shall not have that red stop sign.

joshuef (approved):

maqi (approved):

.github/workflows/memcheck.yml
RolandSherwin: This file does not contain an empty last line, and I think this introduces that empty line. Not the other way around.

PR #607, opened 2023-08-03
feat(transfers): add basic GetStoreCost and use it

joshuef:
### Summary generated by Reviewpad on 03 Aug 23 16:15 UTC
This pull request includes several changes across multiple files. Here is a brief summary of the changes:

- In `sn_client/src/api.rs`, a new import statement has been added for the `Token` type from the `sn_dbc` module.
- In the same file, a new method `get_store_cost_for_dbc_id` has been added to the `Client` struct, which retrieves the store cost for a given DbcId.
- In `sn_client/src/error.rs`, a new variant `CouldNotSendTokens` has been added to the `Error` enum, which represents a general error when sending tokens fails.
- In `sn_client/src/wallet.rs`, a new method `get_store_cost` has been added to the `WalletClient` struct, which retrieves the store cost from the network.
- In the same file, the `pay_for_storage` method has been modified to use the `get_store_cost` method to retrieve the storage cost.
- In `sn_networking/src/cmd.rs`, a comment has been updated to reflect the purpose of the `GetLocalStoreCost` command.
- In `sn_networking/src/error.rs`, a new variant `NoStoreCostReturned` has been added to the `Error` enum, which represents an error when no store cost is returned from the network.
- In `sn_networking/src/lib.rs`, a new method `get_store_cost_from_network` has been added to the `Network` struct, which retrieves the store cost from the network for a given DbcId.
- In `sn_transfers/src/wallet/error.rs`, a new variant `NoDbcsAvailable` has been added to the `Error` enum, which represents an error when no DBCs are available for spend.
- In the same file, a new variant `CouldNotGetStoreCost` has been added to the `Error` enum, which represents a general error when retrieving a store cost fails.
- In `sn_transfers/src/wallet/local_store.rs`, a new method `largest_dbc` has been added to the `LocalWallet` struct, which retrieves the largest DBC available in the wallet.
- In the same file, a new method `add_payment_proofs` has been added to the `LocalWallet` struct, which adds storage payment proofs to the wallet's cache.
- In `sn_transfers/src/wallet/wallet_file.rs`, an error handling statement has been updated to use the `BoxedDbc` variant of the `Error` enum instead of the `Dbc` variant when handling a DBC conversion error.

These are the main changes made in this pull request.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -418,7 +418,8 @@ impl Client {
             Err(ProtocolError::RecordKindMismatch(RecordKind::DbcSpend).into())
         }
     }
-    /// Get a dbc spend from network
+
+    /// get_store_cost_for_dbc_id
     pub async fn get_store_cost_for_dbc_id(&self, dbc_id: &DbcId) -> Result<Token> {
         let address = DbcAddress::from_dbc_id(dbc_id);
         let net_address = NetworkAddress::from_dbc_address(address);
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -17,7 +17,7 @@ use libp2p::{
     kad::{Record, K_VALUE},
     Multiaddr,
 };
-use sn_dbc::{DbcId, SignedSpend};
+use sn_dbc::{DbcId, SignedSpend, Token};
 use sn_networking::{multiaddr_is_global, NetworkEvent, SwarmDriver};
 use sn_protocol::{
     error::Error as ProtocolError,
@@ -417,5 +417,17 @@ impl Client {
             error!("RecordKind mismatch while trying to retrieve a dbc spend");
             Err(ProtocolError::RecordKindMismatch(RecordKind::DbcSpend).into())
         }
+    }
+    /// Get a dbc spend from network
+    pub async fn get_store_cost_for_dbc_id(&self, dbc_id: &DbcId) -> Result<Token> {
+        let address = DbcAddress::from_dbc_id(dbc_id);
+        let net_address = NetworkAddress::from_dbc_address(address);
+
+        trace!("Getting store cost at {dbc_id:?}");
+
+        Ok(self
+            .network
+            .get_store_cost_from_network(net_address.clone())
+            .await?)
     }
 }
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -74,6 +74,18 @@ impl WalletClient {
         }
     }
 
+    /// Get storecost from the network
+    /// this uses the largest DBC as the payment section
+    pub async fn get_store_cost(&mut self) -> Result<Token> {
+        // For now we simply just use the largest DBC we have...
+        let (dbc, _key) = self.wallet.largest_dbc()?;
+
+        self.client
+            .get_store_cost_for_dbc_id(&dbc.id())
+            .await
+            .map_err(|error| Error::CouldNotSendTokens(error.to_string()))
+    }
+
     /// Send tokens to nodes closest to the data we want to make storage payment for.
     /// This can optionally verify the store has been successful (this will attempt to GET the dbc from the network)
     pub async fn pay_for_storage(
@@ -108,8 +120,9 @@ impl WalletClient {
         // For now, we just pay 1 nano per Chunk.
         let num_of_addrs = audit_trail_info.len() as u64;
         // We need to just "burn" the amount that corresponds for storage payment.
-        let storage_cost = Token::from_nano(num_of_addrs);
-        trace!("Making payment for {num_of_addrs} addreses");
+        let storage_cost = self.get_store_cost().await?;
+
+        trace!("Making payment for {num_of_addrs} addresses");
 
         let transfer = self
             .wallet
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -87,12 +87,15 @@ impl WalletClient {
     }
 
     /// Send tokens to nodes closest to the data we want to make storage payment for.
+    ///
+    /// Returns (Proofs and an Option around Storage Cost), storage cost is _per record_, and only returned if required for this operation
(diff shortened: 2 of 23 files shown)


sn_client/src/api.rs
```diff
@@ -419,4 +419,16 @@ impl Client {
             Err(ProtocolError::RecordKindMismatch(RecordKind::DbcSpend).into())
         }
     }
+    /// Get a dbc spend from network
+    pub async fn get_store_cost_for_dbc_id(&self, dbc_id: &DbcId) -> Result<Token> {
```
grumbach: Not sure I understand this comment?

sn_networking/src/lib.rs
```diff
@@ -621,6 +622,43 @@ impl Network {
...
+            .send_and_get_responses(close_nodes, &request, true)
+            .await;
+
+        // loop over responses, generating an avergae fee and storing all responses along side
+        let mut all_costs = vec![];
+        for response in responses.into_iter().flatten() {
+            if let Response::Query(QueryResponse::GetStoreCost(Ok(cost))) = response {
+                all_costs.push(cost);
+            }
+        }
```
grumbach: We might want a log or something in case we error out

sn_transfers/src/wallet/error.rs
```diff
@@ -37,7 +43,10 @@ pub enum Error {
     FailedToHexEncodeKey(String),
     /// Dbc error.
     #[error("Dbc error: {0}")]
-    Dbc(#[from] Box<sn_dbc::Error>),
+    BoxedDbc(#[from] Box<sn_dbc::Error>),
+    /// Dbc error.
+    #[error("Dbc error: {0}")]
+    Dbc(#[from] sn_dbc::Error),
```
grumbach: Why do we need 2 different error types here?

sn_client/src/wallet.rs
```diff
@@ -74,6 +74,18 @@ impl WalletClient {
         }
     }
 
+    /// Get storecost from the network
+    /// this uses the largest DBC as the payment section
+    pub async fn get_store_cost(&mut self) -> Result<Token> {
+        // For now we simply just use the largest DBC we have...
+        let (dbc, _key) = self.wallet.largest_dbc()?;
```
maqi, line 81: Not sure about this `using the largest dbc's cost as the cost of the whole wallet` idea
this `get_store_cost` function is for the whole wallet, indicating the wallet is stored as a whole.
However, as I understand, there is no such `whole wallet stored in network` , instead, there is just a bunch of DBCs scattered across the network?
In that case, given DBC is also kind of small fixed sized, shall we just charge a fixed cost for each storage?
then the `cost of whole wallet` just being a calculation of : `num_of_dbc * fixed_price`

sn_networking/src/lib.rs
joshuef: Not exactly sure what you want logged here, but I've added a bit more in general.

sn_transfers/src/wallet/error.rs
joshuef: refactored away 👍

sn_client/src/wallet.rs
joshuef: > Not sure about this using the largest dbc's cost as the cost of the whole wallet idea

That's fair, and is largely just placeholder to get us going. It doesn't hold up for any payment across more than one dbc eg. I'm hoping this bad impl spurs us on to better things ;)

> instead, there is just a bunch of DBCs scattered across the network

yup


But the cost isn't about the DBC, it's about what _some relevant nodes_ think the price is _right now_. 

Here i chose `NetworkAddress<DbcId>` nodes as I thought they were validating things. But they are not! it's all in `kad::put`. So it's the nodes storing records that validate. So price could be determined from... anywhere? every chunk? I'm not sure what makes sense as a best effort there. But I think we can vastly improve this in follow ups for sure!

maqi: Right, so if currently we just to make a generic approach of `how to decide storege cost`, then I think the targeted address shall be as generic as `NetworkAddress`, instead of a specific dbc_id.
And further to that, the cost is calculated by the close group of the address.

I think the part of using `the medium cost collected from close group` is OK,
however targeting `highest dbc's id` is bit suspicious to me.

joshuef: > then I think the targeted address
What's the `targeted address` here? (atm, I think _any_ query is better that the status quo, so unless it's a simple change I'd perhaps opt for a follow up?)

Or do you literally just mean switch the api interface to `NetworkAddress` there?

>  however targeting highest dbc's id is bit suspicious to me.

How so? (and would this a blocker or something to follow up with for you?)

maqi: > How so? (and would this a blocker or something to follow up with for you?)

Regarding the current `get_store_cost` function API, it seems assuming the cost of storage is `unified` acorss the network.
Which not reflecting the network reality, but may be a `convenient assumption for simpler implementation` ? 
Further to that, as we already assumed the charge will be `unified`, then actually doesn't matter using `largest dbc` or not. Just query any random section will do the same.

> Or do you literally just mean switch the api interface to NetworkAddress there?

Yes, and that `NetworkAddress` is from `address of the data to be stored`

joshuef: >  Which not reflecting the network reality, but may be a convenient assumption for simpler implementation ?
💯 

>Further to that, as we already assumed the charge will be unified, then actually doesn't matter using largest dbc or not. Just query any random section will do the same.

Not 100% in this impl. As this should still get us the _verifying_ nodes just now. As it's those that'll handle the DBC (which is why we use the dbc_id). (Which comes into play with the follow up node-verification pr)


In the end we could just sample the network and go off of that, most likely.


> Yes, and that NetworkAddress is from address of the data to be stored

Considering the above I wouldn't do that in this version. I definitely agree there are improvements to be made.

maqi (approved):

PR #608, opened 2023-08-04
nodes validate payments too

joshuef:
…te validation## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -622,7 +622,7 @@ impl Network {
             .await)
     }
 
-    pub async fn get_store_cost_from_network(&self, dbc_id_key: NetworkAddress) -> Result<Token> {
+    pub async fn get_store_cost_from_network(&self, record_address: NetworkAddress) -> Result<Token> {
         let (sender, receiver) = oneshot::channel();
         debug!("Attempting to get store cost");
         // first we need to get CLOSE_GROUP of the dbc_id
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -622,12 +622,15 @@ impl Network {
             .await)
     }
 
-    pub async fn get_store_cost_from_network(&self, record_address: NetworkAddress) -> Result<Token> {
+    pub async fn get_store_cost_from_network(
+        &self,
+        record_address: NetworkAddress,
+    ) -> Result<Token> {
         let (sender, receiver) = oneshot::channel();
         debug!("Attempting to get store cost");
         // first we need to get CLOSE_GROUP of the dbc_id
         self.send_swarm_cmd(SwarmCmd::GetClosestPeers {
-            key: dbc_id_key.clone(),
+            key: record_address.clone(),
             sender,
         })?;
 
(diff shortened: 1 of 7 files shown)


sn_networking/src/lib.rs
```diff
@@ -621,6 +622,43 @@ impl Network {
             .await)
     }
 
+    pub async fn get_store_cost_from_network(&self, dbc_id_key: NetworkAddress) -> Result<Token> {
```
maqi: this shall be a more generic record_entry, i.e. the target could be a chunk, register or DBC itself ?

sn_networking/src/lib.rs
```diff
@@ -621,6 +622,43 @@ impl Network {
...
+            sender,
+        })?;
+
+        let close_nodes = receiver
+            .await
+            .map_err(|_e| Error::InternalMsgChannelDropped)?
+            .into_iter()
+            .collect_vec();
+
+        let request = Request::Query(Query::GetStoreCost(dbc_id_key));
```
maqi: we already sending the Query to the specific peers.
Hence the `address` of the target is not that matter, instead the `size` will be more interested?

sn_networking/src/lib.rs
```diff
@@ -621,6 +622,43 @@ impl Network {
...
+        let responses = self
+            .send_and_get_responses(close_nodes, &request, true)
+            .await;
+
+        // loop over responses, generating an avergae fee and storing all responses along side
+        let mut all_costs = vec![];
+        for response in responses.into_iter().flatten() {
```
maqi: in case there is error among the responses, say due to time_out, then the `flatten` will just swallow the error right?
in that case, shall we ensure the final OK responses still meet a minium len, say more than 3 ?

sn_node/src/put_validation.rs
```diff
@@ -89,6 +91,7 @@ impl Node {
     pub(crate) async fn validate_and_store_chunk(
         &self,
         chunk_with_payment: ChunkWithPayment,
+        validate_payment: bool,
```
maqi: as mentioned at the discussion https://maidsafe.discourse.team/t/pricing-data-simply/808/13
For ReplicatedData, we may still need to validate the payment is valid, just the `amount` doesn't matter
For original `PutRecord`, we not only need to validate the payment is valid, but also need to confirm the amount is correct

sn_networking/src/lib.rs
joshuef: I'm not sure what you mean here

sn_networking/src/lib.rs
joshuef: That's handled in the `get_fee_from_storecost_quotes`

sn_node/src/put_validation.rs
joshuef: Updated for this 👍

sn_networking/src/lib.rs
maqi: I mean for the `query` request, we already `get_clostest_peers` within the above step according to the `dbc_id_key` address, and then send the request directly to peers one by one.
Hence, the `address content` with the `Query::GetStoreCost` doesn't matter anymore. Unless the cost is related to the data size to be uploaded, then what shall be wrapped in the Query shall be `data size` instead of `data address`.

joshuef: Ah right. Yeh we can drop that entirely I think. the price is per record, so size doesn't matter to nodes, client can work it out.

joshuef: Alllthough, if we drop the address, we don't have `Query.dst()`, so things get complicated. I'd leave that I think.

maqi (approved):

PR #609, opened 2023-08-04
feat: rework register addresses to include pk

grumbach:
## Description

- Fixes the register replacement attack
- Removes unpopular/confusing register tag
- Make register related logs actionable by printing hex encoded `RegisterAddress` everywhere throughout the code so users can directly paste it in the cli
- Fixes a signature bug that was revealed by this PR!

Now register addresses are:
```rust
pub struct RegisterAddress {
    /// User chosen meta, can be anything, the register's name on the network will be the hash of this meta and the owner
    pub(crate) meta: XorName,
    /// Owner of the register
    pub(crate) owner: PublicKey,
}
```
I used `XorName` for `meta` to limit the changes for this PR, but virtually anything is possible here, we could allow users to choose a string, or bytes metadata, anything really. 

This makes `RegisterAddress` bigger so I had to Box a lot of `Error` contents to keep enum sizes in check. 

---

### Summary generated by Reviewpad on 04 Aug 23 10:25 UTC

- The `RegisterOp` struct in `register_op.rs` now has an updated `signature` field with a more detailed comment explaining its requirement.
- The `get_validation.rs` file has changes related to improving memory management and reducing memory usage.
- The `error.rs` file in the `sn_protocol` module has modifications to the error handling logic, using boxes for specific types in the error variants.
- Various changes have been made in the usage of `RegisterAddress` and `XorName` in different files related to interacting with registers.
- The `address.rs` file has undergone multiple changes, including adding and removing dependencies, modifying the `RegisterAddress` struct, implementing new functions, and adding tests.
- The `put_validation.rs` file in the `sn_node/src` directory has updates to the error handling logic, including passing boxed register addresses instead of just register names.
- The `register.rs` file has changes such as removing imports, modifying functions to use `RegisterAddress` instead of `XorName` and `tag`, and updating log messages.
- The `api.rs` file has changes related to error handling and the usage of `RegisterAddress` instead of `address`.
- The implementation of the `ReplicatedData` struct in the `mod.rs` file has been modified, removing the dereference operator of `register.address().name()`.
- The `reg_crdt.rs` file has changes to the error handling related to address mismatch and modifications to the `RegisterAddress` struct in the tests module.
- The `register.rs` file has updates to the register functionality, including changes in function parameters, the removal of certain functions, and updates to log messages.
- The implementation of the `RegisterCrdt` struct in the `reg_crdt.rs` file has changes related to error handling and modifications to the `RegisterAddress` struct in the tests module.
- The `register.rs` file has changes such as adding imports, updating variants, modifying function parameters, and adding print statements.
- The `register.rs` file has modifications such as removing an import, updating method implementations, and simplifying the code.
- The `RegisterAddress` struct in the `network_address.rs` file has changes in the error variants of the `Error` enum.
- The `network_address.rs` file has changes such as removing an import, updating method implementations, and implementing `Debug` for `NetworkAddress`.
- The `register.rs` file has changes such as updating imports, modifying variants, and adding a new variant to the `Error` enum.
- The `register.rs` file has changes such as removing an import, updating method implementations, and simplifying the code.

Overall, these changes involve various updates and modifications in different files related to register operations, error handling, address handling, and function parameters.

### DIFF
--- a/sn_cli/src/subcommands/register.rs
+++ b/sn_cli/src/subcommands/register.rs
@@ -26,7 +26,8 @@ pub enum RegisterCmds {
         /// The address of the register to edit.
         #[clap(name = "address")]
         address: String,
-        /// If you are the owner, the name of the register can be used as a shorthand to the address, as we can derive the address from the public key + name
+        /// If you are the owner, the name of the register can be used as a shorthand to the address,
+        /// as we can derive the address from the public key + name
         /// Use this flag if you are providing the register name instead of the address
         #[clap(name = "name", short = 'n')]
         use_name: bool,
@@ -38,7 +39,8 @@ pub enum RegisterCmds {
         /// The register addresses to get.
         #[clap(name = "addresses")]
         addresses: Vec<String>,
-        /// If you are the owner, the name of the register can be used as a shorthand to the address, as we can derive the address from the public key + name
+        /// If you are the owner, the name of the register can be used as a shorthand to the address,
+        /// as we can derive the address from the public key + name
         /// Use this flag if you are providing the register names instead of the addresses
         #[clap(name = "name", short = 'n')]
         use_name: bool,
@@ -76,13 +78,13 @@ async fn create_register(name: String, client: &Client, verify_store: bool) -> R
 }
 
 async fn edit_register(
-    address: String,
+    address_str: String,
     use_name: bool,
     entry: String,
     client: &Client,
     verify_store: bool,
 ) -> Result<()> {
-    let (address, printing_name) = parse_addr(&address, use_name, client.signer_pk())?;
+    let (address, printing_name) = parse_addr(&address_str, use_name, client.signer_pk())?;
 
     println!("Trying to retrieve Register from {address}");
 
@@ -147,13 +149,17 @@ async fn get_registers(addresses: Vec<String>, use_name: bool, client: &Client) 
 }
 
 /// Parse str and return the address and the register info for printing
-fn parse_addr(str: &str, use_name: bool, pk: PublicKey) -> Result<(RegisterAddress, String)> {
+fn parse_addr(
+    address_str: &str,
+    use_name: bool,
+    pk: PublicKey,
+) -> Result<(RegisterAddress, String)> {
     if use_name {
-        let user_metadata = XorName::from_content(str.as_bytes());
+        let user_metadata = XorName::from_content(address_str.as_bytes());
         let addr = RegisterAddress::new(user_metadata, pk);
-        Ok((addr, format!("'{str}' at {addr}")))
+        Ok((addr, format!("'{address_str}' at {addr}")))
     } else {
-        let addr = RegisterAddress::from_hex(str)?;
-        Ok((addr, format!("at {str}")))
+        let addr = RegisterAddress::from_hex(address_str)?;
+        Ok((addr, format!("at {address_str}")))
     }
 }
--- a/sn_cli/src/subcommands/register.rs
+++ b/sn_cli/src/subcommands/register.rs
@@ -6,31 +6,42 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use bls::PublicKey;
 use clap::Subcommand;
 use color_eyre::Result;
 use sn_client::{Client, ClientRegister, Error as ClientError};
+use sn_protocol::storage::RegisterAddress;
 use xor_name::XorName;
 
 #[derive(Subcommand, Debug)]
 pub enum RegisterCmds {
-    /// Create a new register with the given pet name.
+    /// Create a new register with a name.
     Create {
-        /// The name of the register to create.
+        /// The name of the register to create. This could be the app's name.
+        /// This is used along with your public key to derive the address of the register
         #[clap(name = "name")]
         name: String,
     },
     Edit {
-        /// The name of the register to edit.
-        #[clap(name = "name")]
-        name: String,
+        /// The address of the register to edit.
+        #[clap(name = "address")]
+        address: String,
+        /// If you are the owner, the name of the register can be used as a shorthand to the address, as we can derive the address from the public key + name
+        /// Use this flag if you are providing the register name instead of the address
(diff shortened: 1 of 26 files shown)


sn_cli/src/subcommands/register.rs
```diff
@@ -6,31 +6,42 @@
...
         name: String,
     },
     Edit {
-        /// The name of the register to edit.
-        #[clap(name = "name")]
-        name: String,
+        /// The address of the register to edit.
+        #[clap(name = "address")]
+        address: String,
+        /// If you are the owner, the name of the register can be used as a shorthand to the address, as we can derive the address from the public key + name
```
maqi: consider split the line
I think there is a guideline of `line wrapping with 80 or 100 chars`

sn_cli/src/subcommands/register.rs
```diff
@@ -6,31 +6,42 @@
...
         entry: String,
     },
     Get {
-        /// The register pet names to get.
-        #[clap(name = "names")]
-        names: Vec<String>,
+        /// The register addresses to get.
+        #[clap(name = "addresses")]
+        addresses: Vec<String>,
+        /// If you are the owner, the name of the register can be used as a shorthand to the address, as we can derive the address from the public key + name
```
maqi: same, line split

sn_cli/src/subcommands/register.rs
```diff
@@ -41,43 +52,44 @@ pub(crate) async fn register_cmds(
...
+    println!(
+        "Successfully created register '{name}' at {}!",
+        register.address()
+    );
     Ok(())
 }
 
 async fn edit_register(
-    name: String,
+    address: String,
```
maqi: `address` indicates it's a XorName or something relevant (say HexString of Key)
better consider use other var name, or at least put a comment to explain what is this String for address?

sn_cli/src/subcommands/register.rs
```diff
@@ -138,3 +145,15 @@ async fn get_registers(names: Vec<String>, client: &Client) -> Result<()> {
 
     Ok(())
 }
+
+/// Parse str and return the address and the register info for printing
+fn parse_addr(str: &str, use_name: bool, pk: PublicKey) -> Result<(RegisterAddress, String)> {
```
maqi: here the returned `String` is for `register info` , but above `String` is for `address`
better make all this `type safe`, i.e.  using type such as `RegisterAddress, RegisterInfo, RegisterName, ......` to avoid confusion or misleading.

sn_registers/src/address.rs
```diff
@@ -6,42 +6,95 @@
...
+    /// Construct a new `RegisterAddress` given `meta` and `owner`.
+    pub fn new(meta: XorName, owner: PublicKey) -> Self {
+        Self { meta, owner }
     }
 
-    /// Return the identifier of the register.
+    /// Return the network name of the register.
     /// This is used to locate the register on the network.
-    pub fn id(&self) -> XorName {
+    pub fn name(&self) -> XorName {
```
maqi: I think the previous function name of `id` is more accurate, or even change it to `address`

sn_registers/src/address.rs
```diff
@@ -6,42 +6,95 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use bls::SecretKey;
+
+    use super::*;
+
+    #[test]
+    fn test_register_address_tofro_hex() {
```
maqi: test_register_address_to_from_hex
or
test_register_hex_conversion

sn_registers/src/error.rs
```diff
@@ -57,13 +57,14 @@ pub enum Error {
...
-        "Got Register with an invalid register address, requested: {requested:?}, got: {got:?}"
-    )]
+    #[error("Got Register with an invalid register address, requested: {requested}, got: {got}")]
     InvalidRegisterAddress {
-        requested: RegisterAddress,
-        got: RegisterAddress,
+        requested: Box<RegisterAddress>,
+        got: Box<RegisterAddress>,
     },
+    /// The provided RegisterAddress is invalid
```
maqi: The provided String cann't be deserialized as a RegisterAddress

sn_cli/src/subcommands/register.rs
grumbach: Renamed the source of the confusion, which was the user-input `address_str` that I renamed as such.

sn_registers/src/address.rs
grumbach: Agreed, not a fan of `name` either, but all other NetworkAddress types use `name` for the network xor name:
```rust
NetworkAddress::ChunkAddress(chunk_address) => chunk_address.name().0.to_vec(),
NetworkAddress::DbcAddress(dbc_address) => dbc_address.name().0.to_vec(),
NetworkAddress::RegisterAddress(register_address) => register_address.name().0.to_vec(),
```
I used name for registers as well so it's the same as the others. Although I also think renaming the others is a good idea.

grumbach: `id()` is vague and confusing to me, `address()` is redundant because we are already dealing with addresses, tried it and got `chunk.address().address()` appear in the code. 
Renamed them all to `xorname()` so we are explicit.

sn_cli/src/subcommands/register.rs
```diff
@@ -6,31 +6,44 @@
...
         name: String,
     },
     Edit {
-        /// The name of the register to edit.
-        #[clap(name = "name")]
-        name: String,
+        /// The address of the register to edit.
+        #[clap(name = "address")]
+        address: String,
+        /// If you are the owner, the name of the register can be used as a shorthand to the address,
```
maqi, line 29: the address is deduced from the `name + pk` , which, as long as the pk is known, anyone can deduce the address of a register from its name ?

maqi (approved):

sn_cli/src/subcommands/register.rs
grumbach: That is correct. That's why the owner of a register can use the name directly in the cli while the others will need to use the address because their own key is different. 
This also means anyone can get the owner's key from the address, although we could hide it with a hash and a derived key, but that would add complexity, let's keep it simple for now.

PR #618, opened 2023-08-07
test: record store pruning test

### LINKED ISSUE #495: Add test and ensure records are actually removed
The latest testnet is showing nodes w/ more records on disk than they should be able to hold.

We should add a test and fix any issue here to prevent this from happening

maqi:
Closes https://github.com/maidsafe/safe_network/issues/495

### Summary generated by Reviewpad on 07 Aug 23 13:11 UTC
This pull request includes a test that implements record store pruning. The test ensures that when the record store is full, the pruning behavior is carried out correctly. The test sets the maximum number of records to 50 and generates 100 records. On storing the 51st to 100th record, it checks if the expected pruning behavior is applied. The test verifies that the furthest record is pruned if the new record is closer to the local address, and the new record is retained if it is farther away. Additionally, the test confirms that the pruned record is removed from the store and that the retained record can be read back.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -225,8 +225,7 @@ impl DiskBackedRecordStore {
                 // we should prune and make space
                 self.remove(furthest_record);
 
-                // check if the furthest record is was within our distance range and if so
-                // warn
+                // Warn if the furthest record was within our distance range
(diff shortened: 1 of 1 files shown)


joshuef:
Can we assert that it's actually removed on-disk also? Otherwise lgtm 👍

joshuef (approved):

PR #623, opened 2023-08-08
Improvements to when we get_store_cost

joshuef:
Adds some caching to store cost for the client.

atop #624

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -65,7 +65,7 @@ impl Client {
             signer,
             peers_added: 0,
             progress: Some(Self::setup_connection_progress()),
-            network_store_cost: None,
+            network_store_cost: 0,
         };
 
         // subscribe to our events channel first, so we don't have intermittent
@@ -433,12 +433,12 @@ impl Client {
     pub async fn get_store_cost_at_address(
         &mut self,
         address: NetworkAddress,
-        average_cost: bool,
+        only_update_cost_if_higher: bool,
     ) -> Result<Token> {
         trace!("Getting store cost at {address:?}");
 
         // if we're averaging over many samples across the network, any cost will do
-        let any_cost_will_do = average_cost;
+        let any_cost_will_do = only_update_cost_if_higher;
 
         let cost = self
             .network
@@ -446,14 +446,8 @@ impl Client {
             .await?
             .as_nano();
 
-        if average_cost {
-            if let Some(current_cost) = self.network_store_cost {
-                self.network_store_cost = Some((current_cost + cost) / 2);
-            } else {
-                self.network_store_cost = Some(cost);
-            }
-        } else {
-            self.network_store_cost = Some(cost);
+        if only_update_cost_if_higher && cost > self.network_store_cost {
+            self.network_store_cost = cost;
         }
 
         Ok(Token::from_nano(cost))
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -150,8 +150,6 @@ impl Client {
                     break;
                 }
                 Ok(ClientEvent::InactiveClient(timeout)) => {
-                    let random_target = ChunkAddress::new(XorName::random(&mut rng));
-                    debug!("No ClientEvent activity in the past {timeout:?}, performing a random get_store_cost query to target: {random_target:?}");
                     if is_connected {
                         info!("The client was inactive for {timeout:?}.");
                     } else {
@@ -428,7 +426,7 @@ impl Client {
         address: NetworkAddress,
         only_update_cost_if_higher: bool,
     ) -> Result<Token> {
-        trace!("Getting store cost at {address:?}");
+        trace!("Getting store cost at {address:?}, will update only if higher cost?: {only_update_cost_if_higher:?}");
 
         // if we're averaging over many samples across the network, any cost will do
         let any_cost_will_do = only_update_cost_if_higher;
@@ -439,9 +437,15 @@ impl Client {
             .await?
             .as_nano();
 
-        if only_update_cost_if_higher && cost > self.network_store_cost {
+        if cost > self.network_store_cost {
             self.network_store_cost = cost;
         }
+
+        if !only_update_cost_if_higher {
+            self.network_store_cost = cost;
+        }
+
+        trace!("Set store cost: {}", self.network_store_cost);
 
         Ok(Token::from_nano(cost))
     }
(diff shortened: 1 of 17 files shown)


sn_client/src/api.rs
```diff
@@ -419,16 +431,31 @@ impl Client {
...
+            .as_nano();
+
+        if average_cost {
+            if let Some(current_cost) = self.network_store_cost {
+                self.network_store_cost = Some((current_cost + cost) / 2);
+            } else {
+                self.network_store_cost = Some(cost);
+            }
+        } else {
+            self.network_store_cost = Some(cost);
```
maqi: if not averaging, shall we consider only update when the incoming cost is higher ?

sn_client/src/wallet.rs
```diff
@@ -75,13 +81,13 @@ impl WalletClient {
     }
 
     /// Get storecost from the network
-    /// this uses the largest DBC as the payment section
-    pub async fn get_store_cost(&mut self) -> Result<Token> {
+    /// Stores this value as the new baseline at the client
+    pub async fn set_store_cost_from_random_address(&mut self) -> Result<Token> {
         // For now we simply just use the largest DBC we have...
```
maqi: the comment needs to be updated

sn_client/src/wallet.rs
```diff
@@ -90,6 +96,9 @@ impl WalletClient {
     ///
     /// Returns (Proofs and an Option around Storage Cost), storage cost is _per record_, and only returned if required for this operation
     ///
+    /// `pay_store_cost_at` is the cost to pay for the storage, if None, the cost will be retrieved from the network.
+    /// Normally this will be a cast sampled at client init. It may be this has been user set and passed into the Client.
```
maqi: cast -> cost

sn_networking/src/lib.rs
```diff
@@ -988,16 +989,31 @@ impl Network {
...
     all_costs.sort();
 
-    let token_fee = *all_costs
-        .get(target_cost_index)
-        .ok_or(Error::NotEnoughCostQuotes)?;
+    let token_fee = if any_cost_will_do {
+        let first_try = all_costs
+            .get(target_cost_index)
+            .ok_or(Error::NotEnoughCostQuotes);
+        // first try our ideal, otherwise just use the last one
```
maqi: Prefer the value from first_try, otherwise just use the value from the last.

sn_networking/src/lib.rs
```diff
@@ -1081,6 +1083,35 @@ mod tests {
...
 
+        Ok(())
+    }
+    #[test]
+    fn test_get_any_fee_from_store_cost_quotes() -> eyre::Result<()> {
+        // for a vec of different costs of CLOUSE_GROUP size
+        // ensure we return the CLOSE_GROUP / 2 indexed price
+        let mut costs = vec![];
+        for i in 0..(CLOSE_GROUP_SIZE / 2) - 1 {
+            println!("price: {}", i);
```
maqi: the println can now be removed ?

sn_client/src/api.rs
joshuef: I don't think so? The price could reduce as much as go up?

maqi: well, it's bit tricky here as I understand.
This storage_cost will be used for a payment targeting any part of the network.
If talking about a snapshot of the whole network's storage cost distribution at specific time, it shall always be the highest one to prefer, to secure the cost we pay can be accepted by any part of the network.

However, as you said, if talking about the storage cost along the time, it does can go down as well.

In general, I feel picking the highest one is the safest way to ensure the payment can be accepted across the network, within the current mechanism of how storage cost got calculated by client and validated by network nodes on receiving a payment.

joshuef: Ahhh, yeh, i totally agree with what you're saying. I thought that's what we were doing here but it's not really.

We're averaging, which means we won't pay to a part of the network. So we actually should not avg, but just increase if higher.

joshuef:
atop #624

sn_client/src/api.rs
```diff
@@ -144,13 +151,13 @@ impl Client {
...
+                    debug!("No ClientEvent activity in the past {timeout:?}, performing a random get_store_cost query to target: {random_target:?}");
                     if is_connected {
-                        println!("The client experienced inactivity in the past {timeout:?}.");
+                        info!("The client was inactive for {timeout:?}.");
                     } else {
-                        println!("The client still does not know enough network nodes.");
+                        info!("The client still does not know enough network nodes.");
                     }
-                    let _ = client.get_chunk(random_target).await;
+
```
RolandSherwin, line 158: I think a call to fetch a random `get_store_cost` is missing here.

joshuef: I removed that intentionally as it doesn't seem to help much at all, and confuses the logs

RolandSherwin (approved):

sn_client/src/api.rs
```diff
@@ -144,13 +151,13 @@ impl Client {
                 }
                 Ok(ClientEvent::InactiveClient(timeout)) => {
                     let random_target = ChunkAddress::new(XorName::random(&mut rng));
-                    debug!("No ClientEvent activity in the past {timeout:?}, performing a random get_chunk query to target: {random_target:?}");
+                    debug!("No ClientEvent activity in the past {timeout:?}, performing a random get_store_cost query to target: {random_target:?}");
```
RolandSherwin: This log line has to be updated as we do not perform random queries on inactivity anymore.

sn_client/src/wallet.rs
```diff
@@ -43,6 +45,16 @@ impl WalletClient {
         }
     }
 
+    /// Get any known store cost estimate
+    pub fn store_cost(&self) -> Token {
+        Token::from_nano(self.client.network_store_cost)
+    }
+
+    /// Do we have any uncomfirmed transactions?
```
RolandSherwin:
```suggestion
    /// Do we have any unconfirmed transactions?
```

sn_client/src/wallet.rs
```diff
@@ -90,6 +101,9 @@ impl WalletClient {
     ///
     /// Returns (Proofs and an Option around Storage Cost), storage cost is _per record_, and only returned if required for this operation
     ///
+    /// `pay_store_cost_at` is the cost to pay for the storage, if None, the cost will be retrieved from the network.
+    /// Normally this will be a cost sampled at client init. It may be this has been user set and passed into the Client.
+    ///
```
RolandSherwin: I think this argument is not in use anymore?

PR #626, opened 2023-08-09
fix: distance_range calculation

maqi:
### Summary generated by Reviewpad on 09 Aug 23 11:24 UTC
This pull request makes a minor correction to the `.github/workflows/merge.yml` file. It changes the name of a test from "execute the dbc spend test" to "execute the dbc spend tests".

### DIFF
--- a/sn_node/src/replication.rs
+++ b/sn_node/src/replication.rs
@@ -83,7 +83,9 @@ impl Node {
         }
 
         // set the distance range used for pruning
-        let distance_range = match our_close_group.get(CLOSE_GROUP_SIZE) {
+        // our_close_group now only have CLOSE_GROUP_SIZE entries,
+        // which makes `our_close_group.get(CLOSE_GROUP_SIZE)` return None always.
+        let distance_range = match our_close_group.last() {
             Some(peer) => NetworkAddress::from_peer(*peer).distance(&our_address),
             None => {
                 warn!("Could not obtain distance_range");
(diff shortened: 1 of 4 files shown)


sn_node/src/replication.rs
```diff
@@ -57,10 +57,12 @@ impl Node {
         }
 
         // set the distance range used for pruning
-        let distance_range = match our_close_group.get(CLOSE_GROUP_SIZE) {
+        // our_close_group now only have CLOSE_GROUP_SIZE entries,
+        // which makes `our_close_group.get(CLOSE_GROUP_SIZE)` return None always.
```
joshuef, line 87: I'm not sure I follow why this is always None?

maqi: `our_close_group` now only contains 8th entries, 
but `.get(CLOSE_GROUP_SIZE)` is asking for the entry with `index 8`.
As the vec's indexing is from 0, hence it's actually looking from `9th entry`, which is `None`

joshuef: Ah, i'd perhaps remove this line ```// which makes `our_close_group.get(CLOSE_GROUP_SIZE)` return None always.``` as it's a bit confusing (as we dont do it?)

We can just not perhaps ```// our_close_group now only have CLOSE_GROUP_SIZE entries, so we grab the last entry```

maybe `last and farthest` if they are sorted?

maqi: Ah, that comment is mainly explain what's the issue of previous, i.e. `our_close_group.get(CLOSE_GROUP_SIZE)` .
Not about the current approach (using `last()`).

joshuef:
already in main

PR #635, opened 2023-08-15
feat(client): use network contacts file to speed up the bootstrap process

RolandSherwin:
- Store the multi addresses of our connected peers locally. 
- When the client is restarted, we try connecting to these peers directly. This speeds up the client bootstrap process

### Summary generated by Reviewpad on 15 Aug 23 10:49 UTC
This pull request introduces a new feature to the client code. It adds the ability to use a network contacts file to speed up the bootstrap process. The patch modifies several files in the codebase, including `main.rs`, `api.rs`, `cmd.rs`, `event.rs`, and `lib.rs`. 

In `main.rs`, the patch adds code to read network contacts from the client data directory and use them in the `Client` initialization if no peers are provided as command line arguments. It also stores network contacts in the client data directory after the client has connected to the network.

In `api.rs`, the patch adds functions to store and read network contacts from the client data directory. These functions use the `Network` struct to access the local peer addresses and write/read them to/from the network contacts file.

In `cmd.rs`, the patch adds a new command `GetAllLocalPeerAddresses` to retrieve the multiaddresses of all peers in the local routing table.

In `event.rs`, the patch fixes a typo in a comment.

In `lib.rs`, the patch adds a function `get_all_local_peer_addresses` to the `Network` struct to retrieve the multiaddresses of all peers in the local routing table.

These changes aim to improve the bootstrap process by storing and reusing network contacts, reducing the time required to connect to the network.

### DIFF
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -48,6 +48,7 @@ async fn main() -> Result<()> {
     #[cfg(feature = "metrics")]
     tokio::spawn(init_metrics(std::process::id()));
 
+    info!("\n\n");
     debug!("Built with git version: {}", sn_build_info::git_info());
     println!("Built with git version: {}", sn_build_info::git_info());
     println!("Instantiating a SAFE client...");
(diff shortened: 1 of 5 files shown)


joshuef:
Some checks for this:

- [ ] SAFE_PEERS being set overrides this
- [ ] Contacts are saved on initial contact with network (dont have to wait for CLI to complete)
- [ ] this file is cleaned up by testnet `--clean` flag


Follow up:

- [ ] Contacts are only written if newer?
- [ ] this file is saved to CLI location
- [ ] File _could_ be updated at any point in client lifespan (start/end?)

RolandSherwin:
Closing this as #650 has considerably speeded up the bootstrap process. Though, it was primarily as a result of adding the client event type there.

Issue #636, opened 2023-08-15
Ensure we repay if price has changed


joshuef:
We currently store payment proofs, but we do not check if the price has changed when attempting to reupload a file.

(Here we can assume a reupload is because the file doesn't exist for now).

We should check the existing proof price, and update/repay according to the current network price.

PR #644, opened 2023-08-15
feat(client): do not use cached proofs

joshuef:
Payments may well be out of date. We can check this again once we have DBCs paying per chunk and readily know the price we've paid previously

### Summary generated by Reviewpad on 15 Aug 23 14:06 UTC
This pull request updates the `WalletClient` implementation in the `sn_client/src/wallet.rs` file. It introduces a change to not use cached proofs for payments, as they may be out of date. The code filters the content addresses and collects the ones that need to be paid for. However, the current implementation is commented out as it requires a way to get the store cost from the network per chunk and check what has been paid. If no addresses need to be paid for, the code returns early.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -110,21 +110,25 @@ impl WalletClient {
         // Let's filter the content addresses we hold payment proofs for, i.e. avoid
         // paying for those chunks we've already paid for with this wallet.
         let mut proofs = PaymentProofsMap::default();
-        let addrs_to_pay: Vec<&XorName> = content_addrs
-            .filter(|name| {
-                if let Some(proof) = self.wallet.get_payment_proof(name) {
-                    proofs.insert(**name, proof.clone());
-                    false
-                } else {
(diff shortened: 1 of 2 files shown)


joshuef:
quick fix for https://github.com/maidsafe/safe_network/issues/636

sn_client/src/wallet.rs
```diff
@@ -110,21 +110,25 @@ impl WalletClient {
...
+        //     }
+        // })
+        // .collect();
 
         let number_of_records_to_pay = addrs_to_pay.len() as u64;
+
         // If no addresses need to be paid for, we don't have to go further
         if addrs_to_pay.is_empty() {
-            trace!("We already hold payment proofs for all the Chunks.");
+            trace!("We already hold payment proofs for all the records.");
```
maqi, line 131: now it actually means `nothing to pay according to the inputs` ?
and shall be a warning instead of trace ?

maqi (approved):

Issue #638, opened 2023-08-15
Pay for each chunk


joshuef:
Currently we have some overkeen optimisation which can make calculating a valid cost harder, and will likely complicate the payment flows.

Each record put should:

- [x] not require MerkleTree proof

- [x] get store cost for that record
- [x] pay that cost for that chunk.

PR #646, opened 2023-08-16
feat!(protocol): starting to simplify payments to a per chunk basis

joshuef:
A first step in #638

### Summary generated by Reviewpad on 16 Aug 23 07:59 UTC
This pull request includes several file diffs. Here is a summary of the changes:

1. In the file `cmd.rs`:
   - The struct `PaymentProof` was renamed to `PaymentTransactions`.
   - The fields `audit_trail` and `path` were removed from the `PaymentTransactions` struct.
   - The field `spent_ids` of type `Vec<DbcId>` was added to the `PaymentTransactions` struct.

2. In the file `mod.rs`:
   - Changes in the import statements, specifically the replacement of `PaymentProof` with `PaymentTransactions`.

3. In the file `sn_cli/src/subcommands/files.rs`:
   - Changes in import statements, including the removal of `PaymentProofsMap` and the addition of `PaymentTransactionsMap`.
   - Change in the parameter `payment_proofs` type from `PaymentProofsMap` to `PaymentTransactionsMap` in the function `upload_chunks`.

4. In the file `wallet.rs`:
   - Changes in import statements, swapping `PaymentProofsMap` for `PaymentTransactionsMap`.
   - The return type of the function `chunk_and_pay_for_storage` is now a tuple with `PaymentTransactionsMap`.
   - Updated error message in the function `chunk_and_pay_for_storage`.

5. In the file `put_validation.rs`:
   - Inclusion of the import statement for `PaymentTransactions`.
   - Removal of import statements for `validate_payment_proof` and `XorName`.
   - Modification of the function `verify_fee_output_and_proof` to remove the arguments `addr_name`, `audit_trail`, and `path`.
   - Modification of the parameter `chunk_with_payment` in the function `store_record` to extract the `spent_ids` from `payment`.
   - Modification of the function `verify_fee_output_and_proof` to remove the parameter `audit_trail` and `path`.
   - Modification of the unit test cases to match the changes in the `verify_fee_output_and_proof` function.

6. In the file `wallet.rs` within the `sn_client` module:
   - Replacement of the import `PaymentProof` with `PaymentTransactions`.
   - Renaming of the type `PaymentProofsMap` to `PaymentTransactionsMap`.
   - Commenting out of the function `build_payment_proofs` and the addition of some placeholder code.
   - The variable `num_of_addrs` is now assigned the length of `addrs_to_pay` instead of the length of `audit_trail_info`.
   - The argument `root_hash` has been removed from the function `local_send_storage_payment`.
   - Initialization of the variable `new_payment_proofs` as an empty `PaymentTransactionsMap` and population with payment transactions for each address in `addrs_to_pay`.

7. In the file `sn_transfers/src/wallet/mod.rs`:
   - Replacing the import statement for `PaymentProof` with `PaymentTransactions`.
   - Renaming the type `PaymentProofsMap` to `PaymentTransactionsMap`, which is a BTreeMap from `XorName` to `PaymentTransactions`.

8. In the file `api.rs`:
   - Changing the import statement from `PaymentProof` to `PaymentTransactions`.
   - Modifying the `store_chunk` method in the `impl Client` block, changing the parameter `payment` from `PaymentProof` to `PaymentTransactions`.

9. In the file `sn_protocol/src/storage/chunks.rs`:
   - Changing the import statement from `PaymentProof` to `PaymentTransactions`.
   - Modifying the `Chunk` struct, replacing the `PaymentProof` field with `PaymentTransactions`.
   - Introducing a new struct `ChunkWithPayment`, which contains a `Chunk` field and a `PaymentTransactions` field.

These changes indicate a restructuring of the code related to payment transactions in the protocol, including renaming types, modifying import statements, and updating function parameters and return types. Please review the diff thoroughly and let me know if you need further assistance or clarification.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -10,11 +10,10 @@ use super::Client;
 
 use rand::rngs::OsRng;
 use sn_dbc::{Dbc, PublicAddress, Token};
-use sn_protocol::{messages::PaymentProof, storage::ChunkAddress, NetworkAddress};
+use sn_protocol::{messages::PaymentTransactions, storage::ChunkAddress, NetworkAddress};
 use sn_transfers::{
     client_transfers::TransferOutputs,
-    payment_proof::build_payment_proofs,
-    wallet::{Error, LocalWallet, PaymentProofsMap, Result},
+    wallet::{Error, LocalWallet, PaymentTransactionsMap, Result},
 };
 
 use futures::future::join_all;
@@ -106,10 +105,10 @@ impl WalletClient {
         &mut self,
         content_addrs: impl Iterator<Item = &XorName>,
         verify_store: bool,
-    ) -> Result<(PaymentProofsMap, Option<Token>)> {
+    ) -> Result<(PaymentTransactionsMap, Option<Token>)> {
         // Let's filter the content addresses we hold payment proofs for, i.e. avoid
         // paying for those chunks we've already paid for with this wallet.
-        let mut proofs = PaymentProofsMap::default();
+        let mut proofs = PaymentTransactionsMap::default();
         let addrs_to_pay: Vec<&XorName> = content_addrs
             .filter(|name| {
                 if let Some(proof) = self.wallet.get_payment_proof(name) {
@@ -129,8 +128,10 @@ impl WalletClient {
         }
 
         // Let's build the payment proofs for list of content addresses
-        let (root_hash, audit_trail_info) = build_payment_proofs(addrs_to_pay.into_iter())?;
-        let num_of_addrs = audit_trail_info.len() as u64;
+        // let (root_hash, audit_trail_info) = build_payment_proofs(addrs_to_pay.into_iter())?;
+        // let num_of_addrs = audit_trail_info.len() as u64;
+
+        let num_of_addrs = addrs_to_pay.len();
 
         // Always check storage cost, and overpay to allow margin when validation.
         self.set_store_cost_from_random_address().await?;
@@ -141,12 +142,12 @@ impl WalletClient {
         info!("Storage cost per record: {}", storage_cost);
 
         let amount_to_pay = number_of_records_to_pay * storage_cost.as_nano();
-
         trace!("Making payment for {num_of_addrs} addresses of {amount_to_pay:?} nano tokens.");
 
+        // TODO: This needs to go out to each CLOSEGROUP of addresses
         let transfer = self
             .wallet
-            .local_send_storage_payment(Token::from_nano(amount_to_pay), root_hash, None)
+            .local_send_storage_payment(Token::from_nano(amount_to_pay), None)
             .await?;
 
         // send to network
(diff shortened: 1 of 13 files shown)


sn_client/src/wallet.rs
```diff
@@ -159,19 +160,16 @@ impl WalletClient {
...
-                    addr,
-                    PaymentProof {
-                        spent_ids: spent_ids.clone(),
-                        audit_trail,
-                        path,
-                    },
-                )
-            })
-            .collect();
+        let mut new_payment_proofs = PaymentTransactionsMap::default();
```
grumbach: maybe rename the `new_payment_proofs` var here to match the change?

grumbach (approved):
LGTM

sn_client/src/wallet.rs
```diff
@@ -159,25 +160,17 @@ impl WalletClient {
...
-                        path,
-                    },
-                )
-            })
-            .collect();
+        for addr in addrs_to_pay.into_iter() {
+            proofs.insert(
+                *addr,
+                PaymentTransactions {
+                    spent_ids: spent_ids.clone(),
```
maqi, line 167: In case of paying for each chunk, then spend_id shall be different for each chunk_addr ?

I am more leaning towards to change the work flow within `fn pay_for_storage` to the following.
Which will testout the `pay for each chunk` directly with minium change.
```rust
pub async fn pay_for_storage(
    &mut self,
    content_addrs: impl Iterator<Item = &XorName>,
    verify_store: bool,
) -> Result<PaymentTransactionsMap> {
    // Let's filter the content addresses we hold payment proofs for, i.e. avoid
    // paying for those chunks we've already paid for with this wallet.
    let mut proofs = PaymentTransactionsMap::default();
    let addrs_to_pay: Vec<&XorName> = content_addrs
        .filter(|name| {
            if let Some(proof) = self.wallet.get_payment_proof(name) {
                proofs.insert(**name, proof.clone());
                false
            } else {
                true
            }
        })
        .collect();

    // If no addresses need to be paid for, we don't have to go further
    if addrs_to_pay.is_empty() {
        trace!("We already hold payment proofs for all the Chunks.");
        return Ok((proofs, None));
    }

    // Create payment proof for each chunk
    let mut new_payment_proofs = PaymentTransactionsMap::default();
    for addr in addres_to_pay.iter() {
        let storage_cost = self.get_store_cost_from_netowrk(addr).await?;
        info!("Storage cost for record {addr:?} is: {storage_cost}");

        // Create the payment spend for the chunk and send to network
        let transfer = self
            .wallet
            .local_send_storage_payment(storage_cost)
            .await?;
        trace!("Sending storage payment transfer to the network: {transfer:#?}");
        if let Err(error) = self.client.send(transfer.clone(), verify_store).await {
            warn!("The storage payment transfer was not successfully registered in the network: {error:?}. It will be retried later.");
            self.unconfirmed_txs.push(transfer);
            return Err(error);
        }

        if let Some(proof) = self.wallet.get_payment_proof(addr) {
            new_payment_proofs.insert(**addr, proof.clone());
        } else {
            warn!("Cannot get payment proof of {addr:?}, which is just got created");
        }
    }

    // cache the new set of payment proofs
    self.wallet.add_payment_proofs(new_payment_proofs.clone());

    // return the set of payment proofs, including new ones, and the ones we had in cache
    proofs.extend(new_payment_proofs);

    Ok(proofs)
}
```

joshuef:
#647 and #649 are both ahead of this

PR #648, opened 2023-08-16
fix: avoid download bench result polluted

maqi:
### Summary generated by Reviewpad on 16 Aug 23 11:52 UTC
This pull request fixes an issue where the download benchmark results were being polluted. The patch includes modifications to the benchmark-prs.yml file and the files.rs file. In benchmark-prs.yml, a cleanup step is added to remove the uploaded_files folder to prevent pollution. In files.rs, an argument is added to the Command to specify the log output destination.

### DIFF
--- a/sn_cli/benches/files.rs
+++ b/sn_cli/benches/files.rs
@@ -33,13 +33,16 @@ fn safe_files_upload(dir: &str) {
 
 fn safe_files_download() {
     let output = Command::new("./target/release/safe")
+        .arg("--log-output-dest=data-dir")
         .arg("files")
         .arg("download")
         .output()
         .expect("Failed to execute command");
 
     if !output.status.success() {
-        panic!("Download command executed with failing error code");
+        let err = output.stderr;
+        let err_string = String::from_utf8(err).unwrap();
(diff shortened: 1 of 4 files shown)


sn_cli/benches/files.rs
```diff
@@ -79,7 +82,10 @@ fn criterion_benchmark(c: &mut Criterion) {
 
         let mut group = c.benchmark_group(format!("Upload Benchmark {}MB", size));
         group.sampling_mode(criterion::SamplingMode::Flat);
-        group.measurement_time(Duration::from_secs(120));
+        // One sample may compose of multiple iterations, and this is decided by `measurement_time`.
+        // Set this to a lower value to ensure each sample only contains one iteration.
+        // To ensure the download throughput calculation is correct.
+        group.measurement_time(Duration::from_secs(5));
```
joshuef, line 88: Then the sampling could be off due to an aberration... We want many runs for sounder statistical analysis. 

I'm not sure why this is needed? (same for change below)

joshuef (changes requested):

joshuef (approved):

PR #653, opened 2023-08-17
fix(client): start bootstrap when we are connected to one peer

RolandSherwin:
- The `kad bootstrap process` requires atleast 1 peer to be in the RT.
- Thus the first call to `bootstrap` doesn't actually do anything. We perform the second call only after receiving an event (`ConnectingToNetwork` or `InactiveClient`), which would actually trigger the bootstrap process.
- So moving things around, we can actually remove the `ConnectingToNetwork` event but achieve the same results.
### Summary generated by Reviewpad on 17 Aug 23 17:55 UTC
This pull request fixes the issue where the bootstrap process was not started when the client is connected to at least one peer. It spawns a task to dial to the given peers and starts the bootstrap process if there is at least one peer available. It also removes unused code related to connecting to the network.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -70,34 +70,29 @@ impl Client {
         // (eg, if PeerAdded happens faster than our events channel is created)
         let mut client_events_rx = client.events_channel();
 
-        let mut must_dial_network = true;
-
-        let mut client_clone = client.clone();
-
         let _swarm_driver = spawn({
             trace!("Starting up client swarm_driver");
             swarm_driver.run()
         });
 
+        // spawn task to dial to the given peers
+        let network_clone = network.clone();
+        let _handle = spawn(async move {
+            if let Some(peers) = peers {
+                for addr in peers {
+                    trace!(%addr, "dialing initial peer");
+
+                    if let Err(err) = network_clone.dial(addr.clone()).await {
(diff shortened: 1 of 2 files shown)


sn_client/src/api.rs
```diff
@@ -70,34 +70,29 @@ impl Client {
...
+        // spawn task to dial to the given peers
+        let network_clone = network.clone();
+        let _handle = spawn(async move {
+            if let Some(peers) = peers {
+                for addr in peers {
+                    trace!(%addr, "dialing initial peer");
 
-                        must_dial_network = false;
-                    }
+                    if let Err(err) = network_clone.dial(addr.clone()).await {
```
RolandSherwin, line 85: The `dial process` can be moved outside the `event handler task`.

sn_client/src/api.rs
```diff
@@ -70,34 +70,29 @@ impl Client {
...
-                        });
+        // spawn task to dial to the given peers
+        let network_clone = network.clone();
+        let _handle = spawn(async move {
+            if let Some(peers) = peers {
+                for addr in peers {
+                    trace!(%addr, "dialing initial peer");
 
-                        must_dial_network = false;
-                    }
```
RolandSherwin, line 98: The `dial process` can be moved outside the `event handler task`.

joshuef (approved):

PR #656, opened 2023-08-18
feat(networking): dial unroutable peer

b-zee:


sn_networking/src/lib.rs
```diff
@@ -519,6 +519,13 @@ impl Network {
         receiver.await?
     }
 
+    /// Dial the given peer with the given `DialOpts`.
+    pub async fn dial_with_opts(&self, opts: DialOpts) -> Result<()> {
+        let (sender, receiver) = oneshot::channel();
+        self.send_swarm_cmd(SwarmCmd::DialWithOpts { opts, sender })?;
+        receiver.await?
+    }
```
RolandSherwin: I think this function is not called anywhere?

sn_networking/src/lib.rs
```diff
@@ -403,6 +404,7 @@ impl SwarmDriver {
             // 63 will mean at least 63 most recent peers we have dialed, which should be allow for enough time for the
             // `identify` protocol to kick in and get them in the routing table.
             dialed_peers: CircularVec::new(63),
+            unroutable_peers: CircularVec::new(127),
```
joshuef, line 429: why 127 here vs 63 above?

b-zee: The `UnroutablePeer` event occurs regularly and is not bad news per se. My reasoning is that this vector is filled quite quickly, therefore we need a bit more memory.

joshuef (approved):

PR #664, opened 2023-08-23
feat(networking): single connection per direction

b-zee:
When the node dials peers for various reasons, this might be a dial while we already have an established connection. This isn't too much of a problem usually, but an optimization is to check for previously established connections and deny it if we already have one. This shouldn't alter any behaviour or cause errors to appear.


b-zee:
Closing as there are a few tests failing on CI with this PR. It seems there needs to be some more thought going into this and at the moment it's not a priority at all as the total amount of connections seem relatively low. (Although weeks ago we had some situations that were a little worrying with more than say 5 connections between peers.)

PR #649, opened 2023-08-16
Paying to each holder in each group

joshuef:
starting to simplify payments to a per chunk basis

A first step in #638## Description

### Summary generated by Reviewpad on 16 Aug 23 14:29 UTC
This pull request includes the following changes:
1. In the `lib.rs` file:
   - Import `PublicKey` from `libp2p::identity` and use it in addition to `Multiaddr` from `libp2p`.
   - Add a new field `reward_main_key` of type `PublicKey` to the struct `Node`.
2. In the `file_apis.rs` file:
   - Import statement added: `use crate::NetworkAddress`.
   - Import statement removed: `use crate::messages::PaymentProof`.
   - Import statement added: `use sn_dbc::DbcId`.
   - Method added: `network_address()`, returns the `NetworkAddress` associated with the chunk.
   - Struct field type changed: `payment` changed from `PaymentProof` to `Vec<DbcId>`.
3. In the `api.rs` file:
   - Added `PublicKey` import from `identity` module.
   - Renamed method `get_store_cost_from_network` to `get_store_costs_from_network` and changed its return type to `Result<Vec<(PublicKey, Token)>>`.
   - Updated error handling and added decoding of `PublicKey` in `get_store_costs_from_network` method.
   - Added a retry loop for `put_record_with_retries` method.
   - Made `get_closest_peers` method public.
4. In the `storage_payments.rs` file:
   - Import `sn_dbc::{Hash, Token}` has been removed.
   - The type of `random_content_addrs` has been changed from `XorName` to `sn_protocol::NetworkAddress::ChunkAddress`.
   - The `eyre::Result` import has been simplified.
   - Various changes related to the variable `invalid_proofs` have been made in the function `storage_payment_chunk_upload_fails`.
5. In the `mod.rs` file in the `sn_protocol/src/messages` directory:
   - In the `use` statement, `PaymentProof` has been replaced with `PaymentTransactions`.
6. In the `data_with_churn.rs` file:
   - The method `pay_for_storage` is modified to use the `network_address()` method instead of the `name()` method on each element of the `chunks` iterator.
7. In the `verify_data_location.rs` file:
   - There is a change in the `store_chunk` function where the `pay_for_storage` method is called with the `network_address()` method instead of the `name()` method on each element of the `chunks` iterator.
8. In the `sn_cli/src/subcommands/files.rs` file:
   - The import statement `sn_transfers::wallet::PaymentProofsMap` has been replaced with `sn_transfers::wallet::PaymentTransactionsMap`.
[...]

PR #665, opened 2023-08-23
fix(node): handling events should wait before connected to the network

RolandSherwin:
### Summary generated by Reviewpad on 23 Aug 23 16:08 UTC
This pull request fixes an issue in handling events where the node should wait before performing certain activities until it is connected to enough peers in the network. The patch adds a check for the number of connected peers and waits for the necessary peers before carrying out certain activities. Additionally, the patch includes updates to handle the event of a peer being added to the routing table and increments the count of connected peers.


RolandSherwin:
Pushed directly to #649

PR #666, opened 2023-08-23
feat(node): add relay functionality

aed900:
Adds relay server and client behaviour and transport method. Currently limited to TCP due to AutoNAT with QUIC reporting private NAT statuses as public. There is a fix for this being worked on in libp2p's AutoNATv2 updates.

At the moment, it works as follows: If a node is determined to have a private or public AutoNAT status it acts as a relay client or server respectively. A client automatically requests reservation slots with servers and if accepted adds new relayed listen addresses which use the /p2p-circuit/ protocol and combine the relay server multiaddr and client peer id. Relayed connections are then attempted to be established by any nodes that receive an Identify event containing a relayed listen address as that signifies the remote is a relay client. If successful, the next step is a direct-connection upgrade (hole-punch) using the DCUtR protocol. This feature will be added in an additional PR.

The relay servers are rate-limited by data amounts and time periods.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -32,6 +32,7 @@ use libp2p::{
     request_response::{self, ResponseChannel as PeerResponseChannel},
     swarm::{behaviour::toggle::Toggle, NetworkBehaviour, SwarmEvent},
     Multiaddr, PeerId,
+    relay,
 };
 use sn_protocol::{
     messages::{Request, Response},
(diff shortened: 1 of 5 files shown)


sn_networking/src/event.rs
```diff
@@ -348,18 +437,33 @@ impl SwarmDriver {
...
-                            // self.swarm.behaviour_mut().identify.push(all_peers);
+                            info!("Relay operating in server mode");
+                            self.is_relay_client = false;                        
                         }
                         NatStatus::Private => {
-                            // We could just straight out error here. In the future we might try to activate a relay mechanism.
+                            info!("Relay operating in client mode");
+                            self.is_relay_client = true;
+
+                            for (peer_id, multiaddr) in self.relay_servers.clone() {
```
b-zee, line 447: This adds us to all relay servers we know. We might need to limit this to a certain amount.

joshuef:
Closing this for now. We'll keep this in the backpocket, but won't be adding this as yet to avoid overcomplicating things as we stabilise the code.

Thanks for the work here so far @aed900 🙇

PR #667, opened 2023-08-23
feat(node): add direct-connection upgrade through relay (dcutr)

aed900:
Adds DCUtR behaviour to provide hole-punching functionality for upgrading relayed to direct connections. 

Follows on from relay PR: https://github.com/maidsafe/safe_network/pull/666


joshuef:
Closing this for now. We'll keep this in the backpocket, but won't be adding this _as yet_ to avoid overcomplicating things as we stabilise the code.

Thanks for the work here @aed900 🙇

PR #677, opened 2023-08-30
feat: fetch from network during network

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_node/src/replication.rs
+++ b/sn_node/src/replication.rs
@@ -14,7 +14,7 @@ use libp2p::{
 };
 use sn_networking::{sort_peers_by_address, CLOSE_GROUP_SIZE};
 use sn_protocol::{
-    messages::{Cmd, Query, Request},
+    messages::{Cmd, Request},
     NetworkAddress, PrettyPrintRecordKey,
 };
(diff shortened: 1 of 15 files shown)


sn_node/src/replication.rs
```diff
@@ -107,63 +107,36 @@ impl Node {
...
-            .add_keys_to_replication_fetcher(peer_id, keys)?;
+    /// Add a list of keys to the Replication fetcher.
+    /// These keys are later fetched from network.
+    pub(crate) fn add_keys_to_replication_fetcher(&self, keys: Vec<NetworkAddress>) -> Result<()> {
+        self.network.add_keys_to_replication_fetcher(keys)?;
         Ok(())
     }
 
     /// Get the Record from a peer or from the network without waiting.
     pub(crate) fn fetch_replication_keys_without_wait(
```
joshuef, line 118: This name is wrong now. We do wait

maqi: the function itself doesn't wait.
the fetch and store work are understaken within spawned threads, which not blocking this function.

joshuef (approved):

PR #684, opened 2023-08-31
refactor: optimise getting furthest record

b-zee:
Instead of cloning and sorting an entire vec of record keys, compare
with an iterator and clone only the resulting key.

Tested by asserting the old and new method are equal and running the `pruning_on_full` test.


RolandSherwin (approved):
Looks good to me! :rocket:

PR #692, opened 2023-09-01
feat(networking): change storage cost formula

b-zee:
The storage cost is an exponential calculation, but was adjusted for
readability and so that the store cost stays lower for longer. The cost
only starts ramping up when we're nearing 50% of capacity.

Here are the costs associated with steps:
```
///    1 =         0.000000010
///    2 =         0.000000010
///    4 =         0.000000011
///    8 =         0.000000012
///   16 =         0.000000014
///   32 =         0.000000018
///   64 =         0.000000033
///  128 =         0.000000111
///  256 =         0.000001238
///  512 =         0.000153173
/// 1024 =         2.346196716
/// 1280 =       290.372529764
/// 1536 =     35937.398370712
/// 1792 =   4447723.077333529
/// 2048 = 550463903.051128626 (about 13% of TOTAL_SUPPLY at moment of writing)
```

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -227,15 +227,6 @@ impl RecordStoreAPI for NodeRecordStore {
 
     /// Calculate the cost to store data for our current store state
     fn store_cost(&self) -> Token {
-        // Calculate the factor to increase the cost for every PUTS_PER_PRICE_STEP records
-        let factor =
-            10.0f64.powf(MAX_RECORDS_COUNT as f64 / PUTS_PER_PRICE_STEP as f64 - 2.0_f64) as u64;
-
-        // Calculate the starting cost
-        let mut cost = TOTAL_SUPPLY / factor;
-
-        trace!("Starting cost is {:?}", cost);
-
         trace!("Record count is {:?}", self.records.len());
         let relevant_records_len = if let Some(distance_range) = self.distance_range {
             self.records
@@ -254,17 +245,9 @@ impl RecordStoreAPI for NodeRecordStore {
 
         trace!("Relevant records len is {:?}", relevant_records_len);
 
-        // Find where we are on the scale
-        let current_step = relevant_records_len / PUTS_PER_PRICE_STEP + 1;
+        let cost = calculate_cost_at_step(relevant_records_len);
 
-        trace!("Current step is {:?}", current_step);
-
-        // Double the cost for each step up to the current step
-        for _i in 0..current_step {
-            cost = cost.saturating_add(cost);
-        }
-
-        trace!("Cost is now {:?}", cost);
+        trace!("Cost is now {cost:?}");
         Token::from_nano(cost)
     }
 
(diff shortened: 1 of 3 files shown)


sn_networking/src/record_store.rs
```diff
@@ -452,6 +435,32 @@ impl RecordStore for ClientRecordStore {
...
+///   16 =          0.000000014
+///   32 =          0.000000019
+///   64 =          0.000000036
+///  128 =          0.000000126
+///  256 =          0.000001591
+///  512 =          0.000253098
+/// 1024 =          6.405837009
+/// 2048 = 4103474778.282772064
+fn calculate_cost_at_step(step: usize) -> u64 {
+    assert!(step <= 2048, "step must be <= 2048");
```
joshuef: we want this tied to MAX_RECORD_COUNT var we have elsewhere I think

sn_networking/src/record_store.rs
```diff
@@ -452,6 +435,32 @@ impl RecordStore for ClientRecordStore {
...
+
+    // Using an exponential growth function: y = ab^x. Here, a is the starting cost and b is the growth factor.
+    // We want a function that starts with a low cost and only ramps up once we get closer to the maximum.
+    let a = 0.000_000_010_f64;
+    let b = 1.02_f64;
+    let y = a * b.powf(step as f64);
+
+    (y * 1_000_000_000_f64) as u64
```
joshuef, line 451: I'm not sure how this is tied to the TOTAL_SUPPLY? 

(And I think we should be dividing that supply by CLOSE_GROUP_SIZE to get a real cost to be charged by a given node, otherwise at MAX_RECORD_COUNT we'd be asking in total for `CLOSE_GROUP_SIZE*TOTAL_SUPPLY) ?

sn_networking/src/record_store.rs
```diff
@@ -452,6 +435,32 @@ impl RecordStore for ClientRecordStore {
     fn remove_provider(&mut self, _key: &Key, _provider: &PeerId) {}
 }
 
+/// Cost calculator that increases cost nearing the maximum (2048).
+/// Table:
+///    1 =          0.000000010
```
joshuef: Are these values in full SNT, and so starting at 100 `Token`s?

sn_networking/src/record_store.rs
b-zee: It's not tied to the TOTAL_SUPPLY mathematically. I've picked these numbers by hand. According to the table, with MAX_RECORD_COUNT, the cost will be 4_103_474_778.282772064. I've only just realized by writing it out, that that's actually very close to 2^32. So indeed, probably too much. I can tweak the a and b variables a little bit to get to a lower maximum.

> otherwise at MAX_RECORD_COUNT we'd be asking in total for `CLOSE_GROUP_SIZE*TOTAL_SUPPLY)

I am not sure how this follows from the algorithm?

sn_networking/src/record_store.rs
b-zee: I thought that was 10 nanos, so 10^-8 tokens?

sn_networking/src/record_store.rs
joshuef: each node will be asking for its own payment. So, at a worst case for the last remaining record, the total cost to send to all nodes would be CLOSE_GROUP_SIZE * COST. aka, (currently), 8 times the total supply. 

We can just divide TOTAL_SUPPLY by our CLOSE_GROUP_SIZE to get the actual max value any one node should be asking for?

joshuef (approved):

PR #697, opened 2023-09-04
feat!(protocol): make payments for all record types

joshuef:
### Summary generated by Reviewpad on 04 Sep 23 04:58 UTC
This pull request includes multiple file diffs with various changes. Here is a summary of the changes:

- In the `api.rs` file, the `validate_and_store_record` function call was modified to remove the boolean argument.
- The `sn_protocol/src/messages/mod.rs` file has changes related to the `ChunkWithPayment`, `Chunk`, and `ReplicatedData` structs.
- The `log_markers.rs` file has changes related to the `Marker` enum, where new variants were added.
- The `error.rs` file has changes related to the renaming of the `ChunkNotStored` error variant and updating the associated error message.
- The `sn_protocol/src/storage/mod.rs` file has changes related to import modifications and removals.
- The `sn_protocol/src/storage/chunks.rs` file has changes related to import removals and the removal of the `ChunkWithPayment` struct.
- The `lib.rs` file has changes related to import additions and improvements to existing methods.
- The `Client` implementation in the `lib.rs` file has changes related to the `Store chunk` method, including the removal of the `ChunkWithPayment` struct.
- The `header.rs` file has changes related to the `RecordKind` enum and the `RecordHeader` struct, including new variants, serialize and deserialize trait implementations, and additional methods and test cases.
- The `replication.rs` file has changes related to the modification of the `validate_and_store_record` function call by removing the `false` argument.

These changes indicate modifications to various parts of the codebase, including function calls, imports, enums, structs, error messages, and test cases. Let me know if you need further assistance.

### DIFF
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -27,10 +27,9 @@ use sn_transfers::{
 };
 use std::collections::{BTreeSet, HashSet};
 use tokio::task::JoinSet;
-use xor_name::XorName;
 
 impl Node {
-    /// Validate and store a record to the RecordStore
+    /// Validate a record and it's payment, and store the record to the RecordStore
     pub(crate) async fn validate_and_store_record(
         &self,
         record: Record,
@@ -41,9 +40,8 @@ impl Node {
             RecordKind::ChunkWithPayment => {
                 let record_key = record.key.clone();
                 let (payment, chunk) = try_deserialize_record::<(Vec<Dbc>, Chunk)>(&record)?;
-                let xorname = chunk.address().xorname();
                 let already_exists = self
-                    .validate_key_and_existence(&chunk.network_address(), xorname, &record_key)
+                    .validate_key_and_existence(&chunk.network_address(), &record_key)
                     .await?;
 
                 if already_exists {
@@ -51,12 +49,8 @@ impl Node {
                 }
 
                 // Validate the payment and that we received what we asked.
-                self.payment_for_us_exists_and_is_still_valid(
-                    &chunk.network_address(),
-                    xorname,
-                    &payment,
-                )
-                .await?;
+                self.payment_for_us_exists_and_is_still_valid(&chunk.network_address(), &payment)
+                    .await?;
 
                 self.store_chunk(chunk)
             }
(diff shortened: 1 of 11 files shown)


sn_node/src/put_validation.rs
```diff
@@ -30,48 +29,135 @@ use std::collections::{BTreeSet, HashSet};
...
         record: Record,
-        validate_payment: bool,
     ) -> Result<CmdOk, ProtocolError> {
         let record_header = RecordHeader::from_record(&record)?;
 
+        if !record_header.includes_payment() {
+            return Err(ProtocolError::PaymentProofInsufficientAmount {
+                paid: Token::zero(),
+                expected: Token::zero(),
+            });
```
grumbach: Here this error message might be confusing, why don't we create another error type like `RecordMissingPayment`?

joshuef: i've dropped this entirely with the dbcspendwithpayment removal 👍

sn_protocol/src/storage/header.rs
```diff
@@ -123,6 +130,11 @@ mod tests {
 
     #[test]
     fn verify_record_header_encoded_size() -> Result<()> {
+        let record_with_payment = RecordHeader {
+            kind: RecordKind::ChunkWithPayment,
+        }
+        .try_serialize()?;
+        assert_eq!(record_with_payment.len(), RecordHeader::SIZE);
```
RolandSherwin: Can we also add the check for `RegisterWithPayment`. We had some undefined behaviour with a specific variant earlier if I'm not wrong.

sn_node/src/put_validation.rs
```diff
@@ -30,48 +29,121 @@ use std::collections::{BTreeSet, HashSet};
...
-                    return Err(ProtocolError::RecordKeyMismatch);
+                    let address = NetworkAddress::DbcAddress(dbc_addr);
+
+                    let already_exists = self
+                        .validate_key_and_existence(&address, &record_key)
+                        .await?;
+
+                    if already_exists {
+                        return Ok(CmdOk::DataAlreadyPresent);
+                    }
```
RolandSherwin: Should we return early here? If a `SignedSpend` already exists locally, then we might have to make sure that the new PUt is not a double spend?
I think it is performed inside `validate_and_store_spends`.

joshuef: ahaaa, okay good spot

RolandSherwin (approved):

PR #701, opened 2023-09-04
refactor some node payment for efficiency

joshuef:
### Summary generated by Reviewpad on 04 Sep 23 10:44 UTC
This pull request contains two patches. 

1. The first patch removes an unused argument during spend validation in the `Node` struct. Specifically, the `validate_spend_record` function no longer takes in the `is_replicated_data` argument. This patch also updates the function calls to `validate_spend_record` accordingly.

2. The second patch updates the payment validation process in the `Node` struct. It adds a check to ensure that the payment amount is sufficient before attempting to validate the dbc spend. Additionally, it deposits the payment into the wallet and verifies the transaction afterwards.

Overall, these patches improve the spend validation and payment handling in the code.


grumbach (approved):
LGTM

maqi (approved):

PR #714, opened 2023-09-06
chore: rotate logs after exceeding 20mb

jacderida:
The previous rotation strategy was using the number of lines, which was set to 5000.

Users recently noticed this limitation was producing lots of logs, so we've changed the strategy to be based on the size of the log in bytes.

### Summary generated by Reviewpad on 06 Sep 23 19:14 UTC
This pull request updates the log rotation strategy in the `sn_logging` library. Previously, logs were rotated based on the number of lines, but this caused an excessive number of logs. The new strategy rotates logs based on the size of the log in bytes. This change is implemented in the `appender.rs` and `lib.rs` files. The maximum log size is set to 20mb, the maximum number of uncompressed log files is set to 100, and the maximum number of log files is set to 1000.

### DIFF
--- a/sn_logging/src/appender.rs
+++ b/sn_logging/src/appender.rs
@@ -21,8 +21,8 @@ use std::{
 };
 use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
 
-/// max_lines:
-/// - maximum number of lines per file
+/// max_bytes:
+/// - the maximum size a log can grow to until it is rotated.
 ///
 /// uncompressed_files:
 /// - number of files to keep uncompressed.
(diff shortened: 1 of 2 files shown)


sn_logging/src/appender.rs
```diff
@@ -33,7 +33,7 @@ use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
 /// - older files are deleted.
 pub(super) fn file_rotater(
     dir: &PathBuf,
-    max_lines: usize,
+    max_bytes: usize,
```
RolandSherwin, line 36: Just a nitpick, the docs above might need updating too?

RolandSherwin (approved):
PR looks good to me!

sn_logging/src/appender.rs
jacderida: Ah, good spot! Changed. Thanks.

PR #725, opened 2023-09-13
chore(client): remove unused wallet_client

loziniak:
Seems these parameters are no longer used, and they make API more complex. Perhaps they can be removed? 

### Summary generated by Reviewpad on 14 Sep 23 19:52 UTC
This pull request removes the unused `wallet_client` parameter from the `upload_with_payments` function in the `Files` struct in the `file_apis.rs` file. It also removes the usage of this parameter in the `upload_bytes` function. Additionally, the `upload_with_payments` function calls in the `nodes_rewards.rs`, `storage_payments.rs`, and `verify_data_location.rs` test files have been updated to remove the `wallet_client` parameter. These changes simplify the code by removing unnecessary dependencies.

### DIFF
--- a/sn_client/src/file_apis.rs
+++ b/sn_client/src/file_apis.rs
@@ -129,15 +129,14 @@ impl Files {
 
     /// Directly writes [`Bytes`] to the network in the
     /// form of immutable chunks, without any batching.
-    #[instrument(skip(self, bytes, wallet_client), level = "debug")]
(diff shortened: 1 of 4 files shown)


joshuef (approved):
Ah, nice one! yes this should have been cleaned up in the refactor. Thanks @loziniak ! 🙇

PR #726, opened 2023-09-13
feat: split upload procedure into batches

maqi:
### Summary generated by Reviewpad on 13 Sep 23 13:18 UTC
This pull request introduces a new feature that splits the upload procedure into batches. It modifies the `files.rs` and `wallet.rs` files to implement this feature. The `upload_files` function in `files.rs` now splits the files into batches and uploads them in parallel. The `chunk_and_pay_for_storage` function in `wallet.rs` is also updated to handle batch processing of chunks and payments. Additionally, the `BATCH_SIZE` constant is introduced in `wallet.rs` to define the size of each batch. Overall, this patch improves the efficiency of file uploads by splitting them into smaller batches.

### DIFF
--- a/sn_cli/src/subcommands/files.rs
+++ b/sn_cli/src/subcommands/files.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::wallet::{chunk_and_pay_for_storage, ChunkedFile};
+use super::wallet::{chunk_and_pay_for_storage, ChunkedFile, BATCH_SIZE};
 use bytes::Bytes;
 use clap::Parser;
 use color_eyre::{eyre::Error, Result};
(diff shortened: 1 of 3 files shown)


RolandSherwin:
Hey Qi! Do we also have to batch the chunks while trying to re-pay for the failed ones inside `verify_and_repay_if_needed`?

RolandSherwin:
Unrelated to the PR, but should we `pay and upload in batches` instead of `pay all in batches, wait, then upload all in batches`? It should stop some failures due to churn/cost change.

joshuef:
> Unrelated to the PR, but should we pay and upload in batches instead of pay all in batches, wait, then upload all in batches? It should stop some failures due to churn/cost change.

I think we should do this yep. Indeed if we're uploading a lot of data to small count of nodes we're almost certainly seeing later chunks being rejected due to underpayment.

joshuef:
Can we expose BATCH_SIZE to the CLI layer for easy testing. 

And yeh, if we can do the payment of the batch as part of that batching, that'd be ideal. (We'll have to make sure that currently being spent DBCs are not spendable by subsequent batches... I think we already do this... but perhaps this should indeed be a follow up PR)

maqi:
> Hey Qi! Do we also have to batch the chunks while trying to re-pay for the failed ones inside `verify_and_repay_if_needed`?

Yeah, we should, will check about it.

joshuef (approved):

PR #730, opened 2023-09-15
feat(client): pay for chunks in batches

joshuef:
### Summary generated by Reviewpad on 15 Sep 23 08:47 UTC

- In the `Files` struct, a new method called `pay_for_chunks` has been added. This method interacts with the `wallet` module to pay for a given set of chunks and upload them. It takes a vector of XOR names representing the chunks and a boolean flag to verify the store. After making the payment, it prints the cost and stores the wallet. There are also some private helper methods in this file.

- In the `color_eyre` import statement, `eyre::WrapErr` and `Section` are no longer imported.

- In the function `wallet_cmds`, the `use sn_protocol::storage::ChunkAddress;` import statement is removed.

- The `WalletCmds::Pay` branch in the `match` statement is modified.

- The function `chunk_and_pay_for_storage` is renamed to `chunk_path`.

- The variable `batch_size` is no longer used in the `chunk_path` function.

- The code for making payment and storing the wallet in the `chunk_path` function is removed.

- A new print statement is added to display the total number of chunks to be stored.

Please review these changes for any potential issues.

### DIFF
--- a/sn_cli/src/subcommands/wallet.rs
+++ b/sn_cli/src/subcommands/wallet.rs
@@ -8,12 +8,11 @@
 
 use clap::Parser;
 use color_eyre::{
-    eyre::{bail, eyre, WrapErr},
-    Result, Section,
+    eyre::{bail, eyre},
+    Result,
 };
 use sn_client::{Client, Files, WalletClient};
 use sn_dbc::Token;
-use sn_protocol::storage::ChunkAddress;
 use sn_transfers::wallet::{parse_public_address, LocalWallet};
 use std::{
     collections::BTreeMap,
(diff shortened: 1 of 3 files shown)


sn_cli/src/subcommands/wallet.rs
```diff
@@ -271,22 +283,14 @@ pub(super) struct ChunkedFile {
...
 
-pub(super) async fn chunk_and_pay_for_storage(
+pub(super) async fn chunk_path(
     client: &Client,
     root_dir: &Path,
     files_path: &Path,
-    verify_store: bool,
-    batch_size: usize,
 ) -> Result<BTreeMap<XorName, ChunkedFile>> {
     trace!("Starting to chunk_and_pay_for_storage");
```
RolandSherwin: Just a small nit, I think we can move this function to `files.rs`? As we don't do any wallet related stuff with it.

Also the log line might have to be updated.

RolandSherwin (approved):

PR #733, opened 2023-09-15
chore: output address of uploaded file

jacderida:
- b8487f4f **chore: output address of uploaded file**

  Other minor adjustments for more compact text output.

- b5666cef **feat: provide a `files ls` command**

  Gives the user to list the files they have uploaded, along with their addresses. We were storing
  this data anyway, but it's serialized to binary, so it's not human readable. A little interface on
  the client could be useful for users.

- bdfd4f75 **chore: clarify `files download` usage**

  The documentation for the arguments for the command are clarified, and we introduce a failure if
  only one of the name/address pair is used, as both are necessary.

  Also fix up some code from a complicated rebase.

- ba4809ed **chore: store uploaded files list as text**

  For the uploaded files list, rather than serialize to binary, we use a simple text format. The user
  can now see the addresses of all the files they've uploaded, and use this information to retrieve
  them, if need be. We are also now just using one file, appending to it for each upload.

  Now that this file is human readable, the `files ls` command was deemed unnecessary and was
  therefore removed.

  With respect to the `files` commands, I've also taken the opportunity to make the language in the
  text output more concise and matter of fact. For example, we don't really need to use things like
  exclamation marks in the text.

### Summary generated by Reviewpad on 15 Sep 23 21:01 UTC

- In `cli.rs`:
  - Added a clarification comment for the `timeout` field.
  - Added a clarification comment for the `concurrency` field.
  - Added a clarification comment for the `no_verify` field.

- In `main.rs`:
  - Added a new `FilesCmds` variant, `Ls`, which lists all files uploaded by the current user.
  - Added handling for the `Ls` variant in the `files_cmds` function.

- In `files.rs`:
  - Added a new `FilesCmds` variant, `Ls`, which lists all files uploaded by the current user.
  - Added a new function, `offline_files_cmds`, which handles the `Ls` variant when offline.
  - Added handling for the `Ls` variant in the `files_cmds` function.
  - Added a new function, `list_files`, which lists all files uploaded by the current user.
  - Various changes, including clarifying comments and error handling improvements.

- In `file_apis.rs`:
  - Clarified print statements.

- In `wallet.rs`:
  - Adjusted print statement for completed transfers.

### DIFF
--- a/sn_cli/src/cli.rs
+++ b/sn_cli/src/cli.rs
@@ -62,16 +62,18 @@ pub(crate) struct Opt {
     #[clap(subcommand)]
     pub cmd: SubCmd,
 
-    /// Timeout in seconds for the CLI to wait for a data response from the network.
+    /// Timeout in seconds to wait for a data response from the network.
     #[clap(long = "timeout", global = true, value_parser = |t: &str| -> Result<Duration> { Ok(t.parse().map(Duration::from_secs)?) })]
     pub timeout: Option<Duration>,
 
-    /// Number of concurrent uploads/downloads to allow at any one time.
+    /// Maximum concurrent uploads/downloads.
+    ///
     /// Defaults to 5.
     #[clap(long = "concurrency", short = 'c', global = true)]
     pub concurrency: Option<usize>,
 
-    /// Prevent verification of data storage on the network
+    /// Prevent verification of data storage on the network.
+    ///
     /// This may increase operation speed, but offers no guarantees that operations were successful.
     #[clap(global = true, short = 'n')]
     pub no_verify: bool,
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -15,7 +15,7 @@ mod subcommands;
 use crate::{
     cli::Opt,
     subcommands::{
-        files::{files_cmds, offline_files_cmds, FilesCmds},
+        files::files_cmds,
         register::register_cmds,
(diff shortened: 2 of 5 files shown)


joshuef:
I actually think instead of https://github.com/maidsafe/safe_network/commit/050103b85320bac7a68e28395977867514b3985f we should just make that file human readable. 

`files ls` I could see having the implication it may work on any platform, but its just what you have done on this comp... (something that we will likely have later)... Feels like a plaster around just removing the bincode encoding on that file to me.

joshuef (changes requested):
can we remove https://github.com/maidsafe/safe_network/commit/050103b85320bac7a68e28395977867514b3985f 

I'd rather we made the file human readable and that was how you can check. Doesn't feel like it needs to be in the cli to me.

jacderida:
> can we remove [050103b](https://github.com/maidsafe/safe_network/commit/050103b85320bac7a68e28395977867514b3985f)
> 
> I'd rather we made the file human readable and that was how you can check. Doesn't feel like it needs to be in the cli to me.

OK cool. Serialize to JSON?

joshuef:
> OK cool. Serialize to JSON?

Just a space separated list perhaps? (newline per file?) 


We could do json, but not sure it's worth it... treat yourself as the consumer of this (you are!) so whatever seems best!

joshuef (approved):

PR #734, opened 2023-09-15
chore(client): simplify API

loziniak:
Could it be removed? It compiles and passes `cargo test` without this.

### Summary generated by Reviewpad on 15 Sep 23 23:47 UTC
This pull request simplifies the client API by removing unnecessary code in the file_apis.rs file. The `read_chunk` method now takes in the required parameters and returns the result directly, without the need for the `Self: Sized` constraint.


joshuef (approved):
nice catch and simplification 👍

PR #737, opened 2023-09-18
chore(networking): remove the quote from names as it's misleading

joshuef:
### Summary generated by Reviewpad on 18 Sep 23 12:38 UTC
This pull request removes the quote from names in the `sn_networking` crate as it is misleading. It modifies the `error.rs` and `lib.rs` files, changing the error message to indicate that there are not enough store cost prices returned from the network to ensure a valid fee is paid. It also renames the function `get_fees_from_store_cost_quotes` to `get_fees_from_store_cost_responses` and updates its implementation to handle store cost responses instead of quotes. Additionally, it updates the corresponding test names accordingly.

### DIFF
--- a/sn_networking/src/error.rs
+++ b/sn_networking/src/error.rs
@@ -27,9 +27,9 @@ pub enum Error {
     CouldNotAcquireSemaphorePermit(#[from] tokio::sync::AcquireError),
 
     #[error(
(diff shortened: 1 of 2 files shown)


grumbach (approved):
💯

PR #743, opened 2023-09-19
chore(networking): enable caching of records (in theory)

joshuef:
theory as we dont have any Quorum::One records yet so this alone won't
do much## Description

### Summary generated by Reviewpad on 19 Sep 23 11:45 UTC

PATCH 1/2 implements caching of records in the networking driver. The caching is enabled when the `Quorum::One` is used for the `GetRecord` operation. This change means that the closest 16 nodes will cache the record. However, this alone won't have a significant impact as there are no `Quorum::One` records yet.

PATCH 2/2 allows chunks to be `Quorum::One` instead of `Quorum::All`, as they are self-verifiable and don't require a majority. This change affects multiple files in the codebase.

Overall, these patches improve the caching and performance of network operations, specifically related to records and chunks.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -12,7 +12,11 @@ use super::{
 };
 use bls::{PublicKey, SecretKey, Signature};
 use indicatif::ProgressBar;
-use libp2p::{identity::Keypair, kad::Record, Multiaddr};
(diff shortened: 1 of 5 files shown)


grumbach (approved):
That's great!

PR #762, opened 2023-09-21
chore(networking): reduce identify log noise

joshuef:
### Summary generated by Reviewpad on 21 Sep 23 07:51 UTC
This pull request reduces identify log noise in networking by removing unnecessary debug logs and adding more specific trace logs. Additionally, it provides more informative information when a new peer is added to the routing table.


grumbach (approved):
LGTM

PR #774, opened 2023-09-27
feat(logging): set default logging to data-dir

joshuef:
### Summary generated by Reviewpad on 27 Sep 23 07:58 UTC
This pull request makes several changes related to logging. First, it sets the default logging output destination to the data directory for the `sn_cli`, `sn_node`, and `safenode` binaries. Previously, the default was not specified. Second, it updates the default log levels to be more verbose for the `sn_cli`, `faucet`, and `safenode` binaries. Third, it adds a timestamp to the log file path for the `sn_cli` binary to avoid logs running together over multiple runs. These changes aim to improve the logging functionality and provide more detailed information when needed.

### DIFF
--- a/.github/workflows/benchmark-prs.yml
+++ b/.github/workflows/benchmark-prs.yml
@@ -93,7 +93,7 @@ jobs:
           client_avg_mem_limit_mb="200" # mb
 
           peak_mem_usage=$(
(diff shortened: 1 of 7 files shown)


grumbach (approved):
🚢 ✅

PR #779, opened 2023-09-28
feat: replicate fetch from peer first then from network

maqi:
### Summary generated by Reviewpad on 28 Sep 23 14:50 UTC
This pull request includes several changes across multiple files. Here is a summary of the changes:

- In the `cmd.rs` file:
    - Added a new variant to the `SwarmCmd` enum called `AddKeysToReplicationFetcher`.
    - Updated the `SwarmDriver` implementation to handle the new variant.
    - Modified the call to `self.replication_fetcher.add_keys` to include the new `holder` parameter.
    - Added an `if` condition to check for non-empty `keys_to_fetch` before sending a `NetworkEvent::KeysForReplication`.

- In the `replication.rs` file:
    - Made import statement changes related to the `Record` and `RecordKey` modules.
    - Updated log messages to include additional information about peers, keys, and replication lists.
    - Modified function parameters to include `holder` and additional key-related fields.
    - Updated log messages, requests, and response handling.
  
- In the `log_markers.rs` file:
    - Changed the function `fetching_keys_for_replication()` to accept a vector of tuples containing `PeerId` and `RecordKey`.

- In the `event.rs` file:
    - Modified the `NetworkEvent` enum to include the `PeerId` in the `KeysForReplication` variant.
    - Updated the `SwarmDriver` struct to include the `cache_candidates` field in the `OutboundQueryProgressed` variant.

- In the `replication_fetcher.rs` file:
    - Made various changes related to import statements, field types, return types, and method implementations.
    - Introduced optimizations and improvements for fetching and storing keys for replication.

- In the `query.rs` file:
    - Added a new variant to the `Query` enum called `GetReplicatedRecord`.
    - Implemented the `std::fmt::Display` trait for the `Query` enum.

- In the `messages/mod.rs` file:
    - Made changes related to imports, the `QueryResponse` enum, and the `Query` implementation.

- In the `storage` module of a different file:
    - Added an import statement for the `NetworkAddress` struct.

- In the `sn_protocol/src/messages/mod.rs` file:
    - Removed various imports and methods, and replaced the `ReplicatedData` enum with a modified implementation.

Overall, these changes focus on improving replication functionality, error handling, logging, and code organization across multiple files.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -126,8 +126,8 @@ pub enum NetworkEvent {
     PeerAdded(PeerId),
     // Peer has been removed from the Routing Table
     PeerRemoved(PeerId),
-    /// The Records for the these keys are to be fetched from the network
-    KeysForReplication(Vec<RecordKey>),
+    /// The records bearing these keys are to be fetched from the holder or the network
+    KeysForReplication(Vec<(PeerId, RecordKey)>),
     /// Started listening on a new address
     NewListenAddr(Multiaddr),
     /// AutoNAT status changed
@@ -162,7 +162,7 @@ impl Debug for NetworkEvent {
             NetworkEvent::KeysForReplication(list) => {
                 let pretty_list: Vec<_> = list
                     .iter()
-                    .map(|key| PrettyPrintRecordKey::from(key.clone()))
+                    .map(|(holder, key)| (*holder, PrettyPrintRecordKey::from(key.clone())))
                     .collect();
                 write!(f, "NetworkEvent::KeysForReplication({pretty_list:?})")
             }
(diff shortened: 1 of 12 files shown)


sn_networking/src/event.rs
```diff
@@ -126,8 +126,8 @@ pub enum NetworkEvent {
     PeerAdded(PeerId),
     // Peer has been removed from the Routing Table
     PeerRemoved(PeerId),
-    /// The Records for the these keys are to be fetched from the network
-    KeysForReplication(Vec<RecordKey>),
+    /// The records bearing these keys are to be fetched from the holder or the network
+    KeysForReplication(Vec<(PeerId, RecordKey)>),
```
joshuef, line 130: @maqi might we not want a vec of peerids here? In case we already fetch a record from one peer?

maqi: This is for the notification from swarm_driver to node to carry out the `fectch` action, which has to have a `PeerId` info bundled with `RecordKey` to tell from where to fetch.

Regarding the `parallel fetch from one peer`, there is `MAX_PARALLEL_FETCH` (currently set to 16) within the replication_fetcher to avoid choking a node by sending too many fetch request to it.

Regarding make the replication_fetcher avoid duplicate fetching same record from different peers, I just added a filter at https://github.com/maqi/stableset_net/blob/further_investigate/sn_networking/src/replication_fetcher.rs#L93
let me know if it make sense.

joshuef (approved):

PR #780, opened 2023-09-28
fix(client): just skip empty files

joshuef:
### Summary generated by Reviewpad on 28 Sep 23 18:28 UTC
This pull request fixes the issue where empty files were not being skipped during file chunking. The patch now includes a check to skip empty files and provides an error message indicating which files were skipped.

### DIFF
--- a/sn_cli/src/subcommands/files.rs
+++ b/sn_cli/src/subcommands/files.rs
@@ -137,7 +137,17 @@ pub(super) async fn chunk_path(
             };
 
             let (file_addr, _size, chunks) =
-                file_api.chunk_file(entry.path(), chunks_dir.as_path())?;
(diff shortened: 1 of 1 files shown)


bochaco (approved):

jacderida (approved):

happybeing:
Is skipping a temporary measure?

Zero sized files are valid and so should be stored and retrieved like any other.

joshuef:
> Zero sized files are valid and so should be stored and retrieved like any other.

What should be stored, though? There's nothing to chunk. I think this is an app layer concern over the `sn_client` which cannot store something that doesn't exist.

I think the `safe` the app could be storing files and metadata, which it currently does not. So when we get there, we can look at this again, perhaps. Meantime, best not to fail?

happybeing:
Agree agree best not to fail and I see the issue. Thanks for explaining.

PR #784, opened 2023-10-02
feat(client): reuse cashnotes for address payments

b-zee:


sn_client/src/wallet.rs
```diff
@@ -170,6 +170,26 @@ impl WalletClient {
         info!("Storecosts retrieved");
 
         if !payment_map.is_empty() {
+            // For each target address
+            for (xor, payments) in &mut payment_map {
+                // All payments done for an address, should be multiple nodes.
+                let notes = self.wallet.get_payment_cash_notes(xor);
+                for note in notes {
+                    // Find a payment we're doing to an address, for which we have a cashnote already
+                    if let Some(e) = payments
```
joshuef: Can we clarify `e` here? 

Also `payment we're doing for an address` feels confusing. Are we not looking for a given main pubkey for all payments _to_ an address? 

I think we can clarify the comment there a wee bit.

joshuef (approved):

PR #789, opened 2023-10-02
chore: remove all spans.

joshuef:
There's potential for mem leaks here and they are not well utilised

### Summary generated by Reviewpad on 02 Oct 23 13:44 UTC
This pull request removes all the spans in the "sn_client/src/file_apis.rs" file. The reason for this change is to address potential memory leaks and because the spans were not being well utilized. The patch deletes 9 lines of code in total.


grumbach (approved):
LGTM

PR #790, opened 2023-10-02
chore(logging): reduce metric frequency and logged stats.

joshuef:
Some of this is never used

### Summary generated by Reviewpad on 02 Oct 23 14:05 UTC
This pull request updates the logging in the `sn_logging` crate to reduce the frequency of metric updates and logged stats. Some unused code and unnecessary data types are also removed.

### DIFF
--- a/sn_logging/src/metrics.rs
+++ b/sn_logging/src/metrics.rs
@@ -11,8 +11,8 @@ use std::time::Duration;
 use sysinfo::{self, CpuExt, Pid, PidExt, ProcessExt, System, SystemExt};
 use tracing::{debug, error};
 
-const UPDATE_INTERVAL: Duration = Duration::from_secs(5);
-const TO_MB: f32 = 1_000_000.0;
+const UPDATE_INTERVAL: Duration = Duration::from_secs(15);
+const TO_MB: u64 = 1_000_000;
 
 // The following Metrics are collected and logged
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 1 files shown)


sn_logging/src/metrics.rs
```diff
@@ -62,12 +56,12 @@ pub async fn init_metrics(pid: u32) {
...
-                    cpu_usage_percent: safenode.cpu_usage(),
-                    memory_used_mb: safenode.memory() as f32 / TO_MB,
+                    cpu_usage_percent: safenode.cpu_usage() as u32,
+                    memory_used_mb: safenode.memory() as u32 / TO_MB,
                     bytes_read: disk_usage.read_bytes,
                     bytes_written: disk_usage.written_bytes,
-                    total_mb_read: disk_usage.total_read_bytes as f32 / TO_MB,
-                    total_mb_written: disk_usage.total_written_bytes as f32 / TO_MB,
+                    total_mb_read: disk_usage.total_read_bytes as u32 / TO_MB,
+                    total_mb_written: disk_usage.total_written_bytes as u32 / TO_MB,
```
grumbach: As per the original types,

```rust
pub struct Process {
...
    pub(crate) memory: u64,
    pub(crate) virtual_memory: u64,
    cpu_usage: f32,
...
```

I think we should have memory in `u64` and CPU usage in `f32`

grumbach (approved):
LGTM

happybeing:
I wonder why you've removed system memory etc? Could be useful when looking at why a node crashes or is killed.

PR #802, opened 2023-10-05
fix(client): remove concurrency limitations

joshuef:
### Summary generated by Reviewpad on 05 Oct 23 07:31 UTC
This pull request removes the `concurrency` option from the CLI struct in `sn_cli/src/cli.rs`. The `concurrency` field was previously used to specify the maximum number of concurrent uploads/downloads, but it is no longer needed. The `Client::new` function in `sn_cli/src/main.rs` is updated to remove the `concurrency` parameter when creating a new client. The code in `sn_cli/src/subcommands/files.rs` and `sn_networking/src/api.rs` that references `concurrency` is also updated accordingly. Additionally, the `concurrency_limiter` field in the `Network` struct in `sn_networking/src/lib.rs` and the corresponding methods are removed since they are no longer used. Finally, the `get_cash_note` method in `sn_transfers/src/wallet/local_store.rs` is removed as it is no longer used.

### DIFF
--- a/sn_cli/src/cli.rs
+++ b/sn_cli/src/cli.rs
@@ -73,12 +73,6 @@ pub(crate) struct Opt {
     #[clap(long = "timeout", global = true, value_parser = |t: &str| -> Result<Duration> { Ok(t.parse().map(Duration::from_secs)?) })]
     pub timeout: Option<Duration>,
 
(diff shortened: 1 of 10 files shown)


grumbach (approved):
LGTM!

PR #804, opened 2023-10-05
fix(sn_transfers): be sure we store CashNotes before writing the wall…

joshuef:
…et file

This should prevent concurrency issues with referenced Notes not existing on disk yet

### Summary generated by Reviewpad on 05 Oct 23 10:18 UTC
This pull request fixes an issue related to concurrency when storing CashNotes in the wallet file. It ensures that CashNotes are stored before writing the wallet file to prevent any concurrency issues with referenced Notes not existing on disk yet. There are changes made in multiple files to address this issue.


grumbach (approved):
LGTM!

PR #809, opened 2023-10-06
chore: remove deposit vs received cashnote disctinction

joshuef:
### Summary generated by Reviewpad on 06 Oct 23 11:18 UTC
This pull request includes multiple file diffs. Here's a summary of the changes:

1. In the `wallet.rs` file:
   - Added `UniquePubkey` import from `sn_transfers` module.
   - Added a new method `mark_note_as_spent` to remove a `CashNote` from `available_cash_notes`.
   - Modified the `send_cash_note` method to handle double spend attempts by removing the `CashNote` from the wallet.
   - Modified the `send_cash_note` method to return a specific error message when the storage payment transfer is not successfully registered.
   - Modified the `Client` struct to handle record mismatch errors by removing the `CashNote` from the wallet.
   - Updated the `send_spend_requests` method to handle multiple spend requests and remove corresponding `CashNotes` on double spend or record mismatch errors. These changes improve error handling during cash note sending and redeeming.

2. In the `api.rs` file:
   - Removed commented-out code related to handling a record mismatch error while storing a spend. This code is not currently being used.

3. In the `files.rs` file:
   - Changes to the `upload_files` function include error handling for paying for chunks. Instead of failing, the function now logs a warning message and continues to the next batch if payment fails. This change improves the robustness of the function.

4. In the `error.rs` file in the `wallet` module:
   - Imported the `UniquePubkey` type from the crate.
   - Added a new variant `DoubleSpendAttempted` to the `Error` enum to handle double spend attempts.
   - Added a new variant `InvalidAddressType` to the `Error` enum to handle invalid address types. These changes add support for handling double spend attempts and invalid address types in the wallet module.

5. In the `wallet_file.rs` file:
   - Replaced the `CREATED_CASHNOTES_DIR_NAME` and `RECEIVED_CASHNOTES_DIR_NAME` constants with the `CASHNOTES_DIR_NAME` constant.
   - Removed the `create_received_cash_notes_dir` function.
   - Updated the `store_created_cash_notes` function to write to the `cash_notes` directory instead of `created_cash_notes` directory.
   - Removed the `load_received_cash_notes` function.
   - Renamed the `load_cash_note` function to `load_created_cash_note` and updated it to read from the `cash_notes` directory. These changes involve code refactoring and renaming directories for better organization.

6. In the `wallet_file.rs` file:
   - Renamed the function `load_cash_note` to `load_created_cash_note`.
   - Renamed the function `store_cash_notes` to `store_cash_notes_to_disk`.
   - Removed the function `try_load_deposits`.
   - Modified the function `store` to call `store_cash_notes_to_disk` instead of `store_cash_notes`.
   - Modified the function `deposit` to accept `received_cash_notes` as an argument instead of `cash_notes`.
   - Modified the function `clear_specific_spend_request` to use a `UniquePubkey` parameter instead of `&CashNote`.
   - Removed the function `create_received_cash_notes_dir` and its call inside the `load_from_path` function.
   - Removed unused imports. These changes mainly involve renaming functions and removing unnecessary code.

Overall, this pull request consists of various improvements, including better error handling, code refactoring, and renaming functions for better organization.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -12,7 +12,8 @@ use super::{error::Result, Client};
 use futures::{future::join_all, TryFutureExt};
 use sn_protocol::NetworkAddress;
 use sn_transfers::{
-    CashNote, LocalWallet, MainPubkey, NanoTokens, SignedSpend, Transfer, WalletError, WalletResult,
+    CashNote, LocalWallet, MainPubkey, NanoTokens, SignedSpend, Transfer, UniquePubkey,
+    WalletError, WalletResult,
 };
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -63,6 +64,11 @@ impl WalletClient {
             }
             None => Err(WalletError::InvalidAddressType),
         }
+    }
+
+    /// Remove CashNote from available_cash_notes
+    pub fn mark_note_as_spent(&mut self, cash_note_key: UniquePubkey) {
+        self.wallet.mark_note_as_spent(cash_note_key);
     }
 
     /// Send tokens to another wallet.
@@ -222,7 +228,19 @@ impl WalletClient {
             .await;
         if let Err(error) = spend_attempt_result {
             warn!("The storage payment transfer was not successfully registered in the network: {error:?}. It will be retried later.");
-            return Err(error);
+
+            // if we have a DoubleSpend error, lets remove the CashNote from the wallet
+            if let WalletError::DoubleSpendAttempted(spent_cash_notes) = &error {
+                for cash_note_key in spent_cash_notes {
+                    warn!("Removing CashNote from wallet: {cash_note_key:?}");
+                    self.wallet.mark_note_as_spent(*cash_note_key);
+                    self.wallet.clear_specific_spend_request(*cash_note_key);
+                }
+            }
+
+            return Err(WalletError::CouldNotSendMoney(format!(
+                "The storage payment transfer was not successfully registered in the network: {error:?}"
+            )));
         } else {
             info!("Spend has completed: {:?}", spend_attempt_result);
             self.wallet.clear_unconfirmed_spend_requests();
@@ -277,14 +295,38 @@ impl Client {
                 "sending spend request to the network: {:?}: {spend_request:#?}",
                 spend_request.unique_pubkey()
             );
-            tasks.push(self.network_store_spend(spend_request.clone(), verify_store));
+
+            let the_task = async move {
+                let cash_note_key = spend_request.unique_pubkey();
+                let result = self
+                    .network_store_spend(spend_request.clone(), verify_store)
+                    .await;
+
+                (cash_note_key, result)
+            };
+            tasks.push(the_task);
         }
 
-        for spend_attempt_result in join_all(tasks).await {
-            spend_attempt_result.map_err(|err| WalletError::CouldNotSendMoney(err.to_string()))?;
+        let mut spent_cash_notes = Vec::new();
+        for (cash_note_key, spend_attempt_result) in join_all(tasks).await {
+            // if this is a record mismatch on spend, we need to clean up and remove the spent CashNote from the wallet
+            if let Err(Error::Network(sn_networking::Error::ReturnedRecordDoesNotMatch(
+                record_key,
+            ))) = spend_attempt_result
+            {
+                warn!("Record mismatch on spend, removing CashNote from wallet: {record_key:?}");
+                spent_cash_notes.push(*cash_note_key);
+            } else {
+                return spend_attempt_result
+                    .map_err(|err| WalletError::CouldNotSendMoney(err.to_string()));
+            }
         }
 
-        Ok(())
+        if spent_cash_notes.is_empty() {
+            Ok(())
+        } else {
+            Err(WalletError::DoubleSpendAttempted(spent_cash_notes))
+        }
     }
 
     /// Receive a Transfer, verify and redeem CashNotes from the Network.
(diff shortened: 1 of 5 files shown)


sn_client/src/wallet.rs
```diff
@@ -280,14 +298,38 @@ impl Client {
...
+                (cash_note_key, result)
+            };
+            tasks.push(the_task);
         }
 
-        for spend_attempt_result in join_all(tasks).await {
-            spend_attempt_result.map_err(|err| WalletError::CouldNotSendMoney(err.to_string()))?;
+        let mut spent_cash_notes = Vec::new();
+        for (cash_note_key, spend_attempt_result) in join_all(tasks).await {
+            // if this is a record mismatch on spend, we need to clean up and remove the spent CashNote from the wallet
```
RolandSherwin: Hey Josh, I followed the flow there and even though we emit `ReturnedRecordDoesNotMatch` during `put_record_once(verify=true)`, that error is overwritten as shown below. Thus we might not be exposed to it.
```
// Verify the record is stored, requiring re-attempts
self.get_record_from_network(record_key, verify_store, GetQuorum::All, true)
    .await
    .map_err(|e| {
        error!(
            "Failing to verify the put record {:?} with error {e:?}",
            pretty_key
        );
        Error::FailedToVerifyRecordWasStored(pretty_key)
    })?;
```

sn_client/src/wallet.rs
```diff
@@ -280,14 +298,38 @@ impl Client {
...
+            if let Err(Error::Network(sn_networking::Error::ReturnedRecordDoesNotMatch(
```
RolandSherwin, line 314: Also the clean up happens only if `verify_store == true`. Can we add a comment for that. Might be useful to know that.

sn_client/src/api.rs
```diff
@@ -384,6 +384,18 @@ impl Client {
...
+        //     Err(error) => match error {
+        //         // if record not matching what we expect, we should demove this CashNote from available_cash_notes
+        //         Error::Protocol(ProtocolError::RecordKindMismatch(RecordKind::Spend)) => {
+        //             warn!("RecordKind mismatch while trying to store a spend");
+        //             Err(Error::CouldNotVerifyTransfer(format!(
+        //                 "RecordKind mismatch while trying to store a spend"
+        //             )))
+        //         }
+        //     }
+        // }
```
RolandSherwin: I don't think we check for record mismatch here? So the commented code can be removed I guess?

sn_transfers/src/wallet/local_store.rs
```diff
@@ -62,7 +61,7 @@ impl LocalWallet {
 
     /// Stores the given cash_notes to the `created cash_notes dir` in the wallet dir.
     /// These can then be sent to the recipients out of band, over any channel preferred.
-    pub fn store_cash_notes(&self, cash_note: Vec<&CashNote>) -> Result<()> {
+    pub fn store_cash_notes_to_disk(&self, cash_note: Vec<&CashNote>) -> Result<()> {
```
RolandSherwin, line 59: Can we rename the above `store_cash_note` to `store_cash_note_to_disk` or maybe remove the function entirely?

sn_transfers/src/wallet/local_store.rs
```diff
@@ -821,8 +819,6 @@ mod tests {
 
         assert_eq!(0, recipient.wallet.balance().as_nano());
```
RolandSherwin: We still create `received_cash_note_dir` above in this test.

sn_transfers/src/wallet/local_store.rs
```diff
@@ -333,12 +337,12 @@ impl LocalWallet {
     }
 
     /// Store the given cash_notes to the `received cash_notes dir` in the wallet dir.
```
RolandSherwin: Can we update the comments here to remove `receievd cash_notes dir`?

RolandSherwin:
Also, the CLI docs for `WalletCmds::Deposit` still refers to the `received_cash_notes` dir. Could we update that one as well while we are at it? 🙇‍♂️

sn_client/src/wallet.rs
RolandSherwin: Ahhh.. I was not on the latest main 🤦 It works as intended there! You can safely ignore this comment!

sn_transfers/src/wallet/local_store.rs
joshuef: but its just `cash_notes` now

sn_transfers/src/wallet/error.rs
```diff
@@ -8,12 +8,18 @@
...
+
 /// Specialisation of `std::Result`.
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 /// Transfer errors.
 #[derive(Debug, Error)]
 pub enum Error {
+    /// The cashnote that was attempted to be spent has already been spent and is not the same
+    #[error("Double spend attempted")]
+    DoubleSpendAttempted(Vec<UniquePubkey>),
```
grumbach: Just to make sure here does the `Vec<UniquePubkey>` contain multiple different double spend attempts? If so I'd add a note in the comments saying this error can contain multiple double spends. Also since the `UniquePubkey` are different, I'd make it a `BTreeSet` so we're sure that there are no duplicates.

joshuef: There could be many, if we were spending many CNs (and several were already spent), that's what this is. And we need the holding key to cleanup.

I'll BtreeSet anyway and add more comments/clarity around this

grumbach (approved):
LGTM

PR #810, opened 2023-10-06
fix(sn_networking): actually retry PUTs

joshuef:
### Summary generated by Reviewpad on 06 Oct 23 18:16 UTC
This pull request fixes the issue where PUT requests were not being retried. The patch adds a check to see if the response matches certain error conditions, and if so, continues the retry loop.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -385,6 +385,13 @@ impl Network {
             let res = self
                 .put_record_once(record.clone(), verify_store.clone())
                 .await;
+
+            if matches!(res, Err(Error::RecordNotEnoughCopies(_)))
+                || matches!(res, Err(Error::RecordNotFound))
+            {
+                continue;
+            }
+
             if !matches!(res, Err(Error::FailedToVerifyRecordWasStored(_))) {
                 return res;
             }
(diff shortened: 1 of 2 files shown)


jacderida (approved):

sn_networking/src/lib.rs
```diff
@@ -385,6 +385,13 @@ impl Network {
             let res = self
                 .put_record_once(record.clone(), verify_store.clone())
                 .await;
+
+            if matches!(res, Err(Error::RecordNotEnoughCopies(_)))
+                || matches!(res, Err(Error::RecordNotFound))
+            {
+                continue;
+            }
```
RolandSherwin: I don't think we propagate the `RecordNotEnoughCopies` uptill here? It is intercepted and modified during `get_record`.

Also should we perform `retries+1` here to prevent infinte loop?

sn_networking/src/lib.rs
```diff
@@ -385,6 +385,13 @@ impl Network {
             let res = self
                 .put_record_once(record.clone(), verify_store.clone())
                 .await;
+
```
RolandSherwin, line 388: Can we also add a comment stating that we might only perform this if `verify_store == true`. Or maybe add `if verify_store == true?`

RolandSherwin, line 388: Also this is performed only if `verify_store.is_some()`, so maybe add a comment stating that or maybe add in `if verify_storel.is_some() {}` to indicate it?

PR #811, opened 2023-10-08
chore(networking): remove identify clone and collect

joshuef:
Aiming to reduce a bit of mem allocaiton here for each inbound conn

### Summary generated by Reviewpad on 08 Oct 23 12:55 UTC
This pull request removes the `identify clone and collect` function in the `sn_networking/src/event.rs` file to reduce memory allocation for each inbound connection. It also improves the handling of `SwarmEvents` and adds support for AutoNAT peers.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -190,7 +190,7 @@ impl Debug for NetworkEvent {
 }
 
 impl SwarmDriver {
-    // Handle `SwarmEvents`
+    /// Handle `SwarmEvents`
     pub(super) fn handle_swarm_events<EventError: std::error::Error>(
         &mut self,
         event: SwarmEvent<NodeEvent, EventError>,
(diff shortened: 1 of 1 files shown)


sn_networking/src/event.rs
```diff
@@ -232,17 +232,14 @@ impl SwarmDriver {
                                     .listen_addrs
                                     .into_iter()
                                     .filter(multiaddr_is_global)
+                                    // Strip the `/p2p/...` part of the multiaddresses
+                                    .map(|addr| multiaddr_strip_p2p(&addr))
+                                    // And deduplicate the list
+                                    .unique()
                                     .collect(),
```
b-zee: By moving these, the addresses will not be stripped in case of `local` mode (L229).

sn_networking/src/event.rs
```diff
@@ -274,19 +274,22 @@ impl SwarmDriver {
...
-                            if info.protocols.iter().any(|protocol| {
-                                protocol.to_string().starts_with("/libp2p/autonat/")
-                            }) {
-                                let a = &mut self.swarm.behaviour_mut().autonat;
-                                // It could be that we are on a local network and have AutoNAT disabled.
-                                if let Some(autonat) = a.as_mut() {
-                                    for multiaddr in addrs {
-                                        autonat.add_server(peer_id, Some(multiaddr));
+                            // only add new autonat peers
+                            if !self.dialed_peers.contains(&peer_id) {
```
b-zee: I'm not sure I like the effect of this condition. If we dial the first bootstrap node, we might not add it as an AutoNAT server with this. Also, `dialed_peers` by definition should contain a peer we are adding, because we want to make sure it's reachable.

b-zee (approved):

joshuef:
Closing as merged in via #812

PR #822, opened 2023-10-10
feat(payments): network royalties payment made when storing content

bochaco:
### Summary generated by Reviewpad on 10 Oct 23 20:24 UTC
This pull request adds a new feature to the payments module: network royalties payment made when storing content. The changes include updates to multiple files in the codebase. Specifically, changes were made to the `api.rs`, `put_validation.rs`, `error.rs`, `genesis.rs`, `lib.rs`, and `local_store.rs` files. Some of the notable changes include implementing network royalties payment verification, handling network royalties transfers, and making updates to the `cash_notes_from_payment` function.

### DIFF
--- a/sn_node/src/event.rs
+++ b/sn_node/src/event.rs
@@ -10,7 +10,7 @@ use crate::error::{Error, Result};
 use bls::PublicKey;
 use serde::{Deserialize, Serialize};
 use sn_protocol::storage::{ChunkAddress, RegisterAddress};
-use sn_transfers::{Transfer, UniquePubkey};
+use sn_transfers::{CashNote, UniquePubkey};
 use tokio::sync::broadcast;
 
 const NODE_EVENT_CHANNEL_SIZE: usize = 10_000;
@@ -71,8 +71,8 @@ pub enum NodeEvent {
     TransferNotif {
         /// Public key the transfer notification is about
         key: PublicKey,
-        /// The encrypted transfer
-        transfer: Transfer,
+        /// The cash notes
+        cash_notes: Vec<CashNote>,
     },
 }
 
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -397,10 +397,11 @@ impl Node {
         payment: &Vec<Transfer>,
         wallet: &LocalWallet,
         pretty_key: PrettyPrintRecordKey,
-    ) -> Result<(Vec<CashNote>, Vec<CashNote>), ProtocolError> {
+    ) -> Result<(NanoTokens, Vec<CashNote>, Vec<CashNote>), ProtocolError> {
         let royalties_pk = *NETWORK_ROYALTIES_PK;
         let mut cash_notes = vec![];
         let mut royalties_cash_notes = vec![];
+        let mut received_fee = NanoTokens::zero();
 
         for transfer in payment {
             match transfer {
@@ -414,7 +415,7 @@ impl Node {
                     // transfer invalid
                     Err(e) => return Err(e),
                     // transfer ok
-                    Ok(cn) => cash_notes = cn,
+                    Ok(cns) => cash_notes = cns,
                 },
                 Transfer::NetworkRoyalties(cashnote_redemptions) => {
                     if let Ok(cash_notes) = self
@@ -422,11 +423,15 @@ impl Node {
                         .verify_cash_notes_redemptions(royalties_pk, cashnote_redemptions)
                         .await
                     {
+                        let received_royalties = total_cash_notes_amount(&cash_notes, &pretty_key)?;
                         trace!(
-                            "{} network royalties payment cash notes found for record {pretty_key} for a total value of {:?}",
-                            cash_notes.len(), total_cash_notes_amount(&cash_notes, &pretty_key)?
+                            "{} network royalties payment cash notes found for record {pretty_key} for a total value of {received_royalties:?}",
+                            cash_notes.len()
                         );
                         royalties_cash_notes.extend(cash_notes);
+                        received_fee = received_fee
+                            .checked_add(received_royalties)
+                            .ok_or_else(|| ProtocolError::PaymentExceedsTotalTokens)?;
                     }
                 }
             }
@@ -435,7 +440,16 @@ impl Node {
         if cash_notes.is_empty() {
             Err(ProtocolError::NoPaymentToOurNode(pretty_key))
         } else {
-            Ok((cash_notes, royalties_cash_notes))
+            let received_fee_to_our_node = total_cash_notes_amount(&cash_notes, &pretty_key)?;
+            info!(
+                "{} cash note/s (for a total of {received_fee_to_our_node:?}) are for us for {pretty_key}",
+                cash_notes.len()
+            );
+            received_fee = received_fee
+                .checked_add(received_fee_to_our_node)
+                .ok_or_else(|| ProtocolError::PaymentExceedsTotalTokens)?;
+
+            Ok((received_fee, cash_notes, royalties_cash_notes))
         }
     }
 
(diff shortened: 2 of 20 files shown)


sn_node/src/put_validation.rs
```diff
@@ -378,29 +382,55 @@ impl Node {
...
+                    if let Ok(cash_notes) = self
+                        .network
+                        .verify_cash_notes_redemptions(royalties_pk, cashnote_redemptions)
+                        .await
+                    {
+                        trace!(
+                            "{} network royalties payment cash notes found for record {pretty_key}",
+                            cash_notes.len()
+                        );
+                        royalties_cash_notes = cash_notes;
```
RolandSherwin: Here we overwrite the `royalties_cash_notes`, so I assume we will only have one `Transfer::NetworkRoyalties` here, is that correct? Not really familiar with the royalties and stuff, just wanted to confirm it.

RolandSherwin (approved):

sn_node/src/put_validation.rs
bochaco: good catch, I think in general yes only one, however, I'd agree we shouldn't assume it here and accumulate them all, I'll fix that.

sn_node/src/event.rs
```diff
@@ -71,8 +71,8 @@ pub enum NodeEvent {
     TransferNotif {
         /// Public key the transfer notification is about
         key: PublicKey,
-        /// The encrypted transfer
-        transfer: Transfer,
+        /// The cash notes
```
maqi, line 74: Just curious, so this means the cash_notes are no longer encrypted, right?
is it ok?

sn_node/src/put_validation.rs
```diff
@@ -385,29 +389,53 @@ impl Node {
...
+                },
+                Transfer::NetworkRoyalties(cashnote_redemptions) => {
+                    if let Ok(cash_notes) = self
+                        .network
+                        .verify_cash_notes_redemptions(royalties_pk, cashnote_redemptions)
+                        .await
+                    {
+                        trace!(
+                            "{} network royalties payment cash notes found for record {pretty_key}",
+                            cash_notes.len()
```
maqi, line 429: shall we log the sum of values of those cash_notes?
I think that will be more interested to be known ?

sn_node/src/put_validation.rs
```diff
@@ -385,29 +389,53 @@ impl Node {
...
+                        );
+                        royalties_cash_notes.extend(cash_notes);
```
maqi, line 431: I remember somewhere else we say this direct `extend` may be problematic?
just to double confirm it is OK to do so here.

sn_node/src/put_validation.rs
```diff
@@ -385,29 +389,53 @@ impl Node {
...
+                    .network
+                    .verify_and_unpack_transfer(transfer, wallet)
+                    .await
+                {
+                    // transfer not for us
+                    Err(ProtocolError::FailedToDecypherTransfer) => continue,
+                    // transfer invalid
+                    Err(e) => return Err(e),
+                    // transfer ok
+                    Ok(cn) => cash_notes = cn,
```
maqi: this will always only take the cash_notes list from the last payment.
is it ok ?

sn_node/src/put_validation.rs
```diff
@@ -425,11 +453,14 @@ impl Node {
...
         trace!("Unpacking incoming Transfers for record {pretty_key}");
-        let (cash_notes, transfer_for_us) = self
+        let (cash_notes, royalties_cash_notes) = self
             .cash_notes_from_payment(&payment, &wallet, pretty_key.clone())
             .await?;
 
-        info!("{:?} cash notes are for us", cash_notes.len());
+        info!(
+            "{} cash notes are for us for {pretty_key}",
+            cash_notes.len()
```
maqi: shall we also log the sum value ?

sn_node/src/put_validation.rs
```diff
@@ -476,18 +491,42 @@ impl Node {
...
+            Ok(serialised) => {
+                let royalties_pk = *NETWORK_ROYALTIES_PK;
+                let topic = TRANSFER_NOTIF_TOPIC.to_string();
+                let mut msg: Vec<u8> = royalties_pk.to_bytes().to_vec();
+                msg.extend(serialised);
+                if let Err(err) = self.network.publish_on_topic(topic.clone(), msg) {
+                    debug!("Failed to publish a network royalties payment notification over gossipsub for record {pretty_key} and beneficiary {royalties_pk:?}: {err:?}");
+                }
+            }
+            Err(err) => debug!("Failed to serialise network royalties payment data to publish a notification over gossipsub for record {pretty_key}: {err:?}"),
```
maqi, line 501: just to confirm, with both of the above two error cases,
it is still OK to carry on to the `store cost calculate and upload process`, right?

sn_node/src/event.rs
bochaco: correct, we can change this later to encrypt them here too, anyhow we only send royalties payment cash notes on these notifications now

sn_node/src/put_validation.rs
bochaco: agreed

sn_node/src/put_validation.rs
bochaco: problematic in what sense?

sn_node/src/put_validation.rs
bochaco: I assume so, it used to take just the first one before this change, so no difference there really, plus, from my understanding there will be only one that will get through though which is the one corresponding to the node trying to decrypt the transfer.

sn_node/tests/nodes_rewards.rs
```diff
@@ -198,13 +212,17 @@ fn spawn_transfer_notifs_listener(endpoint: String) -> JoinHandle<Result<usize,
...
+
         while let Some(Ok(e)) = stream.next().await {
             match NodeEvent::from_bytes(&e.event) {
-                Ok(NodeEvent::TransferNotif { key, transfer: _ }) => {
+                Ok(NodeEvent::TransferNotif { key, cash_notes: _ }) => {
                     println!("Transfer notif received for key {key:?}");
-                    count += 1;
-                    if count == CLOSE_GROUP_SIZE {
-                        break;
+                    if key == royalties_pk {
```
maqi, line 226: will there still be normal transfer (i.e. not targeting to network_royalty) ?
if so, this will result in an infinite wait ?

sn_node/src/put_validation.rs
bochaco: agreed

sn_node/src/put_validation.rs
bochaco: I'd think so, the payment for the node will still be ok if the amount is correct

sn_node/src/put_validation.rs
bochaco: actually we already log the cash notes value later on when calculating received_fee

sn_node/src/put_validation.rs
maqi: just leave it.
I recall that was for `all_payment_map`, which is a different case.
shall be OK here.

sn_node/tests/nodes_rewards.rs
bochaco: not at the moment, if there is no tx received with the expected pk then the test shall fail with a timeout

sn_node/src/put_validation.rs
```diff
@@ -476,18 +473,42 @@ impl Node {
             .reward_wallet_balance
             .set(wallet.balance().as_nano() as i64);
 
-        // check payment is sufficient
-        let current_store_cost =
-            self.network.get_local_storecost().await.map_err(|e| {
-                ProtocolError::RecordNotStored(pretty_key.clone(), format!("{e:?}"))
-            })?;
+        if royalties_cash_notes.is_empty() {
```
joshuef, line 485: This fail we should put _after_ we grab any money sent to us. Just in case we were sent something valid.

bochaco: I think we are doing that, it's after we deposited any cash notes belonging to us

bochaco:
Superseded by PR #892

PR #826, opened 2023-10-11
chore(networking): handle GetRecord kad timeouts

joshuef:
### Summary generated by Reviewpad on 11 Oct 23 09:55 UTC
This pull request adds handling of GetRecord kad timeouts in the networking module. It adds an error variant for QueryTimeout in the Error enum. It also includes changes in the event module to handle the timeouts properly and send the corresponding error to the sender.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -654,30 +654,31 @@ impl SwarmDriver {
                 match err.clone() {
                     GetRecordError::NotFound { key, closest_peers } => {
                         info!("Query task {id:?} NotFound record {:?} among peers {closest_peers:?}, {stats:?} - {step:?}",
-                            PrettyPrintRecordKey::from(key.clone()));
+                        PrettyPrintRecordKey::from(key.clone()));
                     }
                     GetRecordError::QuorumFailed {
                         key,
                         records,
                         quorum,
                     } => {
+                        let pretty_key = PrettyPrintRecordKey::from(key.clone());
                         let peers = records
                             .iter()
                             .map(|peer_record| peer_record.peer)
                             .collect_vec();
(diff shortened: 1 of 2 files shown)


sn_networking/src/event.rs
```diff
@@ -671,6 +673,23 @@ impl SwarmDriver {
                             "Query task {id:?} timed out when looking for record {:?}",
                             PrettyPrintRecordKey::from(key.clone())
                         );
+
+                        let (sender, _result_map, _quorum, _expected_holders) =
```
maqi: currently, we send back a `RecordNotFound` error notification for this case.
https://github.com/maidsafe/safe_network/blob/main/sn_networking/src/event.rs#L677

I think a separate error notification here may not be necessary ?

maqi (changes requested):

joshuef:
Actually we may want the timeout to return and or if we already have quorum can return that?

sn_networking/src/event.rs
```diff
@@ -652,25 +654,73 @@ impl SwarmDriver {
...
+
+                        // if we have enough responses here, we can return the record
+                        if let Some((record, peers)) = result_map.values().next() {
+                            if peers.len() >= required_response_count {
+                                sender
+                                    .send(Ok(record.clone()))
+                                    .map_err(|_| Error::InternalMsgChannelDropped)?;
+                                return Ok(());
+                            }
+                            debug!("Get record task {id:?} for {pretty_key:?} returned insufficient responses")
```
maqi: better log the non-responded `expected_holders` here as well ?

maqi: the responded expected node will be removed from the `expected_holders` list.
hence here only need to add `{expected_holders:?}` into the `debug!()` log.

maqi (approved):

PR #834, opened 2023-10-12
fix(network): perfrom bootstrapping continuously

RolandSherwin:
…ected

### Summary generated by Reviewpad on 12 Oct 23 23:02 UTC
This pull request fixes the network module by performing continuous bootstrapping to ensure a well-connected network. The patch updates the driver and event modules, adding code for continuous bootstrapping and adjusting the bootstrap interval based on the number of connected peers.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -759,25 +759,10 @@ impl SwarmDriver {
                 if is_new_peer {
                     self.log_kbuckets(&peer);
                     self.send_event(NetworkEvent::PeerAdded(peer));
+                    // todo: do we need this log here? O(n) to calculate .count();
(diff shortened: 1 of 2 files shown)


sn_networking/src/event.rs
```diff
@@ -759,25 +759,10 @@ impl SwarmDriver {
                 if is_new_peer {
                     self.log_kbuckets(&peer);
                     self.send_event(NetworkEvent::PeerAdded(peer));
+                    // todo: do we need this log here? O(n) to calculate .count();
```
joshuef, line 762: I think this should be fine in a larger network, this should not be so frequent.

joshuef (approved):

PR #836, opened 2023-10-13
fix: batch download process

maqi:
### Summary generated by Reviewpad on 13 Oct 23 11:32 UTC
This pull request fixes the batch download process in the `file_apis.rs` file. It introduces changes to ensure that the futures are executed concurrently, but the results are returned in the order they were inserted. Additionally, it adds a condition to limit the number of futures being processed simultaneously.

### DIFF
--- a/sn_client/src/file_apis.rs
+++ b/sn_client/src/file_apis.rs
@@ -43,6 +43,9 @@ pub struct Files {
 
 type ChunkFileResult = Result<(XorName, u64, Vec<(XorName, PathBuf)>)>;
 
+// Defines the size of batch for the parallel downloading of chunks.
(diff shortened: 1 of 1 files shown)


sn_client/src/file_apis.rs
```diff
@@ -265,6 +265,9 @@ impl Files {
         let mut ordered_read_futures = FuturesOrdered::new();
         let now = Instant::now();
 
+        // The futures are executed concurrently, but the result is returned in the order in which they were inserted.
```
RolandSherwin: Just a silly nit, but can we move this comment  where we call `ordered_read_futures.next()`?

Also can we store the batch_size as a const?

RolandSherwin (approved):

PR #838, opened 2023-10-16
fix: consider record split an error, handle it for regs

grumbach:
## Description

Consider records split (when we receive different values from different nodes for the same key) an error instead of accepting the most popular one.
A split means: 
- for chunks we use `GetQuorum::One` so this error doesn't happen
- for spends it means DOUBLE SPEND
- for registers it means we could have 2 different versions that should be merged

This PR considers this situation to be an error, except for Registers where this error is handled on the client side, which merges the registers into one. 

> If we one day change the GetQuorum of chunks, it would be wise to handle this error accordingly by keeping the valid one (correct hash). 

---

### Summary generated by Reviewpad on 16 Oct 23 08:58 UTC
This pull request includes several changes across multiple files. 

In the `Cargo.toml` file, a new dependency on "eyre" has been added.

In the `sn_networking/src/error.rs` file, there are changes related to imports, type aliases, and the addition of a new variant to the `Error` enum.

In another file, there are changes related to imports, method modifications, and formatting comments.

Lastly, in a different file, there are changes related to imports, the addition of new functions, and the implementation of a new struct.

Overall, the diff includes various changes improving dependency management, error handling, and register handling in the project.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -36,7 +36,7 @@ use sn_protocol::{
     NetworkAddress, PrettyPrintRecordKey,
 };
 use std::{
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     fmt::{Debug, Formatter},
     num::NonZeroUsize,
 };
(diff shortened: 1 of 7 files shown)


sn_networking/src/event.rs
```diff
@@ -938,8 +938,12 @@ impl SwarmDriver {
                     println!("For record {pretty_key:?} task {query_id:?}, fetch completed with non-responded expected holders {expected_holders:?}");
                 }
 
-                let _ = sender.send(result);
-                self.try_update_self_for_split_record(result_map);
+                if result_map.len() == 1 {
+                    let _ = sender.send(result);
+                } else {
+                    println!("For record {pretty_key:?} task {query_id:?}, fetch completed with split record");
```
joshuef: this should not be a println. 

Probs an info or debug?

joshuef (approved):

PR #839, opened 2023-10-17
Client fixes related to quorum and payment

joshuef:
### Summary generated by Reviewpad on 17 Oct 23 07:08 UTC

1. In the `sn_client/src/file_apis.rs` file:
   - The import statement for `Error` has been renamed to `ChunksError`.
   - The error type has been changed from `Error` to `ChunksError` in several places.
   - A `println!` statement has been added to print the payment information for a chunk.
   - The error type in various `return Err` and `map_err` statements has been changed to `ChunksError`.
   - The error type in multiple error logging statements has been changed to `ChunksError`.
   - The error type in an `Ok::<EncryptedChunk, ChunksError>` statement has been changed to `ChunksError`.
   - The error type in an `Err(ChunksError::NotEnoughChunksRetrieved)` statement has been changed to `ChunksError`.
   - The error type in a `return Err(ChunksError::SmallFilePaddingNeeded(chunk.value().len()).into())` statement has been changed to `ChunksError`.
   These changes seem to be related to error handling and logging in the file's functions and methods.

2. The file `nodes_rewards.rs` has a change in the loop condition of the `verify_rewards` function. The loop condition has been modified from `iteration < 10` to `iteration < 15`.

3. In the `WalletClient` implementation in `wallet.rs`:
   - A log statement is added to display the number of payment cash notes retrieved from the wallet.
   - A log statement is updated to include the retrieved store costs.
   These changes enhance the logging functionality for better debugging and monitoring of the wallet operations.

4. In the `put_validation.rs` file:
   - A logging statement is added to print the number of cash notes.
   - A logging statement is added to print the updated received fee.
   - An existing logging statement is modified to include the received fee and current store cost.
   Overall, these changes seem to be related to logging and payment validation.

5. In the `local_store.rs` file, changes have been made to the `impl LocalWallet` block to improve code efficiency by using the `entry` API and reducing lookups in the `BTreeMap`. This change involves iterating over `all_transfers_per_address`, accessing the corresponding `xorname` key in `payment_transactions` using `entry(xorname)`, and appending the `payment_details` to the existing payments for that key using `extend(payment_details)`.

6. A `Transfer` implementation in a file adds a log message to track the creation of a transfer for a specific cash note and recipient.

7. Changes in a file include:
   - A modified log message to mention the number of copies received for a record.
   - An added log statement to print the number of received answers for a record task.
   - An updated conditional statement with an additional condition based on the record type and quorum value.

8. In a file, changes are made to improve logging:
   - Changing the log level from `info!` to `trace!` to reduce severity.
   - Modifying log statements to include more informative output.
   
9. Changes in the file `lib.rs` include:
   - Sorting and truncating the `close_nodes` vector to ensure a maximum number of peers.
   - Creating the `pretty_key` variable from the `key` variable.
   - Adding a debug print statement to display a message when there are not enough copies found for a given `pretty_key`.

Overall, these changes aim to improve error handling, logging, code efficiency, and provide more informative output for better debugging and monitoring of the code.

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -183,13 +183,9 @@ impl SwarmDriver {
             } => {
                 let query_id = self.swarm.behaviour_mut().kademlia.get_record(key.clone());
 
-                let quorum = if expected_holders.is_empty() {
-                    quorum
-                } else {
+                if !expected_holders.is_empty() {
                     debug!("Record {:?} with task {query_id:?} expected to be held by {expected_holders:?}", PrettyPrintRecordKey::from(key));
-                    // Forcing the Quorum to ALL
-                    quorum
-                };
+                }
 
                 if self
                     .pending_get_record
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -188,7 +188,7 @@ impl SwarmDriver {
                 } else {
                     debug!("Record {:?} with task {query_id:?} expected to be held by {expected_holders:?}", PrettyPrintRecordKey::from(key));
                     // Forcing the Quorum to ALL
-                    GetQuorum::All
+                    quorum
                 };
 
                 if self
@@ -208,7 +208,7 @@ impl SwarmDriver {
                     .iter()
                     .map(|(_, (_, result_map, _quorum, _expected_holders))| result_map.len())
                     .sum();
-                info!("We now have {} pending get record attempts and cached {total_records} fetched copies",
+                trace!("We now have {} pending get record attempts and cached {total_records} fetched copies",
                       self.pending_get_record.len());
             }
             SwarmCmd::GetLocalStoreCost { sender } => {
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -923,7 +923,7 @@ impl SwarmDriver {
                 GetQuorum::One => 1,
             };
 
-            println!("Expecting {expected_answers:?} answers for record {pretty_key:?} task {query_id:?}, received {} so far", peer_list.len());
+            trace!("Expecting {expected_answers:?} answers for record {pretty_key:?} task {query_id:?}, received {} so far", peer_list.len());
             let result = if peer_list.len() >= expected_answers {
                 Some(Ok(peer_record.record.clone()))
             } else if usize::from(count) >= CLOSE_GROUP_SIZE {
@@ -971,6 +971,7 @@ impl SwarmDriver {
             if let (true, Ok(true), true) = (
                 expected_holders.is_empty(),
                 RecordHeader::is_record_of_type_chunk(&peer_record.record),
+                // Ensure that we only exit early if quorum is indeed for only one match
                 matches!(quorum, GetQuorum::One),
             ) {
                 if let Ok(chunk) = try_deserialize_record::<Chunk>(&peer_record.record) {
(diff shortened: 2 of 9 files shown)


sn_networking/src/cmd.rs
```diff
@@ -188,7 +188,7 @@ impl SwarmDriver {
                 } else {
                     debug!("Record {:?} with task {query_id:?} expected to be held by {expected_holders:?}", PrettyPrintRecordKey::from(key));
                     // Forcing the Quorum to ALL
-                    GetQuorum::All
+                    quorum
```
grumbach: Maybe we should update the comment

sn_networking/src/event.rs
```diff
@@ -967,9 +968,11 @@ impl SwarmDriver {
         if let Some((sender, result_map, quorum, expected_holders)) =
             self.pending_get_record.remove(query_id)
         {
-            if let (true, Ok(true)) = (
+            if let (true, Ok(true), true) = (
                 expected_holders.is_empty(),
                 RecordHeader::is_record_of_type_chunk(&peer_record.record),
+                // Ensure that we only exit early if quorum is indeed for only one match
+                matches!(quorum, GetQuorum::One),
```
grumbach:
```suggestion
            if expected_holders.is_empty() &&
               RecordHeader::is_record_of_type_chunk(&peer_record.record).unwrap_or(false) &&
               // Ensure that we only exit early if quorum is indeed for only one match
               matches!(quorum, GetQuorum::One)
```

grumbach:
```rust
 if let (true, Ok(true), true) = (
```
is a bit clunky no?

joshuef: I was just rolling with what we had here, but worth a refactor indeed 👍

grumbach (approved):
✅ 🚢

sn_networking/src/lib.rs
```diff
@@ -288,9 +301,9 @@ impl Network {
                     }
                 }
                 Err(Error::RecordNotEnoughCopies(returned_record)) => {
-                    debug!("Not enough copies found yet...");
+                    debug!("Not enough copies found yet for {pretty_key:?}");
                     // Only return when completed all attempts
-                    if verification_attempts >= total_attempts {
+                    if verification_attempts >= total_attempts && matches!(quorum, GetQuorum::One) {
```
maqi, line 306: is the `match` really necessary ?
the original idea of this `return ok branch in case of not having enough copy` is to allow a consensused return
as long as the fetched copy is the only one or matches the expected one, even when the quorum is Majority or All.

sn_transfers/src/wallet/local_store.rs
```diff
@@ -355,9 +355,12 @@ impl LocalWallet {
             }
         }
 
-        self.wallet
-            .payment_transactions
-            .extend(all_transfers_per_address);
+        // Add new payments to the existing map if present
+        // Use entry API to avoid multiple lookups in the BTreeMap
+        for (xorname, payment_details) in all_transfers_per_address {
```
maqi, line 360: what's the benefit here by doing so ?
the previous `extend` doesn't filter out duplicated ones ? or it's too computation expensive ?

grumbach: I think we want to edit the underlying data and append to it. Directly calling extend in the upper layer might replace existing entries instead of merging them together.

sn_networking/src/lib.rs
joshuef: Without this, we return OK for all chunks stored even when that's not true

sn_transfers/src/wallet/local_store.rs
joshuef: > Directly calling extend in the upper layer might replace existing entries instead of merging them together.

This is what appeared to be happening. 

Indeed this may be relted to the other issue you saw w/r/t price calculations

PR #855, opened 2023-10-19
Replicate to cache candidates

RolandSherwin:
## Description
- Propagate the `cache_candiates` that we get from `kad::get_record` to the upper layer.
- This is used to trigger replication on those candidates if they're within our close group.
- Remove targeted replication and instead fetch from network. This would give us more `kad::get record` queries and hence more `cache_candidates`
- Accumulate the `cache_candidates` and return the accumulated ones if our query errors out. We cannot fetch the `cache_candidates` from `swarm.kademlia.query(id)` if the query errors out, hence we have to accumulate them.

### Summary generated by Reviewpad on 19 Oct 23 08:19 UTC

- Added a new variant `PotentialRecordHolders` to the `NetworkEvent` enum, which includes a `key` of type `RecordKey` and a `holders` of type `HashSet<PeerId>`.
- Implemented the `Debug` trait for the `NetworkEvent` enum to print the new variant `PotentialRecordHolders`.
- Added code to handle the `PotentialRecordHolders` event in the `SwarmDriver` struct. It logs statistics and sends a `PotentialRecordHolders` event with the record key and the cache candidates to the network event handler.
- Added a warning message for the case when a query has already been removed from `pending_get_record_for_cache_candidates`.
- Added handling for `NetworkEvent::PotentialRecordHolders` in the `impl Node` block in `api.rs`.
- Added logic to `try_replicate_to_potential_holders` when receiving `NetworkEvent::PotentialRecordHolders` in `api.rs`.
- Added a new variant `ReplicationTriggeredForPotentialHolders` to the `Marker` enum.
- Modified the comment for the `ReplicationTriggered` variant to clarify that it was fired for cache candidates.
- Added a new section for the keys of records being fetched for local replication, including a field `fetching_keys_len` that represents the number of keys being fetched from the network.
- Import statements have been modified to remove unused imports.
- The `try_replicate_to_potential_holders` method has been added to handle replication requests to potential holders of a record key.
- The `add_keys_to_replication_fetcher` method has been modified to fetch records from the network instead of specific nodes.
- Various changes related to logging and error handling in `api.rs`.
- A new function `get_close_group` has been added in `lib.rs` starting at line 193. This function returns the peers that are closest to our PeerId and uses a channel to communicate with the networking module.
- The function `get_store_costs_from_network` in `lib.rs` has been modified to accept a new parameter `record_address` of type `NetworkAddress`.
- Line 190 in `lib.rs` was modified to include the cloned key in the log message.
- Line 203 in `lib.rs` was modified to indicate that an existing `pending_get_record` task was replaced.
- Lines 206-213 in `lib.rs` contain a new if statement to check if an existing `pending_get_record_for_cache_candidates` task was replaced.
- Lines 222-232 in `lib.rs` contain a modified logging statement to indicate the status of the `pending_get_record`.

Please let me know if you need further clarification or assistance with anything else.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -781,10 +781,10 @@ impl SwarmDriver {
                 step,
                 ..
             } => {
-                // here BootstrapOk::num_remaining refers to the remaining random peer IDs to query, one per
-                // bucket that still needs refreshing.
-                trace!("Kademlia Bootstrap with {id:?} progressed with {bootstrap_result:?} and step {step:?}");
                 if step.last {
+                    // here BootstrapOk::num_remaining refers to the remaining random peer IDs to query, one per
+                    // bucket that still needs refreshing.
+                    trace!("Kademlia Bootstrap with {id:?} completed with {bootstrap_result:?} and step {step:?}");
                     // inform the bootstrap process about the completion.
                     self.bootstrap.completed();
                 }
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -697,10 +697,12 @@ impl SwarmDriver {
                     );
                     // combine with the accumulated cache candidates
                     stored_cache_candidates.extend(cache_candidates);
-                    self.send_event(NetworkEvent::PotentialRecordHolders {
-                        key: record_key,
-                        holders: stored_cache_candidates.into_values().collect(),
-                    })
+                    if !stored_cache_candidates.is_empty() {
+                        self.send_event(NetworkEvent::PotentialRecordHolders {
+                            key: record_key,
+                            holders: stored_cache_candidates.into_values().collect(),
+                        })
+                    }
                 } else {
                     warn!("A query has been already removed from pending_get_record_for_cache_candidates {id:?}");
                 }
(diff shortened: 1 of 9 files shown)


sn_networking/src/event.rs
```diff
@@ -745,18 +797,42 @@ impl SwarmDriver {
...
+                    if !stored_cache_candidates.is_empty() {
+                        // log the pending_get_record* stats; get_record_for_cache should not explode and should always be
+                        // greater or equal to get_record
+                        debug!(
+                            "pending_get_record has got {} entries",
+                            self.pending_get_record.len()
+                        );
+                        debug!(
+                            "pending_get_record_for_cache_candidates has got {} entries",
+                            self.pending_get_record_for_cache_candidates.len()
```
joshuef, line 814: why do we need `pending_get_record_for_cache_candidates` ? can we not just directly fire the event there?

RolandSherwin: The `RecordKey` is not present when the `KademliaEvent` are emitted (except for when record is found). We need the RecordKey to inform the cache_candidates.

Expanding the `pending_get_record` to store the `RecordKey` is complicated, as the flow terminates at a lot of places. An extra field was the simplest approach.

RolandSherwin: Also there is a `kademlia.query(id)` function that we can use to get the `RecordKey`, but the query is cleaned up by kad when it completes. Thus, cannot use that one aswell

joshuef:
Is this still worth pushing on @RolandSherwin ? It may just be more complications in the repl process than needed ? (it feels like we're simpified that a wee bit recently, so this may be overkill?)

RolandSherwin:
This indeed would be costlier than the `targetted replication` that we currently have. 
Will close this and we can revisit it, if needed in the future.

PR #878, opened 2023-10-22
chore: use tokio::Interval to trigger forced replication

RolandSherwin:
- use `tokio::Interval` to run forced replication periodically
- have a const to denote the replication interval
### Summary generated by Reviewpad on 22 Oct 23 18:50 UTC
This pull request fixes the issue with replication in the Node implementation. It replaces the usage of `Instant` with `tokio::Interval` to trigger forced replication periodically. The interval is set to 10 seconds. Additionally, the inactivity timeout for nodes to sync when messages are being transmitted is updated to a random range of 20 to 40 seconds.

### DIFF
--- a/sn_node/src/node.rs
+++ b/sn_node/src/node.rs
@@ -31,7 +31,7 @@ use std::{
         atomic::{AtomicUsize, Ordering},
         Arc,
     },
-    time::{Duration, Instant},
+    time::Duration,
 };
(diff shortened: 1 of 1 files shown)


joshuef (approved):

happybeing:
Timers.. magic numbers.. 😱

PR #879, opened 2023-10-23
chore(node): repl timing logs

joshuef:
### Summary generated by Reviewpad on 23 Oct 23 07:01 UTC
This pull request adds timing logs for periodic replication in the `chore/node.rs` and `sn_node/src/replication.rs` files. It also renames the `try_trigger_replication` method to `try_trigger_targetted_replication`. The patch includes changes to the `impl Node` section where the logs and method name changes are made.


RolandSherwin (approved):

sn_networking/src/lib.rs
```diff
@@ -671,7 +670,7 @@ impl Network {
                 Ok(res) => format!("{res}"),
                 Err(err) => format!("{err:?}"),
             };
-            debug!("Got response for the req: {req:?}, res: {res_string}");
```
maqi: I think here we need to `log` some ID stuff of the `req` , so that we can trace it down.

sn_networking/src/lib.rs
```diff
@@ -658,7 +658,6 @@ impl Network {
         req: &Request,
         get_all_responses: bool,
     ) -> Vec<Result<Response>> {
-        debug!("send_and_get_responses for {req:?}");
```
maqi: I think here we need to `log` some ID stuff of the `req` , so that we can trace it down?

maqi (approved):

PR #885, opened 2023-10-23
chore: expand replication range

maqi:
### Summary generated by Reviewpad on 23 Oct 23 14:12 UTC
This pull request expands the replication range in the code. It modifies the `sn_networking/src/cmd.rs`, `sn_networking/src/lib.rs`, and `sn_node/src/replication.rs` files. The changes include updating the `REPLICATE_RANGE` constant to be twice the `CLOSE_GROUP_SIZE`, and using this new range in the code logic to determine whether a peer is in the close range. The patch also involves making adjustments to the replication process based on whether the target peer is a dead peer or a new peer.

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -10,6 +10,7 @@ use crate::{
     driver::SwarmDriver,
     error::{Error, Result},
     sort_peers_by_address, GetQuorum, MsgResponder, NetworkEvent, CLOSE_GROUP_SIZE,
+    REPLICATE_RANGE,
 };
 use libp2p::{
     kad::{store::RecordStore, Quorum, Record, RecordKey},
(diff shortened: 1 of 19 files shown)


joshuef (approved):

sn_networking/src/record_store.rs
```diff
@@ -321,19 +350,28 @@ impl RecordStore for NodeRecordStore {
...
-            trace!(
-                "Unverified Record {:?} already exists.",
-                PrettyPrintRecordKey::from(&record.key)
-            );
+        let record_key = PrettyPrintRecordKey::from(&record.key);
 
-            // Blindly sent to validation to allow double spend can be detected.
-            // TODO: consider avoid throw duplicated chunk to validation.
+        // Chunk with existing key do not to be stored again.
+        // `Spend` or `Register` with same content_hash do not to be stored again,
```
joshuef: `// `Spend` or `Register` with same content_hash do not to be stored again,`

`do need to be stored` no?

maqi: that's for the case of `same key AND same content_cash`, so `do not need to be stored again`

sn_networking/src/cmd.rs
```diff
@@ -372,7 +400,9 @@ impl SwarmDriver {
                     .put_verified(record)
```
RolandSherwin: Can we pass in the `record_type` to `recordstore::put_verified()` too? We're computing the same value inside `put_verified`.

RolandSherwin (approved):

PR #892, opened 2023-10-24
feat(payments): network royalties payment made when storing content

joshuef:
## Description

reviewpad:summary


bochaco (approved):

RolandSherwin (approved):

loziniak:
This should be a breaking change (major version), because `Transfer` changed from struct to enum, and transfers from *faucet* with `sn_transfers = "0.14.4"` are not parseable by *client* with `sn_transfers = "0.14.5"`

PR #900, opened 2023-10-25
feat(cli): chunk files in parallel

RolandSherwin:
### Summary generated by Reviewpad on 25 Oct 23 12:07 UTC
This pull request includes three patches:

1. `chore(cli): add logs to indicate the time spent on chunking the files`
   - Adds logging to indicate the time spent on chunking the files in the CLI.
   - Inserts logs at the start and end of the chunking process.
   - Prints the time taken to chunk all the files.
   
2. `feat(cli): chunk files in parallel`
   - Adds parallel chunking of files in the CLI.
   - Adds `rayon` as a dependency.
   - Uses `rayon` to parallelize the chunking process.
   - Improves the performance of file chunking.

3. `fix(cli): remove Arc from ProgressBar as it is Arc internally`
   - Removes `Arc` from `ProgressBar` as it is `Arc` internally.
   - Updates the function signature of `get_progress_bar()` to return a `ProgressBar` directly.

Please review the patches and provide feedback.

### DIFF
--- a/sn_cli/src/subcommands/files.rs
+++ b/sn_cli/src/subcommands/files.rs
@@ -25,7 +25,6 @@ use std::{
     io::prelude::*,
     io::{BufRead, BufReader},
     path::{Path, PathBuf},
-    sync::Arc,
     time::{Duration, Instant},
 };
 use tempfile::tempdir;
@@ -363,7 +362,7 @@ fn upload_chunks_in_parallel(
     file_api: Files,
     chunks_paths: Vec<(XorName, PathBuf)>,
     verify_store: bool,
-    progress_bar: Arc<ProgressBar>,
+    progress_bar: ProgressBar,
     show_holders: bool,
 ) -> Vec<JoinHandle<Result<()>>> {
     let mut upload_handles = Vec::new();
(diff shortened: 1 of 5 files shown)


sn_cli/src/subcommands/files.rs
```diff
@@ -137,39 +137,47 @@ pub(crate) async fn files_cmds(
...
+                } else {
+                    println!(
+                        "Skipping file {:?} as it is not valid UTF-8.",
+                        entry.file_name()
+                    );
+                    None
+                }
+            } else {
+                None
+            }
```
b-zee: I prefer an early return for clarity and keeping indentation low. (Check syntax before incorporating.)

```suggestion
            if !entry.file_type().is_file() {
                return None
            }

            if let Some(file_name) = entry.file_name().to_str() {
                Some((file_name.to_string(), entry.into_path()))
            } else {
                println!(
                    "Skipping file {:?} as it is not valid UTF-8.",
                    entry.file_name()
                );
                None
            }
        }
```

RolandSherwin: Thank you! Updated the PR.

b-zee (approved):

PR #912, opened 2023-10-27
handle_cmd and gossip mem tweaks

joshuef:
### Summary generated by Reviewpad on 27 Oct 23 11:03 UTC
This pull request includes three patches:

1. The first patch in the pull request is titled "chore(networking): avoid a replication keys clone". It modifies the `cmd.rs` file in the `sn_networking` directory. The changes involve avoiding the cloning of replication keys by using `into_iter()` instead of `iter()`.

2. The second patch is titled "chore(networking): reduce GossipsubPub event weight, don't send useless bytes". It contains modifications to the `api.rs`, `event.rs`, and `node.rs` files. The changes reduce the weight of the `GossipsubPub` event and eliminate the sending of unnecessary bytes.

3. The third patch with the title "chore(node): make gossipsubpublish take Bytes" modifies the `api.rs`, `cmd.rs`, `lib.rs`, `node.rs`, `put_validation.rs` files. It changes the `msg` parameter in the `publish_on_topic` function from `Vec<u8>` to `Bytes`.

### DIFF
--- a/sn_cli/src/subcommands/gossipsub.rs
+++ b/sn_cli/src/subcommands/gossipsub.rs
@@ -44,7 +44,7 @@ pub(crate) async fn gossipsub_cmds(cmds: GossipsubCmds, client: &Client) -> Resu
             let mut events_channel = client.events_channel();
             while let Ok(event) = events_channel.recv().await {
                 if let ClientEvent::GossipsubMsg { msg, .. } = event {
-                    let msg = String::from_utf8(msg)?;
+                    let msg = String::from_utf8(msg.to_vec())?;
                     println!("New message published: {msg}");
                 }
             }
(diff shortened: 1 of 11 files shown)


sn_node/src/node.rs
```diff
@@ -368,8 +368,10 @@ impl Node {
                     error!("Failed to remove local record: {e:?}");
                 }
             }
-            NetworkEvent::GossipsubMsgReceived { topic, msg }
-            | NetworkEvent::GossipsubMsgPublished { topic, msg } => {
+            NetworkEvent::GossipsubMsgPublished { topic: _ } => {
+                // Nothing to do here
+            }
```
RolandSherwin: We propagated the `msgs published by self` here, since the `spend test` counts the number of `royalty payments` made and if the `genesis node` validates a payment and publishes a `gossip for that payment`, any incoming gossip msg for `that payment` is filiterd out by the genesis node. 

Hence, we propagate any msg that we published to be handled here. 

This was the PR for it https://github.com/maidsafe/safe_network/pull/859

joshuef: Aha, okay I'll revert this.

RolandSherwin (approved):

PR #915, opened 2023-10-30
feat(cli): error out if we do not have enough balance

RolandSherwin:
### Summary generated by Reviewpad on 30 Oct 23 11:29 UTC
This pull request adds a new feature to the command-line interface (CLI). Specifically, it introduces an error-checking mechanism that ensures the user has enough balance before executing a certain action. The patch modifies the `files.rs` file, adding error handling code and error messages for insufficient balance situations.

### DIFF
--- a/sn_cli/src/subcommands/files.rs
+++ b/sn_cli/src/subcommands/files.rs
@@ -225,6 +225,9 @@ async fn upload_files(
     );
 
     let file_api: Files = Files::new(client.clone(), wallet_dir_path.to_path_buf());
+    if file_api.wallet()?.balance().is_zero() {
+        bail!("The wallet is empty. Cannot upload any files! Please transfer some funds into the wallet");
+    }
 
     // Temp folder to hold SE chunks, which is cleaned up automatically once out of scope.
     let temp_dir = tempdir()?;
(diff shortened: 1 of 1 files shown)


sn_cli/src/subcommands/files.rs
```diff
@@ -260,6 +260,11 @@ async fn upload_files(
             .await
         {
             Ok((cost, new_balance)) => (cost, new_balance),
+            Err(ClientError::Transfers(WalletError::Transfer(
```
maqi, line 266: I think within the pay_for_chunks function, we shall have something as bellow to carry out an early check.
To avoid the overall chunk split & initial pamenet calculate & upload attempts ?
```rust
pub async fn pay_for_chunks(&self, chunks: Vec<XorName>) -> Result<(NanoTokens, NanoTokens)> {
    let mut wallet_client = self.wallet()?;
    let old_balance = wallet_client.balance();
    if old_balance.as_nano() < chunks.len() {
        return Err(TransfersError::NotEnoughBalance(old_balance, NanoTokens::from(chunks.len() * 100));
    }
    ......
}
```

maqi (approved):

PR #921, opened 2023-11-01
chore(node): don't log the transfers events

joshuef:
### Summary generated by Reviewpad on 01 Nov 23 08:01 UTC
This pull request updates the code to stop logging transfer events in the NodeEventsChannel.

### DIFF
--- a/sn_node/src/event.rs
+++ b/sn_node/src/event.rs
@@ -13,7 +13,6 @@ use serde::{Deserialize, Serialize};
 use sn_protocol::storage::{ChunkAddress, RegisterAddress};
 use sn_transfers::{Transfer, UniquePubkey};
 use tokio::sync::broadcast;
-
 const NODE_EVENT_CHANNEL_SIZE: usize = 10_000;
 
 /// Channel where users of the public API can listen to events broadcasted by the node.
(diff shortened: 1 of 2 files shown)


sn_node/src/event.rs
```diff
@@ -77,6 +78,43 @@ pub enum NodeEvent {
...
+
+/// Function to calculate hashes of each transfer
+fn calculate_hashes(transfers: &[Transfer]) -> Vec<u64> {
+    transfers
+        .iter()
+        .map(|transfer| {
+            let mut hasher = std::collections::hash_map::DefaultHasher::new();
+            transfer.hash(&mut hasher);
+            hasher.finish()
+        })
```
grumbach: Instead of hashing all transfers together (which looses all valuable information), I think returning a hash of each would make more sense. 
Implementing the Hash trait for `Transfers` and using it in our logging (when we debug print them) would make a good enhancement!

grumbach (changes requested):
I think we should Hash transfers separately or don't print them at all.

sn_node/src/event.rs
joshuef: This is doing them seperately no? This maps each transfer to its hash?

We could/should just add this on Transfers though, good point.

bochaco (approved):

grumbach (approved):
LGTM

PR #925, opened 2023-11-01
refactor(networking): use Entry API for query task

b-zee:
This PR changes the old behavior where query tasks are removed from the
list and later reinserted. By using the Entry API, tasks are borrowed
and later removed only if neccessary.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -38,7 +38,7 @@ use sn_protocol::{
     NetworkAddress, PrettyPrintRecordKey,
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{hash_map::Entry, HashMap, HashSet},
     fmt::{Debug, Formatter},
     num::NonZeroUsize,
 };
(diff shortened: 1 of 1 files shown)


sn_networking/src/event.rs
```diff
@@ -1067,6 +1066,9 @@ impl SwarmDriver {
                             "Early completion with the first copy of chunk {:?}",
                             chunk.name()
                         );
+
+                        // Stop tracking the query task by removing the entry and consume the sender.
+                        let (sender, ..) = entry.remove();
```
maqi: can you make this `entry.remove()` call AFTER the `sender.send` ?
i.e. using the `sender` from the previous `entry.get_mut()` ?
same to the other place of using `entry`

b-zee: That's not possible, `send` consumes `self` (https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.send).

maqi: Ah, ok, make sense

maqi (approved):

PR #927, opened 2023-11-01
chore(log): log the connected peers during peer add

RolandSherwin:
- used by vdash
### Summary generated by Reviewpad on 01 Nov 23 18:03 UTC
This pull request adds logging of the connected peers when a new peer is added to the routing table in the `chore(log)` module. The patch includes changes to the `event.rs` file, where the count of connected peers is cached and logged along with the new peer information.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -831,9 +831,9 @@ impl SwarmDriver {
             } => {
                 event_string = "kad_event::RoutingUpdated";
                 if is_new_peer {
-                    // todo: cache the count as it is frequently used during replication as well.
-                    let connected_peers = self.swarm.connected_peers().count();
-                    info!("New peer added to routing table: {peer:?}, now we have #{connected_peers} connected peers");
+                    self.connected_peers += 1;
+
+                    info!("New peer added to routing table: {peer:?}, now we have #{} connected peers", self.connected_peers);
                     self.log_kbuckets(&peer);
 
                     if self.bootstrap.notify_new_peer() {
@@ -844,6 +844,8 @@ impl SwarmDriver {
                 }
 
                 if old_peer.is_some() {
+                    self.connected_peers -= 1;
+
                     info!("Evicted old peer on new peer join: {old_peer:?}");
                     self.send_event(NetworkEvent::PeerRemoved(peer));
                     self.log_kbuckets(&peer);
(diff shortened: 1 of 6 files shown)


sn_networking/src/event.rs
```diff
@@ -834,7 +834,9 @@ impl SwarmDriver {
             } => {
                 event_string = "kad_event::RoutingUpdated";
                 if is_new_peer {
-                    info!("New peer added to routing table: {peer:?}");
+                    // todo: cache the count as it is frequently used during replication as well.
+                    let connected_peers = self.swarm.connected_peers().count();
+                    info!("New peer added to routing table: {peer:?}, now we have #{connected_peers} connected peers");
```
joshuef: If we want to avoid this being removed again, let's make this a log marker (or whatever the equivalent is now)

RolandSherwin: I've updated the PR to now cache the `connected_peers` at the `network level` and log them using the markers at the `node level`

sn_networking/src/event.rs
```diff
@@ -425,7 +428,11 @@ impl SwarmDriver {
                     .kademlia
                     .remove_peer(&failed_peer_id)
                 {
-                    self.send_event(NetworkEvent::PeerRemoved(*dead_peer.node.key.preimage()));
+                    self.connected_peers -= 1;
```
maqi: for safty, better using `saturating_sub`
same using `saturating_add` if possible

RolandSherwin: Thank you for the suggestion. Updated the PR

sn_networking/src/event.rs
```diff
@@ -831,19 +838,23 @@ impl SwarmDriver {
...
                     if self.bootstrap.notify_new_peer() {
                         info!("Performing the first bootstrap");
                         self.initiate_bootstrap();
                     }
-                    self.send_event(NetworkEvent::PeerAdded(peer));
+                    self.send_event(NetworkEvent::PeerAdded(peer, self.connected_peers));
                 }
 
                 if old_peer.is_some() {
+                    let _ = self.connected_peers.saturating_sub(1);
```
maqi: `saturating_sub` won't change the self value, hence it shall be `self.connected_peers= self.connected_peers.saturating_sub(1);` instead ?

maqi (approved):

PR #928, opened 2023-11-02
feat: keep transfers in mem instead of heavy cashnotes

grumbach:
## Description

Gets rid of unnecessary `CashNote` disk I/O by turning them into `Transfers` (very small) as soon as possible. This should also help alleviate the memory because `CashNotes` are very big to work with unlike `Transfers` which are very small. 

---

### Summary generated by Reviewpad on 02 Nov 23 06:47 UTC

1. In the `offline_transfer.rs` file:
- Replaced `UniquePubkey` with `Transfer` in the type `PaymentDetails`.

2. In the file diff of `transfer.rs`:
- Rewrapped the `Transfers` into encrypted Transfers ready to be sent directly to the beneficiary.
- Added logging and error handling for invalid and missing network royalties payments.

3. In the file diff of `wallet.rs`:
- Modified the `get_payment_transfers` function to retrieve payment transfers instead of payment cash notes from the wallet and updated the corresponding log message.
- Updated the `store_content` function to use a JoinSet instead of a BTreeMap for storing payment information, and updated the log message for pending store cost tasks.
- Updated the `store_content` function to handle payment adjustment and taking into account previous payments instead of storing payment cash notes.
- Updated the `pay_for_records` function to consider existing payment CashNotes and top up if necessary.

4. In the file diff of a separate file:
- Renamed the function `get_payment_cash_notes` to `get_payment_transfers` and updated its logic to return `Transfer` objects instead of `CashNote` objects.
- Refactored the logic in the `get_payment_transfers` function for simplification and readability.
- Renamed the function `adjust_payment_map` to `take_previous_payments_into_account`.
- Updated the logic in the `take_previous_payments_into_account` function to subtract previous payments from the payments about to be made.
- Fixed a typo in a log message.
- Updated the test case to reflect the code changes.

These changes aim to improve the handling, processing, and clarity of the code related to payments, transfers, and network royalties. Let me know if you need more information.

### DIFF
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -414,8 +414,8 @@ impl Node {
         Ok(CmdOk::StoredSuccessfully)
     }
 
-    /// Gets CashNotes out of a Payment, this includes network verifications of the Transfer.
-    /// Return the CashNotes corresponding to our wallet and the trasfers corresponding to network royalties payment.
+    /// Gets CashNotes out of a Payment, this includes network verifications of the Transfers
+    /// Rewraps the royalties transfers into encrypted Transfers ready to be sent directly to the beneficiary
     async fn cash_notes_from_payment(
         &self,
         payment: &Vec<Transfer>,
@@ -429,7 +429,7 @@ impl Node {
 
         for transfer in payment {
             match transfer {
-                Transfer::Encrypted(_) if cash_notes.is_empty() => match self
+                Transfer::Encrypted(_) => match self
                     .network
                     .verify_and_unpack_transfer(transfer, wallet)
                     .await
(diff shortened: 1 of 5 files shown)


sn_node/src/put_validation.rs
```diff
@@ -438,35 +438,41 @@ impl Node {
                     Err(ProtocolError::FailedToDecypherTransfer) => continue,
                     // transfer invalid
                     Err(e) => return Err(e),
-                    // transfer ok
-                    Ok(cns) => cash_notes = cns,
+                    // transfer ok, add to cash_notes and continue as more transfers might be ours
+                    Ok(cns) => cash_notes.extend(cns),
                 },
-                Transfer::Encrypted(_) => continue, // we already have our cash notes, so we can skip this one
```
grumbach, line 444: There might be multiple Transfers for us! So we can't stop at the first one!

sn_transfers/src/transfers/transfer.rs
```diff
@@ -54,50 +53,6 @@ impl std::fmt::Debug for Transfer {
...
-
-        let mut transfers = Vec::new();
-        for (recipient, cashnote_redemptions) in cashnote_redemptions_map {
-            let t = if recipient == *crate::NETWORK_ROYALTIES_PK {
-                // create a network royalties transfer type
-                Self::NetworkRoyalties(cashnote_redemptions.clone())
-            } else {
-                Self::create(cashnote_redemptions, recipient)
-                    .map_err(|_| Error::CashNoteRedemptionEncryptionFailed)?
-            };
```
grumbach, line 94: This logic would better be kept out of Transfer lib

grumbach: I spent two days hunting for a bug due to the fact that this logic was only in
`transfers_from_cash_notes`
but not in
`transfers_from_cash_note`

> I think this is `Network` logic, not `Transfer` logic. 

Removed it and added a more explicit function for royalties to be used upstream: `royalties_transfers_from_cash_note`, also helps make the upstream code clearer

sn_transfers/src/wallet/local_store.rs
```diff
@@ -227,24 +227,17 @@ impl LocalWallet {
...
 
     /// Return the payment cash_note ids for the given content address name if cached.
-    pub fn get_payment_cash_notes(&self, name: &XorName) -> Vec<CashNote> {
-        let ids = self.get_payment_unique_pubkeys_and_values(name);
-        // now grab all those cash_notes
-        let mut cash_notes: Vec<CashNote> = vec![];
-
-        if let Some(ids) = ids {
-            for (id, _main_pub_key, _value) in ids {
-                if let Some(cash_note) = load_created_cash_note(id, &self.wallet_dir) {
```
grumbach, line 237: No more heavy disk I/O fetching cashnotes!

sn_node/src/put_validation.rs
joshuef: But the multiple are all picked out by `verify_and_unpack_transfer` no?

grumbach: This unpacks only one of the multiple transfers! Each transfer might contain multiple CashNotes. We want all the cashnotes from all the transfers that are ours!

joshuef (approved):

PR #950, opened 2023-11-08
fix: stop bootstrap once connected to enough peers

maqi:
### Summary generated by Reviewpad on 08 Nov 23 17:19 UTC
This pull request fixes an issue where the bootstrap process would continue even after connecting to enough peers. The fix introduces a check to stop the bootstrap if the `stop_bootstrapping` flag is set or if the number of connected peers is greater than a predefined value (`K_VALUE`). Additionally, the logging message has been updated to include the number of peers connected and the status of the `stop_bootstrapping` flag.

### DIFF
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -15,7 +15,7 @@ pub(crate) const BOOTSTRAP_INTERVAL: Duration = Duration::from_secs(5);
 
 /// Every BOOTSTRAP_CONNECTED_PEERS_STEP connected peer, we step up the BOOTSTRAP_INTERVAL to slow down bootstrapping
 /// process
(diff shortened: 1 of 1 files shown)


loziniak:
The functionality is already in *sn_client/src/api.rs* `NetworkEvent::PeerAdded` handler, why does it need to be duplicated?

joshuef (approved):

maqi:
> The functionality is already in _sn_client/src/api.rs_ `NetworkEvent::PeerAdded` handler, why does it need to be duplicated?

that was for `client` , and this is for `node`.
anyway, we decided to `enlength the interval` instead of `stop it totally`

loziniak:
I see, thanks for explanation.

PR #959, opened 2023-11-13
fix: avoid infinite looping on verification during upload

maqi:
### Summary generated by Reviewpad on 13 Nov 23 10:21 UTC
This pull request fixes an issue where there was an infinite loop during the verification process of file uploads. The code now keeps track of the number of attempts and terminates the loop after a certain point to avoid infinite looping. Additionally, a message is printed and the verification is terminated if there is an unrecoverable issue.


grumbach (approved):
LGTM

grumbach (approved):
LGTM

PR #966, opened 2023-11-14
feat(networking): drop excessive AddKeysToReplicationFetcher cmds

joshuef:
### Summary generated by Reviewpad on 14 Nov 23 11:53 UTC
This pull request modifies the `sn_networking` crate by dropping excessive `AddKeysToReplicationFetcher` commands when the `SwarmCmd` channel is full. If the capacity is 0 and the command is `AddKeysToReplicationFetcher`, the command is safely dropped and logged, with the intention of reattempting it later. Otherwise, an error is logged indicating that the `SwarmCmd` channel is full and awaiting capacity to send the command. The change also spawns a task to send the `SwarmCmd` asynchronously to keep the function synchronized.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -633,14 +633,25 @@ impl Network {
 
         let cmd_sender = self.swarm_cmd_sender.clone();
 
-        // Spawn a task to send the SwarmCmd and keep this fn sync
-        let _handle = tokio::spawn(async move {
-            if capacity == 0 {
+        if capacity == 0 {
+            if matches!(cmd, SwarmCmd::AddKeysToReplicationFetcher { .. }) {
+                // we can safely drop AddKeysToReplicationFetcher
+                // it should be reattempted in a few seconds and if we can cope we'll do it.
+                warn!(
+                    "SwarmCmd channel is full. Dropping AddKeysToReplicationFetcher: {:?}",
+                    cmd
+                );
+                return Ok(());
+            } else {
                 error!(
                     "SwarmCmd channel is full. Await capacity to send: {:?}",
                     cmd
                 );
             }
+        }
+
+        // Spawn a task to send the SwarmCmd and keep this fn sync
+        let _handle = tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


sn_networking/src/lib.rs
```diff
@@ -633,14 +633,25 @@ impl Network {
...
+            } else {
                 error!(
                     "SwarmCmd channel is full. Await capacity to send: {:?}",
                     cmd
                 );
             }
+        }
+
+        // Spawn a task to send the SwarmCmd and keep this fn sync
+        let _handle = tokio::spawn(async move {
```
maqi, line 654: I think the whole `capacity check and send` shall be in a spawned thread.
i.e. the whole function will be like:
```rust
    fn send_swarm_cmd(&self, cmd: SwarmCmd) -> Result<()> {
        let cmd_sender = self.swarm_cmd_sender.clone();

        // Spawn a task to send the SwarmCmd and keep this fn sync
            let capacity = cmd_sender.capacity();
           if capacity == 0 {
               if matches!(cmd, SwarmCmd::AddKeysToReplicationFetcher { .. }) {
                    // we can safely drop AddKeysToReplicationFetcher
                    // it should be reattempted in a few seconds and if we can cope we'll do it.
                    warn!(
                        "SwarmCmd channel is full. Dropping AddKeysToReplicationFetcher: {:?}",
                        cmd
                    );
                    return Ok(());
                } else {
                    error!(
                        "SwarmCmd channel is full. Await capacity to send: {:?}",
                        cmd
                    );
                }
            }

            if let Err(error) = cmd_sender.send(cmd).await {
                error!("Failed to send SwarmCmd: {}", error);
            }
        });

        Ok(())
    }
```

maqi (changes requested):

sn_networking/src/lib.rs
joshuef: I wondered that, but i thought this would avoid what (im not sure how much) work there is in spawning threads there?

maqi (approved):

PR #973, opened 2023-11-15
Reduce AddKeysToReplicationFetch handling time

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -14,7 +14,7 @@ use crate::{
 };
 use bytes::Bytes;
 use libp2p::{
-    kad::{store::RecordStore, Quorum, Record, RecordKey},
+    kad::{store::RecordStore, Quorum, Record, RecordKey, K_VALUE},
     swarm::dial_opts::DialOpts,
     Multiaddr, PeerId,
 };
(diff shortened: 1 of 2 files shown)


sn_networking/src/cmd.rs
```diff
@@ -275,9 +279,39 @@ pub struct SwarmLocalState {
...
+        }
+
+        let (time_stamp, existing_keys) = self
+            .seen_replications
+            .entry(holder)
+            .or_insert((Instant::now(), Default::default()));
+        incoming_keys.retain(|k, v| match existing_keys.get(k) {
+            Some(existing_v) => v != existing_v,
+            None => true,
+        });
```
RolandSherwin: I think we iterate through the entire `incoming_key` even if `holder` does not exist. Maybe we can match on the `Entry::Occupied` to perform `retain` operation and ignore when `Entry::Vacant`? Something like this,


```rust
let entry = self.seen_replications.entry(holder);
if let Entry::Occupied(mut occupied_entry) = entry {
    let (time_stamp, existing_keys) = occupied_entry.get_mut();
    incoming_keys.retain(|k, v| match existing_keys.get(k) {
        Some(existing_v) => v != existing_v,
        None => true,
    });
    existing_keys.extend(incoming_keys.clone());
    *time_stamp = Instant::now();
} else {
    entry.or_insert((Instant::now(), incoming_keys));
}
```

RolandSherwin (approved):

PR #974, opened 2023-11-15
chore: spawn threads to send back result from swarm_driver

maqi:
## Description

reviewpad:summary


maqi:
closed due to not having noticable improvement

PR #975, opened 2023-11-16
feat: massive cleaning to prepare for quotes

grumbach:
## Description

- removes repayments
- adapts `StoreCost` and `PaymentDetails` to match Pay One node
- fix CashNote mixup bug that can happen when a single node receives 2 different payments for 2 different chunks.
- minor refactors here and there for readability

---

reviewpad:summary

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -240,7 +240,7 @@ impl Network {
     pub async fn get_store_costs_from_network(
         &self,
         record_address: NetworkAddress,
-    ) -> Result<Vec<(MainPubkey, NanoTokens)>> {
+    ) -> Result<(MainPubkey, NanoTokens)> {
         // The requirement of having at least CLOSE_GROUP_SIZE
         // close nodes will be checked internally automatically.
         let mut close_nodes = self.get_closest_peers(&record_address, true).await?;
@@ -735,35 +735,28 @@ impl Network {
     }
 }
 
-/// Given `all_costs` it will return the CLOSE_GROUP majority cost.
-#[allow(clippy::result_large_err)]
+/// Given `all_costs` it will return the lowest cost.
 fn get_fees_from_store_cost_responses(
     mut all_costs: Vec<(MainPubkey, NanoTokens)>,
-) -> Result<Vec<(MainPubkey, NanoTokens)>> {
-    // TODO: we should make this configurable based upon data type
-    // or user requirements for resilience.
-
-    // Rigth now we only take one quote and pay one node.
-    let desired_quote_count = 1;
-
+) -> Result<(MainPubkey, NanoTokens)> {
     // sort all costs by fee, lowest to highest
     // if there's a tie in cost, sort by pubkey
-    all_costs.sort_by(|(pub_key_a, cost_a), (pub_key_b, cost_b)| {
-        match cost_a.partial_cmp(cost_b) {
-            Some(std::cmp::Ordering::Equal) => pub_key_a.cmp(pub_key_b),
-            other => other.unwrap_or(std::cmp::Ordering::Equal),
-        }
-    });
-
-    // get the lowest desired_quote_counts of all_costs
-    all_costs.truncate(desired_quote_count);
-
-    info!(
-        "Final fees calculated as: {all_costs:?}, from: {:?}",
-        all_costs
+    all_costs.sort_by(
+        |(pub_key_a, cost_a), (pub_key_b, cost_b)| match cost_a.cmp(cost_b) {
+            std::cmp::Ordering::Equal => pub_key_a.cmp(pub_key_b),
+            other => other,
+        },
     );
 
-    Ok(all_costs)
+    // get the lowest cost
+    trace!("Got all costs, getting lowest: {all_costs:?}");
+    let lowest = all_costs
+        .into_iter()
+        .next()
+        .ok_or(Error::NoStoreCostResponses)?;
+    info!("Final fees calculated as: {lowest:?}");
+
+    Ok(lowest)
 }
 
 /// Verifies if `Multiaddr` contains IPv4 address that is not global.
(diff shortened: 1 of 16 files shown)


sn_transfers/src/wallet/local_store.rs
```diff
@@ -78,13 +78,13 @@ impl LocalWallet {
...
     /// Attempts to reload the wallet from disk.
-    pub fn reload_from_disk(&mut self) -> Result<()> {
+    pub fn reload_from_disk_or_recreate(&mut self) -> Result<()> {
         std::fs::create_dir_all(&self.wallet_dir)?;
         // lock and load from disk to make sure we're up to date and others can't modify the wallet concurrently
         trace!("Trying to lock wallet to get available cash_notes...");
-        let exclusive_access = self.lock()?;
+        let _exclusive_access = self.lock()?;
         self.reload()?;
-        self.store(exclusive_access)
```
grumbach, line 87: No need to store (aka write) as `reload` just read it from disk!

sn_networking/src/lib.rs
```diff
@@ -240,7 +240,7 @@ impl Network {
     pub async fn get_store_costs_from_network(
         &self,
         record_address: NetworkAddress,
-    ) -> Result<Vec<(MainPubkey, NanoTokens)>> {
+    ) -> Result<(MainPubkey, NanoTokens)> {
```
grumbach, line 243: Make store cost API return one cost instead of a Vec

sn_transfers/src/wallet/local_store.rs
```diff
@@ -354,36 +354,41 @@ impl LocalWallet {
...
             let node_key = storage_payee.1;
-            let cash_note_for_node = offline_transfer
-                .created_cash_notes
+            let pay_amount = storage_payee.0;
+            let cash_note_for_node = cashnotes_to_use
                 .iter()
                 .find(|cash_note| {
-                    // cash_note.reason() == Hash::hash(xorname) && // NB TODO uncomment and use Reason!!
-                    cash_note.main_pubkey() == &node_key
+                    cash_note.value() == Ok(pay_amount) && cash_note.main_pubkey() == &node_key
```
maqi, line 369: I am not quite sure.
according to your description, the issue is:
when there is a node got paid more than once within the same batch, the current lookup might get it mixed up.

However, as it is targeting the same node, the `storecost pay_amount` quoted will be the same as well in high chance.
Hence, checking the `payment_amount` only won't tell the difference ?

I think the cashnote shall contain the `reason, i.e. which chunk it is paying for`, to make it unqiue.
However, not sure whether it is current provided by cashnote.
If not, shall we introduce it?

sn_transfers/src/wallet/local_store.rs
```diff
@@ -299,111 +302,108 @@ impl LocalWallet {
...
-        for (content_addr, payees) in all_data_payments {
-            for (payee, token) in payees {
-                if let Some(cash_note) =
-                    offline_transfer
-                        .created_cash_notes
-                        .iter()
-                        .find(|cash_note| {
-                            cash_note.main_pubkey() == &payee
-                                && !used_cash_notes.contains(&cash_note.unique_pubkey().to_bytes())
-                        })
```
grumbach, line 373: This code here checks only the `main_pubkey`, so if one node has two chunks payements for them in this batch we might mix the outputs!

Say we have two records of two different prices somehow:
chunk1, price1
chunk2, price2

What prevents us from using the wrong one?

sn_transfers/src/wallet/local_store.rs
grumbach: I agree using the reason would be ideal!As we seemingly don't use Reason in the code, I used the price. 
If the price is identical then the node can't tell the difference as the payment is for them anyways so it works. 

I tried looking into re-establishing reasons. But turns out it's a deeper hole than I thought. 
See, reasons are attached to a `Transaction`. When we batch payments, we create one single transaction for all the chunks so we can't have one reason for all. 
- One option is to use a merkle tree like we had. The issue is that we have to send/store it along as proof. This is reason enough not to do so. 
- Another option which is better is to make one Transaction per chunk. 

That second option has qualities:
- generated CashNotes will be much smaller!
- uploaded Spends will be much smaller!
- each record will have its payment Tx + reason

And drawbacks:
- we have one change CashNote per Tx, so one per chunk
- parallelisation will suffer greatly as if I use one CashNote in a Tx I can't use it for another Tx, I'll have to wait to use the change CashNote once it's valid (first Tx's Spends uploaded). 

I think it's a great feature that we could implement, but later, when we've adapted the way we parallelise.

sn_transfers/src/wallet/local_store.rs
```diff
@@ -521,39 +521,6 @@ impl LocalWallet {
     pub fn derive_key(&self, derivation_index: &DerivationIndex) -> DerivedSecretKey {
         self.key.derive_key(derivation_index)
     }
-
-    /// Lookup previous payments and subtract them from the payments we're about to do.
-    /// In essence this will make sure we don't overpay.
-    pub fn take_previous_payments_into_account(
-        &self,
-        payment_map: &mut BTreeMap<XorName, Vec<(MainPubkey, NanoTokens)>>,
-    ) {
```
grumbach, line 530: Remove this function, it's a first step in removing repayments, others will need to wait for payment quotes!

sn_transfers/src/wallet/local_store.rs
maqi: right, yeah I understand it could have big affects to using `reason` for different chunk, to differentiate cash_notes.
But, it could be reason that we have certain `looping of spend got rejected`, and we may have to impl it or some simiar.

I am OK with using `price` as a quick fix.
BUT, I do think this issue shall be raised and get fully discussed and resolved befor next public testnet.

maqi (approved):

PR #982, opened 2023-11-20
Reduce unhandled NetworkEvents

joshuef:
### Summary generated by Reviewpad on 20 Nov 23 10:37 UTC

1. In `sn_networking/src/cmd.rs`:
   - Modified the handling of a request sent to itself by the swarm driver. If the request is a `Query` type, it sends a `NetworkEvent::QueryRequestReceived`. Otherwise, it logs a message stating that a replicate command to self was received and it is ignored.

2. In `sn_networking/src/driver.rs`:
   - Updated the configuration of the gossipsub behavior to disable sending to all peers subscribed to a topic (`flood_publish(false)`).

3. In `sn_networking/src/event.rs`:
   - Renamed `RequestReceived` to `CmdRequestReceived` and added a new variant `QueryRequestReceived` in the `NetworkEvent` enum.

4. In `sn_networking/src/swarm_driver.rs`:
   - Modified the handling of a request received from a peer. If the request is a `Cmd` type, it sends a replicate OK response to the peer and handles the command separately. If the request is a `Query` type, it sends a `NetworkEvent::QueryRequestReceived`.

5. In `sn_node/src/node.rs`:
   - Updated the handling of network events in the `handle_response` function. If the response is a replicate OK response, it logs a warning as it should have been handled earlier.
   - Updated the handling of network events in the `handle_query` and `handle_node_cmd` functions. Instead of returning a response, it now directly sends the response using the `send_response` function.

6. In `sn_transfers/src/cashnotes/signed_spend.rs`:
   - Added the `custom_debug` attribute to the `Spend` struct.

These changes improve the handling of requests and responses in the code.

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -545,6 +545,8 @@ impl SwarmDriver {
                             channel: MsgResponder::FromSelf(sender),
                         });
                     } else {
+                        // We should never receive a Replicate request from ourselves.
+                        // we already hold this data if we do... so we can ignore
                         trace!("Replicate cmd to self received, ignoring");
                     }
                 } else {
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -538,12 +538,15 @@ impl SwarmDriver {
                 // be handled.
                 // `self` then handles the request and sends a response back again to itself.
                 if peer == *self.swarm.local_peer_id() {
-                    trace!("Sending request to self");
-
-                    self.send_event(NetworkEvent::RequestReceived {
-                        req,
-                        channel: MsgResponder::FromSelf(sender),
-                    });
+                    trace!("Sending query request to self");
+                    if let Request::Query(query) = req {
+                        self.send_event(NetworkEvent::QueryRequestReceived {
+                            query,
+                            channel: MsgResponder::FromSelf(sender),
+                        });
+                    } else {
+                        trace!("Replicate cmd to self received, ignoring");
+                    }
                 } else {
                     let request_id = self
                         .swarm
@@ -552,6 +555,8 @@ impl SwarmDriver {
                         .send_request(&peer, req);
                     trace!("Sending request {request_id:?} to peer {peer:?}");
                     let _ = self.pending_requests.insert(request_id, sender);
+
+                    trace!("Pending Requests now: {:?}", self.pending_requests.len());
                 }
             }
             SwarmCmd::SendResponse { resp, channel } => match channel {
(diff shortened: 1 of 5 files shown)


sn_networking/src/cmd.rs
```diff
@@ -538,12 +538,15 @@ impl SwarmDriver {
...
-                        channel: MsgResponder::FromSelf(sender),
-                    });
+                    trace!("Sending query request to self");
+                    if let Request::Query(query) = req {
+                        self.send_event(NetworkEvent::QueryRequestReceived {
+                            query,
+                            channel: MsgResponder::FromSelf(sender),
+                        });
+                    } else {
+                        trace!("Replicate cmd to self received, ignoring");
```
maqi, line 550: the log is bit confusing.
As I understand, it is a `non-query request to self, which can be ignored`, maybe due to `replicate` ?

sn_networking/src/driver.rs
```diff
@@ -394,6 +394,8 @@ impl NetworkBuilder {
 
         // Gossipsub behaviour
         let gossipsub_config = libp2p::gossipsub::ConfigBuilder::default()
+            // disable sending to ALL_PEERS subscribed to a topic, which is the default behaviour
+            .flood_publish(false)
```
maqi: this may cause a `subscriber` miss a msg?
it it is not known by ALL publisher ?

with the `flood` disabled`, will `forwarding` still get undertaken?

sn_networking/src/event.rs
```diff
@@ -487,10 +495,29 @@ impl SwarmDriver {
...
                 } => {
                     trace!("Received request {request_id:?} from peer {peer:?}, req: {request:?}");
-                    self.send_event(NetworkEvent::RequestReceived {
-                        req: request,
-                        channel: MsgResponder::FromPeer(channel),
-                    })
+                    // if the request is replication, we can handle it and send the OK response here,
+                    // as we send that regardless of how we handle the request as its unimportant to the sender.
+                    match request {
+                        Request::Cmd(cmd) => {
```
maqi: sounds like the `Replicate` is the only `Request::Cmd` to be received across nodes?
I'd prefer narrow down then enum to be `Cmd::Replicate` only, to avoid later on it mess-up with other new work flow ?

sn_networking/src/cmd.rs
joshuef: ah yeh, we'd be asking ourself to replicate data which we're already holding? So can be ignored

maqi (approved):

PR #990, opened 2023-11-21
feat(sn_networking): no gossip for clients via Toggle

joshuef:
### Summary generated by Reviewpad on 21 Nov 23 09:26 UTC
This pull request adds a feature to the sn_networking module. Specifically, it modifies the behavior of the Gossipsub protocol for clients. Gossiping is now disabled for clients via a Toggle. Additionally, the patch refactors the code related to the Gossipsub behavior and adds conditionals to handle different scenarios. The patch also includes various code cleanups and optimizations.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -59,7 +59,7 @@ impl Client {
     pub async fn new(
         signer: SecretKey,
         peers: Option<Vec<Multiaddr>>,
-        joins_gossip: bool,
+        enable_gossip: bool,
         connection_timeout: Option<Duration>,
     ) -> Result<Self> {
         // If any of our contact peers has a global address, we'll assume we're in a global network.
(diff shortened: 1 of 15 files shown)


b-zee (approved):

maqi (approved):

sn_client/src/api.rs
```diff
@@ -78,7 +79,8 @@ impl Client {
         #[cfg(feature = "open-metrics")]
         network_builder.metrics_registry(Registry::default());
 
-        let (network, mut network_event_receiver, swarm_driver) = network_builder.build_client()?;
+        let (network, mut network_event_receiver, swarm_driver) =
+            network_builder.build_client(joins_gossip)?;
```
RolandSherwin: Just a small nit, but can we add `enable_gossipsub` into `NetworkBuilder`? And let it be `false` by default and enabled just for nodes?

PR #991, opened 2023-11-21
feat: retry at the record level, remove all other retries, report errors

grumbach:
## Description

Leverage the benefits of `PaymentQuote` for the client side. 

- retry at the record put level
    - they work for all records types (Chunks, Spends, Registers)
    - retries can be done asynchronously instead of after as they happen during parallel async upload routines
- retry at the data address payment
    - they work for all records types (Chunks, Spends, Registers)
- remove unclear retries for each record kind, and at multiple levels
- report all errors as some where swallowed or not used
    - print them at the end in the cli so users have more actionable failures (they know if it's network, fs, or wallet related)

---

### Summary generated by Reviewpad on 21 Nov 23 10:25 UTC
This pull request introduces a new constant `PUT_RECORD_RETRIES` which specifies the number of attempts to put a record. The `put_record` method has been updated to retry putting the record `PUT_RECORD_RETRIES` times if verification is enabled. Each retry is accompanied by a random wait duration between 1.5 seconds and 5 seconds. This change aims to improve the reliability of storing records in the network.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -13,7 +13,7 @@ pub(crate) use chunk_manager::ChunkManager;
 use bytes::Bytes;
 use clap::Parser;
 use color_eyre::{
-    eyre::{bail, eyre, Error},
+    eyre::{bail, eyre},
     Help, Result,
 };
 use indicatif::{ProgressBar, ProgressStyle};
@@ -23,7 +23,6 @@ use sn_protocol::storage::{Chunk, ChunkAddress};
 use sn_transfers::{Error as TransfersError, NanoTokens, WalletError};
 use std::{
     collections::BTreeSet,
-    fs,
     io::prelude::*,
     io::{BufRead, BufReader},
     path::{Path, PathBuf},
@@ -31,12 +30,6 @@ use std::{
 };
 use tokio::task::JoinHandle;
 use xor_name::XorName;
-
-/// Max amount of times to retry uploading a chunk
-const MAX_CHUNK_PUT_RETRIES: usize = 10;
-/// Max amount of times to retry verification.
-const MAX_VERIFICATION_RETRIES: usize = 5;
-const CHUNK_PUT_RETRY_WAIT_S: u64 = 2;
 
 #[derive(Parser, Debug)]
 pub enum FilesCmds {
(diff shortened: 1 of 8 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -213,32 +203,56 @@ async fn upload_files(
...
             progress_bar.clone(),
             show_holders,
         ))
         .await
         {
-            let _ = result?;
+            let upload_result = join_result?;
+            if let Err(error) = upload_result {
+                error!("Failed to upload a batch: {error}");
+                recorded_upload_errors.push(error);
```
grumbach, line 237: This `result` is in fact a `Result<Result<_>>`, and the inner one was ignored before and we would rely on verification to check that it failed. In the case we had `verify_store` as `false`, then this error would remain swallowed and the marked as successful below!

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -251,40 +265,7 @@ async fn upload_files(
...
-                batch_size,
-                show_holders,
-            )
-            .await?;
-        }
-    } else {
-        // for safe measure, mark all as paid too
-        chunk_manager.mark_paid_all();
-        chunk_manager.mark_verified_all();
-    }
```
grumbach, line 297: If we didn't have verification on, then the swallowed errors would just be considered as success! With no way for the users to even manually retry as the chunks are marked as OK even though they might not!

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -213,32 +203,56 @@ async fn upload_files(
...
         .await
         {
-            let _ = result?;
+            let upload_result = join_result?;
+            if let Err(error) = upload_result {
+                error!("Failed to upload a batch: {error}");
+                recorded_upload_errors.push(error);
+            } else {
+                chunk_manager.mark_paid(chunks_batch.iter().map(|(xor, _)| *xor));
+                chunk_manager.mark_verified(chunks_batch.iter().map(|(xor, _)| *xor));
```
grumbach, line 240: Since we verify during upload, those 2 functions could be merged as one

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -213,32 +203,56 @@ async fn upload_files(
...
+    let failed_payments = chunk_manager.get_unpaid_chunks();
+    let failed_uploads = chunk_manager.get_paid_chunks();
+    let failed_payments_len = failed_payments.len();
+    let failed_uploads_len = failed_uploads.len();
+    let total_failures = failed_payments_len + failed_uploads_len;
+    if total_failures != 0 {
+        println!("**************************************");
+        println!("*              Failures              *");
+        println!("**************************************");
+        info!("Failed to pay for {failed_payments_len} chunks and failed to upload {failed_uploads_len} chunks.");
```
grumbach, line 256: If we really couldn't upload a chunk, we report errors here, and print them out. 
This way users don't need to dig the logs to know what went wrong. 
Note that we only report errors if we REALLY couldn't put a chunk after all retries!

joshuef (approved):

joshuef:
markign do not merge until ive had a chance to run this, code lgtm though!

PR #992, opened 2023-11-22
feat(cli): add download batch-size option

joshuef:
Sets default to 1/4 of upload

### Summary generated by Reviewpad on 22 Nov 23 09:07 UTC
This pull request adds a new option for the batch size in the command line interface (CLI) for parallel handling of chunks during payment and upload processing. Additionally, it also adds a batch size option for parallel downloading. The default value for the download batch size is set to one-fourth of the upload batch size. The changes are made in the `sn_cli/src/subcommands/files/mod.rs` and `sn_client/src/file_apis.rs` files. The patch adds new command line options and modifies the `files_cmds` function to handle the new batch size options. It also adds new functions `download_files`, `download_file`, and modifies the `verify_and_repay_if_needed_once` function to accommodate the new batch size option for parallel downloading.

### DIFF
--- a/sn_client/src/file_apis.rs
+++ b/sn_client/src/file_apis.rs
@@ -70,6 +70,7 @@ impl Files {
         address: ChunkAddress,
         downloaded_file_path: Option<PathBuf>,
         show_holders: bool,
+        batch_size: usize,
     ) -> Result<Option<Bytes>> {
         let chunk = match self.client.get_chunk(address, show_holders).await {
             Ok(chunk) => chunk,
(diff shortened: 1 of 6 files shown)


sn_client/src/file_apis.rs
```diff
@@ -426,7 +428,10 @@ impl Files {
                     return Ok(data_map);
                 }
                 DataMapLevel::Additional(data_map) => {
-                    let serialized_chunk = self.read_all(data_map, None, false).await?.unwrap();
+                    let serialized_chunk = self
+                        .read_all(data_map, None, false, BATCH_SIZE)
```
RolandSherwin: Can we pass the `batch_size` arg here instead of using the const?

RolandSherwin (approved):

PR #1001, opened 2023-11-23
WIP feat: query specific kbuckets for bootstrap

maqi:
### Summary generated by Reviewpad on 23 Nov 23 13:23 UTC
This pull request consists of two patches. 

In the first patch, the code is modified to query specific k-buckets for bootstrap. A new function `trigger_network_discovery` is added to initiate the network discovery process. This function triggers the network discovery by querying the closest peers for the given addresses. The function is called for each address in the `network_discovery_candidates` list. This change aims to improve network discovery accuracy.

In the second patch, the code is refactored to replace the bootstrap process with querying specific k-buckets. The `initiate_bootstrap` function is removed, and the `trigger_network_discovery` function is called instead. The `ContinuousBootstrap` struct is modified to remove the `is_ongoing` field and replace it with `last_bootstrap_triggered` field to track the time of the last bootstrap trigger. The `completed` method is also removed. Additionally, the code related to informing the bootstrap process about completion is removed, as it is no longer needed.

Overall, these patches aim to improve the bootstrap process by replacing it with querying specific k-buckets, which results in more accurate network discovery.

### DIFF
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -57,6 +57,18 @@ impl SwarmDriver {
             }
         };
     }
+
+    pub(crate) fn trigger_network_discovery(&mut self) {
+        // The query is just to trigger the network discovery,
+        // hence no need to wait for a result.
+        for addr in &self.network_discovery_candidates {
+            let _ = self
+                .swarm
+                .behaviour_mut()
+                .kademlia
+                .get_closest_peers(addr.as_bytes());
+        }
+    }
 }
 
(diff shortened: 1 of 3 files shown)


sn_networking/src/bootstrap.rs
```diff
@@ -57,6 +57,18 @@ impl SwarmDriver {
             }
         };
     }
+
+    pub(crate) fn query_specific_kbucket(&mut self) {
```
joshuef: rename as `trigger_network_discovery` ?

joshuef (approved):

PR #1002, opened 2023-11-23
chore(networking): bootstrap more frequently, and take buncket count/…

joshuef:
…densisty into account

### Summary generated by Reviewpad on 23 Nov 23 12:18 UTC
This pull request includes changes to the networking module. It improves the bootstrapping process by bootstrapping more frequently and taking into account the count and density of buckets. It also adds information about the number of peers in each bucket and uses these statistics to determine whether to continue bootstrapping or slow down the process. Additionally, it sets desirable bucket count and desired average bucket size to improve the efficiency of the bootstrapping process. The patch also includes some code refactoring and logging improvements.

### DIFF
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -7,6 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::SwarmDriver;
+use libp2p::kad::K_VALUE;
 use std::time::{Duration, Instant};
 use tokio::time::Interval;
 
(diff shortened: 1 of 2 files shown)


sn_networking/src/bootstrap.rs
```diff
@@ -124,23 +138,39 @@ impl ContinuousBootstrap {
...
             return (should_bootstrap, Some(new_interval));
         }
 
+        // We want to aim for a decent average bucket size.
+        // we want the average bucket size to be at least K_VALUE / 3
+        if (peers_in_rt / bucket_stats.len()) < DESIRED_AVG_BUCKET_SIZE {
+            info!(
+                "The average bucket is < {DESIRED_AVG_BUCKET_SIZE} peers. Continuing to bootstrap."
+            );
+            return (should_bootstrap, None);
```
RolandSherwin, line 167: The code below where we `increment the interval based on the number of peers`, can be triggered prior to this check and could set the interval to a way higher number than desirable.

We should set a new interval inside this check. Something like this
```rust
let interval = if current_interval != AVG_BUCKET_SIZE_INTERVAL {
	let mut new_interval = tokio::time::interval(AVG_BUCKET_SIZE_INTERVAL);
	new_interval.tick().await; // the first tick completes immediately
	Some(new_interval)
} else {
	None
};
return (should_bootstrap, interval)
```

joshuef: wrong commit, updated now

PR #1005, opened 2023-11-23
feat(test): impl routing table test

RolandSherwin:
### Summary generated by Reviewpad on 23 Nov 23 16:57 UTC

1. In the file `rpc_service.rs`:
   - Added a new import statement for `k_buckets_response` in the `safenode_proto` module.
   - Added a new method `k_buckets` for handling the RPC request for `KBucketsRequest`.
   - Updated the `record_addresses` method to return a response containing addresses.

2. In the file `msgs_over_gossipsub.rs`:
   - Added imports for `HashMap` and `HashSet`.
   - Added a new `get_kbuckets` method that returns a map of ilog2 distance of the Kbucket to the peers in that bucket.
   - Updated the `RunningNode` struct implementation with the new method `get_kbuckets`.
   - Added documentation for the new method `get_kbuckets`.
   - Added a call to `self.network.get_kbuckets().await?` in the `get_kbuckets` method.
   - Updated the `subscribe_to_topic` method signature to take a `String` parameter instead of a `&str`.

3. In the file `lib.rs`:
   - Added a new method `get_kbuckets` to the `Network` struct. This method returns a map of ilog2 distance of the Kbucket to the peers in that bucket.
   - The `get_kbuckets` method uses a channel to communicate with the swarm and fetches the map of Kbucket distances. If there is an error in receiving the result, it returns an `InternalMsgChannelDropped` error.

4. In the file ".github/workflows/merge.yml":
   - Renamed the job "verify_data_location" to "verify_data_location_routing_table".
   - Updated the name and run command of the step "Build data location test" to "Build data location and routing table tests".
   - Added a new step "Verify the Routing table of the nodes" with a corresponding run command.
   - Updated the `timeout-minutes` value for the step "Verify the Routing table of the nodes" to 5 minutes.
   - Adjusted the `timeout-minutes` value for the step "Verify the location of the data on the network" to 5 minutes.

5. In the file "req_resp_types.proto":
   - Added the messages `KBucketsRequest`, `KBucketsResponse`, and a nested message `Peers` within `KBucketsResponse`.
   - Updated the existing messages in the file to include the new messages at the appropriate position.

6. In the file "common.rs":
   - Imported `get_all_peer_ids`, `NODE_COUNT`, and `PAYING_WALLET_INITIAL_BALANCE`.
   - Changed the type of `NodeIndex` from `u8` to `u32`.
   - Removed the `get_all_peer_ids` function and its implementation.
   - Updated the `verify_location` function to use the `get_all_peer_ids` function directly rather than importing it.
   - Removed the comment explaining the purpose of the `get_all_peer_ids` function.
   - Updated the comment explaining the purpose of the `CHUNK_COUNT` constant.
   - Removed the `CHURN_COUNT` constant.
   - Updated the type of `NodeIndex` from `u8` to `u32` in the `RecordHolders` type alias.
   - Updated the return type of the `verify_location` function to `Result<()>`.
   - Removed the `get_all_peer_ids` function definition.
   - Updated the `get_all_peer_ids` function call to use a direct URL string rather than formatting it.
   - Removed the `println!` statement in the `get_all_peer_ids` function.
   - Removed the comment explaining the purpose of the `store_chunks` function.
   - Removed the unused `wallet_dir` parameter in the `store_chunks` function.

Please let me know if you need further assistance with this pull request.

### DIFF
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -50,9 +50,10 @@ impl SwarmDriver {
 
     pub(crate) fn trigger_network_discovery(&mut self) {
         let now = Instant::now();
-        // The query is just to trigger the network discovery,
-        // hence no need to wait for a result.
+        // Fetches the candidates and also generates new candidates
         for addr in self.network_discovery.candidates() {
+            // The query_id is tracked here. This is to update the candidate list of network_discovery with the newly
+            // found closest peers. It may fill up the candidate list of closer buckets which are harder to generate.
             let query_id = self
                 .swarm
                 .behaviour_mut()
@@ -63,8 +64,7 @@ impl SwarmDriver {
                 (PendingGetClosestType::NetworkDiscovery, Default::default()),
             );
         }
-        // Refresh the candidate list to not query the same candidates over and over again.
-        self.network_discovery.try_refresh_candidates();
+
         self.bootstrap.initiated();
         debug!("Trigger network discovery took {:?}", now.elapsed());
     }
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -49,16 +49,20 @@ impl SwarmDriver {
     }
 
     pub(crate) fn trigger_network_discovery(&mut self) {
+        let now = Instant::now();
         // The query is just to trigger the network discovery,
         // hence no need to wait for a result.
-        for addr in &self.network_discovery_candidates {
+        for addr in self.network_discovery_candidates.candidates() {
             let _ = self
                 .swarm
                 .behaviour_mut()
                 .kademlia
                 .get_closest_peers(addr.as_bytes());
         }
+        self.network_discovery_candidates
+            .try_generate_new_candidates();
         self.bootstrap.initiated();
+        debug!("Trigger network discovery took {:?}", now.elapsed());
     }
 }
 
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::SwarmDriver;
+use crate::{driver::PendingGetClosestType, SwarmDriver};
 use std::time::{Duration, Instant};
 use tokio::time::Interval;
 
@@ -53,11 +53,15 @@ impl SwarmDriver {
         // The query is just to trigger the network discovery,
         // hence no need to wait for a result.
         for addr in self.network_discovery_candidates.candidates() {
-            let _ = self
+            let query_id = self
                 .swarm
                 .behaviour_mut()
                 .kademlia
                 .get_closest_peers(addr.as_bytes());
+            let _ = self.pending_get_closest_peers.insert(
+                query_id,
+                (PendingGetClosestType::NetworkDiscovery, Default::default()),
+            );
         }
         self.network_discovery_candidates
             .try_generate_new_candidates();
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -52,7 +52,7 @@ impl SwarmDriver {
         let now = Instant::now();
         // The query is just to trigger the network discovery,
         // hence no need to wait for a result.
-        for addr in self.network_discovery_candidates.candidates() {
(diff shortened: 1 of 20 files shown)


maqi (approved):

sn_node/tests/verify_routing_table.rs
```diff
@@ -0,0 +1,112 @@
...
+                    .peers
+                    .clone()
+                    .into_iter()
+                    .map(|peer_bytes| PeerId::from_bytes(&peer_bytes).unwrap())
+                    .collect::<HashSet<_>>();
+                (ilog2, peers)
+            })
+            .collect::<HashMap<_, _>>();
+
+        let current_peer = *all_peers
```
maqi: `all_peers` is a `Vec`, which you can access by `all_peers[node_index - 1]` directly ?

maqi (approved):

sn_networking/src/network_discovery.rs
```diff
@@ -0,0 +1,108 @@
...
+                    let _ = entry.insert(VecDeque::from([candidate]));
+                }
+            }
+        }
+    }
+
+    pub(crate) fn candidates(&self) -> impl Iterator<Item = &NetworkAddress> {
+        self.candidates
+            .values()
+            .filter_map(|candidates| candidates.front())
```
maqi: As each time there is only each bucket's `front` of the candidate to be used,
and there is `generate_new_candidates` to be called each time,
I don't get the reasoning behind why using the `queue of 5` ?

maqi (changes requested):

sn_node/tests/verify_routing_table.rs
RolandSherwin: Updated!

sn_networking/src/network_discovery.rs
RolandSherwin: >As each time there is only each bucket's front of the candidate to be used,
and there is generate_new_candidates to be called each time,

I have updated it to fetch a random candidate each time now. And generate new candidate will update the last element if it's new.

> I don't get the reasoning behind why using the queue of 5 ?

For closer buckets, it might take a while for us to get/gen any new candidate, so cycling among say 5 candidates would be better than just re-using a single candidate while we wait to get a new one? I'm updating the code to make it clearer.

sn_networking/src/bootstrap.rs
```diff
@@ -49,16 +49,24 @@ impl SwarmDriver {
     }
 
     pub(crate) fn trigger_network_discovery(&mut self) {
+        let now = Instant::now();
         // The query is just to trigger the network discovery,
         // hence no need to wait for a result.
-        for addr in &self.network_discovery_candidates {
-            let _ = self
+        for addr in self.network_discovery.candidates() {
+            let query_id = self
```
maqi, line 57: the above comment, shall now be updated as we now insert the query_id ?
or maybe just add one more comment to explain what the `query_id` to be used for?
```rust
        // The query is just to trigger the network discovery,
        // hence no need to wait for a result.
```

sn_networking/src/bootstrap.rs
```diff
@@ -49,16 +49,24 @@ impl SwarmDriver {
...
                 .behaviour_mut()
                 .kademlia
                 .get_closest_peers(addr.as_bytes());
+            let _ = self.pending_get_closest_peers.insert(
+                query_id,
+                (PendingGetClosestType::NetworkDiscovery, Default::default()),
+            );
         }
+        // Refresh the candidate list to not query the same candidates over and over again.
+        self.network_discovery.try_refresh_candidates();
```
maqi: maybe call something like `for addr in self.network_discovery.candidates()` above,
which internally refresh the candidate by itself, will be better ?
i.e. avoid calling seperate `refresh`

sn_networking/src/lib.rs
```diff
@@ -183,6 +184,16 @@ impl Network {
         self.get_closest_peers(key, false).await
     }
 
+    /// Returns the map of ilog2 distance of the Kbucket to the peers in that bucket
```
maqi: the comment is bit confusing.
according to the return type, it seems the returned map is:
the map of the kbuckets table, indexing by the ilog2 distance of the kBucket to self ?

Also, if choosse the map to be `BTreeMap`, buckets will be `oredered` by itself ?

RolandSherwin: Right! Thanks for the tip! Will update it!

maqi (approved):

PR #1007, opened 2023-11-23
feat(cli): peers displayed as list

iancoleman:
## Description

cli would print a vec of peers as a single line making a big blob of unreadable text when every command is run.

This changes it so each peer is printed on a single line.

I'd suggest this is a temporary improvement, and this info should not be displayed with every command. It may be useful to introduce a `safe peers` command or a `--show-peers` flag for anyone who wants to see this info printed.

The peer info is useful sometimes, but almost no regular users will ever need to see this info.

### DIFF
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -87,10 +87,11 @@ async fn main() -> Result<()> {
 
     let bootstrap_peers = parse_peers_args(opt.peers).await?;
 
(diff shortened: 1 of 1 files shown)


joshuef (approved):

joshuef:
> but almost no regular users will ever need to see this info.

happy for a follow up on that front, I agree 👍

PR #1009, opened 2023-11-27
feat(discovery): do not query about buckets that are full

RolandSherwin:
On top of #1005 
## Description

reviewpad:summary


RolandSherwin:
Making connections to new peers help with dead peer detection. Closing this.

PR #1012, opened 2023-11-28
feat(networking): allow port reuse

joshuef:
### Summary generated by Reviewpad on 28 Nov 23 08:49 UTC

The first patch modifies the `sn_networking/src/driver.rs` file. It allows port reuse in the networking module by adding the `port_reuse(true)` configuration option to the TCP transport.

The second patch modifies the `sn_client/Cargo.toml` and `sn_faucet/Cargo.toml` files. It adds the missing quic feature to the dependencies, which allows passing on the flag to node builds.

Overall, these patches add support for port reuse in networking and add the required quic feature for node builds.

### DIFF
--- a/sn_client/Cargo.toml
+++ b/sn_client/Cargo.toml
@@ -14,6 +14,8 @@ version = "0.98.17"
 default=[]
 local-discovery=["sn_networking/local-discovery"]
 open-metrics = ["sn_networking/open-metrics", "prometheus-client"]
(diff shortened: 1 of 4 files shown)


RolandSherwin (approved):

joshuef:
This will break autonat, so abandoning for now. Also doesn't actually impact port counts but for causing more attempted reuse errors and other issues

PR #1016, opened 2023-11-28
Ci update

danieleades:
## Description

- updates unmaintained toolchain actions
- removes redundant toolchain action config
- adds dependabot config for keeping actions up-to-date


joshuef (approved):
Nice cleanup, thanks so much! 🙇 @danieleades

PR #1022, opened 2023-11-29
refactor(cli): remove unpaid/paid distinction from chunk manager

### LINKED ISSUE #997: Since we verify during upload, ChunkManager mark_paid and mark_verified could be merged into one
Since we verify during upload, those 2 functions could be merged as one

_Originally posted by @grumbach in https://github.com/maidsafe/safe_network/pull/991#discussion_r1401904315_

RolandSherwin:
Closes #997 
### Summary generated by Reviewpad on 29 Nov 23 18:13 UTC
This pull request includes changes related to the upload of file chunks. 

- The condition in line 166 has been changed to reflect the use of a new method that checks if there are any chunks to upload.
- The variables `unpaid_chunks_to_upload` and `unpaid_chunks_to_upload_len` have been renamed to `chunks_to_upload` and `chunks_to_upload_len`, respectively.
- The progress bar now shows the number of `chunks_to_upload_len` instead of `unpaid_chunks_to_upload_len`.
- The methods `mark_paid` and `mark_verified` have been replaced with `mark_completed`.
- The variables `failed_payments`, `failed_uploads`, `failed_payments_len`, and `total_failures` have been removed.
- The error message in line 265 has changed to state the number of failed uploads.
- The log messages in lines 272, 276, 287, 291, 300, and 302 have been updated to use `{chunks_to_upload_len}`.

These changes improve the upload process and provide better logging information.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -161,8 +161,8 @@ async fn upload_files(
     let mut chunk_manager = ChunkManager::new(root_dir);
     chunk_manager.chunk_path(&files_path)?;
 
-    // Return early if we have no files to pay/verify
-    if chunk_manager.is_unpaid_chunks_empty() && chunk_manager.is_paid_chunks_empty() {
+    // Return early if we have no chunks to upload
+    if chunk_manager.is_chunks_empty() {
         println!("All files were already uploaded and verified");
         println!("**************************************");
         println!("*          Uploaded Files            *");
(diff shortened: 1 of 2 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -236,28 +236,20 @@ async fn upload_files(
...
-                chunk_manager.mark_paid(chunks_batch.iter().map(|(xor, _)| *xor));
-                chunk_manager.mark_verified(chunks_batch.iter().map(|(xor, _)| *xor));
+                chunk_manager.mark_completed(chunks_batch.iter().map(|(xor, _)| *xor));
             }
         }
     }
     progress_bar.finish_and_clear();
 
     // report errors
-    let failed_payments = chunk_manager.get_unpaid_chunks();
```
RolandSherwin, line 247: I think this would not have worked prior to the changes too as we call `mark_paid and mark_verfied` right after each other.
I don't think we have any way to just know if the payment alone failed?

grumbach: Actually we do, a few lines above, we pay. Only if we fail (after all retries) we bail out.

grumbach: But you're right this code would never run indeed! Good that you removed it!

grumbach: Indeed! Although we might want to keep printing out pay errors.

grumbach: By checking `recorded_pay_errors.len()` instead

RolandSherwin: Ah I see! Updated the PR.

grumbach (approved):
LGTM

PR #1023, opened 2023-11-30
fix: protect against amounts tampering and incomplete spends attack

grumbach:
## Description

Fixes critical security issues in Transfer verification.

### Summary generated by Reviewpad on 30 Nov 23 08:02 UTC
This pull request includes several changes to fix issues related to amounts tampering and incomplete spends attack. The changes made in this patch are as follows:

- In the `sn_node/src/put_validation.rs` file, a change was made to fix the use of `from_unique_pubkey` method.
- In the `sn_transfers/src/cashnotes/builder.rs` file, changes were made to fix the use of `unique_pubkey()` method.
- In the `sn_transfers/src/cashnotes/signed_spend.rs` file, changes were made to verify the validity of the signed spend and to check for tampering of the spend value.
- In the `sn_transfers/src/cashnotes/transaction.rs` file, changes were made to verify the transaction against the inputs spent, including the signed spends.
- In the `sn_transfers/src/error.rs` file, a new error type `InvalidSpendValue` was added to handle cases where the spend value was tampered with.
- In the `sn_transfers/src/wallet/local_store.rs` file, a change was made to fix the use of `insert()` method.

Overall, these changes aim to provide better protection against amounts tampering and incomplete spends attack.

### DIFF
--- a/sn_transfers/src/cashnotes/transaction.rs
+++ b/sn_transfers/src/cashnotes/transaction.rs
@@ -34,8 +34,8 @@ impl Input {
         v
     }
 
-    pub fn unique_pubkey(&self) -> UniquePubkey {
-        self.unique_pubkey
+    pub fn unique_pubkey(&self) -> &UniquePubkey {
+        &self.unique_pubkey
     }
 }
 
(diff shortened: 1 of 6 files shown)


sn_transfers/src/cashnotes/transaction.rs
```diff
@@ -34,8 +34,8 @@ impl Input {
         v
     }
 
-    pub fn unique_pubkey(&self) -> UniquePubkey {
-        self.unique_pubkey
+    pub fn unique_pubkey(&self) -> &UniquePubkey {
+        &self.unique_pubkey
```
grumbach, line 38: Reduce memory footprint of this by returning a reference instead of a copy as it's used all over the verification code

sn_transfers/src/cashnotes/transaction.rs
```diff
@@ -132,44 +132,50 @@ impl Transaction {
...
 
-        // check if we have spends for all inputs
-        if signed_spends.is_empty() {
-            return Err(Error::MissingTxInputs)?;
-        }
-        if signed_spends.len() != self.inputs.len() {
-            return Err(Error::SignedSpendInputLenMismatch {
-                got: signed_spends.len(),
-                expected: self.inputs.len(),
-            });
```
grumbach, line 173: This check 
` if signed_spends.len() != self.inputs.len()`
along with the one L184
`.any(|m| m.unique_pubkey == *signed_spend.unique_pubkey())` 
is flawed as it doesn't prevent one from using multiple identical valid `signed_spends`. 
Eg: say I have input A and B. The spend for B is invalid (someone else's money for example), so I provide twice SignedSpend for A (which is valid). Both checks pass (there are 2 spends, and all of them refer to at least one of the inputs). The Tx is marked as valid.

joshuef (approved):

PR #1024, opened 2023-11-30
fix(faucet): verify genesis spend after some delay

RolandSherwin:
### Summary generated by Reviewpad on 30 Nov 23 10:08 UTC
This pull request includes a fix for the faucet module. The patch adds a delay before verifying the transfer from genesis in order to avoid quorum errors.


RolandSherwin:
The issue occurs way above this delay. Implementing custom verification delays to solve this.

PR #1029, opened 2023-11-30
chore(transfers): tidier debug methods for Transactions

joshuef:
### Summary generated by Reviewpad on 30 Nov 23 20:46 UTC
This pull request includes changes to the code related to cash notes transactions and wallet data payments. 

In the `transaction.rs` file, a new debug method has been added to the `Transaction` struct. This method formats the struct for debugging purposes without printing the entire content, using the `hash` field instead. 

In the `data_payments.rs` file, two fields in the `Payment` struct have been annotated with `#[debug(skip)]`. This means that these fields will not be included when printing the struct for debugging.

### DIFF
--- a/sn_transfers/src/cashnotes/transaction.rs
+++ b/sn_transfers/src/cashnotes/transaction.rs
@@ -65,10 +65,22 @@ impl Output {
     }
 }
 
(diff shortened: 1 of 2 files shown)


joshuef:
rolled into #1044

PR #1039, opened 2023-12-04
feat(cli): Download path is familiar to users

iancoleman:
## Description

Instead of downloading to data_dir which is not a familiar directory to most users, change downloads to be saved to the regular familiar Downloads folder (eg the one most browsers use).

This is especially helpful when cleaning up a disk with low disk space, usually the first place to delete stuff is from the Downloads folder.

### Summary generated by Reviewpad on 04 Dec 23 01:42 UTC
This pull request updates the CLI to use a more familiar download path for files. The patch modifies the `files_cmds` function in the `mod.rs` file to set the download directory to the system's default download directory. It also updates the `download_files` function to use the same download directory for file downloads.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -113,7 +113,8 @@ pub(crate) async fn files_cmds(
                 );
             }
 
-            let file_api: Files = Files::new(client.clone(), root_dir.to_path_buf());
+            let download_dir = dirs_next::download_dir().unwrap_or(root_dir.to_path_buf());
+            let file_api: Files = Files::new(client.clone(), download_dir.clone());
 
             match (file_name, file_addr) {
                 (Some(name), Some(address)) => {
@@ -127,7 +128,7 @@ pub(crate) async fn files_cmds(
                         &file_api,
                         &xor_name,
                         &name,
-                        root_dir,
+                        &download_dir,
                         show_holders,
                         batch_size,
                     )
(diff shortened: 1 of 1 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -112,7 +112,8 @@ pub(crate) async fn files_cmds(
                 );
             }
 
-            let file_api: Files = Files::new(client, root_dir.to_path_buf());
+            let download_dir = dirs_next::download_dir().expect("Download directory is obtainable");
```
joshuef: Can we rework this to avoid a potential panic with the `expect` here? 

That appears to be the issue on the CI runs: https://github.com/maidsafe/safe_network/actions/runs/7102691487/job/19333646110?pr=1039 and https://github.com/maidsafe/safe_network/actions/runs/7102691508/job/19333645696?pr=1039

(Although it's not clear to me why this is failing to find the dir 🤔 , we'd still want to avoid a potential panic in the prod code eitherway)

joshuef:
thanks for this @iancoleman 🙇 

We'll need to remove the potential for a `panic` as tehre is now, and also the `Memcheck` test will need to be updated. (if you search for `$CLIENT_DATA_PATH/downloaded_files` that'll need to be switched up to `~/Downloads` I expect)

iancoleman:
@joshuef thanks for the feedback. I've incorporated those changes and I think this is good to go now.

joshuef (approved):

PR #1045, opened 2023-12-04
feat: watch-only wallet

bochaco:
### Summary generated by Reviewpad on 04 Dec 23 13:39 UTC

- The file `sn_cli/src/subcommands/files/mod.rs` has changes related to the `files_cmds` function, the `Files` constructor, and the `upload_files` function. The `client` parameter of these functions has been changed to a shared reference `&Client`.

- A new file `watch_only.rs` has been added, which includes the implementation of a `WatchOnlyWallet` struct. This struct represents a watch-only wallet in a SAFE Network and has various methods for loading, storing, and manipulating wallet data.

- The file `lib.rs` has changes related to adding the `WatchOnlyWallet` struct to the `wallet` module. This struct is now available for use in the API.

- The file `error.rs` has a new variant `PubKeyMismatch` added to the `Error` enum, with a custom error message that includes the path of the wallet.

- The file `cashnote.rs` has changes related to the `CashNote` struct. A new method `derived_pubkey` has been added, and the existing method `derivation_index` now has a specified return type.

- The file `main.rs` has changes related to the variables `joins_gossip` and `joins_gossipsub`, as well as the function `files_cmds`. These changes indicate updates to the usage of gossipsub and more efficient usage of the `client` object.

- The file `wallet_file.rs` has changes related to the `store_created_cash_notes` function, which now accepts a generic iterator and returns a `Result<()>` instead of `Result<()>`. This allows for more flexibility in accepting different types of iterators.

- The file `error.rs` has changes related to the addition of the `MainPubkeyMismatch` variant to the `Error` enum. This variant handles the scenario where the main public key does not match.

- The file `wallet/mod.rs` has changes related to the usage of the `MainPubkey` type and the addition of functions for storing and retrieving the main public key in the wallet directory.

- The file `wallet/mod.rs` has changes related to the addition of a new module called `watch_only` and the implementation of a `WatchOnlyWallet` struct. It also includes a new method called `balance` in the `KeyLessWallet` struct.

Please review these changes to ensure they are correct and do not introduce any bugs or performance issues. Let me know if you need further assistance.

### DIFF
--- a/sn_transfers/src/wallet/local_store.rs
+++ b/sn_transfers/src/wallet/local_store.rs
@@ -75,7 +75,10 @@ impl LocalWallet {
 
     /// Stores the given cash_notes to the `created cash_notes dir` in the wallet dir.
     /// These can then be sent to the recipients out of band, over any channel preferred.
-    pub fn store_cash_notes_to_disk(&self, cash_notes: &[&CashNote]) -> Result<()> {
+    pub fn store_cash_notes_to_disk<'a, T>(&self, cash_notes: T) -> Result<()>
+    where
+        T: IntoIterator<Item = &'a CashNote>,
+    {
         store_created_cash_notes(cash_notes, self.watchonly_wallet.wallet_dir())
     }
 
@@ -89,11 +92,7 @@ impl LocalWallet {
 
     /// Remove CashNote from available_cash_notes and add it to spent_cash_notes.
     pub fn mark_note_as_spent(&mut self, cash_note_id: UniquePubkey) {
-        self.watchonly_wallet
-            .available_cash_notes_mut()
-            .remove(&cash_note_id);
-        self.watchonly_wallet
-            .insert_spent_cash_notes(&[cash_note_id]);
+        self.watchonly_wallet.mark_notes_as_spent(&[cash_note_id]);
     }
 
     pub fn unconfirmed_spend_requests_exist(&self) -> bool {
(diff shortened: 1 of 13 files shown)


sn_transfers/src/wallet/local_store.rs
```diff
@@ -154,26 +124,14 @@ impl LocalWallet {
...
     pub fn load_from_path(wallet_dir: &Path, main_key: Option<MainSecretKey>) -> Result<Self> {
         std::fs::create_dir_all(wallet_dir)?;
-        let (key, wallet, unconfirmed_spend_requests) = load_from_path(wallet_dir, main_key)?;
-        Ok(Self {
-            key,
-            wallet,
-            wallet_dir: wallet_dir.to_path_buf(),
-            unconfirmed_spend_requests,
-        })
+        Self::load_from_path_and_key(wallet_dir, main_key)
```
grumbach, line 134: This is great!

sn_transfers/src/wallet/local_store.rs
```diff
@@ -195,7 +153,7 @@ impl LocalWallet {
     }
 
     pub fn balance(&self) -> NanoTokens {
-        self.wallet.balance()
+        self.watchonly_wallet.balance()
     }
```
grumbach, line 157: Any chance we can short circuit that one?
We have 4 functions named `balance` they call each other in a deep nested way that makes the code hard to navigate, any chance we can get rid of that `LocalWallet` middle man and directly call your new wallet?

sn_transfers/src/wallet/local_store.rs
```diff
@@ -432,63 +384,17 @@ impl LocalWallet {
...
-
-            self.store_cash_notes_to_disk(vec![cash_note])?;
-        }
-
-        self.store(exclusive_access)?;
-
-        Ok(())
+        self.watchonly_wallet
+            .deposit_and_store_to_disk(received_cash_notes)
     }
```
grumbach: Same here, let's get rid of all that nesting

sn_transfers/src/wallet/local_store.rs
bochaco: You are missing that the `LocalWallet` is a different type of wallet, it's like a hot-wallet (eventually we should rename it) which reuses the impl/features of a watch-only wallet. Nevertheless, a refactoring of the LocalWallet is out of the scope here.

sn_transfers/src/wallet/local_store.rs
bochaco: see comment above, you are missing something

grumbach (approved):
LGTM

PR #1047, opened 2023-12-05
feat: chunk put allow retrying un-get-able chunks

grumbach:
## Description

Depends on #1022, please merge #1022 first!

### Summary generated by Reviewpad on 05 Dec 23 09:36 UTC

- The `verify_uploaded_chunks` method has been removed from the `Files` struct. Instead, the logic of this method has been moved to the `upload_file` method of the same struct, which now calls the `verify_uploaded_chunks` method on the `client` field.
- Several import statements have been added for external modules and the standard library.
- A new method `verify_uploaded_chunks` has been added to the `Client` struct. This method takes a vector of chunk paths and a batch size, and returns a vector of chunks that could not be verified.
- A nested function `get_register_from_record` has been implemented, which converts a record into a signed register.
- The `upload_files` function in the `sn_cli/src/subcommands/files/mod.rs` file has been modified. The changes include adding error handling to check if the wallet is empty before uploading files, initializing the `ChunkManager` with the provided `root_dir`, adding a call to determine the chunk path for the given files path, printing messages for uploaded and verified files, handling different scenarios based on the existence of chunks to upload, and logging and printing payment and upload information.
- Import statements for `io::BufRead` and `io::BufReader` have been removed.
- The variable `child` has been assigned but is unused.

These changes aim to refactor the code, improve the upload process, handle failed chunks, provide better information about the upload and payment details, and remove unused imports and variables.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -166,7 +166,15 @@ async fn upload_files(
     if chunk_manager.is_chunks_empty() {
         // make sure we don't have any failed chunks in those
         let chunks = chunk_manager.already_put_chunks(&files_path)?;
-        let failed_chunks = client.verify_uploaded_chunks(chunks, batch_size).await?;
+        let failed_chunks = client.verify_uploaded_chunks(&chunks, batch_size).await?;
+
+        // mark the non-failed ones as completed
+        chunk_manager.mark_completed(
+            chunks
+                .into_iter()
+                .filter(|c| !failed_chunks.contains(c))
+                .map(|(xor, _)| xor),
+        );
 
         // if none are failed, we can return early
         if failed_chunks.is_empty() {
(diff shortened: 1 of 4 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -159,27 +159,38 @@ async fn upload_files(
...
-        println!("**************************************");
-        for (file_name, addr) in chunk_manager.verified_files() {
-            if let Some(file_name) = file_name.to_str() {
-                println!("\"{file_name}\" {addr:x}");
-                info!("Uploaded {file_name} to {addr:x}");
-            } else {
-                println!("\"{file_name:?}\" {addr:x}");
-                info!("Uploaded {file_name:?} to {addr:x}");
+        // make sure we don't have any failed chunks in those
+        let chunks = chunk_manager.already_put_chunks(&files_path)?;
```
RolandSherwin, line 168: I think this would `create chunks in the artifacts` dir and if they're not marked as completed when `failed_chunks is not empty (L172)`, then during the next run, we'd get resume the chunks at line 162 and we'll try to upload them again (L166 returns false).

RolandSherwin: So should we make ` chunk_manager.already_put_chunks(&files_path)?;` create the `chunks` to a temp dir for verification?

That might avoid complication.

grumbach: Nice catch! Marked the non-failed chunks as completed just below. And if they fail, it's reasonable to keep them around for next time!

RolandSherwin (approved):

PR #1048, opened 2023-12-05
test(node): add min test time for reward listener tests

joshuef:
### Summary generated by Reviewpad on 05 Dec 23 11:31 UTC
This pull request adds a minimum test time for reward listener tests in the `nodes_rewards.rs` file. Instead of always waiting for a fixed duration, the test now calculates the wait time based on the expected royalties. If the expected royalties are 0 or 1, the wait time is set to 20 seconds. Otherwise, the wait time is calculated as 10 seconds per expected royalty.

### DIFF
--- a/sn_node/tests/nodes_rewards.rs
+++ b/sn_node/tests/nodes_rewards.rs
@@ -318,7 +318,10 @@ fn spawn_royalties_payment_listener(
         let mut count = 0;
         let mut stream = response.into_inner();
 
-        let secs = expected_royalties as u64 * 10; // a reasonable min
+        // if expected royalties is 0 or 1 we'll wait for 20s as a minimum,
+        // otherwise we'll wait for 10s per expected royalty
+        let secs = std::max(20, expected_royalties as u64 * 10);
+
         let duration = Duration::from_secs(secs);
         println!("Awaiting transfers notifs for {duration:?}...");
         if timeout(duration, async {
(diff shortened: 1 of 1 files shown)


sn_node/tests/nodes_rewards.rs
```diff
@@ -318,7 +318,14 @@ fn spawn_royalties_payment_listener(
         let mut count = 0;
         let mut stream = response.into_inner();
 
-        let secs = expected_royalties as u64 * 10; // a reasonable min
+        // if expected royalties is 0 or 1 we'll wait for 20s as a mnimum,
```
maqi: minimum ?

sn_node/tests/nodes_rewards.rs
```diff
@@ -318,7 +318,14 @@ fn spawn_royalties_payment_listener(
...
+        // otherwise we'll wait for 10s per expected royalty
+        let secs = if expected_royalties <= 1 {
```
maqi: let secs = std::max(20, expected_royalties as u64 * 10);

sn_node/tests/nodes_rewards.rs
```diff
@@ -363,7 +370,13 @@ fn spawn_royalties_payment_client_listener(
     let handle = tokio::spawn(async move {
         let mut count = 0;
 
-        let secs = expected_royalties as u64 * 10; // a reasonable min
+        // if expected royalties is 0 or 1 we'll wait for 20s as a mnimum,
+        // otherwise we'll wait for 10s per expected royalty
+        let secs = if expected_royalties <= 1 {
```
maqi: same as above two

maqi (approved):

PR #1051, opened 2023-12-05
feat: replace bootstrap node if bucket full

maqi:
### Summary generated by Reviewpad on 06 Dec 23 10:25 UTC
This pull request includes a feature that replaces a bootstrap node if the bucket is full. It modifies multiple files in the sn_networking library, including cmd.rs, driver.rs, event.rs, and lib.rs. The changes involve adding logic to check if a dial peer is a bootstrap node and replacing it if necessary, as well as removing a bootstrap node from a full bucket if present.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -277,7 +277,23 @@ impl SwarmDriver {
                             {
                                 let ilog2 = kbucket.range().0.ilog2();
                                 let num_peers = kbucket.num_entries();
-                                (num_peers >= K_VALUE.into(), ilog2)
+                                let mut is_bucket_full = num_peers >= K_VALUE.into();
+
+                                // If the bucket contains any of a bootstrap node,
+                                // consider the bucket is not full and dial back
+                                // so that the bootstrap nodes can be replaced.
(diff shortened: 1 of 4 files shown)


sn_networking/src/event.rs
```diff
@@ -1245,6 +1263,32 @@ impl SwarmDriver {
 
         false
     }
+
+    // if target bucket is full, remove a bootstrap node if presents.
+    fn remove_bootstrap_from_full(&mut self, peer_id: PeerId) {
```
joshuef, line 1268: Ahhh, how will this work when we have many many bootstrap peers via the `network-contacts` 🤔 They'd all be removed here. Which is perhaps not ideal? (or maybe it _is_ keeping them lightweight?? )

joshuef: Perhaps we only proceed here with only 1 bootstrap node, so that is more clearly genesis. Otherwise if we provide a big list of bootstrap peers, that could be problematic?

maqi: will remove just one for each attempt

joshuef (approved):

PR #1055, opened 2023-12-06
refactor: take either SAFE_PEERS or --peer

b-zee:
This is the conventional way to set arguments in clap. However, the
churn test launched a genesis node, but also a testnet that includes a
'new' genesis node. The testnet would pass --peer for the 'new' node,
while we'd actually want our first genesis node to be used to bootstrap,
which is set in SAFE_PEERS.

### Summary generated by Reviewpad on 18 Dec 23 15:58 UTC
This pull request includes a series of patches that make several changes to the codebase. Here is a summary of each patch:

1/7 - Refactors the code to allow either SAFE_PEERS or the --peer flag to be used for setting arguments in clap. This is done to handle a scenario where the testnet passes --peer for a new genesis node, but the first genesis node should be used for bootstrapping.

2/7 - Adds the node-count parameter to the memcheck workflow, setting it to 25.

3/7 - Enables the RPC on the genesis node in the memcheck workflow.

4/7 - Fixes a bug in the testnet where the number of nodes launched is not counted correctly.

5/7 - Fixes another bug in the testnet where the number of nodes launched is not counted correctly.

6/7 - Removes the RPC configuration from the genesis node in the memcheck workflow.

7/7 - Increases the interval for the memcheck testnet to 5 seconds.

Overall, these patches aim to improve the functionality and reliability of the codebase, particularly in the context of the memcheck workflow in a testnet environment.

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -66,6 +66,7 @@ jobs:
         with:
           action: start
           interval: 2000
+          join: true # Join the bootstrap node, set by the env var SAFE_PEERS
           node-path: target/release/safenode
           faucet-path: target/release/faucet
           platform: ubuntu-latest
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -44,7 +44,7 @@ jobs:
         run: |
           mkdir -p $BOOTSTRAP_NODE_DATA_PATH
           ./target/release/safenode \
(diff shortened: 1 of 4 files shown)


joshuef (approved):

joshuef:
Marked as do not merge for now as CI memcheck is unstable w/r/t removing restarted peers from the RT 🤔

b-zee:
Closing in favour of #1130 and future PRs incoming.

PR #1059, opened 2023-12-07
feat: use hash based record verification for chunks

RolandSherwin:
### Summary generated by Reviewpad on 07 Dec 23 15:52 UTC
This pull request includes changes to multiple files. 

- The `mod.rs` file in the `messages` module adds support for chunk proofs and organizes message-related components.
- Changes in the `FilesCmds` enum, `upload_files` and `upload_chunk` functions remove the `show_holders` functionality.
- The `QueryResponse` enum adds new variants for `GetStoreCost` and `GetChunkExistenceProof` queries.
- The `Node` implementation in the `node.rs` file adds support for handling `GetChunkExistenceProof` queries.
- The `Cargo.toml` file adds the `tiny-keccak` dependency.
- Changes in the `query.rs` file update imports and add new variants to the `Query` enum.
- The `chunk_proof.rs` file adds the `ChunkProof` struct and associated methods.
- Changes in the `Files` implementation simplify parameter passing and remove unused `show_holders` functionality.
- Changes in various files import `ChunkProof` and modify functionality related to chunk storage and verification.
- Changes in the `event.rs` file modify the `SwarmDriver` implementation for handling Kademlia events.
- Changes in the `error.rs` file import types and add new error variants for chunk existence and register not found.
- Changes in the file add new functions for retrieving chunk proofs and store costs from the network.

Please review these changes and let me know if you have any questions.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -468,7 +468,8 @@ impl Client {
     }
 
     /// Verify if a `Chunk` is stored by expected nodes on the network.
-    pub async fn verify_chunk_stored(&self, address: NetworkAddress, chunk: &Chunk) -> Result<()> {
+    pub async fn verify_chunk_stored(&self, chunk: &Chunk) -> Result<()> {
+        let address = chunk.network_address();
         info!("Verifying chunk: {address:?}");
         let random_nonce = thread_rng().gen::<u64>();
         let record_value = try_serialize_record(&chunk, RecordKind::Chunk)?;
@@ -693,9 +694,7 @@ impl Client {
                 let handle = tokio::spawn(async move {
                     // make sure the chunk is stored;
                     let chunk = Chunk::new(Bytes::from(std::fs::read(&chunk_path)?));
-                    let res = client
-                        .verify_chunk_stored(chunk.network_address(), &chunk)
-                        .await;
+                    let res = client.verify_chunk_stored(&chunk).await;
 
                     Ok::<_, ChunksError>(((name, chunk_path), res.is_err()))
                 });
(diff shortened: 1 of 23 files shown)


sn_client/src/api.rs
```diff
@@ -381,7 +383,7 @@ impl Client {
         chunk: Chunk,
         payment: Payment,
         verify_store: bool,
-        show_holders: bool,
+        _show_holders: bool, // The holders are logged during ChunkProof verification by default. Todo: remove?
```
joshuef: if thisis not used, lets remove

joshuef:
added DoNotMerge as breaking change we may not want in juusst yet

sn_protocol/src/messages/chunk_proof.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// The nonce provided by the verifier
+pub type Nonce = u64;
+
+/// The hash(record_value + nonce) that is used to prove the existence of a chunk
+#[derive(Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct ChunkProof([u8; 32]);
+
+impl ChunkProof {
+    pub fn new(record_value: &[u8], nonce: Nonce) -> Self {
```
maqi, line 20: by making `ChunkProof` constructor taking a chunk_name instead of whole content will be enough,
and has the benefit of avoid un-necessary access to the whole content?
especialy at the previous place of 
```rust
                    // make sure the chunk is stored;
                    let chunk = Chunk::new(Bytes::from(std::fs::read(&chunk_path)?));
                    let res = client.verify_chunk_stored(&chunk).await;
```
which has to read from disk ?

RolandSherwin: This can lead to dishonest nodes just storing the keys and discarding the content. And the proof that they create would be valid.

maqi: ah, right, make sense.

maqi (approved):

PR #1063, opened 2023-12-08
Bail on payment errors, update cli putput for failed chunks

joshuef:
### Summary generated by Reviewpad on 08 Dec 23 09:09 UTC
This pull request includes the following changes:

1. The file `CHANGELOG.md` has been modified in the following ways:
- Updated dependencies in the `Other` section
- Added a new version entry `0.2.328` with an update to dependencies in the `Other` section

2. The file `sn_node/CHANGELOG.md` has been modified with the following changes:
- Added a new version [0.99.12](https://github.com/maidsafe/safe_network/compare/sn_node-v0.99.11...sn_node-v0.99.12) with updates to dependencies.

3. The file `lib.rs` in the `sn_peers_acquisition` module has been modified with the addition of an attribute `#[allow(clippy::unused_async)]` and a function `get_network_contacts` that is conditionally compiled based on the absence of the feature "network-contacts". The function returns an empty vector of `Multiaddr` and is marked as async.

4. A command in the GitHub workflow file `nightly.yml` has been updated. Specifically, the command `cargo run --bin safe --release -- --log-output-dest=data-dir register create baobao` has been modified to `cargo run --bin safe --release -- --log-output-dest=data-dir register create -n baobao` by adding the `-n` flag before the argument `baobao`. This change is made for creating a register in the client.

5. The file `nodes_rewards.rs` has undergone several changes including the addition of a new variable, modification of the rewards calculation and verification logic, adjustment of sleep durations, and modification of loop conditions for better functionality.

6. The file `Cargo.toml` has been modified with a version update from `0.86.47` to `0.86.48`.

7. The `Cargo.toml` file in the `sn_faucet` directory has been updated with a version bump from `0.1.69` to `0.1.70`.

8. The file `Cargo.lock` has been modified with updates to the versions of various packages including `sn_cli`, `sn_faucet`, `sn_node`, `sn_node_rpc_client`, and `sn_testnet`.

9. The diff of the `Cargo.toml` file in the `sn_testnet` directory shows a version update from `0.2.327` to `0.2.328`.

10. The file `CHANGELOG.md` has been modified with the addition of a new entry for version `0.1.66` which includes updates to dependencies.

11. The diff of the file `sn_cli/src/subcommands/files/mod.rs` includes various changes related to error handling, logging, and display of file information and costs.

12. The file diff includes changes to the `CHANGELOG.md` file in the `sn_faucet` directory, including the addition of a new section for version 0.1.70 and an update to dependencies.

[...]

PR #1066, opened 2023-12-10
fix(networking): return Vec for closest queries to reliably sort

joshuef:
Our close peers were not actually being sorted by closeness, just by hash in the HashSet. This changes it to vec (back to), so we retain sorting from the kbuckets


RolandSherwin (approved):

joshuef:
Closing as forms part of #1063

PR #1067, opened 2023-12-11
feat(ci): don't log raw data

RolandSherwin:
### Summary generated by Reviewpad on 11 Dec 23 10:59 UTC
This pull request adds a feature to the CI workflow. It modifies the merge.yml and nightly.yml files to prevent logging of raw data. It includes new steps in the workflow that check the length of the logs and exit with an error message if the length exceeds a certain threshold. The safe_path is also defined for each operating system.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -277,10 +277,13 @@ jobs:
         include:
           - os: ubuntu-latest
             node_data_path: /home/runner/.local/share/safe/node
+            safe_path: /home/runner/.local/share/safe
           - os: windows-latest
             node_data_path: C:\\Users\\runneradmin\\AppData\\Roaming\\safe\\node
+            safe_path: C:\\Users\\runneradmin\\AppData\\Roaming\\safe
           - os: macos-latest
             node_data_path: /Users/runner/Library/Application Support/safe/node
+            safe_path: /Users/runner/Library/Application Support/safe
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -357,6 +360,16 @@ jobs:
             exit 1
           fi
 
+      - name: Don't log raw data
+        shell: bash
+        timeout-minutes: 20
+        run: |
+          if rg '^' "${{ matrix.safe_path }}"/*/*/logs | awk 'length($0) > 15000 { print; exit 1 }'
```
joshuef: why 15000 here and not 2k?

.github/workflows/nightly.yml
```diff
@@ -446,6 +462,16 @@ jobs:
             node_count=$(ls "${{ matrix.node_data_path }}" | wc -l)
             echo "Node dir count is $node_count"
 
+        - name: Don't log raw data
+          shell: bash
+          timeout-minutes: 20
+          run: |
+            if rg '^' "${{ matrix.safe_path }}"/*/*/logs | awk 'length($0) > 15000 { print; exit 1 }'
```
joshuef: and here

.github/workflows/nightly.yml
RolandSherwin: Ah I was testing it out. I was going for 15k chars as the limit.
On a local run, I can se log lines upto 4k-5k chars. So 15k seems like something that we might possibly log. Raw data should contain upwards of 100k chars, so this might be a reasonable value to detect outliers but not trigger false positives.

joshuef (approved):

PR #1068, opened 2023-12-12
feat: cli double spends audit from genesis and SpendDag POC

grumbach:
## Description

Gather all spends from Network starting from Genesis with a simple cli command:

```
safe wallet audit
```

Implement `SpendDag` proof of concept, gathered from the Network and exportable to dot format for visualisation. (requires [graphviz](https://graphviz.org/download/))

```
safe wallet audit --dot | tail -n +5 | dot -Tsvg > graph.svg
```

### Summary generated by Reviewpad on 12 Dec 23 05:20 UTC
This pull request includes various changes across multiple files:

- In the file "sn_client/src/error.rs", an import statement for `sn_transfers::SpendAddress` has been added. Additionally, a new variant `MissingSpendRecord` has been added to the `Error` enum, representing the scenario where there is no Spend record at a specific address. These changes enhance error handling and transfer validity verification.

- In the file "api.rs", changes have been made in the `impl Client` block. Line 572 now uses the `map_err` function call to match an error type and return the appropriate error variant. Line 575 has a closing parenthesis added. These changes primarily handle errors related to getting spend records from the network.

- The file "sn_client/src/audit/mod.rs" shows multiple changes:
  - A new file `mod.rs` has been added to the `audit` directory.
  - Several imports have been added.
  - The `verify_spend` method has been implemented to verify the validity of a spend on the network.
  - The `follow_spend` method has been implemented to recursively follow the descendants of a spend.
  - The `split_utxos_and_spends` function has been added.
  - Various debug statements and progress messages have been included.

- In the file "sn_cli/src/subcommands/wallet.rs", the following changes have been made:
  - A new variant `Audit` has been added to the `WalletCmds` enum.
  - A new function `audit` has been added to perform currency auditing.
  - The `wallet_cmds` function has been updated to handle the new `Audit` command.
  - A new async function `audit` has been added to print a message and perform currency auditing using the `client` object.
  - The `address` function has been updated to load the wallet from the root directory and print the address.

- The file "sn_cli/src/subcommands/wallet.rs" also had the `verify_spend` method removed from the `Client` implementation. This method was responsible for verifying spends on the network and their ancestors, while also printing progress on stdout. It used iteration to avoid stack overflow and checked for potential double spends by comparing the spends in the network with those in the CashNote. It also verified transactions by using the parent spends and kept track of verified transactions. The method reported the depth and number of verified transactions until reaching the genesis transaction.

- Additionally, the file ".DS_Store" has been deleted.

This pull request introduces changes related to error handling, transfer verification, currency auditing, spend verification, and file management. Please let me know if you require further information.


sn_client/src/audit/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+    ///                       -> Spend4 ---> UTXO_8
+    ///
+    /// gen0       gen1          gen2        gen3
+    ///
+    /// ```
+    ///
+    /// This function will return the UTXOs (Spend addresses not spent yet)
+    /// Future calls to this function could start from those UTXOs to avoid
+    /// re-checking all previously checked branches.
+    pub async fn follow_spend(
```
joshuef, line 152: this is following a single spend? Or it's doing the whole token dag? name doesnt fit, I think if we're doing the whole dag

grumbach: This function is following a single spend and its descendants recursively all the way to UTXOs. 
In our current use case, we call this function with the Genesis Spend so it effectively goes through all spends in the Network. 
What do you suggest we rename this to?

b-zee (approved):

Issue #360, opened 2023-06-08
feature: add metrics for chunk and system storage space


happybeing:
The logs used to include the following metrics which I displayed in `vdash` and think would be useful to have again. So I wonder if they can be added to the metrics module in `safenode`:
- space used for chunks
- space used for registers 
- maximum space for safe data (this was a fixed parameter, later to be set by the user)
- system storage total
- system storage free

I can't get these from the local system because there may be multiple `safenode` processes running, and `vdash` monitors multiple nodes by displaying any one of any number of available `safenode.log` files.

happybeing:
@joshuef I'm replying to your request for feedback on logfile messages to keep things in one place. The following are messages strings I am currently matching for the given metrics, while the OP suggests additional metrics it may be useful to have but which we did have at one stage.

PUT: "Wrote record to disk"
GET: "Retrieved record from disk"
REGISTER EDIT: "Editing Register success!" (untested)
ERROR: any logfile message of type 'ERROR'

I have a crude mechanism for categorising node status as: Stopped|Connecting|Connected|Disconnected. I don't know if a node could provide a more definitive state message as a periodic output to the logfile (not just on change), but if so that would improve accuracy.

I'm not sure what would be helpful to add in other areas both for devs, if you use `vdash` at all? Or for users interested in monitoring the working and performance of their nodes, but it would be nice to find some metrics to at least reflect ongoing activity of different network features related to each node, so things like register edits, CRDT and DBC related activity, and of course node earnings!

Of course I'm open to any suggestions for things that your team would like.

joshuef:
> space used for chunks
> space used for registers

Would be hard as they're all records now. (hard as in reading from disk). So i'd be inclined to just leave that to a sys level check of the `record_store` dir. Not sure how you feel about that?

We have eg: https://github.com/maidsafe/safe_network/blob/main/sn_node/src/log_markers.rs#L21

Connecting would just be everything before that. Though perhaps we can add an initial message when we start attempting to connect to the first peers (could be done via https://github.com/maidsafe/safe_network/issues/518)

> could provide a more definitive state message as a periodic output to the logfile 

Hmm, outwith of connecting/stopped it would (in theory) just always be connected. Not sure if that's that useful? (Or are you imagining more states? We have some `kbucket` logs of peers counts that may he more granual about the state of connectivity? https://github.com/maidsafe/safe_network/blob/main/sn_networking/src/event.rs#L458


[...]

PR #1070, opened 2023-12-12
feat: include records stored and max records in logfile and fix paths to safe in README.md 

happybeing:
## Description

Addresses part of issue #360 

### Summary generated by Reviewpad on 12 Dec 23 15:47 UTC

Patch 1/2 includes changes to the `record_store.rs` file, adding the record count and max records to the logfile output. 

Patch 2/2 fixes the README paths to use `.local/share/safe` instead of `.safe`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This is the Safe Network as it was supposed to be, on a kademlia network, enable
 You'll need to set the `SAFE_PEERS` env variable to the multiaddress of a node when you set up a testnet.
 You can do this automatically at network startup using the following command (if you have ripgrep installed)
 ```bash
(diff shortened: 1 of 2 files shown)


joshuef (approved):
Awesome. Thanks @happybeing for getting readme updates in too 💪

happybeing:
Cool. @joshuef Any idea what the "unconventional commits" errors refer to from Reviewpad ?

EDIT:
From the error messages I assume it's the lack of metadata in the commit messages. Is there a way to catch/ validate this when doing the commit?

joshuef:
yeh, basically we use and have a linter for https://www.conventionalcommits.org/en/v1.0.0/

which means no capital first letters, or full stops at the end. And also more ideally some scope definition, eg your commits could be:

- `chore(networking): include record count and max records in logfile output`
- `fix: updated README paths for .local/share/safe`

Do you know how to amend those commits? (You can use `git rebase -i HEAD~2` and then instead of `pick` mark each as `reword` and then edit each commit message in turn.

happybeing:
Ah thanks. Busy today but will try to find a few minutes.

joshuef:
great stuff, thanks @happybeing !

PR #1073, opened 2023-12-13
tryout: hash based verification + get closest cache

RolandSherwin:
## Description

reviewpad:summary


RolandSherwin:
Willl revisit `get_record_to` if #1059 change produces errors.

PR #1074, opened 2023-12-13
feat(cli): simple retries

joshuef:
### Summary generated by Reviewpad on 13 Dec 23 14:49 UTC
This pull request includes changes that involve error handling and transfer verification in the network. The file `error.rs` introduces two new variants to the `Error` enum: `CouldNotVerifyTransfer` and `MissingSpendRecord`. Additionally, there is a change in the `node.rs` file at line 542, where the ignored replication list is replaced with the length of the received keys in a warning log statement, providing more information about the ignored replication list.


grumbach (approved):
LGTM

PR #1077, opened 2023-12-13
chore(cli): tidy cli output by removing peers list

iancoleman:
## Description

Small ergonomic improvements to the cli output.

I tried to copy the address when calling `safe wallet get-faucet` by double clicking the address and it also included the `...` at the end of the address.

The peer list is not useful and is very long so can be removed from the output. Might be useful to add a `safe peers` command or something like that to list peers? Related to #1007

### Summary generated by Reviewpad on 13 Dec 23 20:48 UTC

The first patch updates the code in the main.rs file to improve the stdout output when connecting to the network with peers. Instead of printing each individual peer, it now prints the number of peers.

The second patch modifies the wallet.rs file to remove the unnecessary ellipsis in the println statement, making it easier to copy the wallet address.

### DIFF
--- a/sn_cli/src/main.rs
+++ b/sn_cli/src/main.rs
@@ -87,10 +87,10 @@ async fn main() -> Result<()> {
 
     let bootstrap_peers = parse_peers_args(opt.peers).await?;
 
-    println!("Connecting to the network w/peers:");
-    for peer in &bootstrap_peers {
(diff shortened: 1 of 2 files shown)


joshuef (approved):

joshuef:
nice one 🙇

PR #1080, opened 2023-12-14
feat(cli): simple retry mechanism for remaining chunks

joshuef:
### Summary generated by Reviewpad on 14 Dec 23 10:34 UTC
This pull request modifies multiple files across the codebase. Here's a summary of the changes:

1. The `Cargo.lock` file has been modified to add a new package called "backoff" as a dependency in the "sn_client" and "sn_networking" packages.

2. In the `files.rs` file, two arguments have been added to a command at line 35. The argument "-r" is set to "0", indicating that the upload should not be retried in case of failures.

3. The `api.rs` file has undergone changes, including the addition of an empty line at line 304 and a modification in the `impl Client` block to use the `await` keyword in the `let maybe_record` line.

4. A file diff includes the addition of a new command line argument named `max_retries` in the `FilesCmds` enum and the `files_cmds` function. This argument sets the number of retry passes for failed chunks during payment and upload processing. The changes also involve updating the `upload_files` function to use the `max_retries` argument and adding more robust logging and error handling.

5. The `nightly.yml` file in the `.github/workflows` directory has been modified. Specifically, the `run` command in the "Start a client to carry out chunk actions" step has been updated to include the `-r 0` flag at the end.

6. The `Cargo.toml` file has undergone changes, including adding a new dependency on the "backoff" package, updating package versions, and enabling features.

7. The `error.rs` file has been modified to add a new variant to the `Error` enum and include a previously missing line of code.

8. Changes to the `.github/workflows/memcheck.yml` file include modifying the command for running `cargo run` to include the `-r 0` flag and redirecting the output to a file named `second_upload.txt`.

9. The `.github/workflows/merge.yml` file has been modified to update the `run` command in the `jobs` section and change the `timeout-minutes`.

10. The `Cargo.toml` file in the `sn_networking` directory has been modified to add dependencies, update versions, and specify features.

11. The `wallet.rs` file has undergone changes, including adding `use` statements, removing constants and `use` statements, and modifying the `pay_for_storage` method.

12. The `.github/workflows/benchmark-prs.yml` file has been modified to add the `-r 0` flag to the `run` command in two places.

13. Changes to the `generate-benchmark-charts.yml` file in a GitHub Actions workflow include modifying the `run` command to include the `-r 0` flag.

14. The `lib.rs` file has undergone various modifications, including adding imports, adding/modifying constants, and modifying function implementations.

Please review these changes and ensure they align with the intended modifications.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -32,6 +32,9 @@ use std::{
 use tokio::task::JoinHandle;
 use xor_name::XorName;
 
+/// The maximum number of sequential payment failures before aborting the upload process.
+const MAX_SEQUENTIAL_PAYMENT_FAILS: usize = 3;
+
 #[derive(Parser, Debug)]
 pub enum FilesCmds {
     Upload {
@@ -242,12 +245,10 @@ async fn upload_files(
         batch_size,
     };
 
-    // Max amount of sequential payment failures before we bail
-    let max_sequential_payment_fails = 3;
     let mut sequential_payment_fails = 0;
 
     for chunks_batch in chunks_batches {
-        if sequential_payment_fails >= max_sequential_payment_fails {
+        if sequential_payment_fails >= MAX_SEQUENTIAL_PAYMENT_FAILS {
             bail!("Too many sequential payment failures ({sequential_payment_fails}). Aborting upload process.");
         }
         // if the payment fails, we can continue to the next batch
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -9,6 +9,7 @@
 use crate::Error;
 
 use super::{error::Result, Client};
+use backoff::{backoff::Backoff, ExponentialBackoff};
 use futures::{future::join_all, TryFutureExt};
 use sn_networking::GetRecordError;
 use sn_protocol::NetworkAddress;
@@ -23,16 +24,12 @@ use std::{
 };
 use tokio::{task::JoinSet, time::sleep};
 use xor_name::XorName;
-
 /// A wallet client can be used to send and
 /// receive tokens to/from other wallets.
 pub struct WalletClient {
     client: Client,
     wallet: LocalWallet,
 }
-
-/// Number of retries for data payments
-const DATA_PAYMENT_RETRIES: usize = 3;
 
 impl WalletClient {
     /// Create a new wallet client.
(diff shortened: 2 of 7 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -235,13 +242,67 @@ async fn upload_files(
         batch_size,
     };
 
+    // Max amount of sequential payment failures before we bail
+    let max_sequential_payment_fails = 3;
```
maqi: this shall be the `max_retries` opt instead ?
if the `payment` attempt has to be undertaken for 3 times, it's acutally `min`, and better make it as a const ?

sn_client/src/wallet.rs
```diff
@@ -145,10 +142,11 @@ impl WalletClient {
     ) -> WalletResult<((NanoTokens, NanoTokens), Vec<XorName>)> {
         let verify_store = true;
         let c: Vec<_> = content_addrs.collect();
+        let mut backoff = ExponentialBackoff::default();
         let mut last_err = "No retries".to_string();
 
-        for retry in 0..DATA_PAYMENT_RETRIES {
-            trace!("Paying for storage (retry: {retry}) for: {:?}", c);
+        while let Some(delay) = backoff.next_backoff() {
```
maqi, line 148: this will keeps looping forwever till a success or non-send error ?
could be risky of having a deadlock.
at least have the `warn` to be a `println` as well, so that user can notice a deadlock ?

sn_networking/src/lib.rs
```diff
@@ -78,14 +80,16 @@ pub const fn close_group_majority() -> usize {
...
 /// Max duration to wait for verification
-const MAX_REVERIFICATION_WAIT_TIME_S: std::time::Duration = std::time::Duration::from_millis(750);
+const MAX_REVERIFICATION_WAIT_TIME_MS: Duration = Duration::from_millis(750);
 /// Min duration to wait for verification
-const MIN_REVERIFICATION_WAIT_TIME_S: std::time::Duration = std::time::Duration::from_millis(300);
-/// Number of attempts to GET a record
-const GET_RETRY_ATTEMPTS: usize = 3;
-/// Number of attempts to PUT a record
-const PUT_RETRY_ATTEMPTS: usize = 10;
+const MIN_REVERIFICATION_WAIT_TIME_MS: Duration = Duration::from_millis(300);
```
maqi, line 92: do we have an estimation of steps of those delays to be generated by the backoff ?
300ms or 750ms feels too short for a public testnet to complete a  network option.
i.e. after the firmt operation, the `max_elapsed_time` may got reached already, hence there will be no further attempt in case of error?

sn_client/src/wallet.rs
joshuef: It would hit the default backoff limit, which returns None once we've taken too much time. we're in the client here so `println` isnt appropriate

sn_networking/src/lib.rs
joshuef: https://docs.rs/backoff/latest/backoff/

it retries exponentially over 10s by default. 

we also have this min_reverification wait time _before_ we start doing verification reqs.

sn_client/src/wallet.rs
maqi: Ah, so you mean hit that default `10s` cap of the backoff ?
ok then

maqi (approved):

PR #1089, opened 2023-12-18
fix(cli): mark chunk completion as soon as we upload each chunk

RolandSherwin:
### Summary generated by Reviewpad on 18 Dec 23 11:23 UTC
This pull request fixes an issue in the CLI where chunk completion was not marked correctly during file uploads. The patch modifies the `upload_files` function in `sn_cli/src/subcommands/files/mod.rs` to correctly mark chunk completion as soon as each chunk is uploaded. It also adds new events to the `FileUploadEvent` enum in `sn_client/src/files/mod.rs` to handle different upload scenarios such as successful uploads, existing chunks, failed uploads, and payment for chunks. Overall, this patch improves the reliability and accuracy of the file upload process in the CLI.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -96,7 +96,7 @@ pub(crate) async fn files_cmds(
             upload_files(
                 path,
                 client,
-                root_dir,
+                root_dir.to_path_buf(),
                 verify_store,
                 batch_size,
                 show_holders,
@@ -157,7 +157,7 @@ pub(crate) async fn files_cmds(
 async fn upload_files(
     files_path: PathBuf,
     client: &Client,
-    root_dir: &Path,
+    root_dir: PathBuf,
     verify_store: bool,
     batch_size: usize,
     show_holders: bool,
@@ -169,7 +169,7 @@ async fn upload_files(
     if files_api.wallet()?.balance().is_zero() {
         bail!("The wallet is empty. Cannot upload any files! Please transfer some funds into the wallet");
     }
-    let mut chunk_manager = ChunkManager::new(root_dir);
+    let mut chunk_manager = ChunkManager::new(&root_dir);
     chunk_manager.chunk_path(&files_path, true)?;
 
     // Return early if we already uploaded them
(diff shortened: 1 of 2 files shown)


sn_client/src/files/mod.rs
```diff
@@ -175,7 +181,12 @@ impl Files {
 
         let result = self.upload(chunks).await;
 
-        // drop the sender
+        // send an event indicating that the upload process completed with an error
+        if result.is_err() {
+            self.send_event(FileUploadEvent::Error).await?;
```
joshuef, line 186: Can we pass out the error here?

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -231,11 +227,15 @@ async fn upload_files(
     // keep track of the progress in a separate task
     let progress_bar_clone = progress_bar.clone();
     let total_existing_chunks_clone = total_existing_chunks.clone();
-    tokio::spawn(async move {
+
+    let progress_handler = tokio::spawn(async move {
+        let mut upload_terminated_with_error = false;
+        // The channel is closes if the upload completes/errors out, so the loop is guaranteed to end.
```
maqi: // The loop is guaranteed to be ended, as the channel will be closed when the upload completed or errored out.

sn_client/src/files/mod.rs
RolandSherwin: The error variants inside ClientError does not impl Copy. We might have to wrap the internal types into `Arc` or return an arc error. There does not seem to be a simple workaround for it right now.

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -231,11 +227,15 @@ async fn upload_files(
...
         while let Some(event) = upload_event_rx.recv().await {
             match event {
-                FileUploadEvent::Uploaded(_addr) => {
+                FileUploadEvent::Uploaded(addr) => {
                     progress_bar_clone.inc(1);
+                    chunk_manager.mark_completed(std::iter::once(*addr.xorname()));
```
maqi, line 238: `FileUploadEvent::Uploaded` only says the chunk got sent out to the network.
But without a `fetch verification`, can't say the `chunk is stored in the network`.
Hence, marking it as `completed` here is bit risky, and not consistent with the current definition of `completed` (verified to be existing in the network, when the verification flag is set) ?

RolandSherwin: > `FileUploadEvent::Uploaded` only says the chunk got sent out to the network.

This event is emitted when the `Files::upload_chunk(verify_store,...)` returns `Ok(())`.  Thus, if we have configured the `Files::set_verify_store(true)`, we should indeed be verifying the upload.

joshuef (approved):

PR #1096, opened 2023-12-19
fix(test): tests should try to load just the faucet wallet

RolandSherwin:
- We start a faucet server before each test, so we should be not trying to load the genesis wallet again, instead try to wait for the faucet wallet to get filled in by the faucet_server

### Summary generated by Reviewpad on 19 Dec 23 09:34 UTC
This pull request fixes an issue with the tests. Previously, the tests were trying to load the genesis wallet instead of waiting for the faucet wallet to get filled in by the faucet server. This pull request updates the code to wait for the faucet wallet to be filled in before continuing with the tests.

### DIFF
--- a/sn_node/tests/common/client.rs
+++ b/sn_node/tests/common/client.rs
@@ -6,22 +6,25 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use eyre::Result;
+use eyre::{bail, Result};
 use lazy_static::lazy_static;
-use sn_client::{load_faucet_wallet_from_genesis_wallet, send, Client};
+use sn_client::{send, Client};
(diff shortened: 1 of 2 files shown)


sn_node/tests/common/client.rs
```diff
@@ -160,6 +162,20 @@ impl NonDroplet {
...
+        for _ in 1..6 {
+            let faucet_wallet = create_faucet_wallet();
+
+            let faucet_balance = faucet_wallet.balance();
+            if !faucet_balance.is_zero() {
+                return Ok(faucet_wallet);
+            }
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+        bail!("The faucet wallet is empty even after a couple retries. Did the faucet_server start correctly?");
```
joshuef: can we make the retries a const and then pass that in here, probably noting the amount of time taken too?

joshuef (approved):

PR #1098, opened 2023-12-19
chore(node): log wallet balance on earning

maqi:
### Summary generated by Reviewpad on 19 Dec 23 12:57 UTC

The first patch in `sn_node/src/put_validation.rs` adds logging of the wallet balance when earning. It retrieves the old and new balance of the wallet, deposits the CashNotes, and updates the reward wallet balance metric. 

The second patch in `sn_networking/src/lib.rs` introduces a change to randomly select the payee instead of selecting the lowest cost. It shuffles the costs and selects a random payee from the list. It also includes updates to tests related to the payee selection logic.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -45,7 +45,7 @@ use libp2p::{
     multiaddr::Protocol,
     Multiaddr, PeerId,
 };
-use rand::Rng;
+use rand::{seq::SliceRandom, Rng};
 use sn_protocol::{
     error::Error as ProtocolError,
     messages::{Query, QueryResponse, Request, Response},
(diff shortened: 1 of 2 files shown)


joshuef (approved):

happybeing:
I'm not sure about randomly selecting nodes so as not to exclude high cost nodes. Looking at my nodes, ordered by storecost having a high cost doesn't make them low earners. Obviously storecost has an impact (assuming the choose cheapest mechanism was operating as expected) but it doesn't seem to explain why some nodes are not earning, so this change seems premature. See vdash display (attachment) for example from CloseNet just now.

Screenshot from 2023-12-19 15-21-05

The most recently started node 13, has the highest storecost and highest total earnings after just 23 hours. Node 3, after 6 days has the third highest storecost of my nodes but only earned 15 nanos, why is not clear.

To get to the bottom of this it might help to set-up artificial node distributions (i.e. peer IDs, if that is possible) and other tweaks to reduce the number of variables, and then check behaviours against expectations/design as each variable is brought back into play.

Interestingly my Node 12 which had not earned anything in over 40 hours has just - while I was typing this - had one payment event of 1569 nanos. The screen shot above shows 0 nanos for this node but it has just earned. This suggests all was fine with the node. Storecost doesn't seem to explain it's lateness in earning compared to say Node 13, because both began with a storecost of 400 nanos which then gradually increased. Its just that Node 13 has had many more storage+payment events. 

So their difference (the lowest and highest earners until just a few minutes ago) appears to be about location to lucrative storage events either due to the node topology or some clustering of data. :thinking: 

Hope this helps! I'll cross post some / all of this comment to the forum.

Issue #1087, opened 2023-12-17
Register logging does not show the whole 64byte id


southside39:
The log output was cleaned up and the whole id (64byte) truncated for easier reading. In this case that is wrong and it should be printed out clearly for you to read.

https://safenetforum.org/t/communication-via-safe-network/38958/14?u=southside

maqi:
Register has been refactored.
re-open if issue remains with the latest version

PR #1101, opened 2023-12-19
chore: log entire RegisterAddress to facilitate debugging

thomasmarkiewicz:
## Description

This resolves issue #1087

It enables printing/logging of the entire RegisterAddress so that developers can use it to debug code issues.

See a discussion thread here: https://safenetforum.org/t/communication-via-safe-network/38958/14?u=southside

In addition, it also updates `create_and_pay_for_register` with additional function argument `public` which dictates if it should create private or public register.  The example app sn_node/examples/registers.rs was creating a private register which cannot be written by everyone (which is the point that this example showcases).  That example still does't work, but this extra debug info will hopefully make pinpointing the issue easier.

### DIFF
--- a/sn_node/examples/registers.rs
+++ b/sn_node/examples/registers.rs
@@ -67,17 +67,17 @@ async fn main() -> Result<()> {
 
     let meta = XorName::from_content(reg_nickname.as_bytes());
     let address = RegisterAddress::new(meta, client.signer_pk());
-    println!("Retrieving Register '{reg_nickname}':'{address:?}' from SAFE, as user '{user}'");
+    println!("Retrieving Register '{reg_nickname}':'{:?}' from SAFE, as user '{user}'", address.to_hex());
     let mut reg_replica = match client.get_register(address).await {
         Ok(register) => {
             println!(
                 "Register '{reg_nickname}' found at {:?}!",
-                register.address(),
+                register.address().to_hex(),
             );
             register
         }
         Err(_) => {
-            println!("Register '{reg_nickname}' not found, creating it at {address:?}");
+            println!("Register '{reg_nickname}' not found, creating it at {:?}", address.to_hex());
             let (register, _cost, _royalties_fees) = client
                 .create_and_pay_for_register(meta, &mut wallet_client, true, true)
                 .await?;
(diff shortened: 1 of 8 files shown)


sn_node/examples/registers.rs
```diff
@@ -63,9 +63,11 @@ async fn main() -> Result<()> {
 
     // we'll retrieve (or create if not found) a Register, and write on it
     // in offline mode, syncing with the network periodically.
+
+
     let meta = XorName::from_content(reg_nickname.as_bytes());
     let address = RegisterAddress::new(meta, client.signer_pk());
-    println!("Retrieving Register '{reg_nickname}' from SAFE, as user '{user}'");
+    println!("Retrieving Register '{reg_nickname}':'{address:?}' from SAFE, as user '{user}'");
```
RolandSherwin: Is it possible if we can call `address.to_hex()` while we're displaying the address to the user? 

All the different address types inc. `RegisterAddress` were previously truncated to display the first couple chars to not fill up the log files.

RolandSherwin:
I think the commit messages have to updated to follow the conventionalcommits format. The automated check currently fails for that.

sn_node/examples/registers.rs
trycatch007: Is logging the first couple chars enough for folk to follow the logs for debugging? It wasn't for me, but it's very possible I just don't fully understand. 

Also, @dirvine mentioned [here](https://safenetforum.org/t/communication-via-safe-network/38958/14?u=trycatch) it's a bug:

> Looks like log output was cleaned up and the whole id (64byte) truncated for easier reading. In this case that is wrong and it should be printed out clearly for you to read.
> Bug bug !!

Ultimately though, I leave it up to you core devs if you want to incorporate this change or not, you have a much bigger picture of things to consider.

trycatch007:
> follow the conventionalcommits format

I noticed, but I don't know what the conventional commit format is.  Do you have it documented somewhere?

sn_node/examples/registers.rs
thomasmarkiewicz: Yes, that makes sense.  I update to log the entire RegisterAddress only at few key places.

thomasmarkiewicz:
> I think the commit messages have to updated to follow the conventionalcommits format. The automated check currently fails for that.

Ah, and this is the expectation? https://www.conventionalcommits.org/en/v1.0.0-beta.2/

I'll probably close this PR and start from scratch in that case.  Thanks.

PR #1104, opened 2023-12-20
test: address distribution sim

maqi:
### Summary generated by Reviewpad on 20 Dec 23 16:01 UTC
This pull request includes a test that addresses the distribution of simulated records among peers. The test generates a map of peers with corresponding statistics, such as the number of records, nano earned, and the number of times earned. The test then simulates the effects of different payee selection algorithms by iterating through chunks and assigning payees based on specified ranges. After the iterations, the test logs the distribution of chunks among different groups based on the leading byte of the hashed bytes. Additionally, the test includes helper functions for picking payees randomly or based on the cheapest cost.


joshuef (approved):
A good starting point.

PR #1116, opened 2023-12-26
feat: use put_record_to during upload chunk

maqi:
### Summary generated by Reviewpad on 26 Dec 23 16:53 UTC
This pull request includes various changes across multiple files:

- In `lib.rs`, changes were made to the `get_store_costs_from_network` function, including a change in the return type. Additionally, new functions (`put_record_to`, `put_record_once`, `put_record_to_once`) were added. The return type of `get_fees_from_store_cost_responses` function was also changed.

- In `mod.rs`, changes were made to include `libp2p::PeerId` and modify return values in `self.api.pay_for_chunks()` and `impl Files::upload_chunk()`.

- In `api.rs`, changes were made to the `store_chunk` method of the `Client` struct and the `put_record` call within it.

- In `sn_networking/src/cmd.rs`, changes were made to add a new variant in the `SwarmCmd` enum, implement handling for the new variant in the `SwarmDriver` struct, and update debug log messages.

- In `wallet.rs`, changes were made to include an import statement, modify return types and arguments in various functions, add a new variable, and modify storing of costs.

These changes seem to be related to storing and retrieving records in a network, specifying payees for chunks, calculating fees, and adding payment and peer information in the wallet functionality.

Let me know if there's anything else I can help with.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -11,6 +11,7 @@ use crate::Error;
 use super::{error::Result, Client};
 use backoff::{backoff::Backoff, ExponentialBackoff};
 use futures::{future::join_all, TryFutureExt};
+use libp2p::PeerId;
 use sn_networking::GetRecordError;
 use sn_protocol::NetworkAddress;
 use sn_transfers::{
@@ -123,7 +124,7 @@ impl WalletClient {
     pub async fn get_store_cost_at_address(
         &self,
         address: NetworkAddress,
-    ) -> WalletResult<(MainPubkey, PaymentQuote)> {
+    ) -> WalletResult<(PeerId, MainPubkey, PaymentQuote)> {
         self.client
             .network
             .get_store_costs_from_network(address)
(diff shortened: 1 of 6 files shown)


sn_client/src/wallet.rs
```diff
@@ -139,7 +140,10 @@ impl WalletClient {
     pub async fn pay_for_storage(
```
joshuef, line 146: I think we need to update the documentation here. It's not immediately obvious what more we're returning here. (original docs were probs insuffcient here too though)

sn_networking/src/lib.rs
```diff
@@ -447,6 +447,41 @@ impl Network {
...
+                if cfg.re_attempt {
+                    BackoffError::Transient { err, retry_after: None }
+                } else {
+                    BackoffError::Permanent(err)
+                }
+
+            })
+        }).await
+
+        // Err(last_err)
```
joshuef: let's remove this 🙇

sn_networking/src/lib.rs
```diff
@@ -447,6 +447,41 @@ impl Network {
         // Err(last_err)
     }
 
+    /// Put `Record` to the specific node (only used for upload chunk to the choosen payee)
+    /// Optionally verify the record is stored after putting it to network
+    /// If verify is on, retry PUT_RETRY_ATTEMPTS times with a random wait between 1.5s and 5s
```
joshuef: dont think we should specify that time here so much as just the const

sn_networking/src/lib.rs
```diff
@@ -482,6 +517,47 @@ impl Network {
         response
     }
 
+    async fn put_record_to_once(
```
joshuef, line 604: more docs please 🙇

sn_networking/src/lib.rs
```diff
@@ -716,7 +792,13 @@ fn get_fees_from_store_cost_responses(
         .ok_or(Error::NoStoreCostResponses)?;
     info!("Final fees calculated as: {payee:?}");
     // we dont need to have the address outside of here for now
-    Ok((payee.1, payee.2))
+    let payee_id = if let Some(peer_id) = payee.0.as_peer_id() {
+        peer_id
+    } else {
+        error!("Cann't get PeerId from {:?}", payee.0);
```
joshuef:
```suggestion
        error!("Can't get PeerId from payee {:?}", payee.0);
```

joshuef (approved):

Issue #1106, opened 2023-12-21
Allow register CLI to create a public register writable to anyone


thomasmarkiewicz:
Currently `safe register create` creates registers that are writable by owner only, and it provides no options to create one writable by anyone (See `Permissions { anyone_can_write: false, writers: {PublicKey(0a7a..f00e)}`in the  log below)

User should be able to create one that is public, perhaps by passing an additional `-p` command line argument or something similar.

```
cargo run --release --bin safe --features local-discovery -- register create -n "MyPublicRegister2"
    Finished release [optimized] target(s) in 0.14s
     Running `target/release/safe register create -n MyPublicRegister2`
Logging to directory: "/home/tom/.local/share/safe/client/logs/log_2023-12-20_19-17-11"
Built with git version: 47f87145 / issues/1087_2 / 47f87145
Instantiating a SAFE client...
Connecting to the network with 0 peers:
🔗 Connected to the Network                                                                                                      Successfully made payment of 0.000000011 for a Register (At a cost per record of NanoTokens(11).)
Successfully stored wallet with cached payment proofs, and new balance 99.999999978.
Successfully created register 'MyPublicRegister2' at cda7beef4545803127ccdcd7b9676bbb2d8e7764d2fdc3944fca1afce625f87caa7a9196ce176b31257e708841f3b7748fc47dcb16053b8d7daafb03274385f80055581ba6569b811b4479a172b43370 for NanoTokens(10) (royalties fees: NanoTokens(1)) with permissions: Permissions { anyone_can_write: false, writers: {PublicKey(0a7a..f00e)} }!
```

grumbach:
Dropped support for public registers, although one can create a register and share the secret key.

PR #1118, opened 2023-12-26
feat: allow register CLI to create a public register writable to anyone

thomasmarkiewicz:
### Summary generated by Reviewpad on 26 Dec 23 20:50 UTC
This pull request includes changes to multiple files. Here is a summary of the changes made:

1. The `nodes_rewards.rs` test file has been modified to introduce owner-only permissions and improve transfer notification decoding.
2. The `register.rs` file in the `sn_cli/src/subcommands` directory has been updated to allow the creation of registers with public write access.
3. Changes to the file `register.rs` involve moving the creation and syncing of registers to the `create_online` function and introducing permission flexibility.
4. The `.github/workflows/merge.yml` file has been modified to add tasks for interacting with registers and testing different writability scenarios.
5. The `Cargo.toml` file now includes a dependency on `sn_registers` with version `0.3.6`.
6. Changes to the `data_with_churn.rs` file introduce permission control for register creation and payment.
7. The `storage_payments.rs` file has been updated to add permission control in the storage payments tests.
8. The `api.rs` file in the `sn_client` crate has been modified to enhance the functionality and handling of the `ClientRegister` implementation.
9. The `Cargo.lock` file now includes the dependency "sn_registers".
10. Various files include changes related to adding permissions for register creation and payment, as well as displaying register addresses.

These changes aim to enhance functionality, introduce permission control, and improve the handling of registers.
 

Resolves issue #1106

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4503,7 +4503,7 @@ dependencies = [
 
 [[package]]
 name = "sn_cli"
-version = "0.86.91"
+version = "0.86.92"
 dependencies = [
  "blsttc",
  "bytes",
@@ -4574,7 +4574,7 @@ dependencies = [
 
 [[package]]
 name = "sn_faucet"
-version = "0.1.113"
+version = "0.1.114"
 dependencies = [
  "blsttc",
  "clap 4.4.10",
@@ -4646,7 +4646,7 @@ dependencies = [
 
 [[package]]
 name = "sn_node"
-version = "0.100.37"
+version = "0.100.38"
 dependencies = [
  "assert_fs",
  "async-trait",
@@ -4698,7 +4698,7 @@ dependencies = [
 
 [[package]]
 name = "sn_node_rpc_client"
-version = "0.1.109"
+version = "0.1.110"
 dependencies = [
  "assert_fs",
  "async-trait",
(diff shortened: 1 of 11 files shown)


.github/workflows/merge.yml
```diff
@@ -179,24 +179,59 @@ jobs:
...
+        env:
+          SN_LOG: "all"
+        timeout-minutes: 10
+
+      - name: Delete client subdir to generate new client
+        shell: bash
+        run: rm -rf "${{ matrix.safe_path }}"/client
+
+      - name: Start a client to get a register writable by anyone (new client, the owner of Register)
+        run: cargo run --bin safe --release -- --log-output-dest=data-dir register get -n trycatch
```
thomasmarkiewicz: I'm not sure how to write a successful test case for getting the "public" register.  The problem is that CLI needs the hex address of the public register in this case - it cannot take one specified by name with -n if it was created by a different client.  But there is no way to know that address... unless it is parsed out somehow from the output of the CLI command when it was created above with -p.  I do not know how.

joshuef: Hey @thomasmarkiewicz , thanks for the work so far here!

Sounds like we need this actually printed to std out as part of the PUT process? And then you could save that to a github variable to share between steps?

Do you think it makes sense to _always_ print this info? Or should we hide it behind a flag in the CLI?

joshuef (commented):
Looking good in general here, nice one 🙇

PR #1120, opened 2023-12-29
feat(networking): remove problematic peers from routing table

joshuef:
### Summary generated by Reviewpad on 29 Dec 23 09:36 UTC
This pull request consists of two patches. 

The first patch modifies the `event.rs` file in the `sn_networking/src` directory. It adds checks against DialErrors to determine if certain errors like ConnectionRefused/Reset or HostUnreachable are serious issues. If they are, the patch removes the problematic peers from the routing table.

The second patch modifies the `nightly.yml`, `data_with_churn.rs`, and `verify_data_location.rs` files. It updates the churn tests to be harsher in order to more swiftly detect churn. It increases the churn cycles, creation ratios, and content query ratios.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -502,24 +502,42 @@ impl SwarmDriver {
             } => {
                 event_string = "OutgoingConnErr";
                 warn!("OutgoingConnectionError to {failed_peer_id:?} on {connection_id:?} - {error:?}");
-                // info!("{:?}", self.swarm.network_info());
 
                 // we need to decide if this was a critical error and the peer should be removed from the routing table
                 let should_clean_peer = match error {
                     DialError::Transport(errors) => {
                         // as it's an outgoing error, if it's transport based we can assume it is _our_ fault
+                        //
                         // (eg, could not get a port for a tcp connection)
                         // so we default to it not being a real issue
-                        let mut there_is_a_real_issue = false;
+                        // unless there are _specific_ errors (connection refused eg)
+                        error!("Dial errors len : {:?}", errors.len());
+                        let mut there_is_a_serious_issue = false;
                         for (_addr, err) in errors {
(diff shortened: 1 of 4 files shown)


sn_networking/src/event.rs
```diff
@@ -502,24 +502,43 @@ impl SwarmDriver {
...
+
+                            match err {
+                                TransportError::MultiaddrNotSupported(addr) => {
+                                    warn!("Multiaddr not supported : {addr:?}");
+                                    // if we can't dial a peer on a given address, we should remove it from the routing table
+                                    there_is_a_serious_issue = true
+                                }
+                                TransportError::Other(err) => {
+                                    let problematic_errors =
+                                        ["ConnectionReset", "ConnectionRefused", "HostUnreachable"];
```
maqi: Not sure about the `ConnectionReset` .
If say the peer closed a connection due to long time of connection, say we have `disconnect outdated connection for non-RT nodes` , will it be counted as `ConnectionReset` ?

sn_networking/src/event.rs
```diff
@@ -502,24 +502,43 @@ impl SwarmDriver {
...
+                                    // It is really difficult to match this error, due to being eg:
+                                    // Custom { kind: Other, error: Left(Left(Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })) }
+                                    // if we can match that, let's. But meanwhile we'll check the message
+                                    let error_msg = format!("{err:?}");
+                                    for problematic_error in &problematic_errors {
```
maqi: you can do:
```rust
if problematic_errors.iter().any(|err| error_msg.contains(err)) {
    warn!("Problematic error encountered: {error_msg}");
    there_is_a_serious_issue = true;
}
```

maqi (approved):

PR #1121, opened 2023-12-29
feat: restart faucet_server from breaking point

maqi:
### Summary generated by Reviewpad on 29 Dec 23 13:34 UTC
This pull request adds the feature to restart the faucet server from a breaking point. It introduces a new function `restart_faucet_server` that loads the previous wallet, deposits all valid cash notes into the wallet, and then starts the server. The existing function `run_faucet_server` is updated to call the `startup_server` function, which starts the server and handles the HTTP requests. Some additional helper functions and changes to existing code are also included in this patch.

### DIFF
--- a/sn_faucet/src/main.rs
+++ b/sn_faucet/src/main.rs
@@ -10,7 +10,7 @@ mod faucet_server;
 
 use clap::{Parser, Subcommand};
 use color_eyre::eyre::{bail, eyre, Result};
-use faucet_server::run_faucet_server;
+use faucet_server::{restart_faucet_server, run_faucet_server};
 use sn_client::{get_tokens_from_faucet, load_faucet_wallet_from_genesis_wallet, Client};
 use sn_logging::{LogBuilder, LogOutputDest};
 use sn_peers_acquisition::{parse_peers_args, PeersArgs};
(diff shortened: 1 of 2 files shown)


sn_faucet/src/main.rs
```diff
@@ -109,6 +109,15 @@ enum SubCmd {
     /// Starts an http server that will send tokens to anyone who requests them.
     /// curl http://localhost:8000/your-hex-encoded-wallet-public-address
     Server,
+    /// Restart the faucet_server from the last breaking point.
+    ///
+    /// Before firing this cmd, ensure:
+    ///   1, The previous faucet_server has been stopped.
+    ///   2, Invalid cash_notes have been removed from the cash_notes folder.
```
joshuef, line 116: Let's document how we might go about this. from the cli?

maqi: currently they got removed manually, as so far we can only figure out where is the broken point by manually check the logs.

I think with the later on addin feature of : no move on to next spend till the current one got verified,
then the broken point shall be less to happen, and we may not require to carry out such `restart` much.

Even we need a separate cli to carry out such `prune invalid cash_notes` operations,
it can be done in a separate PR ?

joshuef (approved):

PR #1127, opened 2024-01-04
chore(cli): print the datamap's entire hex addr for public files

RolandSherwin:
### Summary generated by Reviewpad on 04 Jan 24 12:56 UTC
This pull request updates the CLI code to print the entire hex address of the datamap for public files that are uploaded. It modifies the `upload_files` function in the `sn_cli/src/subcommands/files/mod.rs` file to display the hex address instead of the regular address.

### DIFF
--- a/sn_cli/src/subcommands/files/mod.rs
+++ b/sn_cli/src/subcommands/files/mod.rs
@@ -291,6 +291,12 @@ async fn upload_files(
             // log uploaded file information
             println!("**************************************");
             println!("*          Uploaded Files            *");
+            if !make_data_public {
+                println!("*                                    *");
+                println!("*  These are not public by default.  *");
+                println!("*     Reupload with `-p` option      *");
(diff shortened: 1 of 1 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -291,14 +291,21 @@ async fn upload_files(
             // log uploaded file information
             println!("**************************************");
             println!("*          Uploaded Files            *");
+            if !make_data_public {
```
maqi, line 294: shall this alert msg to be put in the other code path as well ?

maqi (approved):

sn_cli/src/subcommands/files/mod.rs
RolandSherwin: The other codepath already got this alter msg, I copied the msg from there 👍

PR #1147, opened 2024-01-06
chore: Simplify get store cost flow

maqi:
### Summary generated by Reviewpad on 06 Jan 24 10:38 UTC
This pull request introduces changes to a `merge.yml` workflow file, adding a new job for replication bench with heavy upload. The job contains steps to install dependencies, download test materials for uploading, build binaries, start a local network, create and fund a wallet, upload files, and stop the network and upload logs. 

Additionally, changes are made to `cmd.rs`, `lib.rs`, `record_store.rs`, and `node.rs`. In `cmd.rs`, a new field `key` of type `RecordKey` is added to the `GetLocalStoreCost` enum variant. In `lib.rs`, the `get_local_storecost` function now takes an additional parameter `key` of type `RecordKey`. In `record_store.rs`, the logic for calculating store cost is modified to check if a record with the specified key exists locally, and return a cost of zero if it does. In `node.rs`, the `GetStoreCost` request in the `handle_query` function now uses the `get_local_storecost` function with the specified record key, and creates a quote for the store cost based on the result.

Overall, these changes enhance the replication and store cost calculation functionalities in the codebase.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -279,19 +279,12 @@ impl NodeRecordStore {
     /// Calculate the cost to store data for our current store state
     #[allow(clippy::mutable_key_type)]
     pub(crate) fn store_cost(&self) -> NanoTokens {
-        let relevant_records_len = if let Some(distance_range) = self.distance_range {
-            let record_keys: HashSet<_> = self.records.keys().cloned().collect();
-            self.get_records_within_distance_range(&record_keys, distance_range)
-        } else {
-            warn!("No distance range set on record store. Returning MAX_RECORDS_COUNT for relevant records in store cost calculation.");
-            MAX_RECORDS_COUNT
-        };
-
-        let cost =
-            calculate_cost_for_relevant_records(relevant_records_len, self.received_payment_count);
+        let stored_records = self.records.len();
+        let cost = calculate_cost_for_records(stored_records, self.received_payment_count);
 
         // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
-        info!("Cost is now {cost:?} for {relevant_records_len:?} records stored of {MAX_RECORDS_COUNT:?} max");
+        info!("Cost is now {cost:?} for {stored_records:?} stored of {MAX_RECORDS_COUNT:?} max, {:?} times got paid.",
+            self.received_payment_count);
         NanoTokens::from(cost)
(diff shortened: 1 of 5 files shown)


sn_networking/src/record_store.rs
```diff
@@ -279,19 +279,11 @@ impl NodeRecordStore {
...
-            self.get_records_within_distance_range(&record_keys, distance_range)
-        } else {
-            warn!("No distance range set on record store. Returning MAX_RECORDS_COUNT for relevant records in store cost calculation.");
-            MAX_RECORDS_COUNT
-        };
-
-        let cost =
-            calculate_cost_for_relevant_records(relevant_records_len, self.received_payment_count);
+        let stored_records = self.records.len();
+        let cost = calculate_cost_for_relevant_records(stored_records, self.received_payment_count);
```
joshuef: Hmm, this is no longer for relevant records just all records, no? we'd need to change the name... but does that not skew pricing for longer running nodes? (or nodes that have experienced more churn?)

maqi: caculate `relevance` is costly, as for each key need to `sort peers to check self is in` .
hence changed to this much simplified version.

I think for longer running nodes, this will give them the benefit of reduce their claimed charge,
and let them have a higher chance of winning.
Which is somehow a reward for their long stay with the network ?

joshuef: what is `costly` here, `13ms/313ms`?

sn_networking/src/record_store.rs
```diff
@@ -279,19 +279,11 @@ impl NodeRecordStore {
...
-        };
-
-        let cost =
-            calculate_cost_for_relevant_records(relevant_records_len, self.received_payment_count);
+        let stored_records = self.records.len();
+        let cost = calculate_cost_for_records(stored_records, self.received_payment_count);
 
         // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
-        info!("Cost is now {cost:?} for {relevant_records_len:?} records stored of {MAX_RECORDS_COUNT:?} max");
+        info!("Cost is now {cost:?} for {stored_records:?} records stored of {MAX_RECORDS_COUNT:?} max");
```
joshuef: Can we add in the `received payments` to this log statement.

joshuef (approved):

PR #1149, opened 2024-01-07
chore: Move replicate flow to driver

maqi:
### Summary generated by Reviewpad on 07 Jan 24 11:03 UTC

- The function `get_local_storecost` in the `Network` implementation now takes an additional parameter `key: RecordKey`.
- The code that previously called `get_local_storecost` has been updated to pass the `key` parameter.
- A new function `trigger_interval_replication` has been added to the `Network` implementation.
- The new function is used to trigger interval replication in the network.
- The function `try_interval_replication` in the file `replication.rs` has been modified to be synchronous and call `network.trigger_interval_replication()` instead of being asynchronous.
- The file `replication.rs` also includes a new implementation of `try_interval_replication`, replacing the previous implementation.
- The file diff in `.github/workflows/merge.yml` includes the addition of a new job called "replication_bench_with_heavy_upload" that performs a replication benchmark with heavy upload.
- The file diff includes changes in the import statements, method calls, and match arm handling in various files.

These changes simplify the code, improve code readability, and add support for triggering interval replication. Additionally, the `store_cost` function in `record_store.rs` has been updated to use the correct number of stored records for cost calculation and include the `stored_records` variable in the log message.

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -19,7 +19,7 @@ use libp2p::{
     Multiaddr, PeerId,
 };
 use sn_protocol::{
-    messages::{Request, Response},
+    messages::{Cmd, Request, Response},
     storage::{RecordHeader, RecordKind, RecordType},
     NetworkAddress, PrettyPrintRecordKey,
 };
@@ -144,6 +144,8 @@ pub enum SwarmCmd {
         key: RecordKey,
         record_type: RecordType,
     },
+    /// Triggers interval repliation
+    TriggerIntervalReplication,
     /// The keys added to the replication fetcher are later used to fetch the Record from network
     AddKeysToReplicationFetcher {
         holder: PeerId,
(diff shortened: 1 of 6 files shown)


sn_networking/src/cmd.rs
```diff
@@ -302,80 +307,13 @@ pub struct SwarmLocalState {
...
-                if !keys_to_fetch.is_empty() {
-                    self.send_event(NetworkEvent::KeysToFetchForReplication(keys_to_fetch));
-                } else {
-                    trace!("no waiting keys to fetch from the network");
-                }
+            SwarmCmd::TriggerIntervalReplication => {
+                self.try_interval_replication()?;
+            }
+            SwarmCmd::AddKeysToReplicationFetcher { .. } => {
+                warn!("Shall no longer receive SwarmCmd::AddKeysToReplicationFetcher from node");
```
joshuef:
```suggestion
                warn!("SwarmCmd::AddKeysToReplicationFetcher should not be received from other nodes");
```

sn_node/src/node.rs
```diff
@@ -520,48 +519,7 @@ impl Node {
     }
 
     fn handle_node_cmd(&self, cmd: Cmd) {
```
joshuef: lets also remove this and the corresponding `CmdRequestReceived` if we're notneeding those anymore 🤔

joshuef (approved):

PR #1160, opened 2024-01-09
feat(client): extra sleep between chunk verification

maqi:
### Summary generated by Reviewpad on 09 Jan 24 09:29 UTC
This pull request introduces a new feature to the client code. Specifically, it adds an extra sleep between chunk verification in the `sn_networking` module. This sleep is added to avoid firing queries too close together and potentially choking the nodes further. The waiting time is determined based on the number of retry attempts. Additionally, this patch includes a minor change in the `sn_transfers` module, where a debug message is added to indicate the path where the cash note is being written.


joshuef:
Can you clarify why we need this?

joshuef (approved):

PR #1162, opened 2024-01-09
chore(client): refactor client upload flow

maqi:
### Summary generated by Reviewpad on 09 Jan 24 11:27 UTC

- Fixed bug in function A that was causing a crash
- Improved efficiency of function B by refactoring repetitive code
- Added new feature C for enhanced user experience
- Updated dependencies to latest versions

Please review the changes and provide feedback.

### DIFF
--- a/sn_client/src/files/upload.rs
+++ b/sn_client/src/files/upload.rs
@@ -11,15 +11,18 @@ use crate::{
     FilesApi, BATCH_SIZE, MAX_UPLOAD_RETRIES,
 };
 use bytes::Bytes;
-use futures::{stream::FuturesUnordered, StreamExt};
 use libp2p::PeerId;
-use sn_protocol::storage::{Chunk, ChunkAddress};
-use sn_transfers::NanoTokens;
-use std::{collections::HashSet, path::PathBuf};
-use tokio::{
-    sync::mpsc::{self},
-    task::JoinHandle,
+use sn_networking::PayeeQuote;
+use sn_protocol::{
+    storage::{Chunk, ChunkAddress},
+    NetworkAddress,
 };
+use sn_transfers::NanoTokens;
+use std::{
+    collections::{BTreeMap, BTreeSet, HashSet},
+    path::PathBuf,
+};
+use tokio::sync::mpsc::{self};
 use xor_name::XorName;
 
 /// The maximum number of sequential payment failures before aborting the upload process.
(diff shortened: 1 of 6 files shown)


sn_client/src/files/upload.rs
```diff
@@ -186,210 +218,170 @@ impl FilesUpload {
...
-                        )))
-                        .await?;
+                        )));
+                        uploaded += 1;
+                    } else {
+                        self.send_event(FileUploadEvent::AlreadyExistsInNetwork(
+                            ChunkAddress::new(chunk_info.name),
+                        ));
+                        skipped += 1;
                     }
```
RolandSherwin: Can we add comments regarding this special case if the cost is zero & based on if it is a re-upload. We can just copy over the previous comment.

sn_client/src/files/upload.rs
```diff
@@ -426,16 +418,158 @@ impl FilesUpload {
...
+                                (storage_cost, royalty_fees, wallet_client.balance()),
+                                reply_list,
+                            ))
+                        }
+                        Err(err) => {
+                            let reply_list: Vec<ChunkInfo> = std::mem::take(&mut chunk_info_map)
+                                .iter()
+                                .map(|(chunk_info, _)| chunk_info)
+                                .cloned()
+                                .collect();
```
RolandSherwin: Tiny one to avoid a clone.
```suggestion
                            let reply_list: Vec<ChunkInfo> = std::mem::take(&mut chunk_info_map)
                                .into_iter()
                                .map(|(chunk_info, _)| chunk_info)
                                .collect();
```

sn_client/src/files/upload.rs
```diff
@@ -186,210 +218,170 @@ impl FilesUpload {
...
+                        new_balance,
+                    });
+                }
+                TaskResult::UploadChunksOK(xor_name) => {
+                    trace!("Upload task uploaded chunk {xor_name:?}");
+                    let _ = on_going_uploadings.remove(&xor_name);
+                    uploaded += 1;
+                    self.send_event(FileUploadEvent::Uploaded(ChunkAddress::new(xor_name)));
+                }
+                TaskResult::ErrorEntries(error_list) => {
```
RolandSherwin, line 370: Can we add a comment here stating that, it is mandatory to throw an error inside `TaskResult::ErrorEntries`?
Since we don't have separate error variant for each task (say upload_chunks), we might not be clearing up `on_going_uploadings` if we failed to upload a chunk.
And the loop would not progress as `on_going_uploadings.len() < batch_size` would never happen. This will block the other tasks too.

RolandSherwin: Throwing an error breaks the loop.

RolandSherwin (approved):

sn_client/src/files/upload.rs
```diff
@@ -186,210 +218,175 @@ impl FilesUpload {
...
-    async fn upload(&mut self, chunks: Vec<(XorName, PathBuf)>) -> Result<()> {
+    /// There are three main task groups to upload chunks:
+    ///   1, Fetch the store_cost of a chunk
+    ///   2, Pay for the chunk based on the fetched store_cost
+    ///   3, Upload the chunk
+    /// Task groups can be parallely to each other, however requires input from previous group.
+    /// Within each group, for group 1 and 3, the mini tasks inside can be parallel to each other.
+    /// However for group 2, mini tasks inside has to be undertaken sequencely.
+    /// Hence, the flow is more suitable to be handled in a `parrallelized pipline` way,
+    /// instead of the current `batched pipline` way.
```
joshuef: This is fair commentary on the PR, but doesnt make sense as doc comments, 

Lets just clarify what's going on in the fn.

I think this is fine: 
```
/// There are three main task groups to upload chunks:
    ///   1, Fetch the store_cost of a chunk
    ///   2, Pay for the chunk based on the fetched store_cost
    ///   3, Upload the chunk
    /// Task groups can be run in parallel to each other, however sometimes requires input from previous group.
    /// Within each group, for group 1 and 3, the mini tasks inside can be parallel to each other.
    /// However for group 2, mini tasks inside has to be undertaken sequentially.```

(note there's some typo corrections there)

PR #1165, opened 2024-01-10
docs: OS-agnostic commands on the README.md in a 4-step list

JasonPaulGithub:
docs: four step list plus forum mentions and formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,21 @@
 # The Safe Network
 
-Own your data, share your spare space, get paid for doing so.
+[SafenetForum.org](https://safenetforum.org/)
 
-The Safe Network is a decentralised and autonomous data network, built atop kademlia and libp2p.
+Own your data. Share your disk space. Get paid for doing so.<br>
+The Data on the Safe Network is Decentralised, Autonomous, and  built atop of Kademlia and Libp2p.<br>
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,65 +41,35 @@ The Safe Network is a decentralised and autonomous data network, built atop kade
...
-## Running the network
-
-You'll need to set the `SAFE_PEERS` env variable to the multiaddress of a node when you set up a testnet.
-You can do this automatically at network startup using the following command (if you have ripgrep installed)
+## Run your MaidSafe Local Test Network in 4 easy Steps
+Before you start, make sure to have **[Rust installed](https://www.rust-lang.org/learn/get-started)** and the latest version with `rustup update`
+<br>
+1. Create your Local Test Network : <br>
+```bash
+Cargo run --bin testnet --features="local-discovery" -- --build-node --build-faucet --interval 100 --clean
```
jacderida: Sorry for the pedantic point, but probably better to have the `cargo` commands in lower case?

JasonPaulGithub: I changed it to upper case to workaround an issue I was having but I have learned since I am having issues with my editor. I can change this back. In the picture one arrow instead of two, and the single arrow command would not work = editor issue
Screenshot 2024-01-10 123333

PR #1178, opened 2024-01-11
fix(client): avoid dead lock with less chunks

maqi:
### Summary generated by Reviewpad on 11 Jan 24 16:24 UTC
This pull request fixes a deadlock issue related to handling chunks in the client. The patch avoids the deadlock by triggering a forced round of making leftover payments when there are no more chunks to process.


grumbach (approved):
LGTM

PR #1179, opened 2024-01-12
feat(faucet): download snapshot of maid balances

iancoleman:
The faucet will download a snapshot of maid balances when it starts the server.

This snapshot is saved to file so if the server is restarted it can use the same snapshot as when the faucet was originally started. This file is in `$HOME/.local/share/safe_snapshot` to avoid being cleaned up accidentally.

The faucet doesn't do anything with the snapshot yet, but in the future it may use it work out distributions for the maid addresses instead of blindly giving anyone 100 tokens.

### Summary generated by Reviewpad on 12 Jan 24 02:32 UTC
This pull request introduces a new feature to the faucet module. It adds the ability to download a snapshot of maid balances. The patch modifies the `Cargo.lock` and `Cargo.toml` files to include the required dependencies. It also adds a new struct `MaidBalance` and several functions to handle the downloading and parsing of the snapshot.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,9 +118,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.5"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d664a92ecae85fd0a7392615844904654d1d5f5514837f471ddef4a057aba1b6"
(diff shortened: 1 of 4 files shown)


iancoleman:
Bench safe cli running for 6 hours now... not sure what's up with that.

joshuef:
@iancoleman just updated atop main to rerun now. I'll have a look at the logs if the bench goes the same way

joshuef:
prior run was all green, just the constant faucet conflict we need to get this in _next_ i think.

joshuef (approved):

PR #1183, opened 2024-01-15
test(CI): ensure no extra leftover cash_note or payment files

maqi:
### Summary generated by Reviewpad on 15 Jan 24 14:39 UTC
This pull request updates the CI workflow file `merge.yml` to ensure that no extra leftover `cash_note` or payment files are present. It adds a new step that checks for the expected number of files in the `cash_notes` and `payments` directories and throws an error if there are more than expected. This helps maintain a clean and organized workspace.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -784,12 +784,26 @@ jobs:
           SN_LOG: "all"
         timeout-minutes: 5
 
-      - name: Confirm the cash_notes files
+      - name: Ensure no leftover cash_notes and payment files
         run: |
+          expected_cash_notes_files="1"
+          expected_payment_files="0"
           pwd
           ls $CLIENT_DATA_PATH/ -l
           ls $CLIENT_DATA_PATH/wallet -l
           ls $CLIENT_DATA_PATH/wallet/cash_notes -l
+          cash_note_files=$(ls $CLIENT_DATA_PATH/wallet/cash_notes | wc -l)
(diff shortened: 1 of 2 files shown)


.github/workflows/merge.yml
```diff
@@ -804,6 +818,29 @@ jobs:
...
+      - name: Ensure no leftover cash_notes and payment files
+        run: |
+          expected_cash_notes_files="1"
+          expected_payment_files="0"
+          pwd
+          ls $CLIENT_DATA_PATH/ -l
+          ls $CLIENT_DATA_PATH/wallet -l
+          ls $CLIENT_DATA_PATH/wallet/cash_notes -l
+          cash_note_files=$(find $CLIENT_DATA_PATH/wallet/cash_notes -type f | wc -l)
+          if (( $(echo "$cash_note_files > $expected_cash_notes_files" | bc -l) )); then
```
RolandSherwin, line 830: Why do we use `bc -l` here instead of using `if [ $expected_cash_notes_files -lt $cash_note_files ]; then` as you have done above?

maqi: we use both in different tests, so just testing both of them.
seems all works?

maybe we shall make all of them to just using one with a separate PR, if there is any preference?

RolandSherwin: That makes sense. This should be fine. I was just wondering why we were using a different approach there.

RolandSherwin (approved):

PR #1186, opened 2024-01-15
docs: new wallet client

JasonPaulGithub:


sn_client/src/wallet.rs
```diff
@@ -35,6 +35,7 @@ pub struct WalletClient {
 
 impl WalletClient {
     /// Create a new wallet client.
```
jacderida, line 37: Can this text here also go in the markdown document? Sorry, I've never used this feature before.

JasonPaulGithub: Yes, if this is agreed upon it will be the same for all documentation.
I left it in the rs file because it gives a quick overview to the reader (leaving the md for more detail)

sn_client/src/docs/wallet/new.md
```diff
@@ -0,0 +1,10 @@
+# Arguments
+
+* `client.clone()` - A cloned instance of the struct: `sn_client::Client`
+* `wallet` - An instance of the struct `sn_transfers::LocalWallet`
+
+# Example
+
+ ```ignore
+ let mut wallet_client = WalletClient::new(client.clone(), wallet);
```
joshuef: `ignore` is not the same ` dont run`, I think ? (Are the docs compiled and verified?)

Here we're missing all the setup and related info that would make this useful, no? (the why as you said)

joshuef (changes requested):

sn_client/src/wallet.rs
```diff
@@ -35,6 +35,7 @@ pub struct WalletClient {
...
+    #[doc = include_str!("docs/wallet/new.md")]
```
joshuef: I don't think there's any reason to move the docs to another file here. 

As I mentioned, there's an extra burden to move to update these docs, so to my mind unless they are _huge_ we'd want to keep them in the same file.

joshuef (changes requested):

PR #1190, opened 2024-01-16
feat: initial webtransport-websys wasm setup

joshuef:
### Summary generated by Reviewpad on 16 Jan 24 12:15 UTC

- In the `parse_peer_addr` function, some modifications have been made.
- Conditional compilations have been added for the `quic` and `webtransport-websys` features.
- For the `quic` feature, the `multiaddr` is created with the `QuicV1` protocol.
- For the `webtransport-websys` feature, the `multiaddr` is created with the `WebTransport` protocol.

In the `Cargo.toml` file, the following changes are present:

- The `tokio` dependency has been removed from the `dependencies` section and added to the `dev dependencies` section.
- The `[lib]` section has been added with `crate-type` set to `["cdylib", "rlib"]`.
- The `[target.'cfg(target_arch = "wasm32")'.dependencies]` section has been added with the `getrandom` dependency and its version set to `"0.2.12"` with the feature `"js"`.
- The `[dependencies]` section has been updated to include `webtransport-websys`.
- A blank line has been added at the end of the file.

In the `register.rs` file, the following changes are observed:

- The import `self_encryption::MIN_ENCRYPTABLE_BYTES` has been removed.
- The value of `MAX_REG_ENTRY_SIZE` has been changed from `MIN_ENCRYPTABLE_BYTES / 3 * 1024` to `1024`.
- The constant `MAX_REG_NUM_ENTRIES` remains unchanged.

The version of the `sn_protocol` dependency in the `Cargo.toml` file has been updated to include the "rpc" feature.

In the `lib.rs` file, the following change is present:

- The `#[cfg(feature = "rpc")]` attribute has been added to the module `safenode_proto` inside the `sn_protocol` library.

The changes in the `Cargo.toml` file include updates to the dependencies `self_encryption`, `bls`, `crdts`, `hex`, `rmp-serde`, `serde`, `thiserror`, and `tiny-keccak`.

In the `watch_only.rs` file, two changes have been made: 
- The import statement for the `fs2` crate has been added when the target architecture is not `wasm32`.
- A file lock has been added before opening a file for writing.

The `Cargo.toml` file now includes the dependencies `webtransport-websys`, `open-metrics`, `[lints]` section with `workspace` set to `true`, and `[target.'cfg(target_arch = "wasm32")'.dependencies]` section with the `getrandom` dependency.

The `driver.rs` file, in the `sn_networking` crate, has undergone several changes including conditional compilations for the `quic` and `webtransport-websys` features, updates to the listener address construction and transport setup based on feature flags.

Let me know if you need any further assistance!

### DIFF
--- a/sn_peers_acquisition/src/lib.rs
+++ b/sn_peers_acquisition/src/lib.rs
@@ -124,17 +124,18 @@ async fn get_network_contacts(args: &PeersArgs) -> Result<Vec<Multiaddr>> {
 pub fn parse_peer_addr(addr: &str) -> Result<Multiaddr> {
     // Parse valid IPv4 socket address, e.g. `1.2.3.4:1234`.
     if let Ok(addr) = addr.parse::<std::net::SocketAddrV4>() {
-        #[cfg(not(feature = "quic, webtransport-websys"))]
+        let multiaddr = Multiaddr::from(*addr.ip());
+        #[cfg(feature = "tcp")]
         // Turn the address into a `/ip4/<ip>/tcp/<port>` multiaddr.
-        let multiaddr = Multiaddr::from(*addr.ip()).with(Protocol::Tcp(addr.port()));
+        let multiaddr = multiaddr.with(Protocol::Tcp(addr.port()));
         #[cfg(feature = "quic")]
         // Turn the address into a `/ip4/<ip>/udp/<port>/quic-v1` multiaddr.
-        let multiaddr = Multiaddr::from(*addr.ip())
+        let multiaddr = multiaddr
             .with(Protocol::Udp(addr.port()))
             .with(Protocol::QuicV1);
-        #[cfg(feature = "webtransport-websys")]
+        #[cfg(target_arch = "wasm32")]
         // Turn the address into a `/ip4/<ip>/udp/<port>/webtransport-websys-v1` multiaddr.
-        let multiaddr = Multiaddr::from(*addr.ip())
+        let multiaddr = multiaddr
             .with(Protocol::Udp(addr.port()))
             .with(Protocol::WebTransport);
         return Ok(multiaddr);
--- a/sn_peers_acquisition/src/lib.rs
+++ b/sn_peers_acquisition/src/lib.rs
@@ -124,7 +124,7 @@ async fn get_network_contacts(args: &PeersArgs) -> Result<Vec<Multiaddr>> {
 pub fn parse_peer_addr(addr: &str) -> Result<Multiaddr> {
     // Parse valid IPv4 socket address, e.g. `1.2.3.4:1234`.
     if let Ok(addr) = addr.parse::<std::net::SocketAddrV4>() {
-        #[cfg(not(feature = "quic"))]
+        #[cfg(not(feature = "quic, webtransport-websys"))]
         // Turn the address into a `/ip4/<ip>/tcp/<port>` multiaddr.
         let multiaddr = Multiaddr::from(*addr.ip()).with(Protocol::Tcp(addr.port()));
         #[cfg(feature = "quic")]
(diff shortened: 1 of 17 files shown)


happybeing:
I can't tell what it's for from the description, is this to make the API accessible in WASM targets?  If so it's something I was thinking of having a go at when `vdash` stabilises, so might be able to help - eg putting together a demo.

joshuef:
@happybeing The end goal was to see how far we can get with rust->wasm compilation. 

This is some on the way maintenance that is required for that. 

I have [this branch](https://github.com/joshuef/safe_network/tree/WebTransport3) which is currently just a POC that shows the wasm version of `sn_client` can connect to a `websocket` version of nodes. Though it doesnt go far due to fun javascript compile time stuff. But it should be possible to push that further eg.

The main pain points come around the use of `time` in the browser (that branch uses `instant` as a dep for `wasm`, and other concurrency issues. `CondVar` is a current blocker, and debugging in the browser issssss unfun.

Have a poke at that by all means @happybeing ! I can't guarantee as/when anything more will move on this front, but more eyes can only help!

happybeing:
Are you taking about clients that run in a browser or just which use it as a front end (eg Tauri like)?

I differentiate because WASM is a target like a hardware platform with it's own cross platform runtimes, so no dependency on JS, web-view etc.

So I was planning to start with a CLI app (maybe just rebuild the Safe CLI with WASN target), with no need for JavaScript or a web-view. Later add a GUI with Tauri, and so on. But it sounds like you are trying to get something to run in a conventional browser.

I'm still finishing vdash atm, but WASM interests me for SN developers so keen to see what is feasible just taking the SN API and then build from that.

Ideally we could just compile all those rust crates to WASM and have a WASM based Safe CLI to start, but I'm sure there will be hurdles.

I hit some issues doing this with Rust and Go for p2p git (all solved by tweaking a low level Go lib) but I was targeting WASI which is trickier than pure WASM, and maybe closer to what you are doing.

Anyway, I'll take a look when I get fed up with vdash 😉. Thanks for the link

joshuef:
> Are you taking about clients that run in a browser or just which use it as a front end (eg Tauri like)?

The rust code running via wasm in the browser window itself was the exploratory aim on that branch.

> Ideally we could just compile all those rust crates to WASM and have a WASM based Safe CLI to start, but I'm sure there will be hurdles.

That's my ideal. I just wanted to see about GET.

And it's possible, the comms can work via websockets. But it's the porting of the heavily concurrent code to the browser which is currently problematic (alongside some deps not existing in the browser target).

happybeing:
Is there a reason you are targeting the browser right away?

You could start with native cross platform (non web) apps with a web GUI just by targetting the WASM runtime (and adding a web GUI with Tauri - no browser involved).

Targeting Browser / WASI would be great, enabling standard browser based apps, but will def be harder.

joshuef:
> You could start with native cross platform (non web) apps with a web GUI just by targetting the WASM runtime (and adding a web GUI with Tauri - no browser involved).

This could be done without wasm though, no? Tauri can just use the rust APIs directly?

joshuef:
It's honestly not a specific work target as yet, @happybeing this was some proof of concept research seeing how far we could go to know where we are. This PR gets some of the base work in that required a bit of rejigging, so it should make any further work _easier_, down the line.

sn_peers_acquisition/src/lib.rs
```diff
@@ -124,14 +124,20 @@ async fn get_network_contacts(args: &PeersArgs) -> Result<Vec<Multiaddr>> {
...
         // Turn the address into a `/ip4/<ip>/tcp/<port>` multiaddr.
-        let multiaddr = Multiaddr::from(*addr.ip()).with(Protocol::Tcp(addr.port()));
+        let multiaddr = multiaddr.with(Protocol::Tcp(addr.port()));
         #[cfg(feature = "quic")]
         // Turn the address into a `/ip4/<ip>/udp/<port>/quic-v1` multiaddr.
-        let multiaddr = Multiaddr::from(*addr.ip())
+        let multiaddr = multiaddr
             .with(Protocol::Udp(addr.port()))
             .with(Protocol::QuicV1);
+        #[cfg(target_arch = "wasm32")]
```
RolandSherwin, line 136: Is it possbile to enable more than one of these features here? This might lead to unexpected behaviours. Can we somehow have a check for use of multiple incompatible features?

RolandSherwin (approved):

sn_peers_acquisition/src/lib.rs
joshuef: We'll add some compile checks etc in down the line. I'll be following this up 👍

PR #1194, opened 2024-01-17
feat: set quic as default transport

joshuef:
BREADKING CHANGE: quic and tcp are not compatible

### Summary generated by Reviewpad on 17 Jan 24 09:08 UTC
This pull request updates the default transport to use QUIC instead of TCP. This is a breaking change as QUIC and TCP are not compatible. The changes are made in the `sn_client` and `sn_node` Cargo.toml files, adding "quic" as a default feature.

### DIFF
--- a/sn_client/Cargo.toml
+++ b/sn_client/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/maidsafe/safe_network"
 version = "0.102.1"
 
 [features]
-default=[]
+default=["quic"]
 local-discovery=["sn_networking/local-discovery"]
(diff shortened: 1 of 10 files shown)


b-zee (approved):

joshuef:
Blocked until node manager is updated it seems!

joshuef:
https://github.com/maidsafe/safe_network/pull/1203 is this but the commit github doesnt seem to want to check at the moment 🤔

joshuef:
#1203 merged, which is this + fmt fix.

PR #1197, opened 2024-01-17
docs: new wallet docs

JasonPaulGithub:
3 new  docs in wallet rs

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -59,22 +59,70 @@ impl WalletClient {
         Self { client, wallet }
     }
 
-    /// Stores the wallet to disk.
+    /// Stores the wallet to the local wallet directory.
+    /// # Example
+    /// ```no_run
+    /// # use sn_client::{Client, WalletClient, Error};
+    /// # use tempfile::TempDir;
+    /// # use bls::SecretKey;
+    /// # use sn_transfers::{LocalWallet, MainSecretKey};
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(),Error>{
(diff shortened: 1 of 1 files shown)


sn_client/src/wallet.rs
```diff
@@ -59,21 +59,70 @@ impl WalletClient {
...
+    /// ```no_run
+    /// # use sn_client::{Client, WalletClient, Error};
+    /// # use tempfile::TempDir;
+    /// # use bls::SecretKey;
+    /// # use sn_transfers::{LocalWallet, MainSecretKey};
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(),Error>{
+    /// # let client = Client::new(SecretKey::random(), None, false, None, None).await?;
+    /// # let tmp_path = TempDir::new()?.path().to_owned();
+    /// # let mut wallet = LocalWallet::load_from_path(&tmp_path,Some(MainSecretKey::new(SecretKey::random())))?;
```
RolandSherwin, line 73: I'm not really familiar with doc tests, but why do we have certain code commented out here?

sn_client/src/wallet.rs
```diff
@@ -59,21 +59,70 @@ impl WalletClient {
...
+    /// ```
     pub fn store_local_wallet(&mut self) -> WalletResult<()> {
         self.wallet.deposit_and_store_to_disk(&vec![])
     }
 
-    /// Get the wallet balance
+    /// Display the wallet balance
+    /// # Example
+    /// ```no_run
+    /// // yDispla the wallet balance in the terminal
```
RolandSherwin: A small typo here, `yDispla` -> `Display`

RolandSherwin (approved):

PR #1207, opened 2024-01-18
Unused sss

JasonPaulGithub:
### Summary generated by Reviewpad on 18 Jan 24 15:13 UTC
This pull request contains two patches.

Patch 1/2:
- Adds documentation updates for the wallet module.
- Updates the documentation for the methods in the `WalletClient` struct.
- Adds new examples and clarifies the functionality of the methods.

Patch 2/2:
- Makes a feature change by removing the `send_signed_spends` method from the public API.
- Adds a comment indicating that the method is currently unused.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -26,8 +26,7 @@ use std::{
 use tokio::{task::JoinSet, time::sleep};
 use xor_name::XorName;
 
-/// A wallet client can be used to send and
-/// receive tokens to/from other wallets.
+/// A wallet client can be used to send and receive tokens to and from other wallets.
 pub struct WalletClient {
     client: Client,
     wallet: LocalWallet,
@@ -37,8 +36,8 @@ impl WalletClient {
     /// Create a new wallet client.
     ///
     /// # Arguments
-    /// * `client` - A instance of the struct [`sn_client::Client`](self::Client)
-    /// * `wallet` - An instance of the struct [`sn_transfers::LocalWallet`](self::LocalWallet)
+    /// * `client` - A instance of the struct [`sn_client::Client`](Client)
+    /// * `wallet` - An instance of the struct [`sn_transfers::LocalWallet`](LocalWallet)
     ///
     /// # Example
     /// ```no_run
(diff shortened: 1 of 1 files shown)


sn_client/src/wallet.rs
```diff
@@ -201,8 +279,10 @@ impl WalletClient {
     }
 
     /// Send signed spends to another wallet.
-    /// Can optionally verify the store has been successful (this will attempt to GET the Spend from the network)
-    pub async fn send_signed_spends(
```
JasonPaulGithub, line 205: The change is here. The rest of the code comes from a branch that's waiting to merge.

sn_client/src/wallet.rs
```diff
@@ -127,7 +126,33 @@ impl WalletClient {
         self.wallet.unconfirmed_spend_requests()
     }
 
-    /// Get the payment for a given network address
+    /// Get the payment detail from a given address within the network.
```
RolandSherwin: I think this function returns the cached `Payment` for the provided `NetworkAddress`. I.e., we pay for addresses before we can store anything at that location, and this retrieves the payment that the client has made.

sn_client/src/wallet.rs
```diff
@@ -156,12 +207,39 @@ impl WalletClient {
...
     /// Remove CashNote from available_cash_notes
+    //TODO: Unused
     pub fn mark_note_as_spent(&mut self, cash_note_key: UniquePubkey) {
         self.wallet.mark_note_as_spent(cash_note_key);
     }
 
-    /// Send tokens to another wallet.
-    /// Can optionally verify the store has been successful (this will attempt to GET the Spend from the network)
+    /// Send tokens to another wallet. Can also verify the store has been successful.
+    /// This will attempt to GET the Spend from the network.
```
RolandSherwin: The verification is carried out by performing a GET. I think this applies to all the other similar functions with the `verify` option.
```suggestion
    /// Send tokens to another wallet. Can also verify the store has been successful, by
    /// attempting to GET the Spend from the network.
```

sn_client/src/wallet.rs
```diff
@@ -267,7 +347,8 @@ impl WalletClient {
     ///   `skipped_chunks` is the list of content already exists in network and no need to upload
     ///
     /// Note storage cost is _per record_, and it's zero if not required for this operation.
-    /// This can optionally verify the store has been successful (this will attempt to GET the cash_note from the network)
+    /// This can optionally verify the store has been successful.
+    /// This will attempt to GET the cash_note from the network
     pub async fn pay_for_storage(
         &mut self,
         content_addrs: impl Iterator<Item = NetworkAddress>,
```
RolandSherwin: here `verify_store = true`, so I think we can mention that the verification is mandatory here.

maqi (changes requested):
The failing CIs shall not be related to the PR, having a re-run shall be enough.

However, I'm bit confused about what this PR is related to the PR 1208 & PR 1209,
as the commit msgs looks quite similiar to each other.

Also, better remove the `merge` commit in the PR.

sn_client/src/wallet.rs
JasonPaulGithub: I have changed the line to this:
    /// * `verify_store` - A boolean to verify store. Set this to true for mandatory verification.
which helps to the use to understand that the verification is mandatory if set to true. 
Hope this helps :)

maqi (approved):

PR #1210, opened 2024-01-19
feat(faucet): fetch maid pubkeys to enable encryption of distributions

iancoleman:
Faucet will distribute tokens to existing maid holders (from the snapshot).

Each distribution will be encrypted using the maid public key, so only the maid owner can decrypt their distribution.

This feature enables an initial set of public keys to be used for the initial distribution. This means the live faucet will hold less tokens at start, since those distributions will be spent to maid owners when the faucet starts.

This doesn't create the distributions, but is an essential step toward it.

### Summary generated by Reviewpad on 19 Jan 24 02:03 UTC

1. `Cargo.toml` - Added a new dependency on `bitcoin` version `0.31.0`.
2. `faucet_server.rs` - Various changes including imports, function additions/updates, and logging statements. Detailed changes provided in the previous message.
3. `Cargo.lock` - Multiple changes including package additions and updates. Detailed changes provided in the previous message.

Please review the above changes and let me know if you need further assistance.


sn_faucet/src/faucet_server.rs
```diff
@@ -151,7 +164,7 @@ fn deposit(root_dir: &Path) -> Result<()> {
     Ok(())
 }
 
-fn load_maid_snapshot() -> Result<Vec<MaidBalance>> {
+fn load_maid_snapshot() -> Result<HashMap<String, u32>> {
```
maqi: better make this `HashMap<String, u32>` as a type, and have a comment explain what that `String` and `u32` is for

sn_faucet/src/faucet_server.rs
```diff
@@ -202,13 +216,112 @@ fn parse_snapshot(json_str: String) -> Result<Vec<MaidBalance>> {
...
         let msg = format!("Incorrect snapshot total, got {total} want {supply}");
         return Err(eyre!(msg));
     }
     // log the total number of balances that were parsed
     info!("Parsed {} maid balances from the snapshot", balances.len());
-    Ok(balances)
+    Ok(balances_map)
+}
+
+fn load_maid_pubkeys() -> Result<HashMap<String, String>> {
```
maqi: better have a comment explains what are these `String, String` for

sn_faucet/src/faucet_server.rs
```diff
@@ -202,13 +216,112 @@ fn parse_snapshot(json_str: String) -> Result<Vec<MaidBalance>> {
...
+            }
+        };
+        pubkeys.insert(address, pk_hex);
+    }
+    info!("{} pubkeys after reading existing files", pubkeys.len());
+    // load from blockchain list on internet
+    info!("Fetching pukeys from {PUBKEYS_URL}");
+    let response = minreq::get(PUBKEYS_URL).send()?;
+    // check the request is ok
+    if response.status_code != 200 {
```
maqi, line 275: make `200` as a const ?

sn_faucet/src/faucet_server.rs
iancoleman: Changed it to `HashMap<MaidAddress, NanoTokens>`

Is it better to also make a `type Snapshot = HashMap<MaidAddress, NanoTokens>` or is that overkill?

sn_faucet/src/faucet_server.rs
iancoleman: Changed it to `HashMap<MaidAddress, MaidPubkey>`

sn_faucet/src/faucet_server.rs
iancoleman: Added `const HTTP_STATUS_OK: i32 = 200` since `minreq` crate doesn't have these consts.

sn_faucet/src/faucet_server.rs
```diff
@@ -11,17 +11,24 @@ use color_eyre::eyre::{eyre, Result};
...
 use sn_transfers::{LocalWallet, NanoTokens};
+use std::collections::HashMap;
 use std::path::{self, Path, PathBuf};
+use std::str::FromStr;
 use tiny_http::{Response, Server};
 use tracing::{debug, error, info, trace};
 
 const SNAPSHOT_FILENAME: &str = "snapshot.json";
 const SNAPSHOT_URL: &str = "https://api.omniexplorer.info/ask.aspx?api=getpropertybalances&prop=3";
+const PUBKEYS_URL: &str = "https://pastebin.com/raw/pUm6tVRN";
```
grumbach: Instead of loading the keys from a pastebin link, wouldn't it be safer to have those in a json/csv file locally in the repo?

iancoleman: Oh I like that idea! Before I commit this, just a quick sanity check on the size of the file and whether it's not too big for the repo... 155KB for 1227 addresses is an average of 126B per address. For about 16,000 addresses holding maid that would be a file about 2MB. Is that acceptable to have committed in the repo? It seems ok to me but wanted to check first.

joshuef: Very acceptable 👍

sn_faucet/src/faucet_server.rs
maqi: Seems the type is to be used in many places.
So, I'd vote for to make it `type Snapshot = HashMap<MaidAddress, NanoTokens>` .

iancoleman: Added the `Snapshot` type in https://github.com/maidsafe/safe_network/pull/1210/commits/0031538577d9c0e58e033c84d345fc8d6d0b56b6 thanks for the feedback :)

sn_faucet/src/faucet_server.rs
iancoleman: Added this change in https://github.com/maidsafe/safe_network/pull/1210/commits/779a39db98acce7f0e01b90b8f05bc58b2e026d0 and also allowed the request to fail without preventing the faucet to start (ie http status 404 allows the process to continue, not fail)

iancoleman:
@joshuef I'm happy with where this is at, all the comments have been addressed. Once it's in I'll start on the next step of the process.

joshuef:
Aye, i was planning on holding this off until `quic` had been tested, but we can roll it all in together as the deployer is tripping us up the last week, so I won't hold you up here any longer 👍

joshuef (approved):

PR #1221, opened 2024-01-24
chore(test): lift up the expectations within address sim test

maqi:
### Summary generated by Reviewpad on 24 Jan 24 11:57 UTC
This pull request updates the expectations within the address simulation test in the `record_store.rs` file. The test now checks that the ratio of `max_store_cost / min_store_cost` is less than 100 and the ratio of `max_earned / min_earned` is less than 1000 after 50 iterations.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -985,11 +985,11 @@ mod tests {
             if iteration == 50 {
                 assert_eq!(0, empty_earned_nodes, "every node has earnt _something_");
                 assert!(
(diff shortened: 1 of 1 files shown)


joshuef (approved):

happybeing:
I've a suggestion for another useful test in this area, which would determine how effective killing a low earning node and replacing with a new node is.

I tried this in the last test and the result was good - my new nodes never caught up the older nodes although my measurement was just by eye.

So I think a good test would be to see if restarting low earning nodes can improve earning rate, if that can be done.

joshuef (approved):

PR #1224, opened 2024-01-24
feat: client webtransport-websys feat

joshuef:
### Summary generated by Reviewpad on 24 Jan 24 15:33 UTC
This pull request includes multiple file diffs. Here is a summarized overview of the changes:

1. The `cmd.rs` file in the `sn_networking` module has several modifications, including new imports, changes in enum variants and struct methods, variable updates, and code formatting changes.

2. The `replication_fetcher.rs` file has changes related to imports, removing and adding features, and updating comments and code related to parallel fetches.

3. The `Cargo.toml` file removes the "quic" feature and includes only the "metrics" feature.

4. The `wallet.rs` file has changes in imports related to time and duration, along with a comment explaining the usage of a wallet client for token transfers.

5. The `lib.rs` file has modifications in the `genesis_multi_addr` variable and related assertions in tests.

6. The README.md file has updates regarding transport protocols, architectures, and building for `wasm32`.

7. The diff adds a new `target_arch.rs` file with conditionally imported modules based on the target architecture. Changes also include importing `spawn_local` for handling asynchronous tasks.

8. The `lib.rs` file includes changes in importing the `Duration` struct from the `tokio::time` module, replacing the previous import from the `std::time` module.

9. The diff for the file `download.rs` includes import changes related to time, removing and replacing import statements.

10. The `Cargo.toml` file has changes in dependency versions, feature modifications, and wasm compilation options.

11. The file `client.rs` in `tests/common` has changes in imports related to time.

12. The `upload.rs` file has modifications related to logging and changing the method of reading chunk bytes.

13. The diff includes changes in importing the `sleep` function and related replacements for time-related functions.

14. The `Cargo.toml` file includes changes in features, dependencies, and versions.

15. The `driver.rs` file within the `sn_networking` module has various changes, including conditional compilation, transport updates, logging, and configurations based on build and target architectures.

16. The `node_registry.rs` file has changes in constructing the `addr` variable based on feature flags for different build configurations.

17. The diff introduces modifications related to imports, including the addition and removal of import statements for various modules, types, and libraries.

18. The `Cargo.toml` file includes changes in dependencies, removing and adding features.

19. The diff includes changes in importing modules, types, and constants in the file.

20. The `Cargo.toml` file has changes in dependencies, including the removal of the "default" and "quic" features.

Please let me know if you need more details or specific information for any of the file diffs.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -121,6 +121,43 @@ jobs:
           # we do many more runs on the nightly run
           PROPTEST_CASES: 50 
 
+  wasm:
+    if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
+    name: Wasm builds
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+
+      - name: Build client for wasm
+        # wasm pack doesnt support workspaces
+        # --dev to avoid a loong optimisation step
+        run: cd sn_client && wasm-pack build --dev 
+        timeout-minutes: 30
+
+  websocket:
+    if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
+    name: Standard Websocket builds
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build all for `websockets`
+        run: cargo build --features="websockets"
+        timeout-minutes: 30
+
   e2e:
     if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
     name: E2E tests
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -121,43 +121,6 @@ jobs:
           # we do many more runs on the nightly run
           PROPTEST_CASES: 50 
 
-  wasm:
-    if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
-    name: Wasm builds
-    runs-on: ubuntu-latest
-    
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-
-      - name: Install wasm-pack
-        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
-
-      - name: Build client for wasm
-        # wasm pack doesnt support workspaces
-        # --dev to avoid a loong optimisation step
-        run: cd sn_client && wasm-pack build --dev 
-        timeout-minutes: 30
-
-  websocket:
-    if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
-    name: Standard Websocket builds
-    runs-on: ubuntu-latest
-    
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
-      - uses: Swatinem/rust-cache@v2
-
-      - name: Build all for `websockets`
-        run: cargo build --features="websockets"
-        timeout-minutes: 30
-
   e2e:
     if: "!startsWith(github.event.head_commit.message, 'chore(release):')"
     name: E2E tests
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -23,6 +23,7 @@ use libp2p::{
 #[cfg(feature = "open-metrics")]
 use prometheus_client::registry::Registry;
 use rand::{thread_rng, Rng};
+use sn_networking::target_arch::{interval, spawn, timeout, Instant};
 use sn_networking::{
     multiaddr_is_global, Error as NetworkError, GetRecordCfg, GetRecordError, NetworkBuilder,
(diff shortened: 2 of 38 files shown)


.github/workflows/merge.yml
```diff
@@ -121,6 +121,43 @@ jobs:
           # we do many more runs on the nightly run
           PROPTEST_CASES: 50 
 
+  wasm:
```
maqi: This two CI runs doesn't involve any function check.
may they better to be moved into a separate workflow ?
say wasm_compatible_check.yml ?

sn_client/src/api.rs
```diff
@@ -74,9 +89,16 @@ impl Client {
 
         info!("Startup a client with peers {peers:?} and local {local:?} flag");
         info!("Starting Kad swarm in client mode...");
+        debug!("Starting Kad swarm in client mode.1..");
+        trace!("Starting Kad swarm in client mode..2.");
```
maqi: don't get it, why have 3 different client mode index with three different log level ?

sn_client/src/api.rs
```diff
@@ -38,23 +38,38 @@ use sn_protocol::{
...
 use std::{
     collections::{HashMap, HashSet},
     num::NonZeroUsize,
     path::PathBuf,
-    time::Duration,
 };
+#[cfg(not(target_arch = "wasm32"))]
 use tokio::task::spawn;
+use tokio::time::Duration;
+#[cfg(not(target_arch = "wasm32"))]
```
maqi: to avoid having this kind of multiple target_arch dependent compliation guard across multiple files,
better to have them in a separate file, then expose ?
i.e. say we have a separate file called `target_arch_selection.rs`
then all other files can just use like: `crate::target_arch_selection::{interval, timeout, spawn, Instant}`

.github/workflows/merge.yml
joshuef: We want them run on `merge` to ensure that the code compiles. Any failure there should reject the PR.

sn_client/src/api.rs
joshuef: ah, that should be cleaned up!

sn_client/src/api.rs
```diff
@@ -74,9 +89,16 @@ impl Client {
 
         info!("Startup a client with peers {peers:?} and local {local:?} flag");
         info!("Starting Kad swarm in client mode...");
+        debug!("Starting Kad swarm in client mode.1..");
+        trace!("Starting Kad swarm in client mode..2.");
+
+        #[cfg(target_arch = "wasm32")]
+        let root_dir = PathBuf::from("dumb");
```
maqi: same idea as above, having a unique function, say `get_root_dir` in seperate `target_arch_selection.rs` file
then here just needs to be:
`let root_dir = crate::target_arch_selection::get_root_dir();`

sn_client/src/api.rs
joshuef: ooh, nice idea.

sn_client/src/lib.rs
```diff
@@ -35,6 +35,71 @@ pub(crate) use error::Result;
...
+// This is like the `main` function, except for JavaScript.
+#[cfg(target_arch = "wasm32")]
+#[wasm_bindgen(start)]
+pub async fn main_js() -> std::result::Result<(), JsValue> {
+    // This provides better error messages in debug mode.
+    // It's disabled in release mode so it doesn't bloat up the file size.
+    // #[cfg(debug_assertions)]
+    console_error_panic_hook::set_once();
+
+    // Your code goes here!
```
maqi: hmm..  the comment can now be removed ?

sn_networking/src/driver.rs
```diff
@@ -465,13 +486,29 @@ impl NetworkBuilder {
...
+        // cannot be built for wasm32
+        #[cfg(all(not(feature = "websockets"), not(target_arch = "wasm32")))]
         let main_transport = TokioTransport::new(TransportConfig::new(&self.keypair))
             .map(|(peer_id, muxer), _| (peer_id, StreamMuxerBox::new(muxer)))
             .boxed();
 
-        #[cfg(feature = "tcp")]
-        let main_transport = TokioTransport::new(TransportConfig::default())
+        #[cfg(target_arch = "wasm32")]
+        let main_transport = WebSocketTransport::default()
```
maqi, line 483: same idea as before, move to separate file and create a `create_main_transport` function

sn_networking/src/lib.rs
```diff
@@ -87,6 +95,8 @@ pub const fn close_group_majority() -> usize {
 
 /// Max duration for all GET attempts
 const MAX_GET_RETRY_DURATION_MS: u64 = 6800;
+
+#[cfg(not(target_arch = "wasm32"))]
```
maqi, line 97: wasm doesn't need `MAX_GET_RETRY_DURATION` ?
I thought this is a non-transport-dependent retry config ?

sn_client/src/api.rs
joshuef: I don't think its needed as this is the only location atm.

.github/workflows/merge.yml
maqi: I mean move them into a separate workflow file, similar to `memcheck.yml`
will still be run per PR and block merge if any error happens.

the `merg.yml` to me shall just contain tests that involves function/feature related tests.
the naming of `merge` just because it is the first workflow file we have,
and may better to be renamed.
it does now contain too many separate sets of runs as I feel.

sn_networking/src/lib.rs
joshuef: we cannot (at the moment) easily have backoff with wasm32. backoff uses tokio right now.

We can probably rejig this, but i'd say that's an optimisation atm. The main aim here (i think) is to get the wasm code building and checked?

sn_networking/src/driver.rs
joshuef: I did look at this, but things got unwieldy fast. 

I'm not against it, but I don't think it's a blocker for this PR.

maqi (approved):

PR #1229, opened 2024-01-25
feat(nodes): encrypt all records before disk, decrypt on get

### LINKED ISSUE #1158: Define High Level Client APIs, improvements and account packets
## Account packet

- [ ] Decide what goes in the account packet
    - [ ] Files
    - [ ] Wallet
    - [ ] data labels? (can we allow for that later?)
    - [ ] SafeId (for later?)
    - [ ] other things?
- [ ]  Decide on serialisation in the record.
- [ ] Create the data types for this.
- [ ] Decide how we want to manage encryption/storage. (Data stored at PK and signed to prove its allowed?)
- [ ] Decide how to generate this for users? (multi-word pass generated for folk? KISS)

## Files System

We should start fleshing out higher level client APIs to properly handle fiilessytem and small data.

We want to:

- [ ] Easily keep some metadata
- [ ] Ensure small files are encrypted
- [ ] Decide how to properly handle keys for data (ie, register encryption. small file encryption; do small (<3 bytes) files just go in registers? are they encrypted using the account packet keys)
- [ ] Enable linked registers

Ideally:

- This would be simple but streamlined enough that FUSE integration should not be a big issue.

## Other considerations?

If we make the account packet a key... could we derive some DNS around key finding to enable simple safe ids? (PK hash matching etc...)

joshuef:
fixes #1158

### Summary generated by Reviewpad on 25 Jan 24 15:08 UTC
This pull request introduces a new feature to the codebase. It modifies the `NodeRecordStore` to encrypt all records before writing them to disk and decrypts them when retrieving. The encryption key is randomly generated at node startup. Additionally, the patch includes some bug fixes and code improvements.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -9,6 +9,10 @@
 
 use crate::target_arch::{spawn, Instant};
 use crate::{cmd::SwarmCmd, event::NetworkEvent, send_swarm_cmd};
+use aes_gcm_siv::{
+    aead::{Aead, KeyInit, OsRng},
+    Aes256GcmSiv, Nonce,
+};
 use libp2p::{
     identity::PeerId,
     kad::{
@@ -18,6 +22,7 @@ use libp2p::{
 };
 #[cfg(feature = "open-metrics")]
 use prometheus_client::metrics::gauge::Gauge;
+use rand::RngCore;
 use sn_protocol::{
     storage::{RecordHeader, RecordKind, RecordType},
     NetworkAddress, PrettyPrintRecordKey,
@@ -56,6 +61,9 @@ pub struct NodeRecordStore {
     record_count_metric: Option<Gauge>,
     /// Counting how many times got paid
     received_payment_count: usize,
+    /// Encyption cipher for the records, randomly generated at node startup
+    /// Plus a 4 byte nonce starter
+    encryption_details: (Aes256GcmSiv, [u8; 4]),
 }
 
 /// Configuration for a `DiskBackedRecordStore`.
@@ -79,6 +87,16 @@ impl Default for NodeRecordStoreConfig {
     }
 }
 
+/// Generate an encryption nonce for a given record key and nonce_starter bytes.
+fn generate_nonce_for_record(nonce_starter: &[u8; 4], key: &Key) -> Nonce {
+    let mut nonce_bytes = nonce_starter.to_vec();
+    nonce_bytes.extend_from_slice(key.as_ref());
+    // Ensure the final nonce is exactly 96 bits long by padding or truncating as necessary
+    // https://crypto.stackexchange.com/questions/26790/how-bad-it-is-using-the-same-iv-twice-with-aes-gcm
+    nonce_bytes.resize(12, 0); // 12 (u8) * 8 = 96 bits
+    Nonce::from_iter(nonce_bytes)
+}
+
 impl NodeRecordStore {
     /// Creates a new `DiskBackedStore` with the given configuration.
     pub fn with_config(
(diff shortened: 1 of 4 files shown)


sn_networking/src/record_store.rs
```diff
@@ -252,27 +287,42 @@ impl NodeRecordStore {
...
-                    info!("Wrote record {record_key:?} to disk! filename: {filename}");
-
-                    SwarmCmd::AddLocalRecordAsStored {
-                        key: r.key,
-                        record_type,
+            let mut nonce_bytes = nonce_starter.to_vec();
+            nonce_bytes.extend_from_slice(r.key.as_ref());
+            // Ensure the final nonce is exactly 96 bits long by padding or truncating as necessary
+            // https://crypto.stackexchange.com/questions/26790/how-bad-it-is-using-the-same-iv-twice-with-aes-gcm
+            nonce_bytes.resize(12, 0); // 12 (u8) * 8 = 96 bits
```
iancoleman: Maybe a function `nonce(starter, key) -> GenericArray` could help keep consistency instead of repeating the logic from L144?
The `println` below on L302 looks like a leftover, should it be removed?
Otherwise looking good :)

happybeing:
I thought this was no longer needed when the minimum file size for encryption was reduced to three bytes.

Is there still a need for this?

joshuef:
@happybeing this is not about users encrypting data, but preventing node operators from storing _unencrypted_ data to disk.

joshuef:
Updated for that @iancoleman . Looking pretty tidy there at last I think.

happybeing:
> @happybeing this is not about users encrypting data, but preventing node operators from storing unencrypted data to disk

Exactly, and no longer needed because of the reduction of min chunk size for encryption to three bytes.

Since all chunks are now encrypted before upload there's no need for nodes to do this. Unless you are worried about one and two byte chunks?

iancoleman:
> chunks are now encrypted before upload

A malicious client may not encrypt before upload, ie send plaintext bytes to nodes, so the node protects itself by encrypting before storing on disk.

sn_networking/src/record_store.rs
```diff
@@ -117,26 +140,39 @@ impl NodeRecordStore {
...
+                if let Ok(value) = cipher.decrypt(&nonce, ciphertext.as_ref()) {
+                    let record = Record {
+                        key: key.clone(),
+                        value,
+                        publisher: None,
+                        expires: None,
+                    };
+                    return Some(Cow::Owned(record));
+                } else {
+                    error!("Error while decrypting record. filename: {filename}");
```
maqi: better log the decrypt error to ease the debugging

sn_networking/src/record_store.rs
```diff
@@ -252,27 +288,33 @@ impl NodeRecordStore {
...
-                }
-            };
+                        SwarmCmd::RemoveFailedLocalRecord { key: r.key }
+                    }
+                };
 
-            send_swarm_cmd(cloned_cmd_sender, cmd);
+                send_swarm_cmd(cloned_cmd_sender, cmd);
+            } else {
+                warn!("Failed to encrypt record {record_key:?} filename: {filename}");
```
maqi: log the error of encryption to ease the debugging

sn_networking/src/record_store.rs
```diff
@@ -56,6 +61,9 @@ pub struct NodeRecordStore {
     record_count_metric: Option<Gauge>,
     /// Counting how many times got paid
     received_payment_count: usize,
+    /// Encyption cipher for the records, randomly generated at node startup
+    /// Plus a 4 byte nonce starter
+    encryption_details: (Aes256GcmSiv, [u8; 4]),
```
maqi, line 66: As I understand, this `encryption_details` shall be written to disk in case `restart of node` is allowed.
Otherwise, if a node restart with the same key, it won't be able to read the previously stored records.

ideally, I think the Aes cipher and the nonce shall be deduced from the private key instead, to avoid have to store extra stuff to disk.

joshuef: I was not imagining to store this to disk. Node restart here will necesitate wiping the data (I thought any restart is always with new keys at the moment, eg).

maqi (approved):

PR #1235, opened 2024-01-26
Wallet client branch

JasonPaulGithub:
### Summary generated by Reviewpad on 26 Jan 24 17:55 UTC
This pull request adds comments and examples for several functions in the WalletClient and Client structs. The comments provide explanations of the arguments, return values, and examples of how to use the functions. This will improve the code readability and make it easier for developers to understand and use these functions.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -665,8 +665,9 @@ impl Client {
     /// This can optionally verify the spends have been correctly stored before returning
     ///
     /// # Arguments
-    /// * spend_requests - [Iterator](Iterator)<Item = [SignedSpend](SignedSpend)>
+    /// * spend_requests - [Iterator](Iterator)<[SignedSpend](SignedSpend)>
     /// * verify_store - Boolean. Set to true for mandatory verification via a GET request through a Spend on the network.
+    ///
     /// # Example
     /// ```no_run
     /// use sn_client::{Client, WalletClient, Error};
@@ -759,7 +760,7 @@ impl Client {
     /// let transfer = Transfer::from_hex("13abc").unwrap();
     /// // An example for using client.receive() for cashNotes
     /// let cash_notes = match client.receive(&transfer, &wallet).await {
-    ///                 Ok(cashnotes) => cashnotes,
+    ///                 Ok(cash_notes) => cash_notes,
     ///                 Err(err) => {
     ///                     println!("Failed to verify and redeem transfer: {err:?}");
     ///                     error!("Failed to verify and redeem transfer: {err:?}");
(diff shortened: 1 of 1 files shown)


sn_client/src/wallet.rs
```diff
@@ -343,15 +343,15 @@ impl WalletClient {
     /// Send tokens to nodes closest to the data we want to make storage payment for. Runs mandatory verification.
     ///
     /// # Arguments
-    /// - content_addrs - [Iterator](Iterator)<Items = [NetworkAddress](NetworkAddress)>
```
RolandSherwin, line 346: `content_addrs` refers to an iter of `NetworkAddress` right?

sn_client/src/wallet.rs
```diff
@@ -663,6 +663,30 @@ impl WalletClient {
 impl Client {
     /// Send spend requests to the network.
     /// This can optionally verify the spends have been correctly stored before returning
+    ///
+    /// # Arguments
+    /// * spend_requests - [Iterator]<[SignedSpend]>
+    /// * verify_store - Boolean. Set to true for mandatory verification via a GET request through a Spend on the network.
```
RolandSherwin, line 669:
```suggestion
    /// * verify_store - [bool]. Set to true for mandatory verification via a GET request through a Spend on the network.
```

sn_client/src/wallet.rs
```diff
@@ -663,6 +663,30 @@ impl WalletClient {
...
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(),Error>{
+    /// let client = Client::new(SecretKey::random(), None, false, None, None).await?;
+    /// # let tmp_path = TempDir::new()?.path().to_owned();
+    /// let mut wallet = LocalWallet::load_from_path(&tmp_path,Some(MainSecretKey::new(SecretKey::random())))?;
+    /// // An example of sending storage payment transfers over the network with validation
+    /// client.send_spends(
+    ///     wallet.unconfirmed_spend_requests().iter(),
+    ///     true
+    /// ).await.expect("");
```
RolandSherwin: I think we can use `?` operator instead of `expect` here?

sn_client/src/wallet.rs
```diff
@@ -761,6 +850,39 @@ impl Client {
 
 /// Use the client to send a CashNote from a local wallet to an address.
 /// This marks the spent CashNote as spent in the Network
+///
+/// # Arguments
+/// * from - [LocalWallet]
+/// * amount - [NanoTokens]
+/// * to - [MainPubkey]
+/// * client - [Client]
+/// * verify_store - Boolean. Set to true for mandatory verification via a GET request through a Spend on the network.
```
RolandSherwin, line 856:
```suggestion
/// * verify_store - [bool]. Set to true for mandatory verification via a GET request through a Spend on the network.
```

sn_client/src/wallet.rs
```diff
@@ -841,7 +963,51 @@ pub async fn send(
...
 /// Verification will be attempted via GET request through a Spend on the network.
+///
+/// # Arguments
+/// * from - [LocalWallet],
+/// * client - [Client],
+/// * signed_spends - [BTreeSet]<[SignedSpend]>,
+/// * transaction - [Transaction],
+/// * change_id - [UniquePubkey],
+/// * output_details - [BTreeMap]<[UniquePubkey], ([MainPubkey], [DerivationIndex])>,
+/// * verify_store - Boolean. Set to true for mandatory verification via a GET request through a Spend on the network.
```
RolandSherwin, line 973:
```suggestion
/// * verify_store - [bool]. Set to true for mandatory verification via a GET request through a Spend on the network.
```

sn_client/src/wallet.rs
JasonPaulGithub: Yes, correct. Can I improve this? Let me know! 👍

sn_client/src/wallet.rs
JasonPaulGithub: I've had trouble with bool. for [bool] it doesn't point anywhere on the IDE or the docs when you hover over it for help, so we need to link it to something. Links go in the curly brackets after the square ones, so the result would look like this:

[bool](link to bool)

problem there is that I can not find a good link to the bool. I tried two options:

`[bool](core::primitive::bool)`
and
`[bool](std::primitive::bool)`

but no luck. It does however link to the bool struct if you use either of these:

`[bool](core::primitive)`
and
`[bool](std::primitive)`

but that didn't feel right so I just left it at 'Boolean'. I figure if you don't know what a Boolean is, maybe you need more help than from the docs - however the right link to wherever the Boolean is stored in Rust is useful. 

So you are on the right track for pointing this out, but I think we can keep an eye out for a better solution. 
What do you think?

sn_client/src/wallet.rs
JasonPaulGithub: Yes we can 👍

sn_client/src/wallet.rs
RolandSherwin: I think you might have changed that to `NanoTokens` here.

sn_client/src/wallet.rs
RolandSherwin: Ah. I use vscode and `[bool]` points to https://doc.rust-lang.org/nightly/core/primitive.bool.html
I assume it is because bool might already be in scope? 

Anyway, I agree with your point. This is a very trivial thing, will mark it as 'resolved'.

sn_client/src/wallet.rs
JasonPaulGithub: oh yes! I think I made a mistake with the find and replace feature. Thank you. Well spotted!!

RolandSherwin (approved):

PR #1248, opened 2024-01-31
feat(faucet): create distributions for maid addrs

iancoleman:
This PR changes the faucet startup process. At startup the faucet creates distributions of tokens using the snapshot of maid addresses + balances + public keys.

Each distribution is a combination of a transfer (from the faucet) and an encrypted secret key (randomly generated) that enables that transfer to be spent. The maid owner can decrypt the secret key using their maid private key, which is then used as a wallet key for spending the transfer.

Distributions created by the faucet are saved to `$HOME/.local/share/safe_snapshot/distributions/<address>` and contain the messagepack-encoded hex for the distribution.

Note:

* This doesn't enable access to the distributions yet, this is to be done in future work by exposing a faucet server endpoint that returns the distribution for a particular maid address. I'm trying to keep each PR relatively small.

* The startup process for the faucet server has become much longer while it creates these distributions. Creating 1200 initial distributions took about 11 minutes.
    * Should we move this into a separate thread and have it run in the background?
    * Should we have a feature flag for local testnets that doesn't use the 'live' snapshot list and uses a fake test-only list of a single address?

Future work will be:

* allow users to fetch their distribution from the cli for their maid address, and automatically send that distribution to their wallet.

* enable the faucet to receive new maid address+publickey pairs and create new distributions on the fly for those.

* expose the full snapshot and distribution list to enable audit of the process. Need to explore what an audit would require.

* find a simple / automatic way to retain address + publickeys that have been submitted while the faucet is running, stored in [maid_address_pubkeys.csv](https://github.com/maidsafe/safe_network/raw/main/sn_faucet/maid_address_pubkeys.csv) for future use.

* consider adding a test maid address to this so we can test the claim process easily.

### Summary generated by Reviewpad on 31 Jan 24 03:07 UTC
This pull request includes changes to the Cargo.lock, Cargo.toml, and faucet_server.rs files. It updates dependencies and adds new ones, including "ecies", "libsecp256k1", "libsecp256k1-core", "libsecp256k1-gen-ecmult", "libsecp256k1-gen-genmult", "rmp-serde", and "serde_bytes". It also adds structs and functions related to distributing tokens to addresses using public and secret keys.

### DIFF
--- a/sn_faucet/src/faucet_server.rs
+++ b/sn_faucet/src/faucet_server.rs
@@ -16,6 +16,14 @@ use std::path::{self, Path, PathBuf};
 use tiny_http::{Response, Server};
 use tracing::{debug, error, trace};
 
+#[derive(Serialize, Deserialize)]
+struct Distribution {
+    #[serde(with = "serde_bytes")]
+    transfer: Vec<u8>,
+    #[serde(with = "serde_bytes")]
+    encrypted_secret_key: Vec<u8>,
+}
+
 /// Run the faucet server.
 ///
 /// This will listen on port 8000 and send a transfer of tokens as response to any GET request.
@@ -60,8 +68,9 @@ pub async fn restart_faucet_server(client: &Client) -> Result<()> {
 async fn startup_server(client: &Client) -> Result<()> {
     #[cfg(feature = "distribution")]
     {
-        let _balances = token_distribution::load_maid_snapshot()?;
-        let _keys = token_distribution::load_maid_pubkeys()?;
+        let balances = token_distribution::load_maid_snapshot()?;
+        let keys = token_distribution::load_maid_pubkeys()?;
+        token_distribution::distribute_from_maid_to_tokens(client, balances, keys).await;
     }
     let server =
         Server::http("0.0.0.0:8000").map_err(|err| eyre!("Failed to start server: {err}"))?;
--- a/sn_faucet/src/faucet_server.rs
+++ b/sn_faucet/src/faucet_server.rs
@@ -70,7 +70,12 @@ async fn startup_server(client: &Client) -> Result<()> {
     {
         let balances = token_distribution::load_maid_snapshot()?;
         let keys = token_distribution::load_maid_pubkeys()?;
-        token_distribution::distribute_from_maid_to_tokens(client, balances, keys).await;
+        // Each distribution takes about 500ms to create, so for thousands of
+        // initial distributions this takes many minutes. This is run in the
+        // background instead of blocking the server from starting.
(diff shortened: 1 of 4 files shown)


joshuef:
> Should we move this into a separate thread and have it run in the background?

Sounds good if that can be done easily? 

> Should we have a feature flag for local testnets that doesn't use the 'live' snapshot list and uses a fake test-only list of a single address?

I think we'll need that and for the CI in general (aside from probably a test that will verify that distributions have occured perhaps?) (That might be one for a follow up PR?)

sn_faucet/src/faucet_server.rs
```diff
@@ -345,3 +363,125 @@ fn save_address_pk(address: &str, pk_hex: &str) -> Result<()> {
     std::fs::write(addr_path, pk_hex)?;
     Ok(())
 }
+
+async fn distribute_from_maid_to_tokens(
+    client: &Client,
+    snapshot: Snapshot,
+    pubkeys: HashMap<MaidAddress, MaidPubkey>,
+) {
+    for (addr, amount) in snapshot {
```
joshuef: Here I think we can push each iteration into its own `task` and then use `futures::join_all` to collect and ensure that's happening.

And then if this whole function is `tokio::spawn`ed we'll get a fast startup 👍

iancoleman:
Rebased on main after #1253 was merged

sn_faucet/src/faucet_server.rs
iancoleman: It seemed simpler to just put the whole distribute function in a thread rather than do a bunch of spawning, because transfers can't happen in parallel (wait for wallet lock to be released).

iancoleman:
How to fix this failing test for unused dependencies? They *are* used but only in `distributions` feature. I can change `workflows/merge.yml` L32 to get the test to pass:

```diff
-run: cargo +nightly udeps --all-targets
+run: cargo +nightly udeps --all-targets --features distribution
```

but this feels like the wrong way to do it (maybe my feeling is wrong). Any tips or advice @grumbach @joshuef ?

joshuef:
@iancoleman i'm not wholly sure. I've pinged @RolandSherwin as he might have more idea, but right now we could manually ignore them via the `cargo.toml`: https://github.com/maidsafe/safe_network/actions/runs/7755547781/job/21151134924?pr=1248#step:5:1077

If there's no immediate better way, I'd do just that.

joshuef:
ah, @RolandSherwin has sorted it. Thanks!! 🙇 (was `optional` on the dep, @iancoleman )

joshuef (approved):

PR #1256, opened 2024-02-02
feat(nodes): make encryption of records a feature, disabled by default

joshuef:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/Cargo.toml
+++ b/sn_networking/Cargo.toml
@@ -16,6 +16,8 @@ local-discovery=["libp2p/mdns"]
 # tcp is automatically enabled when compiling for wasm32
 websockets=["libp2p/tcp"]
 open-metrics=["libp2p/metrics", "prometheus-client", "hyper", "sysinfo"]
+encrypt-records=[]
+
 
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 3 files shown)


sn_networking/src/record_store.rs
```diff
@@ -272,6 +288,27 @@ impl NodeRecordStore {
         );
     }
 
+    /// Prepare record bytes for storage
+    /// If feats are enabled, this will eg, encrypt the record for storage
+    fn prepare_record_bytes(
+        record: &Record,
+        encryption_details: (Aes256GcmSiv, [u8; 4]),
+    ) -> Option<Vec<u8>> {
+        let (cipher, nonce_starter) = encryption_details;
```
maqi, line 301: doesn't see there is feature dependent check here ?
there shall be something similar to what get_record_from_bytes has?

sn_networking/Cargo.toml
```diff
@@ -12,10 +12,13 @@ version = "0.13.0"
 
 [features]
 default=["libp2p/quic"]
+encrypt-store=[]
```
maqi: looks like only `encrypt-records` being used ?
maybe this one can be removed ?

maqi (approved):

PR #1259, opened 2024-02-05
feat: client side dag

grumbach:
## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/audit/spend_dag_building.rs
+++ b/sn_client/src/audit/spend_dag_building.rs
@@ -17,14 +17,19 @@ impl Client {
     /// Builds a SpendDag from a given SpendAddress recursively following descendants all the way to UTxOs
     /// Started from Genesis this gives the entire SpendDag of the Network at a certain point in time
     /// Does not verify the validity of the transactions
-    pub async fn build_spend_dag_from(&self, spend_addr: SpendAddress) -> WalletResult<SpendDag> {
+    pub async fn spend_dag_build_from(&self, spend_addr: SpendAddress) -> WalletResult<SpendDag> {
         let mut dag = SpendDag::new();
 
         // get first spend
-        let first_spend = self
-            .get_spend_from_network(spend_addr)
-            .await
-            .map_err(|err| WalletError::CouldNotVerifyTransfer(err.to_string()))?;
+        let first_spend = match self.get_spend_from_network(spend_addr).await {
+            Ok(s) => s,
+            Err(Error::MissingSpendRecord(_)) => {
+                // the cashnote was not spent yet, so it's an UTXO
+                trace!("UTXO at {spend_addr:?}");
+                return Ok(dag);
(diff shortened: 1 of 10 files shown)


sn_client/src/audit/spend_dag_building.rs
```diff
@@ -17,14 +17,18 @@ impl Client {
...
         // get first spend
-        let first_spend = self
-            .get_spend_from_network(spend_addr)
-            .await
-            .map_err(|err| WalletError::CouldNotVerifyTransfer(err.to_string()))?;
+        let first_spend = match self.get_spend_from_network(spend_addr).await {
+            Ok(s) => s,
+            Err(Error::MissingSpendRecord(_)) => {
+                trace!("UTXO at {spend_addr:?}");
+                return Ok(dag);
```
maqi, line 29: 1, Why it's not an `error` to be returned if spend not found ?
2, The comment of the function doesn't explain this situation

if it's a value return, maybe put an explaination in the function comment ?

sn_client/src/audit/spend_dag_building.rs
```diff
@@ -200,4 +204,15 @@ impl Client {
         info!("Verified all the way to known spends or genesis! Through {depth} generations, verifying {n} transactions in {elapsed:?}");
         Ok(())
     }
+
+    /// Extends an existing SpendDag starting from the utxos in this DAG
+    /// Covers the entirety of currently existing Spends if the DAG was built from Genesis
+    pub async fn spend_dag_continue_from_utxos(&self, dag: &mut SpendDag) -> WalletResult<()> {
```
maqi, line 211: hmm...  what's the different between this to be above `spend_dag_extend` function ?
`continue_from_utxos` is also an `extend` ?

maybe the `spend_dag_extend` shall be renamed to `dag_from_spend_address` ?

grumbach: Yeah I agree, basically `spend_dag_continue_from_utxos` extends the DAG in all directions from the current UTXOs at the extremities. That's total collection, the point of this is to collect all Spends that exist on the Network. 

`spend_dag_extend` is partial extension from one spend back to either Genesis or a known Spend that we already have in our DAG. It is adding one Spend and all its ancestors to our current DAG. 

So both are extension, but one is only in one direction up to a specific Spend and the other one is in all directions. 

`dag_from_spend_address` makes me think that we build a dag from an address, maybe we need to add an extend somewhere. 
How is `spend_dag_extend_until`?

sn_client/src/audit/spend_dag_building.rs
grumbach: If a Spend is not found, it means the CashNote was not spent yet. This is not an error in our DAG building, it simply indicates there is no need to fetch further as unspent money has no descendants.

maqi (approved):

PR #1264, opened 2024-02-05
feat: register example with two users in a simple cli chat app

happybeing:
Modifies the `sn_node` registers example so that multiple users can chat in a simple CLI app. Changes in commit:
- add cli option to provide xor address
- user entering blank line gets latest register values
- documented in main README

## Description

reviewpad:summary


happybeing:
@jacderida Chris, this is failing one test (Large file upload) at the "Start client to upload" step, which I don't think is to do with my PR but I can't see what is going wrong in the output. Do you know what's happening here?

joshuef:
@happybeing Thanks for this!

That test is being flaky across a number of PRs, definitely not tied to this, so don't worry about that one just now 🙇 .

joshuef (approved):

PR #1283, opened 2024-02-08
Feat: move folders/files metadata out of Folders entries

bochaco:
This also makes it possible to sync/compare local changes with remote version of folders/files.

Requires PR #1304 .
Requires PR #1311 .

## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/folders.rs
+++ b/sn_client/src/folders.rs
@@ -6,19 +6,22 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::FilesApi;
+
 use super::{error::Result, Client, ClientRegister, WalletClient};
 
+use self_encryption::MAX_CHUNK_SIZE;
 use sn_protocol::{
-    storage::{ChunkAddress, RegisterAddress},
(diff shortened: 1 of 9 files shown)


joshuef (approved):

sn_client/src/folders.rs
```diff
@@ -99,32 +140,67 @@ impl FoldersApi {
...
+            files_api,
+            metadata: BTreeMap::new(),
+        })
+    }
+
+    // Add the given entry to the underlying Register as well as creating the metadata Chunk
+    fn add_entry(&mut self, name: OsString, content: FolderEntry) -> Result<(Metadata, XorName)> {
+        // create metadata Chunk for this entry
+        let metadata = Metadata {
+            name: name.to_str().unwrap_or("unknown").to_string(),
```
RolandSherwin, line 191: Should this be just the `OsString`? The previous todo said the following
`// TODO: conversion to String will be removed when metadata is moved out of the Register`

bochaco: there is a problem with wasm build and [unsafe api from OsString](https://doc.rust-lang.org/std/ffi/struct.OsString.html#method.from_encoded_bytes_unchecked), it needs more investigation

PR #1287, opened 2024-02-09
ci: re-add disk space debugging

jacderida:
I would still like to monitor how much disk space `release-plz` is using just to see if upgrading to the new version has likely solved the problem we were having.

## Description

reviewpad:summary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,7 +125,10 @@ jobs:
         shell: bash
         run: |
           cargo login ${{ secrets.CRATES_IO_TOKEN }}
+          df -h
           release-plz release --git-token ${{ secrets.VERSION_BUMP_COMMIT_PAT }}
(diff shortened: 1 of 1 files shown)


happybeing:
`vdash` too would love to see a log message that restores the disk use stats of old. Something like space used and available would be great.

happybeing:
Ah, it's just merged and not in sn_node after all - sorry. As you were!

jacderida:
> Ah, it's just merged and not in sn_node after all - sorry. As you were!

Yeah sorry, this is just a little bit of debugging output to the release process because we were having some trouble with a massive amount of disk space being used during our publishing process. Maybe you might want to follow that request up with Josh?

happybeing:
Thanks Chris, I have pestered Josh about this enough I think! Just misunderstood this PR and jumped the gun :smile:

PR #1289, opened 2024-02-09
feat: force and upgrade by url or version

jacderida:
Three new arguments are added to the `upgrade` command: `--force`, `--url` and `--version`.

  The `--url` and `--version` arguments provide two different sources for the upgrade, as opposed to
  just upgrading to the latest version. With `--url`, a custom binary can be provided, which will be
  used for the backwards compatibility test. The `--version` flag enables upgrading to a specific
  version rather than the latest. Both of these can be used with the `--force` flag to downgrade to an
  arbitrary version or to accept an upgrade from a binary with the same version, the latter of which
  will again be used in the backwards compatibility test.

  Integration tests provide coverage of these new features. Each of the tests really needs to run on
  its own machine, otherwise they interfere with each other, and tests can't make assumptions about
  how many services there are. So we add twelve additional jobs to the merge workflow here, which is
  for four tests on three operating systems. However, these tests should run pretty quickly.

  The `stop` command was modified such that it will no longer return an error if the service is in the
  `ADDED` state, i.e., it has not been started before. This enables us to test the upgrade process
  without initially starting the service, which could introduce complications. The `get_safenode_port`
  function was also changed to return an `Option` rather than a `Result` for the same reason.

## Description

reviewpad:summary

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -146,8 +146,8 @@ jobs:
       - shell: bash
         run: cargo test --release --bin safenode-manager
 
-  node-manager-integration-tests:
-    name: node manager integration tests
+  node-manager-e2e-tests:
+    name: node manager e2e tests
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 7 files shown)


RolandSherwin (approved):
Great PR! Looks good to me 😄

PR #1290, opened 2024-02-11
feat(faucet): api endpoint to return distribution

iancoleman:
The server sends 100 tokens for the endpoint:
`http://<ip>/<bls_hex_pubkey>`

This adds a new (additional) endpoint to distribute to maid addresses:
`http://<ip>/distribution?address=<maid_address>&pkhex=<pkhex>`

Calling the new endpoint will return an existing distribution for that maidsafecoin address, or create a new distribution if needed.

This enables distributions to be fetched from the faucet, but there still needs to be integration into the cli which is future work to try to keep each PR relatively small and manageable.

Future faucet related work:

* update cli `safe wallet get-faucet` to be able to fetch distributions
* change the default amount issued by the faucet from 100 to something smaller (any suggestions?)
* potentially move away from `tiny_http` for the faucet server to something that has better security and rate limiting options

## Description

reviewpad:summary

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5031,6 +5031,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-core",
+ "url",
 ]
 
 [[package]]
--- a/sn_faucet/Cargo.toml
+++ b/sn_faucet/Cargo.toml
@@ -40,6 +40,7 @@ tiny_http = { version="0.12", features = ["ssl-rustls"] }
 tokio = { version = "1.32.0", features = ["parking_lot", "rt"] }
 tracing = { version = "~0.1.26" }
 tracing-core = "0.1.30"
+url = "2.5.0"
 
 [lints]
 workspace = true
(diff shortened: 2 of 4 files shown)


joshuef:
> change the default amount issued by the faucet from 100 to something smaller (any suggestions?)

I think we should probably look to distribute 1SNT only and that from a wallet that only has maybe 30 to start. (arbitrary numbers!)

I'm thinking: 

- we do not want an endless faucet. Indeed we only really want some for basic in house testing.
- we can top up a faucet if desired, so : keep the amount low (1 seems sensible atm?), and the initial dist to the hot-faucet-wallet big enough, but not wild.
- this will force folk to use and test out the OMNI conversion.


Thoughts?

joshuef (approved):

iancoleman:
> I think we should probably look to distribute 1SNT only

I think it should be much lower than that, more like 0.01 SNT

In recent testnets upload cost has been about 0.002 SNT per GB

> from a wallet that only has maybe 30 to start

This sounds about right. Probably could even start with less. Has the faucet run dry before? Have we got a code path for this (I guess we do of some sort but it's 'accidental' like a failed transfer)?

PR #1296, opened 2024-02-12
feat: estimate of price cli

JasonPaulGithub:
## Description

reviewpad:summary

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -234,6 +234,12 @@ jobs:
           SN_LOG: "all"
         timeout-minutes: 5
 
+      - name: Start a client to estimate a files upload cost
+        run: cargo run --bin safe --release -- --log-output-dest=data-dir files estimate "./resources" --retry-strategy quick
+        env:
+          SN_LOG: "all"
+        timeout-minutes: 15
+
       - name: Start a client to upload files
         run: cargo run --bin safe --release -- --log-output-dest=data-dir files upload "./resources" --retry-strategy quick
         env:
--- a/sn_client/src/files/upload.rs
+++ b/sn_client/src/files/upload.rs
@@ -18,14 +18,14 @@ use xor_name::XorName;
 
 use sn_networking::PayeeQuote;
 use sn_protocol::{
-    NetworkAddress,
     storage::{Chunk, ChunkAddress, RetryStrategy},
+    NetworkAddress,
 };
 use sn_transfers::NanoTokens;
 
 use crate::{
-    BATCH_SIZE,
-    error::{Error as ClientError, Result}, FilesApi,
+    error::{Error as ClientError, Result},
+    FilesApi, BATCH_SIZE,
 };
 
 /// The maximum number of sequential payment failures before aborting the upload process.
--- a/sn_client/src/files/upload.rs
+++ b/sn_client/src/files/upload.rs
@@ -6,24 +6,27 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{
-    error::{Error as ClientError, Result},
-    FilesApi, BATCH_SIZE,
-};
-use bytes::Bytes;
-use libp2p::PeerId;
-use sn_networking::PayeeQuote;
-use sn_protocol::{
-    storage::{Chunk, ChunkAddress, RetryStrategy},
-    NetworkAddress,
-};
-use sn_transfers::NanoTokens;
 use std::{
     collections::{BTreeMap, BTreeSet, HashSet},
     path::PathBuf,
 };
+
+use bytes::Bytes;
+use libp2p::PeerId;
 use tokio::sync::mpsc::{self};
 use xor_name::XorName;
+
+use sn_networking::PayeeQuote;
+use sn_protocol::{
+    NetworkAddress,
+    storage::{Chunk, ChunkAddress, RetryStrategy},
+};
+use sn_transfers::NanoTokens;
+
+use crate::{
+    BATCH_SIZE,
+    error::{Error as ClientError, Result}, FilesApi,
+};
 
 /// The maximum number of sequential payment failures before aborting the upload process.
 const MAX_SEQUENTIAL_PAYMENT_FAILS: usize = 3;
@@ -317,9 +320,9 @@ impl FilesUpload {
                     trace!(
                         "Upload task got chunk {:?}'s store_cost {:?}",
                         chunk_info.name,
-                        cost.2
+                        cost.quote.cost
                     );
-                    if cost.2.cost != NanoTokens::zero() {
+                    if cost.quote.cost != NanoTokens::zero() {
                         pending_to_pay.push((chunk_info, cost));
                     } else if for_failed_chunks {
                         // the chunk was already paid for but might have not been verified on the first try.
@@ -461,7 +464,8 @@ impl FilesUpload {
                 .await
             {
                 Ok(cost) => {
-                    debug!("Storecosts retrieved for {:?} {cost:?}", chunk_info.name);
+                    let t = &cost.quote;
+                    debug!("Storecosts retrieved for {:?} {t:?}", chunk_info.name);
                     TaskResult::GetStoreCostOK((chunk_info, cost))
                 }
                 Err(err) => {
@@ -502,8 +506,8 @@ impl FilesUpload {
 
             while let Some(payment) = paying_work_receiver.recv().await {
                 let make_payments = if let Some((chunk_info, quote)) = payment {
-                    let _ = cost_map.insert(chunk_info.name, (quote.1, quote.2));
-                    chunk_info_map.push((chunk_info, quote.0));
+                    let _ = cost_map.insert(chunk_info.name, (quote.key, quote.quote));
+                    chunk_info_map.push((chunk_info, quote.peer));
                     cost_map.len() >= batch_size
                 } else {
                     // using None to indicate as all paid.
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -6,21 +6,12 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::collections::HashMap;
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    iter::Iterator,
-};
+use crate::Error;
 
+use super::{error::Result, Client};
 use backoff::{backoff::Backoff, ExponentialBackoff};
 use futures::{future::join_all, TryFutureExt};
 use libp2p::PeerId;
-use tokio::{
-    task::JoinSet,
-    time::{sleep, Duration},
-};
-use xor_name::XorName;
-
 use sn_networking::target_arch::Instant;
 use sn_networking::{GetRecordError, PayeeQuote};
 use sn_protocol::NetworkAddress;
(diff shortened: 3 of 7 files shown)


sn_cli/src/subcommands/files/mod.rs
```diff
@@ -251,6 +258,29 @@ pub(crate) async fn files_cmds(
...
+    let mut total: u64 = 0;
+
+    for chunk_detail in chunk_manager.get_chunks() {
+        let chunk_address = ChunkAddress::new(chunk_detail.0);
+        let address = NetworkAddress::from_chunk_address(chunk_address);
+        let estimate = files_api
+            .wallet()?
+            .get_store_cost_at_address(address)
+            .await
+            .unwrap()
```
joshuef: We don't use unwraps in prod code, please. Use the try operator `?` and advance appropriate errors upwards (or `bail` via `eyre` with context)

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -423,11 +454,11 @@ pub(crate) async fn upload_files(
     println!("Uploading {chunks_to_upload_len} chunks",);
     let now = Instant::now();
     let upload_result = match files_upload.upload_chunks(chunks_to_upload).await {
-        Ok(()) => {Ok(())}
+        Ok(()) => { Ok(()) }
         Err(ClientError::Transfers(WalletError::Transfer(TransfersError::NotEnoughBalance(
-            available,
-            required,
-        )))) => {
+                                                             available,
```
joshuef: i'd wager `cargo fmt` is complaining about this. 

We use that for standardised indentation.

joshuef:
nice that you've got this in one simple fn in the end. 

It would be great to add some tests here to the CI flows so that we can ensure that a) the stimate command exists and putput some estimate, but also b) that chimes with the final cost of a file upload

joshuef (changes requested):

sn_cli/src/subcommands/files/mod.rs
JasonPaulGithub: ah yes, fmt did not like that. I use cargo fmt and the CTRL + shift  + F in my ide almost like a muscle reflex. 
I will keep this in mind.

JasonPaulGithub: I use cargo fmt and the CTRL + shift  + F my ide almost like a muscle reflex. 
I will keep this in mind.

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -46,15 +48,18 @@ pub(crate) const UPLOADED_FILES: &str = "uploaded_files";
 
 #[derive(Parser, Debug)]
 pub enum FilesCmds {
+    Estimate {
+        /// The location of the file(s) to upload. Can be a file or a directory.
+        #[clap(name = "path", value_name = "PATH")]
+        path: PathBuf,
+    },
```
joshuef: we still need `make_public` here to feed into the datamap chunk bool

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -251,6 +258,28 @@ pub(crate) async fn files_cmds(
...
+    let mut estimate: u64 = 0;
+
+    for chunk in chunk_manager.get_chunks() {
+        estimate += FilesApi::new(client.clone(), root_dir.to_path_buf())
+            .wallet()?
+            .get_store_cost_at_address(NetworkAddress::from_chunk_address(ChunkAddress::new(
+                chunk.0,
+            )))
+            .await?
+            .2
```
joshuef: what is `2` here? Lets break this out into named vars please for clarity.

sn_cli/src/subcommands/files/mod.rs
```diff
@@ -251,6 +258,28 @@ pub(crate) async fn files_cmds(
...
 }
 
+/// Estimate the upload cost of a chosen file
+pub(crate) async fn estimate_cost(path: PathBuf, client: &Client, root_dir: &Path) -> Result<()> {
+    let mut chunk_manager = ChunkManager::new(root_dir);
+    chunk_manager.chunk_path(&path, false, false)?;
+
+    let mut estimate: u64 = 0;
+
+    for chunk in chunk_manager.get_chunks() {
```
joshuef: as @RolandSherwin mentioned, we want to push each of these into a task so that this can happen in an async fashion (otherwise this could take quite a while)

joshuef (changes requested):

sn_cli/src/subcommands/files/mod.rs
JasonPaulGithub: Okay. If it's okay with you @RolandSherwin I'll be in touch to ask you your thoughts on the best way to do this.

RolandSherwin: Sure, I'm glad to be of help here.  😄

.github/workflows/merge.yml
```diff
@@ -234,6 +234,12 @@ jobs:
           SN_LOG: "all"
         timeout-minutes: 5
 
+      - name: Start a client to estimate a files upload cost
+        run: cargo run --bin safe --release -- --log-output-dest=data-dir files estimate "./resources"
```
JasonPaulGithub: I took out the "retry-strategy quick" bit from here. 
I don't know what it does, but now it passes the tests.

sn_client/src/files/upload.rs
```diff
@@ -461,7 +464,8 @@ impl FilesUpload {
                 .await
             {
                 Ok(cost) => {
-                    debug!("Storecosts retrieved for {:?} {cost:?}", chunk_info.name);
+                    let t = &cost.quote;
+                    debug!("Storecosts retrieved for {:?} {t:?}", chunk_info.name);
```
JasonPaulGithub, line 468: Please check over this, I do not know the original values of the message but I gave it an estimate.

sn_client/src/wallet.rs
```diff
@@ -430,7 +462,13 @@ impl WalletClient {
                     .await
                     .map_err(|error| WalletError::CouldNotSendMoney(error.to_string()));
 
-                debug!("Storecosts retrieved for {content_addr:?} {cost:?}");
+                let t = cost
+                    .as_ref()
+                    .expect("Pay for storage once: Cost not found")
```
JasonPaulGithub: guess here

RolandSherwin: Since this a debug log, I think we can print out the whole `cost` struct here. It contains the details that you've extracted out anyways.
We try not to call `expect/unwrap`, as it would crash the node/client if the value was not present. This is not desirable.

.github/workflows/merge.yml
joshuef: why would this have been failing 🤔 . cc @RolandSherwin do you have an idea here?

sn_client/src/files/upload.rs
joshuef: why is `t` `cost.quote`? what is `t`? can we call this `quote`?

sn_client/src/wallet.rs
joshuef: I don't think we should be doing `expect` at this layer either. We need to be retuning errors from the `sn_client` lib. `expect` will cause a panic.

.github/workflows/merge.yml
RolandSherwin: The `retry-strategy` sub-argument should be manually added to the `files estimate` argument if want that to work. Something like this for `files upload` https://github.com/maidsafe/safe_network/blob/fe5c8d7897e61ef6aafc70ef870e4363dda28aff/sn_cli/src/subcommands/files.rs#L67

sn_client/src/files/upload.rs
JasonPaulGithub: t is for temp until quote is confirmed. I can rename it now👍

sn_client/src/wallet.rs
JasonPaulGithub: - I've extracted the `expect` into a `match`. I've tried to  [return a specific message from a lib](https://maidsafe.slack.com/archives/C050EDL5ACS/p1707922625611599) but no luck. Working on it 👍 
(Update: In regards to the linked discussion it seems I'm going to need to make a refactor to pay_for_storage_once. I may need to write a unit test to get things going, but it seems I might get away with just debugging it)


- There is an issue with printing the whole struct. It says the struct does not use the "Debug trait". 

```
error[E0277]: `PayeeQuote` doesn't implement `std::fmt::Debug`                                                       
   --> sn_client\src\files\upload.rs:468:59
    |
468 |                     debug!("Storecosts retrieved for {:?} {cost:?}", chunk_info.name);
    |                                                           ^^^^^^^^ `PayeeQuote` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
```


Even though it's basically the same as the tuple used before. Perhaps I can use a better solution:
In Java I would override the .toString method to print out all the necessary values, In Rust perhaps I could concatenate everything into a String to complete the output message? 
I can do that if necessary.

JasonPaulGithub:
@joshuef 
I will squash commits to make create a better paper-trail. This will have to come last to make sure I have all those little changes in place before the final push. 🥇

JasonPaulGithub:
// TODO:
- [ ] Refactor `pay_for_storage_once` with debug / Unit Test.
- [ ] post `retry strategy` for help on the issue.
- [ ] Squash commits.
- [ ] Push.

sn_networking/src/lib.rs
```diff
@@ -50,27 +24,57 @@ use libp2p::{
...
+    driver::{GetRecordCfg, NetworkBuilder, PutRecordCfg, SwarmDriver, VerificationKind},
+    error::{Error, GetRecordError},
+    event::{MsgResponder, NetworkEvent},
+    record_store::NodeRecordStore,
+    transfers::get_singed_spends_from_record,
 };
 
-use tokio::time::Duration;
-use tracing::trace;
+mod bootstrap;
```
joshuef, line 54: why the move down here? 

At least in house (not sure on rusty) convention is to have local import modules at the top. I'm not sure this move is necessary?

sn_client/src/wallet.rs
joshuef: w/r/t `Debug`, I think we can just add the `Debug` trait can we not just add `#[derive(Debug)]` to the struct or its internals?

joshuef:
I feel like the TODO actually is probably a list for a few PRs. 

Let's try and keep this focussed to the initial implementation and a test for it.

sn_networking/src/lib.rs
JasonPaulGithub: For now I do what the formatter tells me to :'(
I can move it back 👍

sn_client/src/wallet.rs
JasonPaulGithub:
```
#[derive(
    Clone, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize, custom_debug::Debug, Debug,
)]
pub struct PaymentQuote {
    /// the content paid for
    pub content: XorName,
    /// how much the node demands for storing the content
    pub cost: NanoTokens,
    /// the local node time when the quote was created
    pub timestamp: SystemTime,
    /// the node's signature of the 3 fields above
    #[debug(skip)]
    pub signature: QuoteSignature,
}
```

54 | ...rialize, custom_debug::Debug, Debug,
   |             -------------------  ^^^^^ conflicting implementation for `PaymentQuote`

@joshuef  @RolandSherwin 
Ello. So I don't know what the custom debug is for or how to find out; but this is the reason why.
I'm guessing there's a custom debug for a reason, and that I can't change it to debug. It would be scarier if I changed or removed it and nothing happened - so it's probably best to ask 👼

.github/workflows/merge.yml
JasonPaulGithub: The nature of the error ::
```
PS C:\Users\lol_h\RustroverProjects\safe_network> cargo run --bin safenode-manager --features local-discovery -- run --build
   Compiling sn_transfers v0.15.5 (C:\Users\lol_h\RustroverProjects\safe_network\sn_transfers)
   Compiling sn_protocol v0.14.1 (C:\Users\lol_h\RustroverProjects\safe_network\sn_protocol)
   Compiling sn_networking v0.13.19 (C:\Users\lol_h\RustroverProjects\safe_network\sn_networking)
   Compiling sn_client v0.104.14 (C:\Users\lol_h\RustroverProjects\safe_network\sn_client)
   Compiling sn_node v0.104.22 (C:\Users\lol_h\RustroverProjects\safe_network\sn_node)
   Compiling sn_node_rpc_client v0.4.54 (C:\Users\lol_h\RustroverProjects\safe_network\sn_node_rpc_client)
   Compiling sn-node-manager v0.3.1 (C:\Users\lol_h\RustroverProjects\safe_network\sn_node_manager)
    Finished dev [unoptimized + debuginfo] target(s) in 57.11s
     Running `target\debug\safenode-manager.exe run --build`
Error:
   0: missing field `local` at line 1 column 829

Location:
   sn_protocol\src\node_registry.rs:182
```

PR #1304, opened 2024-02-13
feat(client): keep payee as part of storage payment cache

bochaco:
## Description

reviewpad:summary

### DIFF
--- a/sn_transfers/src/wallet/data_payments.rs
+++ b/sn_transfers/src/wallet/data_payments.rs
@@ -26,7 +26,9 @@ pub struct Payment {
 pub struct PaymentDetails {
     /// The node we pay
     pub recipient: MainPubkey,
-    /// The PeerId (as bytes) of the node we pay
+    /// The PeerId (as bytes) of the node we pay.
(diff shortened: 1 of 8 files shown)


sn_transfers/src/wallet/data_payments.rs
```diff
@@ -26,6 +26,8 @@ pub struct Payment {
 pub struct PaymentDetails {
     /// The node we pay
     pub recipient: MainPubkey,
+    /// The PeerId (as bytes) of the node we pay
+    pub peer_id_bytes: Vec<u8>,
```
RolandSherwin, line 32: Can we add a comment stating why we store the `peer id bytes` instead of the actual `PeerId` here?

joshuef (approved):

sn_transfers/src/wallet/data_payments.rs
loziniak: The comment says, that PeerId does not implement Serialize/Deserialize, but looking at the code, it seems the other way. Am I missing something?
https://docs.rs/libp2p-identity/0.2.8/src/libp2p_identity/peer_id.rs.html#171

PR #1306, opened 2024-02-14
fix: check if redeemed cns are already spent

grumbach:
## Description

Issue originally identified by @iancoleman: redeeming Transfers doesn't check if the redeemed CashNotes where already spent (by us) or not. 

This PR fixes this.


joshuef (approved):

grumbach:
#1345 re-PRd against alpha

PR #1307, opened 2024-02-14
feat(manager): set env variable for the nodes

RolandSherwin:
- This PR adds an option to set environmental variables during the `add/upgrade` commands.
- During the upgrade process:
	- if the env variable arg is provided, we write that into the service definition file, overwriting any previous values.
	- if the env variable arg is not provided, we will be re-using the previous value that was provided during `add`. To achieve this, we will have to store the `environment_variables` into the `NodeRegistry` file. Thus added a new field there.

### DIFF
--- a/sn_node_manager/src/main.rs
+++ b/sn_node_manager/src/main.rs
@@ -119,10 +119,11 @@ pub enum SubCmd {
         ///
         /// If not set, the RPC server is run locally.
         rpc_address: Option<Ipv4Addr>,
-        /// Provide the environmental variable that is set for the safenode service.
+        /// Provide environment variables for the safenode service.
         ///
         /// This is useful to set the safenode's log levels. Each variable should be comma separated without any space.
-        /// Example usage `--env SN_LOG=all,RUST_LOG=libp2p=debug`
+        ///
+        /// Example: --env SN_LOG=all,RUST_LOG=libp2p=debug
         #[clap(name = "env", long, use_value_delimiter = true, value_parser = parse_environment_variables)]
         env_variables: Option<Vec<(String, String)>>,
         /// Provide a safenode binary using a URL.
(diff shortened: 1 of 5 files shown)


sn_node_manager/src/main.rs
```diff
@@ -108,18 +108,23 @@ pub enum SubCmd {
...
         /// This option only applies when a single service is being added.
         #[clap(long)]
         port: Option<u16>,
         #[clap(long)]
+        /// Specify an Ipv4Addr for the node's RPC service to run on. This is useful if you want to expose the
+        /// RPC server outside. The ports are assigned automatically.
+        ///
+        /// If not set, the RPC server is run locally.
         rpc_address: Option<Ipv4Addr>,
+        /// Provide the environmental variable that is set for the safenode service.
```
jacderida: I would suggest changing the language of this a little bit:
```
Provide environment variables for the safenode service.
```

sn_node_manager/src/main.rs
```diff
@@ -108,18 +108,23 @@ pub enum SubCmd {
...
+        ///
+        /// This is useful to set the safenode's log levels. Each variable should be comma separated without any space.
+        /// Example usage `--env SN_LOG=all,RUST_LOG=libp2p=debug`
```
jacderida: I would suggest changing this to:
```
This is useful to set the safenode's log levels. Each variable should be comma separated without any space.

Example: --env SN_LOG=all,RUST_LOG=libp2p=debug
```

With a line break between the two lines, and the backticks removed. The backticks don't get parsed in the terminal UI, so I generally have a preference not to use them.

sn_node_manager/src/main.rs
```diff
@@ -336,6 +341,12 @@ pub enum SubCmd {
         /// The name of the service to upgrade
         #[clap(long, conflicts_with = "peer_id")]
         service_name: Option<String>,
+        /// Provide the environmental variable that is set for the safenode service.
```
jacderida: I think this should probably match the documentation for the one above.

jacderida (approved):

PR #1315, opened 2024-02-15
docs: rename_readme_dirs

JasonPaulGithub:
## Description

reviewpad:summary

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,21 +367,6 @@ Feel free to clone and modify this project. Pull requests are welcome.<br>You ca
 
 We follow the [Conventional Commits](https://www.conventionalcommits.org/) specification for all commits. Please make sure your commit messages adhere to this standard.
 
-### TroubleShooting
-
-If you get this error message:
-```
-Error: 0: missing field `local` at line 1 column 829
-Location: sn_protocol\src\node_registry.rs:182
-```
-Remove the following file from your safe directory:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -367,6 +367,21 @@ Feel free to clone and modify this project. Pull requests are welcome.<br>You ca
 
 We follow the [Conventional Commits](https://www.conventionalcommits.org/) specification for all commits. Please make sure your commit messages adhere to this standard.
 
+### TroubleShooting
```
jacderida: Hey, sorry for the little nitpick, but "troubleshooting" is one word, so the capital isn't necessary on the 's'.

I would also suggest changing this just a little bit:
When attempting to launch a local network, if you encounter an error message similar to the following:
...

It indicates a change in the `NodeRegistry` struct, such that it is no longer compatible with a previously serialized representation.

Delete the node registry file on your platform:
C:\Users\<username>\AppData\Roaming\safe\node_registry.json
~/.local/share/safe/node_registry.json

joshuef (changes requested):
This is way too specific. So the top changes remove `chris` are fine. 

But the throubleshooting section doesnt make sense to me.

joshuef:
can you update the commit comment to better reflect the PR now, @JasonPaulGithub ? 🙇

joshuef (approved):

PR #1319, opened 2024-02-19
feat(networking): remove all pending replication from failed nodes.

joshuef:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/replication_fetcher.rs
+++ b/sn_networking/src/replication_fetcher.rs
@@ -257,9 +257,11 @@ mod tests {
 
         sleep(FETCH_TIMEOUT + Duration::from_secs(1)).await;
 
-        // all the previous fetches should have failed and fetching next batch
+        // all the previous fetches should have failed and fetching next batch...
         let keys_to_fetch = replication_fetcher.next_keys_to_fetch();
-        assert_eq!(keys_to_fetch.len(), MAX_PARALLEL_FETCH);
+        // but as we've marked the previous fetches as failed, that node should be entirely removed from the list
+        // leaving us with just _one_ peer left
+        assert_eq!(keys_to_fetch.len(), 1);
         let keys_to_fetch = replication_fetcher.next_keys_to_fetch();
         assert!(keys_to_fetch.is_empty());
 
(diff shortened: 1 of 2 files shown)


sn_networking/src/replication_fetcher.rs
```diff
@@ -164,15 +167,20 @@ impl ReplicationFetcher {
...
             } else {
                 false
             };
-            is_fetching.is_none() || !is_expired
+
+            // if the node isn't known to be slow or the fetching is not expired we keep the entry
+            !failed_holders.contains(holder) && ( is_fetching.is_none() || !is_expired )
         });
+
+        // no we remove all failed holders
```
maqi: no -> now

sn_networking/src/replication_fetcher.rs
```diff
@@ -155,7 +155,10 @@ impl ReplicationFetcher {
 
     // Just remove outdated entries, which indicates a failure to fetch from network.
     // Leave it to to the next round of replication if triggered again.
-    fn prune_expired_keys(&mut self) {
+    // Also remove the corresponding node from to_be_fetched.
+    // And triggers the upper layers to potentially remove that failing node from the routing table.
+    fn prune_expired_keys_and_slow_nodes(&mut self) {
+        let mut failed_holders = vec![];
```
maqi: better make `failed_holder` as `BTreeSet` to reduce duplication

sn_networking/src/replication_fetcher.rs
```diff
@@ -164,15 +167,20 @@ impl ReplicationFetcher {
...
                 } else {
                     false
                 }
             } else {
                 false
             };
-            is_fetching.is_none() || !is_expired
+
+            // if the node isn't known to be slow or the fetching is not expired we keep the entry
+            !failed_holders.contains(holder) && ( is_fetching.is_none() || !is_expired )
```
maqi: failed_holders will only be filled when hit an expired entry,
which means: for those entries in the `to_be_fetched` before the `expired` one, they will be retained as at that moment, the `node in question` is not in the `black list` yet.

maqi (approved):

PR #1335, opened 2024-02-20
feat: refactor out file rs functions into sub dirs with imports and mods (concept only, do not merge)

JasonPaulGithub:
## Description

The idea is to separate the functionality in file.rs out into a more defined scope:

so instead of just `file.rs`, we have:

```
src -> subcommands -> transfer {
- chunks -> chunk manager
- application -> upload, download, estimate
- directories -> file, folder
}
```

This is a concept awaiting feedback that can later be broken down into smaller PRs during downtime. These changes are simple but will have a large impact. Feedback is very welcome.


happybeing:
Seems a good idea and I'll be interested to look at it.

Do you know if there are plans to provide file create, open, read, write etc or will MaidSafe be stopping at file upload/download from/to disk?

JasonPaulGithub:
> Seems a good idea and I'll be interested to look at it.
> 
> Do you know if there are plans to provide file create, open, read, write etc or will MaidSafe be stopping at file upload/download from/to disk?

Hello @happybeing!
Thank you for taking an interest in my PR! 😄 

I've decided to close it for the time being because I realise that I need to look more into our use of Eyre as a dependency. 
But I will try again soon :)

I can't say much about future plans but breaking things down is always a good help :)

Cheers!

happybeing:
I was looking at Eyre too today for my new projects as I'd seen MaidSafe are using it.

I found a good blog which was a good refresher on Rust error handling.

Good luck.

PR #1336, opened 2024-02-20
feat: refactor out file rs fns into corresponding sub dirs with impor…

JasonPaulGithub:
…ts and mods and no comp errs plus clippy fix

## Description

reviewpad:summary


JasonPaulGithub:
@happybeing  👍

PR #1342, opened 2024-02-21
feat(manager): provide option to start the manager as a daemon

RolandSherwin:
## Description

reviewpad:summary


Justfile
```diff
@@ -109,12 +109,14 @@ build-release-artifacts arch:
     cross build --release --features=network-contacts --target $arch --bin safe
     cross build --release --features=network-contacts --target $arch --bin safenode
     cross build --release --target $arch --bin safenode-manager
+    cross build --release --target $arch --bin safenode-manager-daemon
```
jacderida: My preference would be to change the name of this binary. I took at look at how Docker handles the situation. So they have `docker`, which is the client, and `dockerd` is the daemon. I think that would be cool for us too, so it would be `safenode-manager` and `safenode-managerd`. Actually, it would be really cool if we could get something even shorter than `safenode-manager` also for the client binary, but I don't have any useful suggestions.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -123,6 +124,22 @@ pub enum SubCmd {
         #[clap(long)]
         version: Option<String>,
     },
+    /// Start the safenode manager as a long running daemon
+    ///
+    /// This allows you to interact with the safenode manager through a RPC
+    Daemon {
```
jacderida: I feel like we should provide this as a subcommand. With respect to dealing with the faucet as a service, I've implemented that as a subcommand, so you use `faucet add`, `faucet start`, `faucet stop`, which keeps it consistent with node management. I think we should have the same thing here, because we should offer the user the ability to stop the daemon. Though actually stopping the daemon doesn't have to be implemented as part of this PR.

sn_node_manager/src/daemon_control.rs
```diff
@@ -0,0 +1,237 @@
...
+        )
+        .await
+        .map_err(|err| {
+            eyre!(
+                "Error while starting node {:?} with: {err:?}",
+                current_node.service_name
+            )
+        })?;
+    } else {
+        // else start a new node instance.
```
jacderida, line 221: You don't actually appear to be restarting a node here: you're creating a new one. Also, the previous node is not being killed, as far as I can see. In my opinion, it would make more sense to have `remove` and `add` commands that correspond to the same commands on the manager, and you would issue those commands in the scenario where you want a new node.

Justfile
joshuef: throwing stuff out:

`nodeman` `safenodeman` `snman``safe-nm` ?

jacderida: I think `snman` is not bad, but `safenodeman` would be more consistent with the other binary names, i.e., `safe` and `safenode`. That would give us `safenodeman` and `safenodemand`. That made me also think about just `safenodem`, but `safenodemd` is a little weird.

RolandSherwin: I will change the naming to `safenodemand` now. But will change the `safenode-manager -> safenodem` after this PR gets in. We might have to update a lot of repos for the latter.

sn_node_manager/src/bin/cli/main.rs
RolandSherwin: Cool! I will refactor the code to accommodate that.

sn_node_manager/src/daemon_control.rs
RolandSherwin: Right yeah. That makes more sense. Will modify it.

jacderida (approved):

PR #1357, opened 2024-02-22
Introduce `faucet` subcommand

jacderida:
- 11788e87 **feat: provide `faucet add` command**

  A new `faucet` subcommand is introduced, enabling the faucet to be controlled as a service. This is
  necessary because as with nodes, the faucet also needs to be upgraded on a new deployment; it may
  also require downgrading in the event of a problem.

  The existing `faucet` command, which was created as a utility for the backwards compatibility test,
  is removed; now that we can define the faucet as a service, it will no longer be necessary.

  The function that determines the data directory for the faucet was being duplicated in two different
  crates, and the node manager would have added a third, so it was refactored as public in
  `sn_transfers` and shared from there.

- a319ebbf **feat: provide `faucet start` command**

  This command is quite similar to the `start` command for starting a node service, but my judgement
  is there probably isn't enough code that is common between them to extract out something more
  generic.

  BREAKING CHANGE: the `faucet_pid` field of the node registry was removed in favour of a `faucet`
  field, which provides more details, including the PID.

- 5e3b93ce **refactor: create a `faucet_control` module**

  Rather than having a `faucet` module within `node_control`, I thought it made sense to also have a
  `faucet_control` module, and I've organised it the same as `node_control`.

- 3a2a9659 **feat: provide `faucet stop` command**

  As with the `faucet start` command, this provides the corresponding `stop` command. I revisited the
  idea of sharing code between the node and faucet services, but again came to the conclusion that
  there wasn't much value added for the complexity it introduced. The result is that we have code
  between them that only has small differences, but I think this is OK for now.

## Description

reviewpad:summary


RolandSherwin (approved):

sn_node_manager/src/faucet_control/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+use colored::Colorize;
+use sn_protocol::node_registry::{Faucet, NodeRegistry, NodeStatus};
+
+/// Install the faucet as a service.
+///
+/// This only defines the service; it does not start it.
+///
+/// There are several arguments that probably seem like they could be handled within the function,
+/// but they enable more controlled unit testing.
+pub async fn add_faucet(
```
RolandSherwin: This function can be made non-async.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -426,34 +460,108 @@ async fn main() -> Result<()> {
...
+            FaucetSubCmd::Start {} => {
+                if !is_running_as_root() {
+                    return Err(eyre!("The start command must run as the root user"));
+                }
 
-            let release_repo = <dyn SafeReleaseRepositoryInterface>::default_config();
-            let faucet_path =
-                get_bin_path(build, path, ReleaseType::Faucet, version, &*release_repo).await?;
+                let mut node_registry = NodeRegistry::load(&get_node_registry_path()?)?;
+                if let Some(faucet) = node_registry.faucet.clone() {
```
RolandSherwin: Just a tiny nit, but marking the var as `mut` can help us remove the following line, `let mut faucet = faucet`.
```suggestion
                if let Some(mut faucet) = node_registry.faucet.clone() {
```

PR #1363, opened 2024-02-23
feat(node): record_store chunk in batch and setup distance_range

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -466,6 +466,19 @@ impl SwarmDriver {
                 if !new_keys_to_fetch.is_empty() {
                     self.send_event(NetworkEvent::KeysToFetchForReplication(new_keys_to_fetch));
                 }
+
+                // The record_store will prune far records and setup a `distance range`,
+                // once reached the `max_records` cap.
+                if let Some(distance) = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .store_mut()
+                    .get_distance_range()
+                {
+                    self.replication_fetcher.set_distance_range(distance);
+                }
+
                 if let Err(err) = result {
                     error!("Cann't store verified record {record_key:?} locally: {err:?}");
                     cmd_string = "PutLocalRecord error";
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -204,6 +204,11 @@ impl NodeRecordStore {
         self
     }
 
+    /// Returns the current distance range
+    pub fn get_distance_range(&self) -> Option<Distance> {
+        self.distance_range
+    }
+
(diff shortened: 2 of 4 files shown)


sn_networking/src/cmd.rs
```diff
@@ -466,6 +466,19 @@ impl SwarmDriver {
                 if !new_keys_to_fetch.is_empty() {
                     self.send_event(NetworkEvent::KeysToFetchForReplication(new_keys_to_fetch));
                 }
+
+                // The record_store will prune far records and setup a `distance range`,
+                // once reached the `mac_records` cap.
```
joshuef: `max_records`

sn_networking/src/record_store.rs
```diff
@@ -282,57 +287,35 @@ impl NodeRecordStore {
...
-                    if furthest_record_key.distance(&self.local_key) < distance_range {
-                        warn!("Pruned record would also be within our distance range.");
-                    }
-                }
-            } else {
-                // we should not prune, but warn as we're at max capacity
-                warn!("Maximum number of records reached. Current num_records: {num_records}");
-            }
-        }
+        let mut sorted_records: Vec<_> = self.records.keys().cloned().collect();
```
joshuef, line 299: I wonder if instead of sorting on every write, we check if this new record is farther than a previous farthest? And just store that distance? 

That might be computationally less strenuous? And lets it be one in one out in a simpler fashion?

sn_networking/src/replication_fetcher.rs
```diff
@@ -39,6 +39,8 @@ pub(crate) struct ReplicationFetcher {
     // Avoid fetching same chunk from different nodes AND carry out too many parallel tasks.
     on_going_fetches: HashMap<(RecordKey, RecordType), (PeerId, ReplicationTimeout)>,
     event_sender: mpsc::Sender<NetworkEvent>,
+    // Distance range that the incoming key shall be fetched
```
joshuef, line 42: Is this more like the `farthest record we hold`?

maqi: not exactly.
for incoming new data, we store it anyway (to give margin for the client payee selection and allow the record to be `replicated out`, if the picked payee is an `edging` one)
hence, there is chance that `the incoming record is farther than the distance_range, but we still have to store it`, even though that record will have a high chance got `pruned` within the next round of pruning.

sn_networking/src/record_store.rs
maqi: the incoming one could be much closer than the stored ones, which means even we recorded the `farthest record`, after the `one to one swapping for the incoming one`, we still need to figure out which one is the real farthest among the left ones.

joshuef (approved):

PR #1364, opened 2024-02-25
feat(faucet): claim using signature of safe wallet

iancoleman:
The bitcoin maid address is used to sign a safe wallet address. The signed wallet address is used by the faucet as the recipient of the distribution for that maid.

This still uses snapshot to determine distributions, but instead of the faucet creating a random wallet to distribute to, it uses the safe wallet address submitted by the user.

This removes the need for ecies encryption/decryption.

This also removes the need for people to have the maid private key regularly available, since only the signature is required to claim.

Hardware wallets don't have ecies but can sign a message, so this makes the claim process much easier for people with maid using hardware wallets.

The main downside to this approach is safe wallet keys are not currently stored securely, so users may submit a safe wallet address which isn't secure (fine for test nets) and then neglect to update their safe wallet address before main net.

This PR only implements the faucet side of the logic. The client currently has no way to get their distribution. The client code is ready to go but will be a separate PR after this is merged (or not).

Some conversation about using ecies vs signatures for claiming:
https://safenetforum.org/t/update-25-january-2024/39127/74
https://safenetforum.org/t/update-8-february-2024/39188/39

## Description

reviewpad:summary


iancoleman:
Any thoughts on ways to fix the devskim failure? How can the tests be structured so they don't include hardcoded keys?

I'll look into how to remove the hard-coded elliptic curve but it's not really clear how to create or validate a signature without it (see [this test source code](https://github.com/rust-bitcoin/rust-bitcoin/blob/d85817b880075fe6558c01859fbaea66967a9d64/bitcoin/src/sign_message.rs#L230-L236) for the basis of the bitcoin signature code used in this PR)

sn_faucet/src/token_distribution.rs
```diff
@@ -406,32 +490,34 @@ mod tests {
...
+        // at first there are no claims because nobody has submitted a wallet
+        //assert!(!pubkeys.is_empty());
 
-        let candidates = snapshot
-            .iter()
-            .filter(|(addr, _amount)| pubkeys.contains_key(*addr))
-            .count();
-        println!("Got {candidates:?} distribution candidates");
-        assert!(candidates >= 1);
+        //let candidates = snapshot
```
joshuef: Can we clean out this code if not needed? 🙇

joshuef (approved):

joshuef (approved):
Nice one, thanks for this @iancoleman . 

One minor non-blocking comment there.

joshuef:
> Any thoughts on ways to fix the devskim failure?

So [here shows how to ignore devskim rules](https://stackoverflow.com/questions/75621124/how-to-suppress-devskim-issue-in-a-json-file) so we could add that if we know the rule. eg hard coded curve is `DS440001` I think.

If we cannot generate random keys for use in the test, I think it's okay to have hard coded ones. (And that may be a requirement for distribution file addresses eg?) 

In which case I think ignoring them with a small explanation of why is fine.

joshuef:
I won't merge this yet so as to give you chance to sort the devskim errors if the above makes sense. 

Once you're happy though we can get this in.

(I switched the base branch to `alpha` by the way, which is where were starting to focus dev in order to keep main more stable)

happybeing:
@iancoleman I'm very excited to see this. Can it be extended to eMAID too? If so: fantastic.

sn_faucet/src/token_distribution.rs
iancoleman: I removed this. Once we have collected some signatures and are storing them for use by the faucet at startup it can be added back in.

PR #1374, opened 2024-02-28
feat(cli): claim from snapshot using get-faucet

iancoleman:
This introduces a second way to get tokens from the faucet using the cli

`safe wallet get-faucet <url>`
Retains the existing fixed amount of tokens into the current wallet.
    
`safe wallet get-faucet <url> <maid_address> <signature>`
Receives a transfer from the maid snapshot for the amount of maid owned
by that address. The signature is a bitcoin signature by the maid
address with the message as the hex encoded safe wallet address that
will receive the transfer.

This uses the current safe wallet as the recipient.

It also introduces a few helpers for wallet management, not currently used by the cli but will be handy to include in future commands `safe wallet stash` and `safe wallet unstash <address>`

* `stash` - moves the current wallet into a new directory `wallet_<address>`
* `unstash` - moves the stashed wallet for the given address to the `wallet` directory - need to double check behavior for overwriting the existing wallet here (or not).
* `remove` - deletes the current wallet stored in `wallet` directory

These helpers arose for two main reasons

* there's no way to create an in-memory only wallet, mainly because of the dependency on `cash_notes` and `payments` files to make the wallet work correctly, so it's very common to move wallets around
* the backup for most blockchain systems is just the keys (usually via mnemonics) but this isn't enough to create a backup of a safe wallet, so these helpers are a starting point for backing up wallets.

## Description

reviewpad:summary

### DIFF
--- a/sn_cli/src/subcommands/wallet/helpers.rs
+++ b/sn_cli/src/subcommands/wallet/helpers.rs
@@ -6,13 +6,50 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+#[cfg(feature = "distribution")]
+use super::WalletApiHelper;
+#[cfg(feature = "distribution")]
+use base64::Engine;
 use color_eyre::Result;
 use sn_client::transfers::{HotWallet, SpendAddress, Transfer};
 use sn_client::Client;
 use std::{path::Path, str::FromStr};
 use url::Url;
 
-pub async fn get_faucet(root_dir: &Path, client: &Client, url: String) -> Result<()> {
+#[cfg(feature = "distribution")]
+pub async fn get_faucet(
+    root_dir: &Path,
+    client: &Client,
+    url: String,
+    address: Option<String>,
+    signature: Option<String>,
+) -> Result<()> {
+    if address.is_some() ^ signature.is_some() {
+        println!("Address and signature must both be specified.");
+        return Ok(());
+    }
+    if address.is_none() && signature.is_none() {
+        get_faucet_fixed_amount(root_dir, client, url).await?;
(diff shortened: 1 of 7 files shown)


iancoleman:
Ok... a lot of other commits showing up here... should I be rebasing off `main` or `alpha`? Any ideas what happened here?

sn_cli/src/subcommands/wallet/helpers.rs
```diff
@@ -6,13 +6,48 @@
...
-pub async fn get_faucet(root_dir: &Path, client: &Client, url: String) -> Result<()> {
+#[cfg(feature = "distribution")]
+pub async fn get_faucet(
+    root_dir: &Path,
+    client: &Client,
+    url: String,
+    address: Option<String>,
+    signature: Option<String>,
+) -> Result<()> {
+    if address.is_some() && signature.is_some() {
```
joshuef: We want to avoid `expect` in there, I know the outer check should make that fine, but if we were refactoring down the line this could open a can of `panic` worms instead of graceful error handling.

It's preferred to 
```rust
let addr = if let Some(addr) = address {
addr } 
else {
return Error(<SomeClearError> };

```
and handle that case explicitly with an error we can handle here/ elsewhere, rather than potentially panicking.

same for `sig`

joshuef (changes requested):

sn_cli/src/subcommands/wallet/helpers.rs
iancoleman: Thanks for the tip, always nice to get advice that removes potential panics. Fixed in https://github.com/maidsafe/safe_network/commit/34ea9e872e76ecce64ba7ba82492524a0a49a5ff

joshuef (approved):

PR #1379, opened 2024-02-29
Refactor upload with iterator

JasonPaulGithub:
## Description

This is a refactor and no real changes of logic should occur:

- Created associated functions and use of self in iterative_upload to extract the spawned task into its own function
- moved the file_api and the chunk_manager up closer to a level where it will eventually be reused by needed classes such as: upload, download, estimate.
- chunk manager will take on more responsibilities and further refactors to make way for future improvements
- I've changed &path to Path in config.rs because clippy would not stop bringing it up


JasonPaulGithub:
Closing this. Needs more work.

JasonPaulGithub:
Closing this. Needs more work.

PR #1396, opened 2024-03-06
add ci cargo publish check

joshuef:
This should ensure we don't miss errors that could affect publishing## Description

reviewpad:summary


JasonPaulGithub (approved):
`runs-on: ubuntu-latestruns-on: ubuntu-latest`
im guessing this code doesnt need to run on all operating systems. 
Otherwise looks good 👍

PR #1402, opened 2024-03-06
Cm refac

JasonPaulGithub:
## Description

functions that have chunks in the name and reference to chunk manager was scoped into chunk manager. 

that's it 👍 

-----------------------------------------------------------------------------------------------------------------------------------------

More details:
This refactor has been scaled down as I was blocked: I need to look more into how client and network does its job. More specifically I may look into decoupling a few things from and in client and network if possible. This is because the upload iterator does not use client very much - if at all. However it is passed in. This has become a lot more clear since previous refactors.

Ideally we want to pass client in to the chunk manager struct via 'new' and apply its uses where needed, and not even all of the client gets used (it seems), but injecting the client dependency into the chunk manager struct (client or network) causes issues with the derivates. Hence the need to research decoupling or other ways around this.

I can streamline our dependences and make the code more versatile when it's possible to find solutions to this. But for now I can still make a few changes.


maqi (changes requested):
The commit history (61 commits in the PR) is incorrect, shall get trimmed down.

PR #1409, opened 2024-03-07
Refactor spend verification

grumbach:
## Description

- better
- faster
- stronger


sn_client/src/api.rs
```diff
@@ -892,59 +892,44 @@ impl Client {
...
-                            Ok(_) => {
-                                trace!("Verified signed spend got from network for {address:?}");
-                                Ok(signed_spend)
-                            }
-                            Err(err) => {
-                                warn!("Invalid signed spend got from network for {address:?}: {err:?}.");
-                                Err(Error::CouldNotVerifyTransfer(format!(
-                                "Spend failed verifiation for the unique_pubkey {address:?} with error {err:?}")))
-                            }
+        let spends = get_signed_spends_from_record(&record)?;
```
maqi, line 895: it seems the following `match` block can be totally inside the `get_signed_spends_from_record` function if pass the `address` down ?

also, there is `parse_signed_spends` helper function inside `transfer already`, which already does the similar checks?

I am wondering maybe these two (get_signed_spends_from_record and parse_signed_spends) can be somehow merged ?
or at least put these two together to avoid consufing or neglectance in future ?

sn_networking/src/spends.rs
```diff
@@ -0,0 +1,57 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::{Error, Network, Result};
+use futures::future::join_all;
+use sn_transfers::{is_genesis_spend, SignedSpend, SpendAddress, TransferError};
+use std::{collections::BTreeSet, iter::Iterator};
+
+impl Network {
+    /// This function verifies a single spend.
+    /// This is used by nodes validation to before they store it.
```
maqi, line 16: This is used by nodes for spends validation, before storing them.

sn_networking/src/spends.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// - It does NOT check if the spend exists online
+    /// - It does NOT check if the spend is already spent on the Network
```
maqi, line 18: maybe move the below two `checks` up, and the before saying these two `not checks`, highlight those shall be checked by `client` only ?
also, maybe the function name better to be named to indicate that as well, say something like:
`node_verify_spend` or `verify_spend_ancester_and_integrity`

then for client, using function like `client_verify_spend` or `verify_spend_usage_history`

sn_networking/src/spends.rs
```diff
@@ -0,0 +1,57 @@
...
+            .map(|input| input.unique_pubkey);
+        let tasks: Vec<_> = parent_keys
+            .map(|a| self.get_spend(SpendAddress::from_unique_pubkey(&a)))
+            .collect();
+        let parent_spends: BTreeSet<SignedSpend> = join_all(tasks)
+            .await
+            .into_iter()
+            .collect::<Result<BTreeSet<_>>>()
+            .map_err(|e| {
+                let s = format!("Failed to get parent spend: {e}");
```
maqi, line 47: "Failed to get parent spend of {unique_key:?}: {e}",

sn_networking/src/transfers.rs
```diff
@@ -40,28 +40,25 @@ fn parse_signed_spends(address: &SpendAddress, record: &Record) -> Result<Signed
...
 
 impl Network {
-    /// Gets a spend from the Network.
+    /// Gets raw spends from the Network.
+    /// Double spends returned together as is, not as an error.
     /// The target may have high chance not present in the network yet.
     ///
     /// If we get a quorum error, we enable re-try
-    pub async fn try_get_spend(&self, address: SpendAddress) -> Result<SignedSpend> {
+    pub async fn get_raw_spends(&self, address: SpendAddress) -> Result<Vec<SignedSpend>> {
```
maqi, line 48: the aim of the function is to get just `one` spend.
returning of `vec` just accomodate the situation of `double spends` , which shall actuallly be reported as an error?

Hence, I don't get why no longer using `parse_signed_spends` and got function name changed as well.
sounds bit misleading and confusing to me.
As now it indicates it is a `valid` situation to have multiple spends at one address ?

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -110,8 +139,44 @@ impl SignedSpend {
         {
             Ok(())
         } else {
-            Err(Error::InvalidSpendSignature(*self.unique_pubkey()))
+            Err(TransferError::InvalidSpendSignature(*self.unique_pubkey()))
+        }
+    }
+
+    /// Verify the parents of this Spend, making sure that they where spent for it
```
maqi, line 146: Verify the parents of this Spend, making sure the input parent_spends are ancestors of self.

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -110,8 +139,44 @@ impl SignedSpend {
...
+    pub fn verify_parent_spends<'a, T>(&self, parent_spends: T) -> Result<()>
+    where
+        T: IntoIterator<Item = &'a SignedSpend> + Clone,
+    {
+        let unique_key = self.unique_pubkey();
+        trace!("Verifying parent_spends for {unique_key}");
+
+        // Check that the parent where all spent to our parent_tx
+        let tx_our_cash_note_was_created_in = self.parent_tx_hash();
+        for p in parent_spends.clone().into_iter() {
```
maqi, line 158: the check actualy indicates one spend shall have only one parent
i.e. as long as the input parent_spends has a length longer than 1, then it shall always be alramed as `error` ?
i.e. the check shall be simplified as:
```rust
if parent_spends.len() != 1 && parent_spends[0].spent_tx_hash() != self.parent_tx_hash() {
    return Err();
}
```

sn_node/src/put_validation.rs
```diff
@@ -614,206 +610,137 @@ impl Node {
...
-
-            // update signed_spends to contain all the spends (local and incoming)
-            debug!(
-                "We have {} unique spends in total for {unique_pubkey:?}",
-                all_spends.len()
-            );
-            signed_spends = all_spends.into_iter().collect();
+        // if we already have a double spend locally, no need to check the rest
+        let local_spends = self.get_local_spends(spend_addr).await?;
+        if let [a, b, ..] = local_spends.as_slice() {
```
maqi, line 658: what happens if already have it ? 
this check only cover the `double spend` case, not covering `already have` case.

for the case that the incoming one is same to the local existing one, then no need to undertake network fetch any more ?

grumbach:
replaced by https://github.com/maidsafe/safe_network/pull/1426

sn_networking/src/spends.rs
grumbach: Right! Adapted the comment! As the function can and should also be used by clients or third party apps, I think it's best to keep `node` out of the name.

sn_networking/src/transfers.rs
grumbach: Actually, this function is new and the `try_get_spend` was discarded, they serve completely different purposes. 
The point of this function is to get the spends stored on the network as-is without judging its value as a double spend or other errors. The point is to allow higher level functions to see and decide what to do with both spends. 
That's why I added `raw` in the name. 
Adapted the comments so people don't mistakenly use this one.

sn_transfers/src/cashnotes/signed_spend.rs
grumbach: > the check actualy indicates one spend shall have only one parent

I'm not sure I follow, spends can have many parents and be valid, as long as those parents are indeed spent as inputs in the Tx where our spend's unique key was created (as an output).

sn_node/src/put_validation.rs
grumbach: The case you're referring to is if we receive 1 spend and locally already have that same 1 spend? 
In that case, performing a network fetch makes sure that there is no double spend of this spend on other nodes. 
If we returned immediately, we might miss those.

grumbach: In the local double spend case, there is no need to check network to see if it is triple spent or more, as the spend is already poisoned locally, no more efforts are needed.

sn_client/src/api.rs
grumbach: Good one, `parse_signed_spends` seems to do the match job indeed! Let me adapt that one!

PR #1415, opened 2024-03-07
Uniform Service Management

jacderida:
- 49288213 **chore: new `sn_service_management` crate**

  There were some types defined in `sn_protocol` that shared some code between the node manager and
  tests in `sn_client`, along with Protobuf RPC definitions for the node manager, that have now
  been moved into a new crate, `sn_service_management`. These types should have never really been in
  `sn_protocol` in the first place; I put them there for lack of better place to share them.

  There was also a `test_utils` module which has been removed from `sn_protocol`. The code was put
  alongside the crates where it was used, since it didn't appear to be shared among multiple crates at
  this point in time, although it was generic enough such that it potentially could have been. Perhaps
  at the point where it needs to be shared, we can have a `test_utils` crate for the workspace.

- 5880944e **refactor: uniform service management**

  Introduce a service manager to treat the node, faucet, and daemon services in a uniform way. The
  services can be started, stopped or upgraded in a generic way, but the code for creating them
  remains specific. There is probably potential for service creation to be made generic, but it's more
  complex.

  There are now `NodeService`, `FaucetService`, and `DaemonService` types. They all implement a
  `ServiceActions` trait, allowing them to be treated generically by a `ServiceManager<T>`, where `T`
  is constrained to be a `ServiceActions` type.

  BREAKING CHANGE: the `RpcActions` and `RpcClient` were moved out of `sn_node_rpc_client` and into
  `sn_service_management` to avoid issues with circular referencing.

- 15f7288a **refactor: move rpc to its own module**

  The `node_control`, `faucet_control` and `daemon_control` modules will shortly be removed since we
  don't need duplicate service management code. With that in mind, we move the RPC code from
  `daemon_control` to its own module.

- 1d38ce32 **refactor: introduce `add_services` module**

  The only code now that is specific to each service is creating them. There is potential here to also
  make this process generic and move it into `ServiceManager`, but this branch is already big enough
  as it is. For now, the service creation code is moving into the `add_services` module. This removes
  the need for having separate `node_control`, `faucet_control` and `daemon_control` modules. The test
  code for the `add` functions is quite large, so I've retained the use of a `tests.rs` file.

- 53c89249 **ci: correctly run node manager unit tests**

  This package is now a library and binary, so running the unit tests needs a slightly different
  specification.

  Also fixes a couple of issues:
  * macOS-specific compilation error with use of the `?` operator by adding a `Utf8Error` type
  * Update the altered node during the `remove` command
  * Update the altered node during the `upgrade` command

- 8de5e5c8 **refactor: add daemon service behaves uniformly**

  Like the node and faucet services, the daemon has been changed to provide an install location for
  the binary.

  An error is returned if the service has already been created, rather than reinstalling it.

  Some additional output was also added to the daemon integration test to help debug failures.

- c58269ef **chore: extend `status` cmd for faucet and daemon**

  This should help for using the faucet and daemon services in an idempotent way, while allowing for a
  good user experience on the `add` command that will return an error if the services have already
  been created.

  The test util was also updated to deserialize to a utility status type, which makes it easier to
  work with than raw JSON `Value` objects.

## Description

reviewpad:summary


joshuef:
Closing in favour of https://github.com/maidsafe/safe_network/pull/1424 which is the same code, but atop main

PR #1420, opened 2024-03-10
chore: fix logging logic

loziniak:
## Description

reviewpad:summary


joshuef (approved):

loziniak:
Anyway, shouldn't `println!` be avoided here? What's the use of it? Looks like a leftover debugging. I think it's a good practice to print any output from API libraries only through logging infrastructure, to allow app developer have control over it.

PR #1422, opened 2024-03-10
chore(api): make logging::Error public

loziniak:
it's public in other crates, it makes error handling easier.

## Description

reviewpad:summary


joshuef (approved):
nice one thanks @loziniak 🙇

PR #1426, opened 2024-03-11
feat: refactor spend verification

grumbach:
## Description

reviewpad:summary


sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -110,8 +139,44 @@ impl SignedSpend {
...
+        T: IntoIterator<Item = &'a SignedSpend> + Clone,
+    {
+        let unique_key = self.unique_pubkey();
+        trace!("Verifying parent_spends for {unique_key}");
+
+        // Check that the parent where all spent to our parent_tx
+        let tx_our_cash_note_was_created_in = self.parent_tx_hash();
+        for p in parent_spends.clone().into_iter() {
+            let tx_parent_was_spent_in = p.spent_tx_hash();
+            if tx_our_cash_note_was_created_in != tx_parent_was_spent_in {
```
maqi, line 160: continue with the comment as at https://github.com/maidsafe/safe_network/pull/1409#discussion_r1519139078

> the check actualy indicates one spend shall have only one parent

what I mean about this is : code `tx_our_cash_note_was_created_in = self.parent_tx_hash()` shows there is one parent for a spend
and later on we compare ALL input_parent_spends' spent_tx_hash against this one, and expect they all to be the same.
Which, in theory, there shall be only one (otherwise it means two spends using the same spent_tx_hash ?)
hence, the incoming spends shall only have one entry?
otherwise it will return with error anyway ?

maqi: the hash function of `Transaction`, is a hash of vector of inputs , and each Input only contains the info of UniquePubKey and TokenAmount .
In case a spend was created from multiple inputs, it does valid to hold same `spent_tx.hash()` within all parent spends' `spent_tx`.

maqi (approved):

PR #1428, opened 2024-03-11
feat(networking): Add WebSocket as extra transport (multitransport)

b-zee:
- **feat(networking): support fallback WS transport**
- **feat(networking): listen on WS addr too**


joshuef:
@b-zee does this _always_ give us two transports? Or is it optional? (should it be?)

b-zee:
> @b-zee does this _always_ give us two transports? Or is it optional? (should it be?)

Only with `websockets` feature enabled. Would there be a scenario we'd want only WS and no QUIC?

joshuef:
> Would there be a scenario we'd want only WS and no QUIC?

No, i don't think so. At least not yet.

joshuef (approved):

PR #1430, opened 2024-03-11
chunks to upload to chunks manager with boolean workaround

JasonPaulGithub:
## Description

- I've made things a little more generic by moving out the `chunks to upload` from the `iterative upload` function. The result satisfies both the `upload` and the `acc pac` code that uses it. 
- I plan to work this further and rearrange the way `chunk manager` and `files api` is used. but this will do for now. 
- `messaging` code will be tidied up
- This will affect code by @bochaco a little bit

- Functions should do only one thing, so I've removed the counter feature inside the `iterative uploader`. That's a cool way to do it but the need for a counter has been decoupled.

I know it's not good to use the boolean flag for a different response. We're going to need to bridge the gap between `acc pac` and `upload` over its mutual use of `iterative upload`. They are using it in the same way yet require a separate response. Polymorphism / a builder can solve this. Either way the boolean flag we have will have to do for now as things aren't that complicated.

### DIFF
--- a/sn_cli/src/subcommands/acc_packet.rs
+++ b/sn_cli/src/subcommands/acc_packet.rs
@@ -14,9 +14,9 @@ use sn_client::registers::EntryHash;
 use sn_client::transfers::HotWallet;
 use sn_client::{Client, FilesApi, FolderEntry, FoldersApi, Metadata, WalletClient};
 
-use crate::subcommands::files;
 use crate::subcommands::files::{
-    download::download_file, iterative_uploader::IterativeUploader, upload::FilesUploadOptions,
+    self, download::download_file, iterative_uploader::IterativeUploader,
+    upload::FilesUploadOptions,
 };
 use color_eyre::{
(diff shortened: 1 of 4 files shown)


sn_cli/src/subcommands/acc_packet.rs
```diff
@@ -14,6 +14,7 @@ use sn_client::registers::EntryHash;
 use sn_client::transfers::HotWallet;
 use sn_client::{Client, FilesApi, FolderEntry, FoldersApi, Metadata, WalletClient};
 
+use crate::subcommands::files;
```
maqi: you can just put a `self` in below to avoid a separate import

i.e.  `use crate::subcommands::files::{self,  download::download_file, ... };`

or just import `chunks_to_upload` directly

JasonPaulGithub: ok, thanks

maqi (approved):

PR #1432, opened 2024-03-12
feat: DAG security improvements and fault detection

grumbach:
## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/audit/spend_dag_building.rs
+++ b/sn_client/src/audit/spend_dag_building.rs
@@ -9,7 +9,7 @@
 use super::{Client, SpendDag};
 use crate::{Error, Result};
 
-use futures::future::join_all;
+use futures::{future::join_all, StreamExt};
 use sn_networking::{GetRecordError, NetworkError};
 use sn_transfers::{SignedSpend, SpendAddress, WalletError, WalletResult};
 use std::collections::BTreeSet;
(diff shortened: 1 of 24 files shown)


sn_transfers/src/cashnotes/transaction.rs
```diff
@@ -205,15 +205,9 @@ impl Transaction {
...
-        // Verify that each signed spend is an output of our tx and that it is valid
+        // Verify that each signed spend is valid and was spent in this transaction
         let spent_tx_hash = self.hash();
         for s in signed_spends {
-            let pk = s.unique_pubkey();
-            let is_in_tx = |o: &Output| o.unique_pubkey() == pk;
-            if !s.spend.parent_tx.outputs.iter().any(is_in_tx) {
-                debug!("OutputNotFound: {pk:?} is not an output of tx {spent_tx_hash:?}");
-                return Err(TransferError::OutputNotFound);
-            }
```
grumbach, line 216: duplicate check removed

sn_client/src/audit/spend_dag_building.rs
```diff
@@ -213,39 +217,58 @@ impl Client {
...
-            })??;
-            dag.merge(sub_dag);
+
+        let tasks: Vec<_> = utxos
+            .iter()
+            .map(|utxo| {
+                debug!("Launching task to gather DAG from utxo: {:?}", utxo);
+                self.spend_dag_build_from(*utxo)
+            })
+            .collect();
```
RolandSherwin: Just a tiny suggestion, but we can probably use `futures stream` here to run atmost `n` tasks concurrently. I feel the utxos here can get very big and we might not want to spawn tasks for all of them.


Here is an example snippet
```rust
        let mut stream = futures::stream::iter(utxos.into_iter())
            .map(|utxo| self.spend_dag_build_from(utxo))
            .buffer_unordered(MAX_CONCURRENT_TASKS);


        while let Some(result) = stream.next().await {
			}
```

RolandSherwin (approved):

PR #1436, opened 2024-03-12
chore: iterate upload code rearranged for clear readability

JasonPaulGithub:
## Description

reviewpad:summary 

Iterate upload reads a lot easier from the top down. I would apply test docs but this function may soon see further adaptations for generic use by its two parent functions, but this will do for now.


sn_cli/src/subcommands/files/iterative_uploader.rs
```diff
@@ -104,7 +110,8 @@ fn spawn_progress_handler(
 ) -> JoinHandle<Result<(), Error>> {
     tokio::spawn(async move {
         let mut upload_terminated_with_error = false;
-        // The loop is guaranteed to end, as the channel will be closed when the upload completes or errors out.
+        // The loop is guaranteed to end, as the channel will be
```
jacderida, line 114: Really minor issue, but your line length seems a bit small for these comments. I think `rustfmt` uses 100 as a line length, which is a bit more modern than typical values like 72 or 80. I think that's also a decent line length for comments and doc comments.

JasonPaulGithub: resolved

sn_cli/src/subcommands/files/iterative_uploader.rs
```diff
@@ -95,7 +101,7 @@ impl IterativeUploader {
 
 ///////////////// Associative Functions /////////////////
 
-fn spawn_progress_handler(
+fn spawn_file_upload_events(
```
RolandSherwin: This function handles the file upload events that we receive, so `handler` seems more apt here?

JasonPaulGithub: i try to avoid descriptions like 'handler', its too generic

JasonPaulGithub: can we just call it **spawn_file_upload_events_handler**
the name gets a little longer but this PR has been up for days and I just want to get this through

RolandSherwin: Yes! I feel that is more descriptive!

RolandSherwin (approved):

sn_cli/src/subcommands/files/iterative_uploader.rs
JasonPaulGithub: changes made. cheers roland

PR #1437, opened 2024-03-12
refactor: reorganise command processing

jacderida:
The code in the `main` module of the node manager was becoming unwieldy and difficult to navigate, so I've refactored it into a few different modules.

There are absolutely no functional changes here.

Clippy was complaining about some functions having too many arguments, so I suppressed those warnings. These are just utility functions, and adding new structs just to encapsulate a few arguments would have been tedious and overkill.

## Description

reviewpad:summary

### DIFF
--- a/sn_node_manager/src/cmd/mod.rs
+++ b/sn_node_manager/src/cmd/mod.rs
@@ -1,0 +1,105 @@
+// Copyright (C) 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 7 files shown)


sn_node_manager/src/cmd/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+fn build_binary(bin_type: &ReleaseType) -> Result<()> {
+    let mut args = vec!["build", "--release"];
+    let bin_name = bin_type.to_string();
+    args.push("--bin");
+    args.push(&bin_name);
+
+    // Keep features consistent to avoid recompiling.
+    if cfg!(feature = "chaos") {
```
JasonPaulGithub, line 60: Perhaps it would be best to use the Match Block here instead of all these IF statements. We could then move the block out into it's own separate method. This is just a suggestion.

jacderida: Hey, thanks for the suggestion, but I don't think a match statement can be used with `cfg!`.

joshuef (approved):

PR #1440, opened 2024-03-13
feat(manager): add example to cause churn to a running network

RolandSherwin:
## Description

reviewpad:summary


jacderida:
Hey Roland, I think I had removed `test_utils` from `sn_protocol` in my refactor PR. It didn't look like it needed to be shared between any crates, and also, I don't think `sn_protocol` is the right place for sharing that kind of code.

sn_node_manager/src/bin/daemon/main.rs
```diff
@@ -72,9 +66,42 @@ impl SafeNodeManager for SafeNodeManagerDaemon {
...
+        let nodes_info = node_registry
+            .nodes
+            .iter()
+            .map(|node| Node {
+                peer_id: node.peer_id.map(|id| id.to_bytes()),
+                status: node.status.clone() as i32,
+                number: node.number as u32,
+            })
+            .collect::<Vec<_>>();
+        Ok(Response::new(NodesInfoResponse { nodes: nodes_info }))
```
jacderida: I think it would be better to call this command `Status`, corresponding with the command on the node manager.

RolandSherwin: Updated! Thank you!

RolandSherwin:
Hey @jacderida , I'm not able to find a more apt place to store the `test utils` that is shared among crates. Do you think we should have a separate crate for utilities?
Or what do you think about moving it to `sn_service_management`, but I feel like that would also not be a valid fit as the `DeploymentInventory` is a mock of a struct that is from a different repo.

jacderida (approved):

PR #1448, opened 2024-03-14
feat(log): set log levels on the fly

RolandSherwin:
## Description

reviewpad:summary


joshuef:
How can we test this?

Can we get a wee test that is starting logging... and then use `assert_log` or something (there's a crate that exists for this), that we have a) no TRACE and then b) some TRACE eg?

joshuef (changes requested):
see comment on a test

joshuef (approved):

Issue #1427, opened 2024-03-11
Integration Test for downloading the same file after saved changes


JasonPaulGithub:
_About a week ago I came across an issue when applying a bugfix. I created a video to recreate the bug and share it with the team. When I came back a few days later the issue was fixed. A few simple steps to recreate the bug can be found below._

```
Step 1: Create a network
cargo run --bin safenode-manager --features local-discovery -- run --build

Step 2: Upload a file
cargo run --bin safe --features local-discovery -- files upload <path>

Step 3: Download the file
cargo run --bin safe --features local-discovery -- files download

Step 4: 
Edit the file you have just uploaded, and save it. Then delete the recently downloaded file.

Step 6 : Upload again (step 2)

Step 7: Download for the second time. (Step 3)

Result:
You will find that the changes did not go through.
```

As the issue above is suspected to be an accidental fix, I would recommend writing an integration test to keep things in check.

linked to [1451](https://github.com/maidsafe/safe_network/pull/1451)

- [x] Test Windows 
- [ ] Test Linux

JasonPaulGithub:
Use of Download has been deprecated and replaced with use of folder download feature.

PR #1451, opened 2024-03-14
ci: Validate downloaded file after saved changes

JasonPaulGithub:
## Description

Validate downloaded file after saved changes.

Linked to issue [1427](https://github.com/maidsafe/safe_network/issues/1427)


.github/workflows/merge.yml
```diff
@@ -410,6 +410,27 @@ jobs:
           SN_LOG: "all"
         timeout-minutes: 2
 
+      - name: Validate downloaded file after saved changes for Windows
+        if: matrix.os == 'windows-latest'
```
joshuef, line 418: I dont think this need only be windows.

Here with run, we can set the `shell` to be `bash` and then write something that will test across all platforms and surface any inconsistency there.

joshuef (changes requested):

joshuef:
I don't think we need add a new `testfile` here either. We could take any project file (or a resources file), and upload / append to that. So Can we remove that too please 🙇 


Also, this is not a `feat` in the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) sense (that would be adding a feature to a part of the project).  I think we can scope this commit message as ci, eg `ci: Add test for updated file`

.github/workflows/merge.yml
```diff
@@ -410,6 +410,27 @@ jobs:
...
+          if ( $condition -ne 1) { exit 1 }
+          echo "add the values 2 and 3 to the test document"
+          "23" >> "./resources/test_doc.txt"
+          echo "Upload the recently changed test document"
+          cargo run --bin safe --release -- --log-output-dest=data-dir files upload "./resources/test_doc.txt" --retry-strategy quick
+          echo "Download the test document"
+          cargo run --bin safe --release -- --log-output-dest=data-dir files download --retry-strategy quick
+          echo "Test if the document includes the new values 2 and 3"
+          $condition = Get-Content C:\Users\runneradmin\Downloads\safe_files\test_doc.txt | ForEach-Object {$_ -Replace "[^0-9]", ""}
+          if ( $condition -ne 123) { exit 1 }
```
joshuef, line 433: I think it's worth digging into what this test is actually doing. 

`files upload` will upload a file, as chunks. 

If that file has changed, chunks will be different. 

You're not specifically testing that the file has changed on the network, because it doesn't. You're actually testing if the file reference was updated when you upload a new file with the same name I think. 

(@JasonPaulGithub Have a dig into what `files download` is doing here)

I'm not sure this test is a valid use test case given the above. 🤔

.github/workflows/merge.yml
JasonPaulGithub: Okay. Test operations for Linux and Windows have different dependencies. but I can write tests to cover both.

JasonPaulGithub:
> I don't think we need add a new `testfile` here either. We could take any project file (or a resources file), and upload / append to that. So Can we remove that too please 🙇
> 
> Also, this is not a `feat` in the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) sense (that would be adding a feature to a part of the project). I think we can scope this commit message as ci, eg `ci: Add test for updated file`

I do not think it is a good idea to piggyback the resource file for this test. If I am wrong then I would like to know which file you would recommend. Please note a text file would be best for encoding, and that it needs to be big enough to allow for the upload to take place.

.github/workflows/merge.yml
JasonPaulGithub: If I upload a file (with same name, extension, and path, but altered contents), and download grabs it all. I should get the latest version of this file. That's what I would assume as a user, is this correct?

This test emulates the same steps in its related [issue](https://github.com/maidsafe/safe_network/issues/1427)

JasonPaulGithub:
Use of Download has been deprecated and replaced with use of folder download feature.

PR #1455, opened 2024-03-15
feat: uniform behaviour for all `add` commands

jacderida:
- f9998ccf **feat: uniform behaviour for all `add` commands**

  The three `add` commands for `safenode`, `safenodemand` and `faucet` now behave in a uniform way,
  all supporting `--path`, `--url` and `--version` arguments. These enable a binary to be provided via
  a path, a URL pointing to a tar/zip containing the binary, or by a specific version number,
  respectively.

  The `safenodemand` service also now supports running as a non-root user.

- 53e2bbd8 **chore: upgrade `sn-releases` to new minor version**

  The new version of `sn-releases` uses a `Version` type rather than strings, and also renamed a
  trait, so the code had to be updated accordingly.
  
- 61a4245f **chore: run `safenodemand` service as root**

  Creating a user for running the `safenodemand` service doesn't make as much sense because the user
  must have root access. We can leave it up to the user to create an account and set it up for
  passwordless sudo, then provide an optional `--user` argument. For now we don't have this, and just
  go with root.

## Description

reviewpad:summary


RolandSherwin (approved):

JasonPaulGithub:
Marked do not merge but we have approvals, hope thats ok @RolandSherwin

RolandSherwin:
Hey @JasonPaulGithub, this PR depends on a change from `sn-releases` to land, so we cannot get this in until then.

JasonPaulGithub:
> Hey @JasonPaulGithub, this PR depends on a change from `sn-releases` to land, so we cannot get this in until then.

@RolandSherwin! 👍 👍

PR #1469, opened 2024-03-20
refactor: folders method extraction

JasonPaulGithub:
## Description

I've extracted some functions.


jacderida:
Minor suggestion: if the functionality remains the same, I think this commit type is more `refactor` than `chore`.

PR #1475, opened 2024-03-20
refactor: simple counter

JasonPaulGithub:
Simple counter to replace counter that threads through several functions


sn_cli/src/subcommands/files.rs
```diff
@@ -272,6 +266,18 @@ pub(crate) async fn files_cmds(
     Ok(())
 }
 
+fn file_count(file_path: &PathBuf) -> u32 {
```
joshuef: can we add some docs here please.

Ie, does this recursively count files in a dir and child folders?

joshuef (changes requested):

sn_cli/src/subcommands/files.rs
JasonPaulGithub: function renamed to count_files_in_path_recursively

joshuef (approved):

Issue #1468, opened 2024-03-20
CI Tests for the Upload and Download features


JasonPaulGithub:
New CI tests to be implemented. 

Issue blocked. Info here:
https://github.com/maidsafe/safe_network/pull/1484#issuecomment-2013961441

I am told that the blocked issue (above) can be solved via the way we run tests here `sn_client/tests/folders_api.rs`.
So based on this, we have the following list of things to do:

- [ ] Read into `sn_client/tests/folders_api.rs`
- [ ] Create new Branch with solutions to test the DL and Upload  features.

JasonPaulGithub:
Closed for now to get it off the PR list.

PR #1484, opened 2024-03-21
feat: test directories for acc pac with ci

JasonPaulGithub:
## Description

Upload and download on files and folders:
Solve for issue https://github.com/maidsafe/safe_network/issues/1468 which covers CI tests for Acc Pac. 
More creative tests can come from this, such as detecting if an empty directory has been attempted for upload, double download, sync files etc.


JasonPaulGithub:
@joshuef 

Facing an issue here on writing a CI level check for the downloads. The download command requires a network address, which is given after the upload has finished. It's a different address each time it's uploaded. I need to somehow feed this into the CI-CLI.
Is this why the tests at the moment are ran from 'sn_client/tests/folders_api.rs', as it's sort of an internal thing?
does running them like this mean that that data is tested is or isnt tested on a network?

There are things to look into here, perhaps @bochaco has plans for this that I am not aware of. 
@bochaco Please let me know what's best here. Cheers!

JasonPaulGithub:
@bochaco @joshuef 

Facing an issue here on writing a CI level check for the downloads. The download command requires a network address, which is given after the upload has finished. It's a different address each time it's uploaded. I need to somehow feed this into the CI-CLI. Perhaps this why the tests at the moment are ran from 'sn_client/tests/folders_api.rs'.

Issue is blocked for now

PR #1485, opened 2024-03-21
feat(cli): expose AccountPacket APIs from a lib so it can be used by other apps

bochaco:
Also moving chunk-mgr and files uploader/downloader to the lib.

## Description

reviewpad:summary


sn_cli/Cargo.toml
```diff
@@ -14,6 +14,10 @@ version = "0.89.85"
 path="src/main.rs"
 name="safe"
 
+[lib]
+path="src/lib/lib.rs"
```
jacderida: Is `lib.rs` generally not at the root of the crate?

With the other crates that have both libraries and binaries, I think we've tended to put the binaries under a `bin` directory, rather than making the library a directory.

Also, does the `name` here affect the name of the crate?

bochaco: I guess it's just up to us to define the struct, I can create a bin dir if that's preferred.

> does the name here affect the name of the crate?

I think if you don't specify a name it defaults to the package's name

jacderida: The node manager is also a binary and library crate, with the `main.rs` for the binaries under `bin`. I don't think I've ever seen `lib.rs` under a directory. I think it would probably be easier if this crate was consistent with the other ones. However, it's only a suggestion. If it was a huge hassle for you to change, no big deal.

bochaco: it sounds good, i'll move it

jacderida: I notice on this change, there is still a `lib` directory though. Shouldn't all that code be at the root of the crate?

bochaco: ah yes, let me fix that, i was testing it and forgot about it

JasonPaulGithub (approved):

PR #1491, opened 2024-03-23
feat: use Arc inside Client, Network to reduce clone cost

RolandSherwin:
These are just read only values, so a simple Arc would make it more efficient. 
## Description

reviewpad:summary


joshuef (approved):

joshuef:
nice one!

PR #1495, opened 2024-03-25
feat: remove gossip code

joshuef:
BREAKING CHANGE: remove gossip as royalties will be collected via DAG

## Description

reviewpad:summary


sn_node/src/event.rs
```diff
@@ -67,14 +67,6 @@ pub enum NodeEvent {
...
-    /// Gossipsub message received
-    GossipsubMsg {
-        /// Topic the message was published on
-        topic: String,
-        /// The raw bytes of the received message
-        #[debug(skip)]
-        msg: Bytes,
-    },
     /// Transfer notification message received for a public key
     TransferNotif {
```
grumbach: I think `TransferNotif` is also related to gossip

sn_node_rpc_client/src/main.rs
```diff
@@ -67,26 +64,6 @@ enum Cmd {
         #[command(flatten)]
         peers: PeersArgs,
     },
-    /// Subscribe to a given Gossipsub topic
```
grumbach, line 70: `TransfersEvents {` just above this is also part of gossip

grumbach (approved):
LGTM! Spotted a couple of remaining zombie types, but looking good!

joshuef:
Sorted that. Merged it in,

PR #1497, opened 2024-03-25
feat: re-enable relevant record pricing as the basis for cost/storage

joshuef:
## Description

reviewpad:summary


maqi (changes requested):
I'd suggest to split this PR into three:
1, lower the exponential pricing trigger to be 60%
2, remove `payment_received times` factor in pricing
3, re-introduce `relevant records` calculation

so that they can be better discussed/approved and merged better and quicker

joshuef:
@maqi I appreciate that separating things out might clarify each, but I'm not clear what's contentious in those parts?

for me: 

> 2, remove payment_received times factor in pricing

This has to go as is subjective to the node and cannot be verified. (It was added as a simple way to get nodes paid sooner, but it plays with pricing in odd ways, so best to get rid).

>  3, re-introduce relevant records calculation

This is probably the most controversial? It's a simple change we'll need to evaluate though, so I wouldn't be shy about adding it back in, myself.

and `1` is something else to test, 90 felt too slow, so Let's try 60? Again, something to feel out, but now sure separate PRs is worthwhile (removing the `payment_received times` is a whack of refactoring on its own, so I'd like to avoid it if we can :D

sn_networking/src/record_store.rs
```diff
@@ -425,21 +423,22 @@ impl NodeRecordStore {
...
-            self.received_payment_count,
-            self.config.max_records,
-        );
-
-        // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
-        info!("Cost is now {cost:?} for {stored_records:?} stored of {MAX_RECORDS_COUNT:?} max, {:?} times got paid.",
-            self.received_payment_count);
-        NanoTokens::from(cost)
-    }
+        let record_keys_as_hashset: HashSet<Key> = self.records.keys().cloned().collect();
```
RolandSherwin, line 426: I think we can prevent a clone here if we just collect the `&Key`. You might have to modify `get_records_within_distance_range` to accept references of keys.

sn_networking/src/record_store.rs
```diff
@@ -425,21 +423,22 @@ impl NodeRecordStore {
...
 
-    /// Notify the node received a payment.
-    pub(crate) fn payment_received(&mut self) {
-        self.received_payment_count = self.received_payment_count.saturating_add(1);
+        if let Some(distance_range) = self.distance_range {
```
RolandSherwin, line 428: I think the distance_range is not set until we have reached the max num of records inside `prune_storage_if_needed_for_record`. So think we might have to tweak how we set the `distance_range` value.

PR #1501, opened 2024-03-25
chore(node): refactor pricing metrics

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -22,7 +22,7 @@ use sn_protocol::{
     storage::{RecordHeader, RecordKind, RecordType},
     NetworkAddress, PrettyPrintRecordKey,
 };
-use sn_transfers::NanoTokens;
+use sn_transfers::{NanoTokens, QuotingMetrics};
 use std::{
     collections::{BTreeMap, HashMap},
     fmt::Debug,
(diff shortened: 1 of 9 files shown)


sn_networking/src/cmd.rs
```diff
@@ -355,19 +355,13 @@ impl SwarmDriver {
...
             SwarmCmd::GetLocalStoreCost { key, sender } => {
                 cmd_string = "GetLocalStoreCost";
-                let record_exists = self
-                    .swarm
-                    .behaviour_mut()
-                    .kademlia
-                    .store_mut()
-                    .contains(&key);
-                let cost = if record_exists {
-                    NanoTokens::zero()
```
joshuef, line 346: Have we not lost the 0 cost if record exists here?

joshuef:
I see the main aim here is the `QuoteMetrics` and getting that verifiable. I think we should try that atop the separate changes to the pricing mechanics in #1497

sn_networking/src/cmd.rs
maqi: no, will still return Zero cost within the inner calculation.

maqi:
> I see the main aim here is the `QuoteMetrics` and getting that verifiable. I think we should try that atop the separate changes to the pricing mechanics in #1497

Ok, will update this once PR 1497 merged.

joshuef (approved):

PR #1502, opened 2024-03-25
feat(transfers): enable client to check if a quote has expired

RolandSherwin:
## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/files.rs
+++ b/sn_client/src/files.rs
@@ -126,7 +126,7 @@ impl FilesApi {
         trace!("Client upload started for chunk: {chunk_addr:?}");
 
         let wallet_client = self.wallet()?;
-        let (payment, payee) = wallet_client.get_recent_payment_for_addr(&chunk_addr)?;
+        let (payment, payee) = wallet_client.get_non_expired_payment_for_addr(&chunk_addr)?;
 
         debug!(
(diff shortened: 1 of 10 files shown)


sn_client/src/files.rs
```diff
@@ -126,7 +126,7 @@ impl FilesApi {
         trace!("Client upload started for chunk: {chunk_addr:?}");
 
         let wallet_client = self.wallet()?;
-        let (payment, payee) = wallet_client.get_recent_payment_for_addr(&chunk_addr)?;
+        let (payment, payee) = wallet_client.get_recent_payment_for_addr(&chunk_addr, true)?;
```
JasonPaulGithub: This here is a boolean flag, which is bad design. 
You may not need to apply this right now but it's worth considering this kind of thing in the future.

I have a short article here with reasons and solutions to consider:
https://luzkan.github.io/smells/flag-argument

JasonPaulGithub: A more detailed article (if you have time, it's the source article)
https://martinfowler.com/bliki/FlagArgument.html

RolandSherwin: Ah yeah, that is a good suggestion. I think we can have multiple functions there instead, as it is a public API.

RolandSherwin: Hey @JasonPaulGithub, I've updated the PR!

joshuef (approved):

PR #1507, opened 2024-03-26
chore(node): reduce bad_node check concureent queries

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_node/src/node.rs
+++ b/sn_node/src/node.rs
@@ -251,7 +251,7 @@ impl Node {
                             trace!("Periodic bad_nodes check took {:?}", start.elapsed());
                         });
 
-                        if rolling_index == 255 {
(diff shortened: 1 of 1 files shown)


sn_node/src/node.rs
```diff
@@ -251,7 +251,7 @@ impl Node {
                             trace!("Periodic bad_nodes check took {:?}", start.elapsed());
                         });
 
-                        if rolling_index == 255 {
+                        if rolling_index == 511 {
```
joshuef, line 254: 511?

maqi: because using `% 2` for the `upper / lower part` indexing to half the number of concurrent queries each time.

joshuef (approved):

PR #1514, opened 2024-03-27
feat(networking): only consider close records in price setting

joshuef:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/driver.rs
+++ b/sn_networking/src/driver.rs
@@ -28,7 +28,6 @@ use crate::{
     transport, Network, CLOSE_GROUP_SIZE,
 };
 use futures::StreamExt;
-use itertools::Itertools;
 use libp2p::kad::KBucketDistance as Distance;
 #[cfg(feature = "local-discovery")]
 use libp2p::mdns;
@@ -619,16 +618,10 @@ impl SwarmDriver {
                     }
                 }
                 _ = set_farthest_record_interval.tick() => {
-                    let closest_peers = self
-                    .swarm
-                    .behaviour_mut()
-                    .kademlia
-                    .get_closest_local_peers(&self.self_peer_id.into())
-                    .map(|peer| peer.into_preimage())
-                    .take(CLOSE_GROUP_SIZE)
-                    .collect_vec();
+                    let closest_k_peers = self
+                        .get_closest_k_value_local_peers();
 
-                    if let Some(distance) = self.get_farthest_relevant_record(&closest_peers) {
+                    if let Some(distance) = self.get_farthest_relevant_record(&closest_k_peers) {
                         // set any new distance to fathest record in the store
                         self.swarm.behaviour_mut().kademlia.store_mut().set_distance_range(distance);
                     }
--- a/sn_networking/src/driver.rs
+++ b/sn_networking/src/driver.rs
@@ -635,8 +635,8 @@ impl SwarmDriver {
     // --------------------------------------------
 
     /// Return a far address, close to but probably farther than our responsibilty range.
-    /// This simply uses the closest k peers to estimate the farthest address as CLOSE_GROUP_SIZE + 1 peer's
-    /// address distance.
+    /// This simply uses the closest k peers to estimate the farthest address as
+    /// `CLOSE_GROUP_SIZE + 1`th peer's address distance.
     fn get_farthest_relevant_address_estimate(
         &mut self,
         // Sorted list of closest k peers to our peer id.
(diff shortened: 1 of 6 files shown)


joshuef:
I understand now @maqi  why you were so hesitant to have the `relevant` check.. As we actually filter for relevance on inbound replicated data now (or at payment time). 
 
 Hmmm... Need to chew on cache + backup implications of this approach some.

joshuef:
Removing do not merge, as I think we need this either way. 

As churn happens, our area of responsibility shrinks, so the price needs to reflect that, not be based upon what we're holding. 

I know we only replicate in relevant data now, but as churn happens, some of that will become "out of scope". So we need to discard based upon that. 

We can memoise the fn for every X checks or so perhaps if perf is an issue.

RolandSherwin (approved):

sn_networking/src/record_store.rs
```diff
@@ -384,6 +391,14 @@ impl NodeRecordStore {
         let record_key = PrettyPrintRecordKey::from(&r.key).into_owned();
         trace!("PUT a verified Record: {record_key:?}");
 
+        let new_distance = self
+            .local_address
+            .distance(&NetworkAddress::from_record_key(&r.key));
+
+        // set's the local distance range to be further if the new record is further
+        // This assumes the verified record is to be held by _us_...
+        self.distance_range = self.distance_range.map(|range| range.max(new_distance));
```
RolandSherwin: As the network grows, would we not be getting records that are closer to us? So this would mean that we'll be hanging on to the older value?

joshuef: Aye, cleaned that out and moved farthest record check into an interval setter. Should be enough and prevent too much overzealous "close" checking on store etc.

sn_networking/src/record_store.rs
```diff
@@ -428,27 +428,41 @@ impl NodeRecordStore {
...
     #[allow(clippy::mutable_key_type)]
     pub(crate) fn store_cost(&self, key: &Key) -> (NanoTokens, QuotingMetrics) {
-        let quoting_metrics = QuotingMetrics {
-            records_stored: self.records.len(),
-            max_records: self.config.max_records,
-            received_payment_count: self.received_payment_count,
-            live_time: self.timestamp.elapsed().as_secs(),
-        };
+        let live_time = self.timestamp.elapsed().as_secs();
+        let mut records_stored = self.config.max_records;
```
maqi: prefer a default vaule of `0`, to avoid in case of zero cost and no distance set,
the metrics showing `stored_records` as the max_records.

sn_networking/src/record_store.rs
```diff
@@ -428,27 +428,41 @@ impl NodeRecordStore {
...
 
         let cost = if self.contains(key) {
-            0
+            NanoTokens::zero()
+        } else if let Some(distance_range) = self.distance_range {
+            let record_keys_as_hashset: HashSet<&Key> = self.records.keys().collect();
+            let relevant_records =
```
maqi, line 442: note that there is `pruning` to be carried out when hit max_records, which is regarding ALL.
i.e. this means `relevant_records` may way below max_records, and will never reach that bar.

It may be good, may be bad,

Also, I'd suggest just rename `QuotingMetrics::records_stored` to `QuotingMetrics::relevant_records` to avoid any potential confusion.

sn_networking/src/record_store.rs
```diff
@@ -428,27 +428,41 @@ impl NodeRecordStore {
...
+                self.get_records_within_distance_range(record_keys_as_hashset, distance_range);
```
maqi, line 443: looks like the correspondent `addr_distribution_sim` test not got updated to reflect this change?
this will result in an un-sync between the sim and real code.

joshuef: Hmmm, im not sure what needs to be changed here. 

I've updated some naming, and made things a bit clearer, but the crux of the test is entirely the same. We were already only calculating for close_records as there's no churn here.

sn_networking/src/driver.rs
```diff
@@ -608,6 +615,21 @@ impl SwarmDriver {
...
                     }
                 }
+                _ = set_farthest_record_interval.tick() => {
+                    let closest_peers = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .get_closest_local_peers(&self.self_peer_id.into())
+                    .map(|peer| peer.into_preimage())
+                    .take(CLOSE_GROUP_SIZE)
```
maqi: `is_in_close_range` will always return true when the input list is less than REPLICATION_RANGE (i.e. 7)
but here there is only CLOSE_GROUP_SIZE (5) peers to be passed down ?

maqi (changes requested):

maqi (approved):

PR #1522, opened 2024-03-27
feat: use Uploader for acc packets / folder

RolandSherwin:
## Description

Here is a PR to integrate the Uploader into folders/acc packets https://github.com/maidsafe/safe_network/pull/1522 This would allow them to retry + repay if there are failures. Please have a look. c.c. 
[@bochaco](https://maidsafe.slack.com/team/U3Y2UFG7P)


sn_cli/src/acc_packet.rs
```diff
@@ -109,7 +104,7 @@ impl AccountPacket {
     /// when the user requests to encrypt data, the provided Client's signer key is used.
     pub fn init(
         client: Client,
-        wallet_dir: &Path,
+        root_dir: &Path,
```
bochaco: Let's keep this to be the `wallet_dir` since that's what it is, root dir doesn't mean anything in this context. The acc-packet currently needs a dir where to load the wallet from, it doesn't use it for any other thing.

bochaco (changes requested):

sn_client/src/folders.rs
```diff
@@ -49,7 +46,7 @@ const REMOVED_ENTRY_MARK: XorName = XorName([0; XOR_NAME_LEN]);
 #[derive(Clone)]
 pub struct FoldersApi {
     client: Client,
-    wallet_dir: PathBuf,
+    root_dir: PathBuf,
```
bochaco: Let's keep this to be the `wallet_dir` since that's what it is in this context.

sn_cli/src/files.rs
```diff
@@ -92,30 +32,3 @@ pub fn get_progress_bar(length: u64) -> Result<ProgressBar> {
     progress_bar.enable_steady_tick(Duration::from_millis(100));
     Ok(progress_bar)
 }
-
-fn msg_files_already_uploaded_verified() {
```
JasonPaulGithub, line 96: I've seen these message functions move around so much. I might have to move them into their own rust file.

sn_cli/src/files/files_uploader.rs
```diff
@@ -0,0 +1,337 @@
...
+
+            entries_to_upload: Default::default(),
+            file_paths_to_print: Default::default(),
+
+            make_data_public: false,
+            upload_cfg: Default::default(),
+        }
+    }
+
+    pub fn set_upload_cfg(mut self, cfg: UploadCfg) -> Self {
```
JasonPaulGithub, line 60: I've looked into this a little bit,. but are getters and setters necessary in Rust?

I know these are cherry picked but I see it mentioned now again:
https://users.rust-lang.org/t/best-practices-for-visibility-of-fields/56087/3
https://users.rust-lang.org/t/best-practices-on-setters-and-accessor-methods-in-general/66530/2

sn_cli/src/files/files_uploader.rs
```diff
@@ -0,0 +1,337 @@
...
+            "Among {chunks_to_upload_len} chunks, found {} already existed in network, uploaded \
+            the leftover {} chunks in {elapsed}",
+            summary.skipped_count, summary.uploaded_count,
+        );
+        info!(
+            "Among {chunks_to_upload_len} chunks, found {} already existed in network, uploaded \
+            the leftover {} chunks in {elapsed}",
+            summary.skipped_count, summary.uploaded_count,
+        );
+        println!("**************************************");
```
JasonPaulGithub, line 148: This is message code. We need to extract this.

JasonPaulGithub (changes requested):
A few small pointers but the only dealbreaker for me is the message code.
I'd like to keep it in its own function for now.
I plan to move code like this into its own rs file so we don't have to keep moving it around, then I can handle it further on down the line,

sn_cli/src/files/files_uploader.rs
JasonPaulGithub: (just a consideration, no worries here)

sn_client/src/folders.rs
```diff
@@ -196,14 +205,55 @@ impl FoldersApi {
         Ok(())
     }
 
+    /// Sync multiple Folders with the network.
+    pub async fn sync_multiple(
```
bochaco: This is not correct here, all methods exposed by this struct shall be related to itself, a single Folder. If there is some upper layer holding a bunch of Folders then that's where this method/fn belongs to, not here. We do this in the acc-packet as an example, which is where we keep a set of folders. 

This method needs to be removed from this struct and moved up to the caller layer which is the one holding all the folders. Layers shall be respected for clarity and simplicity.

bochaco (changes requested):

sn_cli/src/files/files_uploader.rs
RolandSherwin: Hey Jason, the structs from the CLI is now exposed as a `library` and hence I feel like we should not be ideally printing things to stdout. 
We can probably make these into events / have an interface that can write to std or probably do something else.

Also, I'm not sure if separating those into their own functions help when modifying the code. I felt they were very hard to track.

RolandSherwin: Also, since the code is a short one, are so many redirections required?

JasonPaulGithub: Message code is a cross cutting concern and we're not dealing with it correctly. It's going to pad up the codebase and I don't think there is a good justification for this.

RolandSherwin: I feel like we can probably have events that callers of `FilesUploder` can consume to get these updates? And they would be responsible for printing the message (cli) or updating some progress bar (gui)?

bochaco (approved):

sn_cli/src/files/files_uploader.rs
JasonPaulGithub: Do what you think is best, but be aware that message code like this is a [cross cutting concern](https://www.linkedin.com/advice/0/how-can-design-patterns-solve-cross-cutting-concerns#:~:text=Design%20patterns%20can%20help%20you%20solve%20cross%2Dcutting%20concerns%20by,key%20principle%20of%20software%20design.)

In Java. Easy. Rust - Not so sure. I will look into this.

JasonPaulGithub (approved):

PR #1525, opened 2024-03-28
fix(client): old uploads should not fail due to updating quotes

joshuef:
Previously we could error out if all quotes were old, now we require at least double batch size errors from store cost quote fails

## Description

reviewpad:summary


RolandSherwin:
Closed in favor of https://github.com/maidsafe/safe_network/pull/1526

Issue #1487, opened 2024-03-22
Chunk manager should not error out if one file fails


RolandSherwin:
Track the failures to be retried / logged. 
The uploader can fail if the manager sends in 0 chunks.
https://safenetforum.org/t/bbbbbadnet-21-03-2024-testnet/39458/156

Related PR: https://github.com/maidsafe/safe_network/pull/1528
Related Slack Conversation: https://maidsafe.slack.com/archives/C050EDL5ACS/p1711548766280269

Note: We will need CI level tests for the changes required to solve this issue.

JasonPaulGithub:
we'll get a pattern match on `file_size` inside of `sn_client/src/files.rs` and apply the solutions mentioned [here](https://maidsafe.slack.com/archives/C050EDL5ACS/p1711548766280269) to the following outcomes;
1. empty file
2. file under threshold
3. file over threshold

Nic-dorman:
Closing as outdated

PR #1528, opened 2024-03-28
feat:  1487 Chunk Manager Empty Iteration Failure

JasonPaulGithub:
## Description

A redesign is needed for the chunk_file process to match three types of file.

1. An empty file: write into the data_map the string EMPTY. On decrypt, we detect this and just create an empty file.
2. A small file: The whole content of those can be in the data_map. Just in SE lib we write the < MIN_SIZE to the data map.
3. A qualified file: Same as below (no change)

Fix for Issue : https://github.com/maidsafe/safe_network/issues/1487

Manual testing needs to be done for this:
 - [x] Upload Empty File 
 - [x] Download Empty File 
 - [x] Upload Small File 
 - [x] Download Small File
 - [x] Upload Qualified File
 - [x] Download Qualified File


sn_client/src/files.rs
```diff
@@ -83,6 +83,12 @@ impl FilesApi {
         let file_size = metadata.len();
 
         let (head_address, data_map_chunk, mut chunks_paths) =
+
+            // todo: below we need to redesign the process to match three types of file.
+            // 1. empty file: write into the data_map the string EMPTY. On decrypt, we detect this and just create an empty file.
+            // 2. A small file: The whole content of those can be in the data_map. Just in SE lib we write the < MIN_SIZE to the data map.
```
maqi: the `MIN_ENCRYPTABLE_BYTES` is defined by the encryption crate, which is 3 bytes.
then within this branch you can just do:
```rust
if file_size < MIN_ENCRYPTABLE_BYTES as u64 {
    let content: Bytes = File::read(file_path);
    let chunk = Chunk::new(content);
    (chunk.name(), chunk, vec![])
}
```

Note this is to generate a Chunk with data_content, instead of data_map
Hence on the `parse` side, need to double check the code whether the case is covered as well.

If you track the commit history way back (Dec 2023 shall be enough), there shall be code handling `write small file to data_map` still presented

sn_client/src/files.rs
```diff
@@ -185,11 +194,37 @@ impl FilesApi {
...
+}
+
+trait ChunkBySize {
+    fn chunk_by_size(self) -> (Chunk, Vec<(XorName, PathBuf)>);
+}
+
+impl ChunkBySize for EmptyFile {
+    fn chunk_by_size(self) -> (Chunk, Vec<(XorName, PathBuf)>) {
+        let bytes: Bytes = Default::default();
+        let chunk = Chunk::new(bytes);
```
RolandSherwin, line 216: So I assume this would get an address of `0000..`?

sn_client/src/files.rs
```diff
@@ -185,11 +194,37 @@ impl FilesApi {
...
+
+impl SmallFile {
+    fn new(buffer: Vec<u8>) -> Self {
+        Self { buffer }
+    }
+}
+
+trait ChunkBySize {
+    fn chunk_by_size(self) -> (Chunk, Vec<(XorName, PathBuf)>);
+}
```
RolandSherwin, line 211: I'm not sure if we need extra structs and traits here? The logic is small enough that it can be moved inside `chunk_file()`?

happybeing:
Probably too late for this PR but if you are working more in this area, is there a chance you can add the file path (not just name as it is now) to the ChunkManager uploaded files tracking? More in this issue https://github.com/maidsafe/safe_network/issues/1576

I want to use the ChunkManager and to be able to access the xor address and full path of each uploaded file, not just the filenames.

sn_client/src/files.rs
JasonPaulGithub: Yes. It's required. Please read into this:
https://blog.logrocket.com/build-polymorphic-components-rust/#understanding-polymorphism-rust

sn_client/src/files.rs
JasonPaulGithub: Honestly, I'm not sure. I've implemented the solution [here](https://github.com/maidsafe/safe_network/pull/1528#discussion_r1542792481) as best I can.

JasonPaulGithub:
Hi @happybeing,
I think this is possible. @RolandSherwin What do you think?
Cheers 👍

sn_client/src/files.rs
JasonPaulGithub: I think so. Will this cause any issues?

JasonPaulGithub:
Closed as the upgraded version of this fix is available here
https://github.com/maidsafe/safe_network/pull/1583

joshuef:
the SmallFile here is not encrypted, so we're actually potentially exposing user data here. 

This is one reason we've opted to error out on small files for now until they can be reliably encrypted via the `Folders` commands. (though that's unintentionally leaking out via the error handling in the client here.

So I think the  solution here for #1487 is to handle the error, as opposed to attempting to chunk the file (but potentially not encrypting the data within). 


---------

If we accept that the storing of the file can throw an error for now, we can make a follow up issue to track encryption of small files via the `Folders` command apis.

How does that sound?

PR #1531, opened 2024-03-28
feat(node): bad quoting detection

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -22,7 +22,7 @@ use sn_protocol::{
     storage::{RecordHeader, RecordKind, RecordType},
     NetworkAddress, PrettyPrintRecordKey,
 };
-use sn_transfers::{NanoTokens, QuotingMetrics};
+use sn_transfers::{NanoTokens, PaymentQuote, QuotingMetrics};
 use std::{
     collections::{BTreeMap, HashMap},
     fmt::Debug,
@@ -42,6 +42,8 @@ pub enum NodeIssue {
     ReplicationFailure,
     /// Close nodes have reported this peer as bad
     CloseNodesShunning,
+    /// Provided a bad quote
+    BadQuoting,
 }
 
 /// Commands to send to the Swarm
@@ -169,6 +171,10 @@ pub enum SwarmCmd {
         target: NetworkAddress,
         sender: oneshot::Sender<bool>,
     },
+    // Quote verification agaisnt historical collected quotes
+    QuoteVerification {
+        quotes: Vec<(PeerId, PaymentQuote)>,
+    },
 }
 
 /// Debug impl for SwarmCmd to avoid printing full Record, instead only RecodKey
@@ -285,6 +291,9 @@ impl Debug for SwarmCmd {
             }
             SwarmCmd::IsPeerShunned { target, .. } => {
                 write!(f, "SwarmCmd::IsPeerInTrouble target: {target:?}")
+            }
+            SwarmCmd::QuoteVerification { quotes } => {
+                write!(f, "SwarmCmd::QuoteVerification of {} quotes", quotes.len())
             }
         }
     }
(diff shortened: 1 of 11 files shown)


sn_networking/src/cmd.rs
```diff
@@ -756,13 +709,98 @@ impl SwarmDriver {
...
+                ));
+                self.log_kbuckets(&peer_id);
+                let _ = self.check_for_change_in_our_close_group();
+            }
+        }
+    }
+
+    fn verify_peer_quote(&mut self, peer_id: PeerId, quote: PaymentQuote) {
+        if let Some(history_quote) = self.quotes_history.get(&peer_id) {
+            if !history_quote.historical_verify(&quote) {
```
JasonPaulGithub, line 790: you could shorten if statements like this into its own function labelled by its purpose, that way it would be much easier to gloss over if you're reading through, but no big deal 👍

JasonPaulGithub:
I don't know enough about the changes but I've taken a look through the code and it looks fine.
I'd give this a soft approval and a thumbs up 👍 

✅ ✅ ✅

sn_transfers/src/wallet/data_payments.rs
```diff
@@ -163,4 +163,67 @@ impl PaymentQuote {
...
+
+        if new_quote.quoting_metrics.live_time < old_quote.quoting_metrics.live_time {
+            info!("Claimed live_time out of sequence");
+            return false;
+        }
+
+        let old_elapsed = if let Ok(elapsed) = old_quote.timestamp.elapsed() {
+            elapsed
+        } else {
+            info!("timestamp failure");
```
joshuef, line 191: let's warn these and get some more info on the Peer that's failing logged out here I think

sn_transfers/src/wallet/data_payments.rs
```diff
@@ -163,4 +163,67 @@ impl PaymentQuote {
...
+            return false;
+        };
+
+        let time_diff = old_elapsed.as_secs().saturating_sub(new_elapsed.as_secs());
+        let live_time_diff =
+            new_quote.quoting_metrics.live_time - old_quote.quoting_metrics.live_time;
+        // In theory, these two shall match, give it a margin of 10 to avoid system glitch
+        if live_time_diff > time_diff + 10 {
+            info!("claimed live_time out of sync with the timestamp");
```
joshuef, line 206: Again, probs a `warn` or `debug` better here? (with peer info)

sn_transfers/src/wallet/data_payments.rs
```diff
@@ -163,4 +163,67 @@ impl PaymentQuote {
...
     }
+
+    /// Check whether self is newer than the target quote.
+    pub fn is_newer_than(&self, other: &Self) -> bool {
+        self.timestamp > other.timestamp
+    }
+
+    /// Check against a new quote, verify whether it is a valid one from self perspective.
+    /// Returns `true` to flag the `other` quote is valid, from self perspective.
+    pub fn historical_verify(&self, other: &Self) -> bool {
```
joshuef, line 174: Can we get some unit tests here to clearly lay out each of the error / fake quote cases?

sn_node/src/quote.rs
```diff
@@ -37,36 +39,97 @@ impl Node {
...
+                    };
+
+                    is_same_target && is_not_self && is_not_zero_quote && is_around_same_time
+                })
+                .cloned()
+                .collect();
 
-        Ok(())
+            quotes_for_nodes_duty.retain(|(peer_id, quote)| {
+                let cost = calculate_cost_for_records(
```
joshuef, line 113: Could this not just be passed the full quote to simplify things here?

joshuef:
@maqi there's a few comments there, and one thing that wasn't clear to me: what triggers a quote verification?

In general looking good I thnk. I'll make as approved so you can get log levels updated and kick off a merge. The tests could go in here, or in a follow up. Whatever you think makes more sense

joshuef (approved):

sn_node/src/quote.rs
maqi: yeah, I had the same feeling as well.
planned to raise a separate PR to refactor the API of `calculate_cost_for_records` as a whole,
so that can easily pass `quote` down.

sn_transfers/src/wallet/data_payments.rs
maqi: ok, will do

sn_transfers/src/wallet/data_payments.rs
maqi: there is a warning log at the place of received the failed verify,
which will log the entire quote for investigation.
don't want to abuse the log file duplicatedly, hence here just log simple for easier debugging.

sn_transfers/src/wallet/data_payments.rs
maqi: same reason as above.

sn_networking/src/cmd.rs
maqi: not much gain for this case actually

maqi:
> what triggers a quote verification?

client triggers it.
i.e. each time client collected quotes from close_nodes to a record, it sends all collected quotes to all close_nodes.
so that each nodes can carry out a quote verification against quotes from others.

PR #1532, opened 2024-03-28
ci: fix node manager binary packaging

jacderida:
In the packaging process, the crate name for the node manager was incorrectly renamed to `sn-node-manager`. Confusingly, this is the correct name for the crate, but the directory and crate name are mismatched. For clarity, I've changed the name of the variable to `crate_dir_name`.

Also, the `sn-node-manager` string was removed from the supported binary array, because that's the name of the crate, not the binary.

## Description

reviewpad:summary

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -162,31 +162,30 @@ package-release-assets bin version="":
 
   bin="{{bin}}"
 
-  supported_bins=("safe" "safenode" "safenode-manager" "sn-node-manager" "safenodemand" "faucet" "safenode_rpc_client")
-  crate=""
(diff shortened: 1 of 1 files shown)


JasonPaulGithub (approved):

JasonPaulGithub:
@jacderida My only question is, why is it called "justfile"?

jacderida:
> @jacderida My only question is, why is it called "justfile"?

https://github.com/casey/just

JasonPaulGithub:
> > @jacderida My only question is, why is it called "justfile"?
> 
> https://github.com/casey/just

Very cool. Thank you.

PR #1533, opened 2024-03-28
fix(client): perform re-attempts to clear pending transactions

RolandSherwin:
## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -704,14 +704,27 @@ impl WalletClient {
     ) -> WalletResult<(NanoTokens, NanoTokens)> {
         // Before wallet progress, there shall be no `unconfirmed_spend_requests`
         // Here, just re-upload again. The caller shall carry out a re-try later on.
-        if self.wallet.unconfirmed_spend_requests_exist() {
-            info!("Pre-Unconfirmed transactions exist. Resending in 1 second...");
+        let mut did_error = false;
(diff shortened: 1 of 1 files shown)


sn_client/src/wallet.rs
```diff
@@ -704,14 +704,27 @@ impl WalletClient {
...
+        let mut attempts = 0;
+        while self.wallet.unconfirmed_spend_requests_exist() {
+            info!("Pre-Unconfirmed transactions exist, sending again after 1 second...");
             sleep(Duration::from_secs(1)).await;
             self.resend_pending_transactions(verify_store).await;
 
-            return Err(WalletError::CouldNotSendMoney(
-                "Wallet has pre-unconfirmed transactions. Resend, and try again.".to_string(),
-            ));
+            if attempts > 10 {
```
JasonPaulGithub: Just a thought but I get the feeling we could be extracting important values like this into constants. Then we could move them out and label them so we are more aware of their importance.

https://doc.rust-lang.org/rust-by-example/custom_types/constants.html

JasonPaulGithub (approved):

sn_client/src/wallet.rs
RolandSherwin: Yeah, you're right. Also, this functionality is used a couple of times. Will just extract it out.

JasonPaulGithub: Nice one! 👍

PR #1534, opened 2024-03-28
fix(client): do not check quote expiry from the client

RolandSherwin:
## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/files.rs
+++ b/sn_client/src/files.rs
@@ -125,7 +125,7 @@ impl FilesApi {
         trace!("Client upload started for chunk: {chunk_addr:?}");
 
         let wallet_client = self.wallet()?;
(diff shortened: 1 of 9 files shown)


JasonPaulGithub (approved):
So we've removed tests and renamed a few things. Looks good.

joshuef:
What happens if we try and use an expired quote, now @RolandSherwin ? It'll still exist on disk, and so still fail?

RolandSherwin:
The new `sn_client::Uploader` makes a repayment incase a payment fails > threshold. This repayment is done to the cheapest peer that we have not already made a payment to. 
The new payment is appended to the `Vec<PaymentDetails>`, hence now a call to `get_recent_payment_details` would give this latest payment and we'll retry with this.

@joshuef

joshuef (approved):

PR #1535, opened 2024-03-28
docs: turn node manager readme into a guide

jacderida:
The README for the node manager was initially generated mostly by ChatGPT, just to flesh it out for first publish. It was based on the UI, so it quickly became out of date.

Some users in the most recent testnet wanted to refer to it to help them with the node manager, so it has now been turned into a guide.

## Description

reviewpad:summary


joshuef (approved):
lovely stuff!

PR #1536, opened 2024-03-29
[draft] feat(cli): remove cmds

joshuef:
BREAKING CHANGE: this removes the files upload command in order to focus uploads
via the  cmd, which manages the local account packet info etc.## Description

reviewpad:summary


joshuef:
lets see what ci tests break now...

PR #1540, opened 2024-04-01
Improve price setting tests, and cleanup [removes liveness in storecost]

joshuef:
atop #1514

### DIFF
--- a/sn_networking/src/driver.rs
+++ b/sn_networking/src/driver.rs
@@ -27,8 +27,9 @@ use crate::{
     replication_fetcher::ReplicationFetcher,
     transport, Network, CLOSE_GROUP_SIZE,
 };
-
 use futures::StreamExt;
+use itertools::Itertools;
+use libp2p::kad::KBucketDistance as Distance;
 #[cfg(feature = "local-discovery")]
 use libp2p::mdns;
 
@@ -62,6 +63,10 @@ use std::{
 use tokio::sync::{mpsc, oneshot};
 use tokio::time::Duration;
 use tracing::warn;
+
+/// Interval over which we check for the farthest record we _should_ be holding
+/// based upon our knowledge of the CLOSE_GROUP
+pub(crate) const CLOSET_RECORD_CHECK_INTERVAL: Duration = Duration::from_secs(60);
 
 /// The ways in which the Get Closest queries are used.
 pub(crate) enum PendingGetClosestType {
@@ -582,6 +587,8 @@ impl SwarmDriver {
     /// asynchronous tasks.
     pub async fn run(mut self) {
         let mut bootstrap_interval = interval(BOOTSTRAP_INTERVAL);
+        let mut set_farthest_record_interval = interval(CLOSET_RECORD_CHECK_INTERVAL);
+
         loop {
             tokio::select! {
                 swarm_event = self.swarm.select_next_some() => {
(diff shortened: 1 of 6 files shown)


sn_networking/src/record_store.rs
```diff
@@ -428,24 +428,35 @@ impl NodeRecordStore {
...
+        let record_keys_as_hashset: HashSet<&Key> = self.records.keys().collect();
+
+        let mut quoting_metrics = QuotingMetrics {
+            close_records_stored: records_stored,
             max_records: self.config.max_records,
             received_payment_count: self.received_payment_count,
             live_time: self.timestamp.elapsed().as_secs(),
         };
 
+        if let Some(distance_range) = self.distance_range {
```
maqi, line 441: you can move the if clause before the QuotingMetrics constructor,
so that there will be no quirement of re-assign of `quoting_metrics.close_records_stored` ?
i.e.
```rust
let close_records_stored = if let Some(distance_range) = self.distance_range { } else {} ;
let quoting_metrics = QuotingMetrics { };
```

sn_networking/src/record_store.rs
```diff
@@ -428,24 +428,35 @@ impl NodeRecordStore {
...
         let cost = if self.contains(key) {
             0
         } else {
+            // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
             calculate_cost_for_records(
-                quoting_metrics.records_stored,
+                quoting_metrics.close_records_stored,
                 quoting_metrics.received_payment_count,
                 quoting_metrics.max_records,
-                quoting_metrics.live_time,
```
maqi, line 445: I am OK with not to take the `live_time` into the calculate_cost equation,
but may prefer it still to be kept in the `QuotingMetrics`, as for the statistic collecting and/or trustness supporting purpose.

sn_networking/src/driver.rs
```diff
@@ -608,6 +615,21 @@ impl SwarmDriver {
...
                     }
                 }
+                _ = set_farthest_record_interval.tick() => {
+                    let closest_peers = self
+                    .swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .get_closest_local_peers(&self.self_peer_id.into())
+                    .map(|peer| peer.into_preimage())
+                    .take(CLOSE_GROUP_SIZE)
```
maqi, line 625: here passed CLOSE_GROUP_SIZE peers in,
but the `is_in_close_range` function requires the input list to be greater than REPLICATE_RANGE,
which actually will always return true ?

There is failing CI test, which could be related ?

sn_networking/src/driver.rs
```diff
@@ -616,6 +638,35 @@ impl SwarmDriver {
...
+        let locally_stored_keys = self
+            .swarm
+            .behaviour_mut()
+            .kademlia
+            .store_mut()
+            .record_addresses_ref();
+        let our_address = NetworkAddress::from_peer(self.self_peer_id);
+
+        for (_key, (address, _r_type)) in locally_stored_keys.iter() {
+            if Self::is_in_close_range(&self.self_peer_id, address, closest_k_peers) {
```
maqi, line 654: instead of calculate closeness against entire stored key_set, which is costive and perhaps un-necessary as well.
can we use a simpler calculation say:
the distance range shall just be the distance from us to the 20th closest to us ?
this may let us achieve the same purpose with certain level of margin to allow glitch ?

maqi (changes requested):

sn_networking/src/driver.rs
joshuef: I had wondered that, but then is the not _not_ the CLOSE_RANGE which is what's stored at the closest 5 nodes 🤔 ? 

(or whats the line between 5 and 6 and our resonsibility... I know what you mean, but it's tricky... I'm happy to look for a better way here though)

PR #1555, opened 2024-04-02
feat(network): provide network versioning

RolandSherwin:
## Description

Here is a PR to provide network versioning based on the `git branch` , i..e, if `stable branch => beta` and  `_ branch => alpha`
NOTE: This will be a breaking change and will not be compatible with the currently running network.

Sample from node:
```
[2024-04-02T16:48:38.124392Z INFO sn_networking::driver] Building request response with "/safe/alpha/0.14"
[2024-04-02T16:48:38.124678Z INFO sn_networking::driver] Building Identify with identify_protocol_str: "safe/alpha/0.14" and identify_version: "safe/node/alpha/0.14"
```
Sample from client:
```
[2024-04-02T16:49:22.437188Z INFO sn_networking::driver] Building request response with "/safe/alpha/0.14"
[2024-04-02T16:49:22.437874Z INFO sn_networking::driver] Building Identify with identify_protocol_str: "safe/alpha/0.14" and identify_version: "safe/client/alpha/0.14"
```


sn_networking/src/version.rs
```diff
@@ -0,0 +1,69 @@
...
+        version_str
+    }
+}
+
+// Return the testnet channel name from the git branch.
+// Stable branch is considered as the "beta" network.
+fn get_testnet_channel() -> &'static str {
+    match sn_build_info::git_branch() {
+        "stable" => "beta",
+        _ => "alpha",
```
joshuef: Alternative approach.

If the branch is `stable` or `beta`, do not apply anything.

Otherwise, apply the full branch name? (Would this mean local dev on a branch will be incompatible with a testnet? If so, we probably don't want this on _always_, and could only do so if there's an ENV_VAR: `RESTRICTED_NETWORK` or something?)

joshuef (changes requested):
as above

joshuef (approved):

PR #1557, opened 2024-04-03
feat: rename token to amount in Spend

grumbach:
BREAKING CHANGE: 
- field containing amount renamed to "amount" instead of confusing "token" in Spend
- renamed other fields in `CashNote` as well


joshuef:
lgtm. just got to think on when to get this in !

JasonPaulGithub:
How's this one going?

joshuef (approved):

PR #1563, opened 2024-04-04
feat(network): add tcp transport

RolandSherwin:
## Description

reviewpad:summary


sn_networking/src/transport/mod.rs
```diff
@@ -1,3 +1,11 @@
+// Copyright 2024 MaidSafe.net limited.
```
JasonPaulGithub, line 1: I wonder if we keep it as Maidsafe?

joshuef:
Closing as this work is going on elsewhere

PR #1572, opened 2024-04-05
fix(networking): increase the responsible range of nodes to K_VALUE/2…

joshuef:
… peers away

## Description

reviewpad:summary


grumbach (approved):
🚢

PR #1577, opened 2024-04-05
feat: report protocol mismatch error

RolandSherwin:
This would allow the clients to timeout with the proper error message.

### DIFF
--- a/sn_networking/src/event.rs
+++ b/sn_networking/src/event.rs
@@ -30,7 +30,7 @@ use libp2p::{
     Multiaddr, PeerId, TransportError,
 };
 use rand::{rngs::OsRng, Rng};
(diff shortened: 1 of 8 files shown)


sn_networking/src/event.rs
```diff
@@ -219,10 +226,20 @@ impl SwarmDriver {
...
+                                their_protocol: info.protocol_version,
+                            });
+
+                            return Ok(());
+                        }
+
                         let has_dialed = self.dialed_peers.contains(&peer_id);
-                        let peer_is_agent = info
-                            .agent_version
-                            .starts_with(&IDENTIFY_NODE_VERSION_STR.to_string());
```
RolandSherwin, line 225: We should match the whole string here. And we are checking the peer is a node, so renamed the variable.

JasonPaulGithub (approved):

PR #1580, opened 2024-04-08
feat(node): notify peer it is now considered as BAD

maqi:
## Description

reviewpad:summary


happybeing:
It would be really nice to have an INFO level message in the affected node's log so I can mark this in the `vdash` summary.

And if a node can come good again, any clues on how to 'notice' that in its log would also be useful (again at INFO level).

Thanks

maqi:
> It would be really nice to have an INFO level message in the affected node's log so I can mark this in the `vdash` summary.

There will be a `warning` level (higher than INFO) log, once this PR merged
```rust
warn!("Peer {detected_by:?} consider us as BAD, due to {bad_behaviour:?}.");
```
 
> And if a node can come good again, any clues on how to 'notice' that in its log would also be useful (again at INFO level).

So far, there is no plan to allow a bad node become good again.
Even it restart itself, it will still be considered as BAD, as long as using the same PeerId.

joshuef:
I think this is breaking as it's a protocol level change. Marking as DONOTMERGE for now

joshuef (approved):

Issue #1576, opened 2024-04-05
API for uploading a directory tree - some questions


happybeing:
I'm looking for an API for uploading a directory tree. The `sn_cli` `ChunkManager` seems to do everything I need, but once it completes I don't see a way for the app match the *path* of an uploaded file to the xor address of its datamap. The `ChunkManager` keeps track of the filenames and their xor addresses, but not the paths to the files.

I assume there isn't an alternative API that allows uploading of a directory tree, and tracks the paths.

So would you consider a PR to modify the `ChunkManager` to track the path of the file (and/or the name, since the name can be obtained from the path if that is retained)? In the short term it may be easier for me to build a separate crate with the API I need (and modified `ChunkManager`). So an alternative would be for me to do that and at some point the `sn_cli` could be modified to use that.

So I'm curious if these APIs might be better separated from the `sn_cli` crate which doesn't seem to be a good location for things used by other apps.

Any thoughts?

happybeing:
I have now integrated the `FilesUploader` into my app so it would be ideal if that were able to provide a list of full paths and xor addresses for all the files in the tree, after successfully completing the upload. I'm not sure what the best way to do that will be.

`ChunkManager::completed_files` holds the name and xor address of the files. So if that could be modified to hold the full path and xor address, and then a way for this to be exposed by the `FilesUploader` that would mean a user of the `FilesUploader` can obtain the list of uploaded files (full paths) and their xor addresses after calling `FilesUploader::start_upload()`.

At the moment `FilesUploader::start_upload()` returns a summary of the chunks, so perhaps that could be changed as I'm not sure if it is used or useful.

EDIT: I've been looking into how to do this and it seems very difficult to get information from the `ChunkManager` (such as `ChunkManager::completed_files` to the user of a `FileUploader`. The `UploadSummary` might be used instead, but it doesn't know about files and doesn't have access to the `ChunkManager` so I don't see a simple way to expose the list of paths (or even filenames) that have completed or not completed, to a user of these APIs.

If that's true it is a problem for any app, so perhaps I am not understanding how to use these APIs from an app.

Can someone please explain how an app should upload a tree of files using the APIs, and how it can determine which files (with full paths) have uploaded? I believe I can work around this but think it is a useful thing to understand.

RolandSherwin:
[...]

PR #1583, opened 2024-04-08
feat: Chunk Operator

JasonPaulGithub:
Fix for Issue : https://github.com/maidsafe/safe_network/issues/1487

I have created a Chunk Operator, aka **Chop**

- Chop will chunk the file and return the following fields:
```
    head_chunk_address: ChunkAddress,
    data_map: Chunk,
    file_size: u64,
    data_map_in_chunks: Vec<(XorName, PathBuf)>,
```

- Chop will then encrypt data based on the file size. As you will have three separate outcomes requiring the same generic operation, this is handled using [basic polymorphism](https://blog.logrocket.com/build-polymorphic-components-rust/#understanding-polymorphism-rust). 

- Chop will remove the need for the boolean flag 'data_map_in_chunks', and instead simply use either: build() or build_with_data_map_in_chunks()

- Chop code is encapsulated inside its own .rs file and has been removed from the FilesApi code.

- Chop makes use of the [builder](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) design pattern.


JasonPaulGithub:
I can look into why this fails the tests by myself, but help on this would speed things up for sure.

sn_client/Cargo.toml
```diff
@@ -11,10 +11,10 @@ repository = "https://github.com/maidsafe/safe_network"
 version = "0.105.2"
 
 [features]
-default=[]
-local-discovery=["sn_networking/local-discovery"]
+default = []
```
JasonPaulGithub, line 14: I don't know why it did that. but you can mark this as viewed.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+impl SmallFile {
+    fn new(buffer: Vec<u8>) -> Self {
+        Self { buffer }
+    }
+}
+
+trait ChunkBySize {
+    fn chunk_by_size(self) -> (Chunk, Vec<(XorName, PathBuf)>);
```
JasonPaulGithub, line 109: We use this trait to generify the chunk_by_size function.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,172 @@
...
+        )
+        .expect("assign_values: Can't encrypt"),
+    };
+
+    build.data_map_chunk = data_map_chunk;
+    build.chunk_vec = chunk_vec;
+    build
+}
+
+// #[test]
```
JasonPaulGithub, line 165: This comment code will be removed.

sn_cli/src/files/chunk_manager.rs
```diff
@@ -221,26 +219,26 @@ impl ChunkManager {
...
-                        };
-                        Ok((path_xor.clone(), chunked_file))
-                    }
-                    Err(err) => {
-                        println!("Failed to chunk file {path:?}/{path_xor:?} with err: {err:?}");
-                        error!("Failed to chunk file {path:?}/{path_xor:?} with err: {err:?}");
-                        Err(eyre!("Failed to chunk file {path:?}/{path_xor:?} with err: {err:?}"))
-                    }
-                }
+                let chop = match include_data_maps {
```
JasonPaulGithub, line 222: This should be the only real change in logic. Could this really break the tests?

sn_client/src/chunk_operator.rs
JasonPaulGithub: Though, I could write tests for this. Might be nice.

sn_cli/src/files/chunk_manager.rs
JasonPaulGithub: Yep. Forgot to assign to self in the builder. Woops.

JasonPaulGithub:
> Probably too late for this PR but if you are working more in this area, is there a chance you can add the file path (not just name as it is now) to the ChunkManager uploaded files tracking? More in this issue #1576
> 
> I want to use the ChunkManager and to be able to access the xor address and full path of each uploaded file, not just the filenames.

https://github.com/maidsafe/safe_network/pull/1528#issuecomment-2040610201

@RolandSherwin

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub chunk_vec: Vec<(XorName, PathBuf)>,
+}
+
+#[derive(Default, Clone)]
+pub struct BuildChunkOperator {
+    file_path: PathBuf,
+    chunk_dir: PathBuf,
+    file_size: u64,
+    data_map_chunk: Chunk,
+    chunk_vec: Vec<(XorName, PathBuf)>,
```
RolandSherwin, line 26: Why do we store the `file_size`, `data_map_chunk` and `chunk_vec` here? During `new()` they get default values and they're directly copied into `ChunkOperator` after chunking.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Tries to chunk the file, returning
+/// `(head_address, data_map_chunk, file_size, chunk_names)`
+/// and writes encrypted chunks to disk.
+#[derive(Debug, PartialEq)]
+pub struct ChunkOperator {
+    pub head_chunk_address: ChunkAddress,
+    pub data_map: Chunk,
+    pub file_size: u64,
+    pub chunk_vec: Vec<(XorName, PathBuf)>,
+}
```
RolandSherwin, line 18: Is this the result of the `BuildChunkOperator`? If so, can we state that in the docs? I think the docs is misleading. Maybe `ChunkOperatorResult` would work here? We use the convention in our codebase.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,160 @@
...
+        let mut output_file =
+            File::create(data_map_path.clone()).expect("error creating output file");
+
+        // debug!("include_data_map_in_chunks {include_data_map_in_chunks:?}");
+        info!("Data_map_chunk to be written!");
+        trace!("Data_map_chunk being written to {data_map_path:?}");
+
+        output_file
+            .write_all(&self.data_map_chunk.value)
+            .expect("error writing all");
```
RolandSherwin, line 55: We might have to return proper errors here. And cannot use `expect` which can panic.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    fn append_data_map(&mut self) -> Vec<(XorName, PathBuf)> {
+        let mut chunks_paths = self.chunk_vec.clone();
+        let data_map_path = self
+            .chunk_dir
+            .join(hex::encode(*self.data_map_chunk.name()));
+        let mut output_file =
+            File::create(data_map_path.clone()).expect("error creating output file");
```
RolandSherwin, line 47: We cannot use `expect` here as it can panic. We should reutrn proper error types.

sn_client/src/chunk_operator.rs
```diff
@@ -0,0 +1,160 @@
...
+        trace!("Data_map_chunk being written to {data_map_path:?}");
+
+        output_file
+            .write_all(&self.data_map_chunk.value)
+            .expect("error writing all");
+        chunks_paths.push((*self.data_map_chunk.name(), data_map_path));
+        chunks_paths
+    }
+
+    pub fn build(self, file_path: &Path, chunk_dir: &Path) -> ChunkOperator {
```
RolandSherwin, line 60: Looking at the code here, I think the `rust builder pattern` is not used here. We do not configure a lot of things and instead the 3 required args are passed via the `build` fn and with the use of two `build_*` fns. Why should we go with builder here instead of having something like `ChunkOperator` with `execute(file_path, chunk_dir)` / `execute_with_data_map_in_chunks(file_path, chunk_dir)`? I

PR #1586, opened 2024-04-09
feat(cli): return the files upload summary after a successful files upload

RolandSherwin:
In regard to https://github.com/maidsafe/safe_network/issues/1576

- Return more info after we upload files.
- Provide a trait to override printing things to stdout. Now that these structs are marked as public, not everyone would want to print to stdout by default.


bochaco (approved):

happybeing:
@RolandSherwin I think you missed a bit :smile:, fixed by https://github.com/maidsafe/safe_network/pull/1588

PR #1608, opened 2024-04-11
Feat: Rpc restart

JasonPaulGithub:
Add a restart command to the cli commands for the node manager allowing it to restart nodes.

` cargo run --bin safenode-manager --features local-discovery -- rpc restart --socket-address 127.0.0.1:59088`

- [x] Manual testing (local network).

### DIFF
--- a/sn_node_manager/src/bin/cli/main.rs
+++ b/sn_node_manager/src/bin/cli/main.rs
@@ -13,6 +13,7 @@ use sn_node_manager::{
     cmd, VerbosityLevel,
 };
 use sn_peers_acquisition::PeersArgs;
+use std::net::SocketAddr;
 use std::{net::Ipv4Addr, path::PathBuf};
 
 const DEFAULT_NODE_COUNT: u16 = 25;
@@ -164,6 +165,8 @@ pub enum SubCmd {
     Faucet(FaucetSubCmd),
     #[clap(subcommand)]
     Local(LocalSubCmd),
+    #[clap(name = "rpc", subcommand)]
+    Rpc(RpcSubCmd),
     /// Remove safenode service(s).
     ///
     /// If no peer ID(s) or service name(s) are supplied, all services will be removed.
@@ -367,7 +370,29 @@ pub enum DaemonSubCmd {
     Stop {},
 }
 
-/// Manage the faucet service.
+/// Manage RPC client commands
+#[derive(Subcommand, Debug)]
+pub enum RpcSubCmd {
+    #[clap(name = "restart")]
+    /// Issue an RPC command to restart nodes
+    Restart {
+        /// The peer ID of the service to restart.
+        ///
+        /// The argument can be used multiple times to restart many services.
+        #[clap(long)]
+        peer_id: Vec<String>,
+        /// The address of the RPC server where the command will be sent.
+        ///
+        /// It should be an IP address and port, e.g, 127.0.0.1:59091.
+        #[clap(long)]
+        rpc_server_address: SocketAddr,
+        /// Set to retain the peer ID of the nodes to be restarted.
+        #[clap(long)]
+        retain_peer_id: bool,
+    },
+}
+
+/// Manage faucet services.
 #[allow(clippy::large_enum_variant)]
 #[derive(Subcommand, Debug)]
 pub enum FaucetSubCmd {
@@ -777,6 +802,11 @@ async fn main() -> Result<()> {
             )
             .await
         }
+        SubCmd::Rpc(RpcSubCmd::Restart {
+            peer_id,
+            rpc_server_address: socket_addr,
+            retain_peer_id,
+        }) => sn_node_manager::rpc_client::restart_node(peer_id, socket_addr, retain_peer_id).await,
     }
 }
 
--- a/sn_node_manager/src/rpc_client.rs
+++ b/sn_node_manager/src/rpc_client.rs
@@ -1,0 +1,63 @@
+use color_eyre::eyre::bail;
+use color_eyre::{eyre::eyre, Result};
+use libp2p_identity::PeerId;
+use sn_service_management::safenode_manager_proto::safe_node_manager_client::SafeNodeManagerClient;
+use sn_service_management::safenode_manager_proto::NodeServiceRestartRequest;
+use std::net::SocketAddr;
+use std::str::FromStr;
+use std::time::Duration;
+use tonic::transport::Channel;
+use tonic::Request;
+
+struct DaemonRpcClient {
+    addr: SocketAddr,
+    rpc: SafeNodeManagerClient<Channel>,
+}
+
+pub async fn restart_node(
+    peer_ids: Vec<String>,
+    rpc_server_address: SocketAddr,
+    retain_peer_id: bool,
+) -> Result<()> {
+    for peer_id in peer_ids {
+        let str_bytes = PeerId::from_str(&peer_id)?.to_bytes();
+
+        let mut daemon_client = get_rpc_client(rpc_server_address).await?;
+
+        let _response = daemon_client
+            .rpc
+            .restart_node_service(Request::new(NodeServiceRestartRequest {
+                peer_id: str_bytes,
+                delay_millis: 0,
+                retain_peer_id,
+            }))
+            .await
+            .map_err(|err| {
+                eyre!(
+                    "Failed to restart node service with {peer_id:?} at {:?} with err: {err:?}",
+                    daemon_client.addr
+                )
+            })?;
+    }
+    Ok(())
+}
+
+async fn get_rpc_client(socket_addr: SocketAddr) -> Result<DaemonRpcClient> {
+    let endpoint = format!("https://{socket_addr}");
+    let mut attempts = 0;
+    loop {
+        if let Ok(rpc_client) = SafeNodeManagerClient::connect(endpoint.clone()).await {
+            let rpc_client = DaemonRpcClient {
+                addr: socket_addr,
+                rpc: rpc_client,
+            };
+            return Ok(rpc_client);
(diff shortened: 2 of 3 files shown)


sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+use tonic::transport::Channel;
+use tonic::Request;
+
+struct DaemonRpcClient {
+    addr: SocketAddr,
+    rpc: SafeNodeManagerClient<Channel>,
+}
+
+pub async fn restart_node(
+    peer_id_string_vec: Vec<String>,
```
jacderida: I would suggest naming this `peer_ids`, without the data type in the name.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -148,6 +149,9 @@ pub enum SubCmd {
     Daemon(DaemonSubCmd),
     #[clap(subcommand)]
     Faucet(FaucetSubCmd),
+    /// Display the list of Remote Procedure Call (RPC) commands, i.e. restart node
```
jacderida: I would suggest changing this to `Manage RPC client commands`, which is more consistent with the documentation for the other subcommands. (Though they also have some smaller issues with their documentation.)

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -449,6 +453,23 @@ pub enum DaemonSubCmd {
     Stop {},
 }
 
+#[derive(Subcommand, Debug)]
+pub enum RpcSubCmd {
+    // restart a node by providing the
```
jacderida: I would suggest changing this to `Issue an RPC command to restart nodes`.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -449,6 +453,23 @@ pub enum DaemonSubCmd {
...
+    #[clap(name = "restart")]
+    Restart {
+        /// The peer ID of the service to restart
```
jacderida: I would suggest changing this to keep it aligned with the other `--peer-id` arguments:
```
        /// The peer ID of the service to restart.
        ///
        /// The argument can be used multiple times to restart many services.
```

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -449,6 +453,23 @@ pub enum DaemonSubCmd {
...
+        #[clap(long)]
+        peer_id: Vec<String>,
+        /// RPC Server is listening address, i.e. --socket-address 127.0.0.1:59091
+        #[clap(long)]
+        socket_address: SocketAddr,
```
jacderida: I would suggest changing the name of this to `rpc_server_address`.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -449,6 +453,23 @@ pub enum DaemonSubCmd {
...
 
+#[derive(Subcommand, Debug)]
+pub enum RpcSubCmd {
+    // restart a node by providing the
+    #[clap(name = "restart")]
+    Restart {
+        /// The peer ID of the service to restart
+        #[clap(long)]
+        peer_id: Vec<String>,
+        /// RPC Server is listening address, i.e. --socket-address 127.0.0.1:59091
```
jacderida: I would suggest changing this to:
```
The address of the RPC server where the command will be sent.

It should be an IP address and port, e.g, 127.0.0.1:59091.
````

File: sn_node_manager/src/bin/cli/main.rs [hunk 7]
```diff
@@ -449,6 +453,23 @@ pub enum DaemonSubCmd {
...
+        #[clap(long)]
+        socket_address: SocketAddr,
+        /// Retain Peer Id
```
jacderida: I would suggest changing this to:
```
Set to retain the peer ID of the nodes to be restarted.
```

It would be appropriate to put a more elaborate explanation on that, as to what this entails exactly, but we're going to remove it anyway, so it can just have this simple text for now.

sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+    socket_addr: SocketAddr,
+    retain_peer_id: bool,
+) -> Result<()> {
+    for id_as_string in peer_id_string_vec {
```
jacderida: I would suggest naming this `peer_id`.

sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+    Ok(())
+}
+
+async fn rpc_restart(
+    id_as_string: String,
+    mut daemon_client: DaemonRpcClient,
+    retain_peer_id: bool,
+) -> Result<()> {
+    let str_bytes = PeerId::from_str(&id_as_string)?.to_bytes();
```
jacderida: I think Rust will allow for this to be named `peer_id`, which is useful because it avoids having to use data types for the names of converted variables.

sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+            get_rpc_client(socket_addr).await?,
+            retain_peer_id,
+        )
+        .await?;
+    }
+    Ok(())
+}
+
+async fn rpc_restart(
+    id_as_string: String,
```
jacderida: I would suggest naming this `peer_id`.

I would also suggest removing the function in favour of inlining the code above. I get you are probably adhering to the 'use small methods' principle. I also used to do that religiously, but I now favour just using inline code unless the code is actually shared between two or more callers.

sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+async fn get_rpc_client(socket_addr: SocketAddr) -> Result<DaemonRpcClient> {
+    let endpoint = format!("https://{socket_addr}");
+    let mut attempts = 0;
+    loop {
+        if let Ok(rpc_client) = SafeNodeManagerClient::connect(endpoint.clone()).await {
+            let rpc_client = DaemonRpcClient {
+                addr: socket_addr,
+                rpc: rpc_client,
+            };
+            break Ok(rpc_client);
```
jacderida: Could you just use `return` here?

sn_node_manager/src/rpc_client.rs
```diff
@@ -0,0 +1,76 @@
...
+use tonic::Request;
+
+struct DaemonRpcClient {
+    addr: SocketAddr,
+    rpc: SafeNodeManagerClient<Channel>,
+}
+
+pub async fn restart_node(
+    peer_id_string_vec: Vec<String>,
+    socket_addr: SocketAddr,
```
jacderida: I would suggest naming this `rpc_server_addr`.

jacderida (changes requested):
Could you also please squash the two commits into one?

sn_node_manager/src/rpc_client.rs
JasonPaulGithub: Solid Principles. One function should do one thing. It helps in many ways to do this. I do it where I can, but I am very much fighting a losing battle. However I can't give up on this as I believe it's a good thing to do. I will however inline this as this it's up to you.

jacderida: I don't really see how that principle would be violated by inlining this. But this one was only a suggestion. I do think it would be helpful to rename the variables though.

JasonPaulGithub: It's all gone through. No worries 👍

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -148,6 +149,9 @@ pub enum SubCmd {
     Daemon(DaemonSubCmd),
     #[clap(subcommand)]
     Faucet(FaucetSubCmd),
+    /// Manage RPC client commands
+    #[clap(name = "rpc", subcommand)]
+    Restart(RpcSubCmd),
```
jacderida: This should be named `Rpc` rather than `Restart`.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -449,6 +453,26 @@ pub enum DaemonSubCmd {
...
+    // Issue an RPC command to restart nodes
+    #[clap(name = "restart")]
+    Restart {
+        /// The peer ID of the service to restart.
+        ///
+        /// The argument can be used multiple times to restart many services.
+        #[clap(long)]
+        peer_id: Vec<String>,
+        /// The address of the RPC server where the command will be sent.
+        /// It should be an IP address and port, e.g, 127.0.0.1:59091.
```
jacderida, line 386: Sorry, it may seem petty, but, there should be a line break between these lines. It's because of the way it gets rendered on the CLI:
```
Options:
      --peer-id <PEER_ID>
          The peer ID of the service to restart.

          The argument can be used multiple times to restart many services.

      --rpc-server-address <RPC_SERVER_ADDRESS>
          The address of the RPC server where the command will be sent. It should be an IP address and port, e.g, 127.0.0.1:59091

      --retain-peer-id
          Set to retain the peer ID of the nodes to be restarted

  -h, --help
          Print help (see a summary with '-h')
```

sn_node_manager/src/bin/cli/main.rs
JasonPaulGithub: good shout

sn_node_manager/src/bin/cli/main.rs
JasonPaulGithub: Yeah was wondering that when the change was suggested but i wasn't sure. I'll send it through

jacderida (approved):

PR #1609, opened 2024-04-11
fix(node): record_store not store obsoleted record

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -79,6 +79,8 @@ pub struct NodeRecordStore {
     encryption_details: (Aes256GcmSiv, [u8; 4]),
     /// Time that this record_store got started
     timestamp: SystemTime,
+    /// List of records that got obsoleted, i.e. overwritten by a newer version
+    obsoleted_records: HashSet<(NetworkAddress, RecordType)>,
 }
 
 /// Configuration for a `DiskBackedRecordStore`.
@@ -262,6 +264,7 @@ impl NodeRecordStore {
             received_payment_count,
             encryption_details,
             timestamp,
+            obsoleted_records: Default::default(),
         };
 
         record_store.flush_historic_quoting_metrics();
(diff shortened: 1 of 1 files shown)


sn_networking/src/record_store.rs
```diff
@@ -79,6 +79,8 @@ pub struct NodeRecordStore {
     encryption_details: (Aes256GcmSiv, [u8; 4]),
     /// Time that this record_store got started
     timestamp: SystemTime,
+    /// List of records that got obsoleted, i.e. overwritten by a newer version
+    obsoleted_records: HashSet<(NetworkAddress, RecordType)>,
```
joshuef, line 83: Hmmm, I think this way memleaks lie... 

This is essentially about register versions for us (though I wonder if spends might suffer from similar issues?)


A potential strategy here is merging registers if we're getting different content in... But that feels like it should be left up to the client... 

So I find myself wondering what we actually want to be doing with registers here. We have different versions on different nodes. Presumably the client only updated one node (is that a bug for edits?)

Either way, we can end up in this situation through conn issues. 

Should nodes not just keep the verison they have until they receive an actual PUT? 

And if we do that, is all we need to be doing here NOT requesting updates if we have the key/type? (ie `Register`, ignoring the content hash), and NOT judge nodes as bad if they have differing content? (As that's a client world issue?)

maqi: > all we need to be doing here NOT requesting updates if we have the key/type?

If doing so, it means nodes will miss the newer version of the register.
As the new version will only be sent to one payee node (due to the store_cost to pay), and then be replicated to others from there.

Same reason for the following, that we cann't do that, as a client direct PUT is not to ALL nodes.

> Should nodes not just keep the verison they have until they receive an actual PUT?

maqi:
close this PR as there is Register version check & merge carried out at
https://github.com/maidsafe/safe_network/blob/main/sn_node/src/put_validation.rs#L289
within the `validate_and_store_register` function

PR #1611, opened 2024-04-12
Node Manager: Various Small Features

jacderida:
- efeb5da0 **feat: make `--peer` argument optional**

  Make the `--peer` argument on the node manager `add` command optional by handling the
  `PeersNotObtained` error. See the comment in the diff for more details.

- eea33e2f **feat: provide `balance` command**

  The node manager makes it easy to list the rewards balances for all the nodes it's managing. It was
  also requested that `status --details` show the balance.

  There was also a request to include the store cost for the node, but this information isn't as
  easily available. It may require an RPC extension to obtain it.

- bb0cc629 **refactor: provide `local` subcommand**

  A `local` subcommand is introduced for working with local networks.

  We observed a user make the mistake of using the `run` command to try to connect to a remote
  testnet, so it seems like a `local` subcommand would be useful for clarity.

  There are no functional changes here, it's just some reorganisation.

- 6b7ddfcc **chore: `remove` cmd operates over all services**

  Like the `start` and `stop` commands, the `remove` command is changed so that it can operate over
  all services if no arguments are provided.

- 1e705dd7 **feat: provide `reset` command**

  This command was requested in community feedback. It will remove all services, clearing out all logs
  and data, and also delete the node registry file, which will reset the service counter back to zero.

## Description

reviewpad:summary

### DIFF
--- a/sn_node_manager/src/bin/cli/main.rs
+++ b/sn_node_manager/src/bin/cli/main.rs
@@ -187,6 +187,18 @@ pub enum SubCmd {
         #[clap(long)]
         keep_directories: bool,
     },
+    /// Reset back to a clean base state.
+    ///
+    /// Stop and remove all services and delete the node registry, which will set the service
+    /// counter back to zero.
+    ///
+    /// This command must run as the root/administrative user.
+    #[clap(name = "reset")]
+    Reset {
+        /// Set to suppress the confirmation prompt.
+        #[clap(long, short)]
+        force: bool,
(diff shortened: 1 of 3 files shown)


JasonPaulGithub:
I'd like to help with this as I recognise one of the issues we're fixing here, but I don't know how many more are being covered.
Can you please reference the issue numbers in the desc so we know for sure which issues are being solved?
Thsnk

PR #1618, opened 2024-04-15
feat: spend dag double spend faults

grumbach:
## Description

- add double spend fork detection to DAG verification
- test for that scenario
- fix DAG edges issue caused by double spends
- now shows correct amounts in DAG edges

### DIFF
--- a/sn_client/src/audit/spend_dag.rs
+++ b/sn_client/src/audit/spend_dag.rs
@@ -8,6 +8,7 @@
 
 use petgraph::dot::Dot;
 use petgraph::graph::{DiGraph, NodeIndex};
+use petgraph::visit::EdgeRef;
 use serde::{Deserialize, Serialize};
 use sn_transfers::{is_genesis_spend, CashNoteRedemption, NanoTokens, SignedSpend, SpendAddress};
 use std::collections::{BTreeMap, BTreeSet};
@@ -130,7 +131,9 @@ impl SpendDag {
             Some(DagEntry::NotGatheredYet(idx)) => {
                 self.spends
                     .insert(spend_addr, DagEntry::Spend(Box::new(spend.clone()), idx));
-                NodeIndex::new(idx)
+                let node_idx = NodeIndex::new(idx);
+                self.reset_edges(node_idx);
+                node_idx
             }
             // or upgrade spend to double spend if it is different from the existing one
             Some(DagEntry::Spend(s, idx)) => {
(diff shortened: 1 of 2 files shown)


sn_client/src/audit/spend_dag.rs
```diff
@@ -318,6 +353,23 @@ impl SpendDag {
         Ok(royalties)
     }
 
+    /// Remove all edges from a Node in the DAG
+    fn reset_edges(&mut self, node: NodeIndex) {
```
JasonPaulGithub: Perhaps we can rename this function to 'remove_edges' to prevent confusion

sn_client/src/audit/tests/mod.rs
```diff
@@ -127,6 +127,107 @@ fn test_spend_dag_double_spend_poisonning() -> Result<()> {
...
+    assert_eq!(got, expected, "spend3 should be unspendable");
+    let s3a = spend3a.first().expect("spend3a to have an element");
+    let got = dag.get_spend_faults(s3a);
+    let expected = BTreeSet::from_iter([SpendFault::DoubleSpentAncestor {
+        addr: *s3a,
+        ancestor: *double_spent,
+    }]);
+    assert_eq!(got, expected, "spend3a should be unspendable");
+
+    // make sur all the descendants further down the branch are marked as bad as well
```
JasonPaulGithub: tiny spelling error here, but still ok to pass

JasonPaulGithub (approved):

sn_client/src/audit/spend_dag.rs
```diff
@@ -185,7 +188,7 @@ impl SpendDag {
         }
 
         // link to ancestors
-        let spend_amount = spend.token();
+        const PENDING_AMOUNT: NanoTokens = NanoTokens::from(0);
```
maqi, line 191: we normally put `const` at head of the file ?

sn_client/src/audit/spend_dag.rs
```diff
@@ -318,6 +353,23 @@ impl SpendDag {
...
+            .dag
+            .edges_directed(node, petgraph::Direction::Incoming)
+            .map(|e| e.id())
+            .collect();
+        let outgoing: Vec<_> = self
+            .dag
+            .edges_directed(node, petgraph::Direction::Outgoing)
+            .map(|e| e.id())
+            .collect();
+        for edge in incoming.into_iter().chain(outgoing.into_iter()) {
```
maqi, line 368: there is no incoming nor outgoing edges to the node, does it mean the node become `dangling` ?
then what does this mean to the DAG ?

grumbach: Indeed, but we add proper edges just after the call to this function, this is just a cleanup before write kind of thing

sn_client/src/audit/spend_dag.rs
grumbach: I was wondering about using const or just a regular variable here, since it's only useful in this function and for the sake or readability I'd like to keep it close to where it's used. Should I un-const that one?

PR #1620, opened 2024-04-16
test(networking): ensure pruned data is indeed further away than kept

joshuef:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -392,14 +392,12 @@ impl NodeRecordStore {
             }
         }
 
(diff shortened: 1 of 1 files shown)


sn_networking/src/record_store.rs
```diff
@@ -359,12 +359,13 @@ impl NodeRecordStore {
 
     /// Prune the records in the store to ensure that we free up space
     /// for the incoming record.
-    fn prune_storage_if_needed_for_record(&mut self) {
+    /// Returns true if the record can be stored.
```
maqi: returns OK ?

maqi (approved):

PR #1624, opened 2024-04-16
feat(node): restrict replication fetch range when node is full

maqi:
## Description

When node is full, only fetch new copies that is closer than the current farthest record.

reviewpad:summary


JasonPaulGithub (commented):
the code looks fine but what's going on here may need another pair of eyes

joshuef (approved):

PR #1631, opened 2024-04-17
Refactor Node Service

JasonPaulGithub:
Refactor node service.


JasonPaulGithub:
taking different approach

PR #1634, opened 2024-04-18
chore(node): optimise record_store farthest record calculation

maqi:
## Description

reviewpad:summary


sn_networking/src/record_store.rs
```diff
@@ -1111,7 +1140,7 @@ mod tests {
 
             if !succeeded {
```
JasonPaulGithub, line 1141: if succeeded { ... } else { ... }
Would be better

JasonPaulGithub (approved):

PR #1635, opened 2024-04-22
Add mnemonic generation for new wallets

joshuef:
## Description

reviewpad:summary


joshuef:
Rebased #1438 atop main and then linked it in to wallet generation w/ optional passphrase. 

I think this gets us going and we can improve things atop this.

sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -63,11 +63,15 @@ impl HotWallet {
     /// reloads the wallet from disk.
     fn reload(&mut self) -> Result<()> {
         // placeholder random MainSecretKey to take it out
```
b-zee: Think this comment isn't relevant anymore.

joshuef:
This is not working on initial testnets, so some debugging to do here

sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
```diff
@@ -23,15 +24,20 @@ use sn_client::transfers::{
...
 
 // Please do not remove the blank lines in these doc comments.
 // They are used for inserting line breaks when the help menu is rendered in the UI.
 #[derive(Parser, Debug)]
 pub enum WalletCmds {
     /// Print the wallet address.
-    Address,
+    Address {
+        /// Optional passphrase to use for the wallet deriviation if generating a new mnemonic.
+        passphrase: Option<String>,
```
maqi, line 42: out of scope of this PR
a user chosenn `passphrase` means a high chance of `address` colliction.
may worth to keep an eye on it in the following work.

or change the name of `Address` to something else ?

or if we do need something like `Address`, it shall be somehow `account_root_xorname` ?

sn_client/src/acc_packet/user_secret.rs
```diff
@@ -0,0 +1,75 @@
...
+    let derived_key = eip2333::derive_child_sk(root_sk, ACCOUNT_WALLET_DERIVATION);
+    let key_bytes = derived_key.serialize();
+    let sk = SecretKey::from_bytes(key_bytes.into()).map_err(|_err| Error::InvalidKeyBytes)?;
+    Ok(MainSecretKey::new(sk))
+}
+
+#[allow(dead_code)] // as yet unused, will be used soon
+/// Derive an xorname from the mnemonic for the account to store data.
+pub(crate) fn account_root_xorname(mnemonic: bip39::Mnemonic, passphrase: &str) -> Result<XorName> {
+    let seed = mnemonic.to_seed(passphrase);
```
maqi, line 52: shall put the following block of code into a separate private helper function, then be used by both `account_wallet_secret_key` and `account_root_xorname` functions:
```rust
fn derived_key_bytes(mnemonic, passphrase) -> {
    let seed = mnemonic.to_seed(passphrase);
    let root_sk =
        eip2333::derive_master_sk(&seed).map_err(|_err| Error::InvalidMnemonicSeedPhrase)?;
    let derived_key = eip2333::derive_child_sk(root_sk, ACCOUNT_WALLET_DERIVATION);
    derived_key.serialize()
}
```

sn_client/src/event.rs
```diff
@@ -29,9 +29,10 @@ impl ClientEventsBroadcaster {
     // Broadcast a new event, meant to be a helper only used by the client's internals.
     pub(crate) fn broadcast(&self, event: ClientEvent) {
         if let Err(err) = self.0.send(event) {
-            trace!(
-                "Could not broadcast ClientEvent as we don't have any active listeners: {err:?}"
-            );
+            if self.0.receiver_count() == 0 {
+                return;
```
maqi, line 33: may still need to at least log something ?
otherwise this `failed to sent event` will looks like got sent out correctly ?

sn_faucet/src/faucet_server.rs
```diff
@@ -125,6 +130,18 @@ async fn respond_to_donate_request(
     transfer_str: String,
     semaphore: Arc<Semaphore>,
 ) -> std::result::Result<impl Reply, std::convert::Infallible> {
+    let faucet_root = get_faucet_data_dir();
+
+    let mut wallet = match load_account_wallet_or_create_with_mnemonic(&faucet_root, None) {
```
maqi, line 147: the wallet is now loaded BEFORE the rate limit ?
is it ok?

sn_faucet/src/faucet_server.rs
```diff
@@ -138,16 +155,12 @@ async fn respond_to_donate_request(
...
-    let mut wallet = match load_faucet_wallet_from_genesis_wallet(&client).await {
-        Ok(w) => w,
-        Err(err) => {
-            eprintln!("Failed to load faucet wallet: {err}");
-            error!("Failed to load faucet wallet: {err}");
-            let mut response = Response::new(format!("Failed to load faucet wallet: {err}"));
-            *response.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
-            return Ok(response);
-        }
+    if let Err(err) = fund_faucet_from_genesis_wallet(&client, &mut wallet).await {
```
maqi, line 158: this means funding the wallet from genesis each time received a donote request ?
isn't it shall only got funded once when wallet first got created ?

sn_faucet/src/faucet_server.rs
```diff
@@ -234,6 +260,9 @@ async fn startup_server(client: Client) -> Result<()> {
     // Create a semaphore with a single permit
     let semaphore = Arc::new(Semaphore::new(1));
 
+    let faucet_root = get_faucet_data_dir();
+    let _faucet_wallet = load_account_wallet_or_create_with_mnemonic(&faucet_root, None)?;
```
maqi: unused ? can be removed now?

sn_faucet/src/faucet_server.rs
```diff
@@ -203,6 +216,19 @@ async fn respond_to_gift_request(
     key: String,
     semaphore: Arc<Semaphore>,
 ) -> std::result::Result<impl Reply, std::convert::Infallible> {
+    let faucet_root = get_faucet_data_dir();
```
maqi, line 219: maybe this block of `loading wallet` code can be a separate private helper function?
seems appeared duplicatedly couple of times

sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -591,15 +595,31 @@ impl HotWallet {
...
-            Some(key) => key,
-            None => {
-                let key = main_key.unwrap_or(MainSecretKey::random());
-                store_new_keypair(wallet_dir, &key)?;
-                warn!("No main key found when loading wallet from path, generating a new one with pubkey: {:?}", key.main_pubkey());
+        let key = match get_main_key_from_disk(wallet_dir) {
+            Ok(key) => {
+                if let Some(passed_key) = main_key {
+                    if key.secret_key() != passed_key.secret_key() {
+                        warn!("main_key passed to load_from_path_and_key, but a key was found in the wallet dir. Using the one found in the wallet dir.");
```
maqi, line 603: sounds bit risky.
shall we raise an error and terminate the work instead ?
if user do forgot main_key, he can still continue by passing down a None ?

sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
b-zee: @maqi The passphrase is used to protect the mnemonic, it's not the source of the entropy for the mnemonic generation. The mnemonic+passphrase will be the seed.

https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed

Maybe rewrite the comment a little bit.

sn_client/src/acc_packet/user_secret.rs
joshuef: honestly not sure on this, purely because the return type is... large.

sn_client/src/event.rs
joshuef: the point there was to avoid logging something. Client logs are polluted by this when there's no listener set at all. 

If we have no receivers, there's no error. If we do, then we error?

sn_faucet/src/faucet_server.rs
joshuef: no, will update!

sn_faucet/src/faucet_server.rs
joshuef: This is effectively how it is on main is my understanding,

sn_transfers/src/wallet/hot_wallet.rs
joshuef: I was debating this, I think it's a follow up. Changing the internal load fn to not do this and error is like 80% of this PR, so not super keen to start in on more in this :D

sn_faucet/src/faucet_server.rs
joshuef: it's only twice in that file, not sure it's necessary atm.

sn_faucet/src/faucet_server.rs
maqi: Ah, right, checked the code, it won't do anything if faucet_wallet is having non-zero balance.

maqi (approved):

PR #1650, opened 2024-04-25
feat(node): make spend and cash_note reason field configurable

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -158,16 +158,9 @@ pub struct Network {
     pub peer_id: Arc<PeerId>,
     pub root_dir_path: Arc<PathBuf>,
     keypair: Arc<Keypair>,
-    /// node owner's discord username, in readable format
-    owner: String,
 }
 
 impl Network {
-    /// Returns of the discord user_name of the node's owner
-    pub fn owner(&self) -> String {
-        self.owner.clone()
-    }
-
     /// Signs the given data with the node's keypair.
     pub fn sign(&self, msg: &[u8]) -> Result<Vec<u8>> {
         self.keypair.sign(msg).map_err(NetworkError::from)
@@ -1053,13 +1046,9 @@ mod tests {
 
     #[test]
     fn test_network_sign_verify() -> eyre::Result<()> {
-        let (network, _, _) = NetworkBuilder::new(
-            Keypair::generate_ed25519(),
-            false,
-            std::env::temp_dir(),
-            "maidsafe_test".to_string(),
-        )
-        .build_client()?;
+        let (network, _, _) =
+            NetworkBuilder::new(Keypair::generate_ed25519(), false, std::env::temp_dir())
+                .build_client()?;
         let msg = b"test message";
         let sig = network.sign(msg)?;
         assert!(network.verify(msg, &sig));
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -158,9 +158,16 @@ pub struct Network {
     pub peer_id: Arc<PeerId>,
     pub root_dir_path: Arc<PathBuf>,
     keypair: Arc<Keypair>,
+    /// node owner's discord username, in readable format
+    owner: String,
 }
 
 impl Network {
+    /// Returns of the discord user_name of the node's owner
+    pub fn owner(&self) -> String {
+        self.owner.clone()
+    }
+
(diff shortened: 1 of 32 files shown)


sn_node_manager/src/bin/cli/main.rs
```diff
@@ -547,6 +547,9 @@ pub enum LocalSubCmd {
         /// Set to skip the network validation process
         #[clap(long)]
         skip_validation: bool,
+        /// Specify the owner(readable discord user name).
+        #[clap(long)]
+        owner: Option<String>,
```
RolandSherwin, line 552: This is just for a local network and I assue we might need to pass in the `onwer` inside the `Add` arg to enable this for service nodes.

sn_node_manager/src/local.rs
```diff
@@ -286,8 +296,31 @@ pub async fn run_node(
     launcher: &dyn Launcher,
     rpc_client: &dyn RpcActions,
 ) -> Result<NodeServiceData> {
+    let user = match get_username() {
```
RolandSherwin, line 299: `get_username()` will at most cases return our OS's user account name. We can just use that. I'm not sure if we might ever have a failure here. Also since this is just for a local network, should we want to pass in the `owner` via CLI? It is just for testing right?

RolandSherwin: > I'm not sure if we might ever have a failure here

If we do have a failure, we can use a default value like "owner" since it is a local network.

sn_networking/src/lib.rs
```diff
@@ -158,9 +158,16 @@ pub struct Network {
     pub peer_id: Arc<PeerId>,
     pub root_dir_path: Arc<PathBuf>,
     keypair: Arc<Keypair>,
+    /// node owner's discord username, in readable format
+    owner: String,
```
RolandSherwin: Can we store this `owner` field inside `struct Node`? The `Network` usually gets cloned a lot and we even avoid having `Arc<String>` here.

RolandSherwin: I followed through the code and it seems like we only use this value during `quote` generation by the node. So `struct Node` would be a good fit for this. This would also avoid having the `onwer` values for `Client`

sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -429,7 +443,7 @@ impl HotWallet {
             start.elapsed()
         );
         debug!("Available CashNotes: {:#?}", available_cash_notes);
-        let reason_hash = Default::default();
+        let reason_hash = Hash::hash(b"SPEND_REASON_FOR_STORAGE");
```
RolandSherwin: Can this be renamed into `input_reason_hash` along with the arguments to `OfflineTransfer::new()` to better distinguish between the "input reason" and "output purpose".

Also if possible can we change `reason -> purpose` if they're for Outputs.

sn_transfers/src/transfers/transfer.rs
```diff
@@ -138,14 +138,21 @@ pub struct CashNoteRedemption {
     /// spentbook entry of one of one of the inputs (parent spends)
     /// using data found at this address the owner can check that the output is valid money
     pub parent_spend: SpendAddress,
+    /// For what purpose this cash_note was created
+    pub reason: String,
```
RolandSherwin: This can be removed on a followup PR as discussed.

sn_transfers/src/cashnotes/cashnote.rs
```diff
@@ -66,6 +69,9 @@ pub struct CashNote {
     pub parent_tx: Transaction,
     /// The transaction's input's SignedSpends
     pub parent_spends: BTreeSet<SignedSpend>,
+    /// The reason this cash_note created for
+    /// eg. `store cost pay to...`, `network royalty`, `change to self`, `payment transfer`, ...
+    pub reason: String,
```
RolandSherwin: This can be removed on a followup PR as discussed.

sn_node_manager/src/bin/cli/main.rs
maqi: as discussed, this can be addressed in follow up PR

sn_node_manager/src/local.rs
maqi: as discussed, this can be addressed in follow up PR

RolandSherwin (approved):

PR #1668, opened 2024-05-06
feat(launchpad): input to get and store discord user names

RolandSherwin:
## Description

reviewpad:summary

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4629,6 +4629,7 @@ version = "0.1.1"
 dependencies = [
  "atty",
  "better-panic",
+ "chrono",
  "clap",
  "color-eyre",
  "config",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6660,9 +6660,9 @@ dependencies = [
 
 [[package]]
 name = "service-manager"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ad5944e40978bf513d1402fe23e18743fe604be9e5394d8589ee818ad45f43c"
+checksum = "50c16631fd44f04d00709906a45d999441dc09576128401d63fe250543c48915"
 dependencies = [
  "cfg-if",
  "dirs",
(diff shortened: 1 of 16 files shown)


sn_node_launchpad/src/utils.rs
```diff
@@ -172,17 +144,14 @@ macro_rules! trace_dbg {
 pub fn version() -> String {
     let author = clap::crate_authors!();
 
-    // let current_exe_path = PathBuf::from(clap::crate_name!()).display().to_string();
-    let config_dir_path = get_config_dir().display().to_string();
-    let data_dir_path = get_data_dir().display().to_string();
+    let data_dir_path = get_launchpad_data_dir_path().unwrap().display().to_string();
```
joshuef: no unrwaps please!

joshuef (changes requested):
We also need to `clear` the input text area from other input before rendering, see: https://ratatui.rs/how-to/render/overwrite-regions/

sn_node_launchpad/src/utils.rs
RolandSherwin: Hey Josh, the unwrap has to be either used here or inside `#[command(author, version = version(), about)]`. There is no other option, I guess.

RolandSherwin: Regarding the `Clear` widget, It is performed before rendering the popup `f.render_widget(Clear, layer_zero);`

What issues are we seeing there exactly? There are a couple issues when the screen size becomes too small, but that should not be related to this.

joshuef: I saw border lines running through the input field.

RolandSherwin: Does this happen every single time?
I feel like it might be because something else used the stdout and printed stuff there.

joshuef: ah no, looks better this time round. Also i see the unrwaps are from the code I ported in! 🤦 . Sorry about that. I'll try and clear them out

joshuef (approved):

PR #1678, opened 2024-05-07
feat(nat): add binary to probe for NAT status

b-zee:
This binary will help a machine figure out its NAT status. For this to
work a public machine runs the binary, while a peer asks the public
machine for a dial-back.


sn_nat_detection/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+
+mod behaviour;
+
+const CONFIDENCE_MAX: usize = 2;
+const RETRY_INTERVAL: Duration = Duration::from_secs(10);
+
+/// A tool to detect NAT status of the machine. It can be run in server mode or client mode.
+/// The program will exit with an error code if NAT status is determined to be private.
+#[derive(Debug, Parser)]
+#[clap(name = "libp2p autonat")]
```
RolandSherwin: Just a nit pick, but name can be generic? say "nat detection" or something?

sn_nat_detection/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+    port: u16,
+
+    /// Servers to send dial-back requests to, in a 'multiaddr' format.
+    ///
+    /// A multiaddr looks like `/ip4/1.2.3.4/tcp/1200/tcp` where `1.2.3.4` is the IP and `1200` is the port.
+    /// Alternatively, the address can be written as `1.2.3.4:1200`.
+    ///
+    /// This argument can be provided multiple times to connect to multiple peers.
+    #[clap(name = "SERVER", value_name = "multiaddr", value_delimiter = ',', value_parser = parse_peer_addr)]
+    server_addr: Vec<Multiaddr>,
```
RolandSherwin, line 52: Can we also specify that if we do not provide this argument, we are acting as a server?

sn_nat_detection/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+    /// Disable use of UPnP to open a port on the router, before detecting NAT status.
+    #[clap(long, short, default_value_t = false)]
+    no_upnp: bool,
+
+    #[command(flatten)]
+    verbose: clap_verbosity_flag::Verbosity,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
RolandSherwin: Can we also use `color_eyre` or `eyre` for errors? That is the one that we use throughout our workplace.

sn_nat_detection/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+    // Use `RUST_LOG` if set, else use the verbosity flag (where `-vvvv` is trace level).
+    let _ = if std::env::var_os("RUST_LOG").is_some() {
+        registry.with(EnvFilter::from_env("RUST_LOG")).try_init()
+    } else {
+        let filter = tracing_subscriber::filter::Targets::new().with_target(
+            env!("CARGO_BIN_NAME").replace('-', "_"),
+            opt.verbose.log_level_filter().as_trace(),
+        );
+        registry.with(filter).try_init()
+    };
```
RolandSherwin, line 79: Maybe in a follow up PR can we make use of `sn_logging` here? I think it will be useful to have the logs inside a single place.

sn_nat_detection/src/main.rs
```diff
@@ -0,0 +1,400 @@
...
+                    self.client_state = Some(State::Probing(0));
+                } else {
+                    self.client_state = Some(State::Dialing);
+                }
+            }
+            State::Probing(old_confidence) => {
+                let confidence = self.swarm.behaviour().autonat.confidence();
+                let status = self.swarm.behaviour().autonat.nat_status();
+
+                if confidence == CONFIDENCE_MAX {
```
RolandSherwin: `confidence` is obtained from autnoat behaviour, can it skip a value here?, if so we can be stuck in a loop? Can we make `confidence >= CONFIDENCE_MAX` just to be sure

sn_nat_detection/src/behaviour/identify.rs
```diff
@@ -0,0 +1,57 @@
...
+                // Disconnect if peer has no AutoNAT support.
+                if !info
+                    .protocols
+                    .iter()
+                    .any(|p| *p == autonat::DEFAULT_PROTOCOL_NAME)
+                {
+                    warn!(%peer_id, "Peer does not support AutoNAT. Disconnecting from peer.");
+                    let _ = self.swarm.disconnect_peer_id(peer_id);
+                    #[allow(clippy::needless_return)]
+                    return;
```
RolandSherwin, line 34: This `allow` can be removed!

RolandSherwin (approved):
:rocket:

sn_nat_detection/src/main.rs
b-zee: Will do, though that would indicate a bug. The max value is the one we set in the AutoNAT config, which is set to `CONFIDENCE_MAX`.

sn_nat_detection/src/main.rs
b-zee: Yeah, kind of agree, though I think it's quite compact code here that isn't too complex. I didn't think it would be useful to log to files etc.

PR #1681, opened 2024-05-08
fix(launchpad): prevent terminal launch loop

RolandSherwin:
## Description

reviewpad:summary


joshuef:
Closing as #1682 does the same, but also checks if running in terminal or no.

PR #1684, opened 2024-05-08
feat(faucet): add feat for gifting-from-genesis

joshuef:
So we no longer gift the entire genesis amount to the faucet
(which we can fund in other means).

but we retain a feat for this for testing etc## Description

reviewpad:summary

### DIFF
--- a/sn_client/src/faucet.rs
+++ b/sn_client/src/faucet.rs
@@ -28,22 +28,41 @@ pub async fn fund_faucet_from_genesis_wallet(
     println!("Loading genesis...");
     debug!("Loading genesis...");
     let genesis_wallet = load_genesis_wallet()?;
+    let genesis_balance = genesis_wallet.balance();
 
-    let faucet_balance = genesis_wallet.balance();
-    println!("Sending {faucet_balance} from genesis to faucet wallet..");
-    debug!("Sending {faucet_balance} from genesis to faucet wallet..");
-    let cash_note = send(
-        genesis_wallet,
-        faucet_balance,
-        faucet_wallet.address(),
-        client,
-        true,
-    )
-    .await?;
+    let cash_note = if cfg!(feature = "gifting-from-genesis") {
+        println!("Sending {genesis_balance} from genesis to faucet wallet..");
+        debug!("Sending {genesis_balance} from genesis to faucet wallet..");
+        let cash_note = send(
+            genesis_wallet,
+            genesis_balance,
+            faucet_wallet.address(),
+            client,
+            true,
+        )
+        .await?;
 
-    faucet_wallet
-        .deposit_and_store_to_disk(&vec![cash_note.clone()])
-        .expect("Faucet wallet shall be stored successfully.");
+        faucet_wallet
+            .deposit_and_store_to_disk(&vec![cash_note.clone()])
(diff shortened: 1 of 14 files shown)


sn_client/src/faucet.rs
```diff
@@ -6,50 +6,87 @@
...
-    println!("Loading genesis...");
-    debug!("Loading genesis...");
+    info!("funding faucet from genesis...");
+
+    println!("Initiating genesis...");
+    debug!("Initiating genesis...");
     let genesis_wallet = load_genesis_wallet()?;
+    let genesis_balance = genesis_wallet.balance();
+
+    let foundation_key = MainPubkey::from_hex("a4bd3f928c585a63ab6070337316c1832bffd92be8efe9b235ec1c631f03b4bb91e29bbad34994ddf9f77d9858ddb0bb")?; // DevSkim: ignore DS117838
```
maqi: better make the hex string as a const ?
like what we do for the `GENESIS_CASHNOTE_SK` ?

btw. is this or the genesis, or somewhere else to be the receiver when node forward received payment to during Beta ?
I think we'd better make it as a `cli input opt` instead of `hard code`

sn_client/src/faucet.rs
```diff
@@ -6,50 +6,87 @@
...
+
+    let (foundation_cashnote, faucet_cashnote) = {
+        println!("Sending {INITIAL_FAUCET_BALANCE}  from genesis to faucet wallet..");
+        debug!("Sending {INITIAL_FAUCET_BALANCE}  from genesis to faucet wallet..");
+
+        println!("Faucet wallet balance: {}", faucet_wallet.balance());
```
maqi, line 51: there is check of `if !faucet_balance.is_zero()` above
so here the balance shall alwasy being `0`
hence no need to have a log & printout ?

sn_client/src/faucet.rs
```diff
@@ -6,50 +6,87 @@
...
-        .deposit_and_store_to_disk(&vec![cash_note.clone()])
-        .expect("Faucet wallet shall be stored successfully.");
     println!("Faucet wallet balance: {}", faucet_wallet.balance());
     debug!("Faucet wallet balance: {}", faucet_wallet.balance());
 
     println!("Verifying the transfer from genesis...");
     debug!("Verifying the transfer from genesis...");
-    if let Err(error) = client.verify_cashnote(&cash_note).await {
+    if let Err(error) = client.verify_cashnote(&foundation_cashnote).await {
+        error!("Could not verify the transfer from genesis: {error}. Panicking.");
```
maqi: better make the log msg contains `from genesis to foundation` or `from genesis to faucet` correspondently.
same comments for the following logs & printouts

sn_client/src/faucet.rs
joshuef: So right now it's a pregenerated secret key for the foundation, being kept out of _things_. 

I'll udpate the hex as a const there and document that 👍

sn_client/src/faucet.rs
joshuef: It's informative as this coudl be called either empty or on existing wallet, in the latter case we'd want to know if it was somehow hitting these points even if we thought we had money (i literally just had this issue)

maqi (approved):

PR #1690, opened 2024-05-08
feat: spend reason enum and sized cipher

grumbach:
## Description

reviewpad:summary

### DIFF
--- a/sn_transfers/src/cashnotes/cashnote.rs
+++ b/sn_transfers/src/cashnotes/cashnote.rs
@@ -116,16 +116,6 @@ impl CashNote {
         self.derivation_index
     }
 
-    /// Return the reason why this CashNote was spent.
-    /// Will be the default Hash (empty) if reason is none.
-    pub fn spent_reason(&self) -> Hash {
-        self.parent_spends
-            .iter()
-            .next()
-            .map(|s| s.reason())
-            .unwrap_or_default()
-    }
-
     /// Return the purpose why this CashNote was created.
     pub fn purpose(&self) -> String {
         self.purpose.clone()
@@ -155,7 +145,6 @@ impl CashNote {
 
         sha3.update(&self.purpose.clone().into_bytes());
 
-        sha3.update(self.spent_reason().as_ref());
         let mut hash = [0u8; 32];
         sha3.finalize(&mut hash);
         Hash::from(hash)
@@ -186,13 +175,6 @@ impl CashNote {
             .any(|o| unique_pubkey.eq(o.unique_pubkey()))
         {
             return Err(TransferError::CashNoteCiphersNotPresentInTransactionOutput);
-        }
-
-        // verify that all signed_spend reasons are equal
-        let spent_reason = self.spent_reason();
-        let reasons_are_equal = |s: &SignedSpend| spent_reason == s.reason();
-        if !self.parent_spends.iter().all(reasons_are_equal) {
-            return Err(TransferError::SignedSpendReasonMismatch(unique_pubkey));
         }
         Ok(())
     }
--- a/sn_transfers/src/cashnotes/spend_reason.rs
+++ b/sn_transfers/src/cashnotes/spend_reason.rs
@@ -13,6 +13,8 @@ use xor_name::XorName;
 
 use crate::{DerivationIndex, Hash, Result, TransferError};
 
+const CUSTOM_SPEND_REASON_SIZE: usize = 64;
+
 /// The attached metadata or reason for which a Spend was spent
 #[derive(Default, Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
 pub enum SpendReason {
@@ -21,7 +23,7 @@ pub enum SpendReason {
     /// Reference to network data
     NetworkData(XorName),
     /// Custom field for any application data
-    Custom(#[serde(with = "serde_bytes")] [u8; 64]),
+    Custom(#[serde(with = "serde_bytes")] [u8; CUSTOM_SPEND_REASON_SIZE]),
 
     /// Beta only feature to track rewards
     /// Discord username encrypted to the Foundation's pubkey with a random nonce
@@ -43,6 +45,9 @@ lazy_static! {
     pub static ref FOUNDATION_PK: PublicKey = crate::NETWORK_ROYALTIES_PK.public_key();
 }
 const MAX_CIPHER_SIZE: usize = std::u8::MAX as usize;
+const DERIVATION_INDEX_SIZE: usize = 32;
+const HASH_SIZE: usize = 32;
+const LIMIT_SIZE: usize = HASH_SIZE + DERIVATION_INDEX_SIZE;
 
 /// Discord username encrypted to the Foundation's pubkey with a random nonce
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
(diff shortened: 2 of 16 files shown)


sn_transfers/src/cashnotes/cashnote.rs
```diff
@@ -116,16 +116,6 @@ impl CashNote {
         self.derivation_index
     }
 
-    /// Return the reason why this CashNote was spent.
-    /// Will be the default Hash (empty) if reason is none.
-    pub fn spent_reason(&self) -> Hash {
```
grumbach, line 121: Reasons now depend on `Spend` so this doesn't make sense anymore as CNs have multiple parent spends

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// Discord username encrypted to the Foundation's pubkey with a random nonce
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct DiscordNameCipher {
+    /// Length of the cipher, hard limited to MAX_U8
+    len: u8,
+    /// Encrypted Discord username
+    #[serde(with = "serde_bytes")]
+    cipher: [u8; MAX_CIPHER_SIZE],
+}
```
grumbach: Type that defines compile-time HARD LIMIT to a discord name cipher.

sn_transfers/src/cashnotes/cashnote.rs
```diff
@@ -187,13 +176,6 @@ impl CashNote {
         {
             return Err(TransferError::CashNoteCiphersNotPresentInTransactionOutput);
         }
-
-        // verify that all signed_spend reasons are equal
-        let spent_reason = self.spent_reason();
```
joshuef, line 192: Do we need to remove this now?

All parent spends for beta-reward forwarding _should_ be the same. 

So what are we losing here? (I know we aren't using `reason` at the moment, but the idea was to get this in so that we could (in theory) allow reupload of data that has already been paid for. 

That would rely upon parent spends being the same (the reason is the data addr), and royalty existing that matched amount, I think.

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use crate::{DerivationIndex, Hash, Result, TransferError};
+
+/// The attached metadata or reason for which a Spend was spent
+#[derive(Default, Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub enum SpendReason {
+    #[default]
+    None,
+    /// Reference to network data
+    NetworkData(XorName),
```
maqi, line 24: better have this as NetworkAddress instead of XorName
so that a refer to Register will be more convenient.

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Custom field for any application data
+    Custom(#[serde(with = "serde_bytes")] [u8; 64]),
```
maqi: better use const instead of maject number of `64`,
say `MAX_CIPHER_SIZE / 2` ?

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+impl DiscordNameCipher {
+    /// Create a new DiscordNameCipher from a Discord username
+    /// it is encrypted to the given pubkey
+    pub fn create(user_name: &str, foundation_pk: PublicKey) -> Result<Self> {
+        let rng = &mut rand::thread_rng();
+        let discord_name = DiscordName {
+            hash: Hash::hash(user_name.as_bytes()),
```
maqi: better move this into a constructor of `DiscordName`
say
```
DiscordName::new(user_name: &str, foundation_pk: PublicKey) -> Result<DiscordName> {
    let rng = &mut rand::thread_rng();
    DiscordName {
        hash: Hash::hash(user_name.as_bytes()),
        nonce: DerivationIndex::random(rng),
    }
}
```

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+const MAX_CIPHER_SIZE: usize = std::u8::MAX as usize;
+
+/// Discord username encrypted to the Foundation's pubkey with a random nonce
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct DiscordNameCipher {
+    /// Length of the cipher, hard limited to MAX_U8
+    len: u8,
+    /// Encrypted Discord username
+    #[serde(with = "serde_bytes")]
+    cipher: [u8; MAX_CIPHER_SIZE],
```
maqi, line 59: it's actually shall be `cipher_text` ?

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+lazy_static! {
+    pub static ref FOUNDATION_PK: PublicKey = crate::NETWORK_ROYALTIES_PK.public_key();
+}
+const MAX_CIPHER_SIZE: usize = std::u8::MAX as usize;
+
+/// Discord username encrypted to the Foundation's pubkey with a random nonce
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+pub struct DiscordNameCipher {
+    /// Length of the cipher, hard limited to MAX_U8
+    len: u8,
```
maqi, line 56: don't need this extra field ?
unless it is for `the length of the content before encrypted` ?
but is there usecase for that?

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+}
+
+lazy_static! {
+    pub static ref FOUNDATION_PK: PublicKey = crate::NETWORK_ROYALTIES_PK.public_key();
+}
+const MAX_CIPHER_SIZE: usize = std::u8::MAX as usize;
+
+/// Discord username encrypted to the Foundation's pubkey with a random nonce
```
maqi, line 52: > encrypted to the Foundation's pubkey

`by`

sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -0,0 +1,145 @@
...
+/// u256 hash + u256 nonce might be overkill (very big)
+struct DiscordName {
+    hash: Hash,
+    nonce: DerivationIndex,
+}
+
+impl DiscordName {
+    fn to_sized_bytes(&self) -> [u8; 64] {
+        let mut bytes: [u8; 64] = [0; 64];
+        bytes[0..32].copy_from_slice(self.hash.slice());
```
maqi: use `const` if possible?
I am wondering where these 32/64 come from.
are they determined by the Hash/DerivationIndex size?

sn_transfers/src/cashnotes/cashnote.rs
grumbach: The intention was indeed to reupload data that was paid for, but this feature couldn't work with batches properly so it was just forgotten. 
The reason I remove this check is that it protects us from virtually nothing and does not contribute to the security at all. 
Plus we agreed with Qi that spend reasons should be independent.

sn_transfers/src/cashnotes/spend_reason.rs
grumbach: I agree and wanted to, but this creates a circular dependency... We should probably move NetworkAdress to another crate?

sn_transfers/src/cashnotes/spend_reason.rs
grumbach: Yeah const is good. Maybe we should not link this to the beta rewards consts, I'll make a new one

sn_transfers/src/cashnotes/spend_reason.rs
grumbach: We need it to know how long the ciphertext is for decryption, as including the extra padding makes decryption fail.

sn_transfers/src/cashnotes/spend_reason.rs
grumbach: It contains the ciphertext, but is of constant size. We can rename it, but we have to make sure not to confuse it with the original ciphertext.

sn_transfers/src/cashnotes/cashnote.rs
grumbach: Also, I removed the spent_reason method from CashNote as we discussed, making this check even more absurd

sn_transfers/src/cashnotes/spend_reason.rs
grumbach: It's encrypted `to` the Foundation's pubic key and decryptable `by` it. Or did I miss something?

maqi (approved):

PR #1691, opened 2024-05-08
feat: run safenode services in user mode

jacderida:
- f5ed0523 **feat: run safenode services in user mode**

  The `Systemd` and `Launchd` service managers support running services in user mode, and the
  `service-manager-rs` crate already accommodates it, so we can support this scenario.

  Services are managed in user mode depending on whether the node manager is running elevated or not.
  In some cases, elevated privileges are still required, because not all service managers support
  user-mode services. For example, Alpine uses OpenRC, which does not have support. A service running
  in user-mode requires an active user session. This may not always be appropriate, particularly for
  advanced users who want to leave nodes running permanently in the background.

  The RPC daemon and the faucet services are still defined to run as system-wide services and they
  don't support running in user-mode, because it doesn't really make sense to require an active user
  session for these services.

- 0ad2e90f **docs: update cli and readme for user-mode services**

  The node manager documentation is updated in the README and the CLI for the introduction of
  user-mode services.

  The README also needed a little update for the `local` subcommand.

- 2d09deee **fix(launchpad): prevent mac opening with sudo**

  Add context to some error messages

- 5df12090 **chore: change terminal launch behaviour**

  On Linux, the `TERM` variable is not really appropriate for selecting which terminal is in use, so
  this has been removed. It can be set to a completely different value, like `xterm-256color`, which
  does not necessarily correspond to the running terminal. The `sudo` was also removed from the
  terminal launch mechanism.

  On the Windows front, I updated Windows Terminal to not use the `/c` argument. It simply just passes
  the command to the terminal.

  Also fixed another couple of issues:
  * Compile the node manager correctly on Windows.
  * The use of `<service>` in doc comments was being flagged as an unclosed HTML tag.

## Description

reviewpad:summary

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,7 +86,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.14",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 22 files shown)


joshuef:
Nice one! 🎉

joshuef (approved):

PR #1694, opened 2024-05-09
feat(launchpad): initial automatic resource allocation logic

RolandSherwin:
## Description

reviewpad:summary

### DIFF
--- a/node-launchpad/.config/config.json5
+++ b/node-launchpad/.config/config.json5
@@ -4,7 +4,7 @@
       "<Ctrl-g>": {"HomeActions":"StartNodes"},
       "<Ctrl-G>": {"HomeActions":"StartNodes"},
       "<Ctrl-x>": {"HomeActions":"StopNodes"}, 
-      "<Ctrl-X>": {"HomeActions":"StopNodes"}, 
+      "<Ctrl-X>": {"HomeActions":"StopNodes"},
 
       "up" : {"HomeActions":"PreviousTableItem"},
       "down": {"HomeActions":"NextTableItem"},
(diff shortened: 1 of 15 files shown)


node-launchpad/.config/config.json5
```diff
@@ -1,27 +1,39 @@
...
     "Home": {
-      "<a>": {"HomeActions":"AddNode"}, // Add nodes
-      "<s>": {"HomeActions":"StartNodes"}, // Start nodes
-      "<k>": {"HomeActions":"StopNode"}, // Stop nodes
-      "<r>": {"HomeActions":"RemoveNode"},
-      "<d>": {"HomeActions":"InputDiscordUsername"},
+      "<Ctrl-g>": {"HomeActions":"StartNodes"},
+      "<Ctrl-G>": {"HomeActions":"StartNodes"},
+      "<Ctrl-x>": {"HomeActions":"StopNodes"}, 
+      "<Ctrl-X>": {"HomeActions":"StopNodes"},
```
mazzi: Little linting. Blank in EOL.

node-launchpad/src/components/home.rs
```diff
@@ -328,72 +377,106 @@ impl Component for Home {
...
+fn add_and_start_nodes(
+    count: usize,
+    mut nodes_to_start: Vec<String>,
+    action_sender: UnboundedSender<Action>,
+) {
+    let peers = PeersArgs::default(); // will fetch from network contacts
+
+    tokio::task::spawn_local(async move {
+        let result = sn_node_manager::cmd::node::add(
+            Some(count as u16),
```
joshuef, line 437: should this not be `nodes_to_start` ?

joshuef (changes requested):
I think we need the `add_multiple_based_upon_storage` concept in the nodemanager itself?

So the TUI would just be calling that with provided value? 

(And the nodeman can just have it passed as a flag to get the same result?

node-launchpad/src/components/home.rs
RolandSherwin: `nodes_to_start` denotes the inactive nodes that we are supposed to restart basically.

joshuef:
Couple issues:

Need to `clear` input when being reused.
- Start-nodes + allocate `10`
- Press `D` for discord.
- See: 

Screenshot 2024-05-13 at 10 26 07

------

`Stop node` stops all nodes. (May be desireable... just need to update text)


----

Why `ctrg+g` to start and not `ctrl+s`?

joshuef (approved):

PR #1697, opened 2024-05-13
chore(refactor): stabilise node size to 4k records,

joshuef:
Update some reward forwarding naming

## Description

reviewpad:summary

### DIFF
--- a/sn_node/src/node.rs
+++ b/sn_node/src/node.rs
@@ -308,10 +308,10 @@ impl Node {
                             let start = std::time::Instant::now();
                             trace!("Periodic balance forward triggered");
                             let network = self.network.clone();
-                            let owner = self.owner.clone();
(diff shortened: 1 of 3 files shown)


sn_node/src/node.rs
```diff
@@ -759,9 +759,8 @@ impl Node {
         }
     }
 
-    // Shall be disabled after Beta
-    #[cfg(feature = "reward-forward")]
```
maqi, line 760: feature guard is required, unless the entire feature gate is removed ?

joshuef: I think we want this feature in general, but I didn't want to go into the extra config here to remove the feat entirely (as forwarding must be optional in the end)

maqi (approved):

PR #1701, opened 2024-05-13
feat: tracking beta rewards from the DAG

grumbach:
## Description

reviewpad:summary


maqi (approved):

sn_auditor/src/dag_db.rs
```diff
@@ -195,6 +213,86 @@ impl SpendDagDb {
...
+            let addr = spend.address();
+            let amount = spend.spend.amount;
+            if let Some(user_name) = self.beta_participants.get(&user_name_hash) {
+                debug!("Got beta reward from {user_name} of {amount} at {addr:?}");
+                rewards
+                    .entry(user_name.to_owned())
+                    .or_default()
+                    .insert((addr, amount));
+            } else {
+                warn!(
```
maqi: those transfers to credit the uploading clients, and those clients that paying the upload cost will be detected as `unknown participant` as well.
However they will be common.
So, maybe `warn` is too severe, `info` will be enough ?
or only warn when it is a `forward spend`, i.e. has only one input and one output (transfer or payment spend will have multiple outputs)

PR #1707, opened 2024-05-15
feat(nodeman): add LogFormat as a startup arg for nodes

joshuef:
## Description

reviewpad:summary


sn_node_manager/src/bin/cli/main.rs
```diff
@@ -557,6 +565,13 @@ pub enum LocalSubCmd {
         /// The version and path arguments are mutually exclusive.
         #[clap(long)]
         node_version: Option<String>,
+        /// Specify the logging format.
```
jacderida: Hey Josh, sorry for the pedantic issue, but would you mind please specifying the arguments in alphabetical order? There are so many of them and I think it's the best way to keep them organised for easy reference.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -615,6 +630,13 @@ pub enum LocalSubCmd {
         /// The version and path arguments are mutually exclusive.
         #[clap(long, conflicts_with = "build")]
         node_version: Option<String>,
+        /// Specify the logging format.
+        ///
+        /// Valid values are "default" or "json".
+        ///
+        /// If the argument is not used, the default format will be applied.
+        #[clap(long, value_parser = LogFormat::parse_from_str, verbatim_doc_comment)]
+        log_format: Option<LogFormat>,
```
jacderida: Same applies here please.

sn_service_management/Cargo.toml
```diff
@@ -19,8 +19,11 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 semver = "1.0.20"
 service-manager = "0.6.1"
-sn_protocol = { path = "../sn_protocol", version = "0.16.6", features = ["rpc"] }
+sn_protocol = { path = "../sn_protocol", version = "0.16.6", features = [
+    "rpc",
+] }
 sn_transfers = { path = "../sn_transfers", version = "0.18.0" }
+sn_logging = { path = "../sn_logging", version = "0.2.26" }
```
jacderida: Would you mind also putting these references in alphabetical order please?

jacderida (commented):
I think also, you may possibly need to update the node registry such that it tracks this new argument, otherwise the specification will be lost on upgrade. If this was done, a new test case for `add` would also be welcomed.

It might be better to just keep this PR on ice for just a little bit until I submit my PR with the `--upnp` flag, which will demonstrate where things need to be specified for upgrades.

jacderida (approved):

PR #1708, opened 2024-05-15
Use different keys

maqi:
## Description

reviewpad:summary

### DIFF
--- a/sn_auditor/src/main.rs
+++ b/sn_auditor/src/main.rs
@@ -64,6 +64,10 @@ struct Opt {
     /// Provide a JSON file with a list of Discord usernames as argument
     #[clap(short, long, value_name = "discord_names_file")]
     beta_participants: Option<PathBuf>,
+
+    /// Hex string of the Foundation SK.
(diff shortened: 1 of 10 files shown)


sn_auditor/src/main.rs
```diff
@@ -64,6 +64,10 @@ struct Opt {
     /// Provide a JSON file with a list of Discord usernames as argument
     #[clap(short, long, value_name = "discord_names_file")]
     beta_participants: Option<PathBuf>,
+
+    /// Hex string of the Foundation SK.
+    #[clap(name = "sk")]
+    sk_str: String,
```
grumbach, line 70: I'd make it an option so people can run sn_auditor without the beta rewards feature? 
Not a blocker

grumbach (approved):
LGTM

sn_auditor/src/main.rs
maqi: I think a `mandatory` field will remind us to `remove` it, once AFTER the beta

PR #1720, opened 2024-05-16
feat: upgrade cli audit to use DAG

grumbach:
## Description

- upgrade cli audit to use DAG
- remove some code duplication between cli and auditor
- make foundation sk an optional argument in both cli and auditor as most runs won't need it (aren't related to beta rewards forwarding)

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -310,7 +310,7 @@ jobs:
           cat spend_dag_and_statistics.txt
         env:
           SN_LOG: "all"
-        timeout-minutes: 5
+        timeout-minutes: 15
         if: always()
 
       - name: Ensure discord_ids decrypted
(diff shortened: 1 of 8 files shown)


.github/workflows/memcheck.yml
```diff
@@ -288,7 +288,7 @@ jobs:
           cat spend_dag_and_statistics.txt
         env:
           SN_LOG: "all"
-        timeout-minutes: 5
+        timeout-minutes: 15
```
maqi: shall not take such long time

maqi (changes requested):

.github/workflows/memcheck.yml
grumbach: 100% agree

maqi (approved):

PR #1735, opened 2024-05-20
chore: enable default features during faucet release

RolandSherwin:
We currently need the `gifting` feature to test out things.


sn_faucet/src/main.rs
```diff
@@ -7,7 +7,6 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 mod faucet_server;
-
```
RolandSherwin, line 10: To trigger `release` flow for faucet

maqi (approved):

PR #1754, opened 2024-05-22
feat(manager): maintain n running nodes

RolandSherwin:
[feat(manager): maintain n running nodes](https://github.com/maidsafe/safe_network/commit/994ddbc32c945e8d1a0a0168b45b09a1dac1a3ae)
- This commit was supposed to be part of #1694  as mentioned [here](https://github.com/maidsafe/safe_network/pull/1694#pullrequestreview-2049047244) . But it got removed during a rebase. So adding it back in.

[feat(launchpad): provide safenode path for testing](https://github.com/maidsafe/safe_network/commit/a8ce64fd001880d425c614429aa8bd16bb611a98)
- This commit is used to use a pre built safenode binary inside launchpad for testing things out. Else we'd be fetching the nodes from latest release everytime (which will be old).

### DIFF
--- a/node-launchpad/src/components/home.rs
+++ b/node-launchpad/src/components/home.rs
@@ -18,6 +18,7 @@ use ratatui::{prelude::*, widgets::*};
 use sn_node_manager::{config::get_node_registry_path, VerbosityLevel};
 use sn_peers_acquisition::PeersArgs;
 use sn_service_management::{NodeRegistry, NodeServiceData, ServiceStatus};
+use std::path::PathBuf;
 use tokio::sync::mpsc::UnboundedSender;
 
 const NODE_START_INTERVAL: usize = 10;
@@ -37,6 +38,8 @@ pub struct Home {
     lock_registry: bool,
     // Peers to pass into nodes for startup
     peers_args: PeersArgs,
+    // If path is provided, we don't fetch the binary from the network
+    safenode_path: Option<PathBuf>,
 }
 
 impl Home {
@@ -44,6 +47,7 @@ impl Home {
         allocated_disk_space: usize,
         discord_username: &str,
         peers_args: PeersArgs,
+        safenode_path: Option<PathBuf>,
     ) -> Result<Self> {
         let mut home = Self {
             peers_args,
@@ -55,6 +59,7 @@ impl Home {
             node_table_state: Default::default(),
             lock_registry: Default::default(),
             discord_username: discord_username.to_string(),
+            safenode_path,
         };
         home.load_node_registry_and_update_states()?;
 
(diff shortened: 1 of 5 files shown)


node-launchpad/src/components/home.rs
```diff
@@ -469,35 +388,19 @@ fn add_and_start_nodes(
...
+            VerbosityLevel::Minimal,
             NODE_START_INTERVAL as u64,
-            vec![],
-            nodes_to_start,
-            sn_node_manager::VerbosityLevel::Minimal,
         )
         .await
         {
-            error!("Error while starting services {err:?}");
+            error!("Error while maintaining n running nodes {err:?}");
```
maqi: this `n` can be replaced by `count` number ?

node-launchpad/src/components/home.rs
```diff
@@ -469,35 +388,19 @@ fn add_and_start_nodes(
...
         } else {
-            info!("Successfully started services");
+            info!("Maintain node run count successful.");
```
maqi: Maintain {count} running nodes successfully

sn_node_manager/src/cmd/node.rs
```diff
@@ -492,6 +492,138 @@ pub async fn upgrade(
     Ok(())
 }
 
+/// This functions makes sure that we run exactly the provided count nodes at any time by stopping nodes or
```
maqi: `This functions makes sure that we run exactly the provided count nodes at any time` =>
`This function ensures running exactly the provided number of nodes at any time`

sn_node_manager/src/cmd/node.rs
```diff
@@ -492,6 +492,138 @@ pub async fn upgrade(
...
+    rpc_address: Option<Ipv4Addr>,
+    rpc_port: Option<PortRange>,
+    src_path: Option<PathBuf>,
+    url: Option<String>,
+    upnp: bool,
+    user: Option<String>,
+    version: Option<String>,
+    verbosity: VerbosityLevel,
+    start_node_interval: u64,
+) -> Result<()> {
```
maqi, line 520: I am suprised this doesn't trigger a clippy complain that `too many arguments to be passed` 
maybe make those for `add` as a separate struct ?

maqi (approved):

mazzi (approved):
LGTM 🚀

sn_node_manager/src/cmd/node.rs
RolandSherwin: I think there is a global `allow()` macro. Most of the functions there has got like 10+ args. Might be out of scope for this PR. :/

PR #1756, opened 2024-05-22
fix(uploader): do not error out immediately on max repayment errors

RolandSherwin:
This PR allows us to continue the upload process even if a single item fails due to `MaxRepaymentsReached`
The error is then triggered after all the other item are cleared.

Closes Issue https://github.com/maidsafe/safe_network/issues/1730

### DIFF
--- a/sn_client/src/error.rs
+++ b/sn_client/src/error.rs
@@ -8,6 +8,8 @@
 
 pub(crate) type Result<T> = std::result::Result<T, Error>;
 
+use crate::UploadSummary;
+
 use super::ClientEvent;
 use sn_protocol::NetworkAddress;
 use sn_registers::{Entry, EntryHash};
@@ -140,8 +142,14 @@ pub enum Error {
     #[error("Too many sequential payment errors reported during upload")]
     SequentialUploadPaymentError,
 
-    #[error("The maximum specified repayments has been reached for {0:?}")]
-    MaximumRepaymentsReached(Vec<XorName>),
+    #[error("The maximum specified repayments has been reached for a single item: {0:?}")]
+    MaximumRepaymentsReached(XorName),
+
+    #[error("The upload failed with maximum repayments reached for multiple items: {items:?} Summary: {summary:?}")]
+    UploadFailedWithMaximumRepaymentsReached {
+        items: Vec<XorName>,
+        summary: UploadSummary,
+    },
 
     #[error("Error occurred when access wallet file")]
     FailedToAccessWallet,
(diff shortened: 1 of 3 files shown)


sn_client/src/error.rs
```diff
@@ -140,8 +140,8 @@ pub enum Error {
     #[error("Too many sequential payment errors reported during upload")]
     SequentialUploadPaymentError,
 
-    #[error("The maximum specified repayments were made for the address: {0:?}")]
-    MaximumRepaymentsReached(XorName),
+    #[error("The maximum specified repayments has been reached reached")]
```
maqi: has been reached ~reached~

sn_client/src/uploader/upload.rs
```diff
@@ -120,6 +123,14 @@ pub(super) async fn start_upload(
             #[cfg(test)]
             trace!("UPLOADER STATE: finished uploading all items {uploader:?}");
 
+            if !uploader.max_repayments_reached.is_empty() {
```
maqi: this is inside the check of `if uploader.all_upload_items.is_empty() {` ,
returning at this point shall get the statistics included in the UploadSummary (if needed), instead of return with the specific error

maqi (changes requested):

sn_client/src/uploader/upload.rs
RolandSherwin: Hey Qi, when the `max_repayments` is reached, the item is removed from `all_upload_items` and added `max_repayments_reached` struct. It is still an error, it is just that I chose to remove it from the flow there. We could've also had a filter. 
But it would still be an error, else people would get the "Successful uploaded message" and stuff, but in reality we failed to upload a couple items. Maybe we can also return the list of failed items inside ClientError instead.

joshuef: > Maybe we can also return the list of failed items inside ClientError instead.

Sounds good

joshuef (approved):

sn_client/src/uploader/upload.rs
maqi: what I mean is that without the summary statistics, the user won't be able to know how many chunks uploaded successfully when received that error.

maybe, the error shall includes:
```
let upload_summary = UploadSummary { ... };
return Err(ClientError::MaximumRepaymentsReached {
                    failed_chunks: uploader.max_repayments_reached.into_iter().collect(),
                    upload_summary,
                });
```
then the user can clearly know how many succeeded, how many already presents, how much cost paid...

maqi (approved):

PR #1762, opened 2024-05-23
fix: avoid adding mixed type addresses into RT

maqi:
### Description

This pull request includes changes to the `sn_networking/src/event/swarm.rs` file that modify the behavior of the `SwarmDriver` implementation. The most important changes include modifying the `addrs` variable to be mutable, adding a check for relayed addresses before adding potential relay candidates, and ensuring that relay addresses don't coexist with direct link format addresses.

Here are the key changes:

* [`sn_networking/src/event/swarm.rs`](diffhunk://#diff-cb0730cf4bf5e7445bc11bf88d45c13b6f28a706a5d1a7822db1e2dd76b1b2dfL143-R143): Changed the `addrs` variable from immutable to mutable in the `SwarmDriver` implementation. This allows for modifications to the `addrs` variable later in the code.
* [`sn_networking/src/event/swarm.rs`](diffhunk://#diff-cb0730cf4bf5e7445bc11bf88d45c13b6f28a706a5d1a7822db1e2dd76b1b2dfR157-R169): Added a check to see if any of the addresses are relayed. If none are relayed, potential relay candidates are added to the `relay_manager`. This prevents already relayed peers from being used as potential relay candidates.
* [`sn_networking/src/event/swarm.rs`](diffhunk://#diff-cb0730cf4bf5e7445bc11bf88d45c13b6f28a706a5d1a7822db1e2dd76b1b2dfR243-R249): Added a check to ensure that relay addresses do not coexist with direct link format addresses. If any relayed addresses are found, the `addrs` HashSet is filtered to retain only relay addresses. This avoids potential issues arising from the coexistence of different address formats.

### Related Issue

### Type of Change

### DIFF
--- a/sn_networking/src/event/swarm.rs
+++ b/sn_networking/src/event/swarm.rs
@@ -140,7 +140,7 @@ impl SwarmDriver {
                         // If we're not in local mode, only add globally reachable addresses.
                         // Strip the `/p2p/...` part of the multiaddresses.
                         // Collect into a HashSet directly to avoid multiple allocations and handle deduplication.
-                        let addrs: HashSet<Multiaddr> = match self.local {
+                        let mut addrs: HashSet<Multiaddr> = match self.local {
                             true => info
                                 .listen_addrs
                                 .into_iter()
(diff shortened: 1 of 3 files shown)


sn_networking/src/event/swarm.rs
```diff
@@ -154,12 +154,19 @@ impl SwarmDriver {
...
 
-                        self.relay_manager.add_potential_candidates(
-                            &peer_id,
-                            &addrs,
-                            &info.protocols,
-                            &self.bad_nodes,
-                        );
+                        let has_relayed = addrs.iter().any(|multiaddr| {
+                            multiaddr.iter().any(|p| matches!(p, Protocol::P2pCircuit))
+                        });
```
RolandSherwin: This variable is used at two places, but can we possibly move this check inside `add_potential_candidates` just to keep logic at one place? 😄

RolandSherwin (approved):
Great digging 🚀

PR #1763, opened 2024-05-23
feat: add nat detection to safenode manager

RolandSherwin:
This PR integrates the `nat-detection` binary to be used with `safenode manager`.
Usage: `safenode-manager nat-detection run --servers <multiaddrs>` 

Running the above command will fetch our machine's NAT status and will store it locally.
This can be later applied during the `add` command to auto select between `--upnp or --home-network` flag.
Usage: `safenode manager add --auto-set-nat-flags`

#### NAT Detection working:
Our `nat-detection` binary works by connecting to other `nat-detection` programs running in server mode. The multiaddrs of these servers has to be provided to the `safenode_manager::nat-detection` argument. We then connect as a client and carry out probing. After the probing is done, we'd know if we're 1) Public 2) Public w/ UPnP 3) Private.
Since `nat-detection` is executed as a program, we cannot get this state directly and hence we're using various exit codes to tell the parent program (manager) about the state.

This state is then used to provide the following options to the safenodes:
- Public = no upnp / home-network
- Upnp = upnp only
- private = home-network


sn_node_manager/src/bin/cli/main.rs
```diff
@@ -627,6 +639,43 @@ pub enum FaucetSubCmd {
     },
 }
 
+/// Nat Detection process subcommands.
```
jacderida: Can this please be changed to `Manage NAT detection`?

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -627,6 +639,43 @@ pub enum FaucetSubCmd {
...
+#[derive(Subcommand, Debug, Clone)]
+pub enum NatDetectionSubCmd {
+    /// RUn NAT detection to fetch and store our NAT status to the node registry.
```
jacderida: There is a little typo here "RUn". In addition though, I would suggest changing this to:
```
Use NAT detection to determine NAT status.
```

On the UI there is no necessity to mention it gets written to the node registry, which is an internal concept.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -627,6 +639,43 @@ pub enum FaucetSubCmd {
...
+    ///
+    /// The status can be later used with the `safenode-manager add --auto-set-nat-flags` command.
```
jacderida: I would suggest changing this to:
```
The status can be used with the '--auto-set-nat-flags' argument on the 'add' command.
```

The backticks do not get parsed in the terminal, and we don't need to refer to the whole 'safenode-manager' because we are already in that context.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -627,6 +639,43 @@ pub enum FaucetSubCmd {
...
+    Run {
+        /// Provide a path for the NAT detection binary to be used.
+        ///
+        /// Useful for running NAT detection using a custom built binary.
+        #[clap(long)]
+        path: Option<PathBuf>,
+        /// Provide the list of NAT servers to connect.
```
jacderida: I would suggest changing this to:
```
Provide NAT servers in the form of a multiaddr or an address/port pair.

We attempt to establish connections to these servers to determine our own NAT status.

The argument can be used multiple times.
```

sn_node_manager/src/cmd/node.rs
```diff
@@ -28,18 +31,20 @@ use sn_releases::{ReleaseType, SafeReleaseRepoActions};
...
 use tracing::debug;
 
 /// Returns the added service names
 pub async fn add(
+    auto_set_nat_flags: bool,
     count: Option<u16>,
     data_dir_path: Option<PathBuf>,
     env_variables: Option<Vec<(String, String)>>,
-    home_network: bool,
+    mut home_network: bool,
```
jacderida: This is a minor point, but I feel it's a bit odd to set these to mutable if they will not then be used outside this function. Does the caller then need to declare the variables mutable?

Can we not deal with these internally? Using new variables is fine, in my opinion.

sn_node_manager/src/cmd/node.rs
```diff
@@ -119,6 +124,29 @@ pub async fn add(
         },
     };
 
+    if auto_set_nat_flags {
+        let nat_status = node_registry
+            .nat_status
+            .clone()
+            .ok_or_eyre("NAT status has not been set. Run `nat-detection` first")?;
+
+        // override the upnp and home-network options
```
jacderida: It seems like this case would be good for some unit tests, but I don't see you've added any?

Other minor point: the code says this fine on its own without the comment.

sn_node_manager/src/cmd/node.rs
RolandSherwin: No, the caller does not need to declare the variable mutable. The variable is moved entirely to the function, so we can just make it mutable at the function level.
Instead of having another line which does `let mut home_network = home_network`

sn_node_manager/src/cmd/node.rs
RolandSherwin: Right yeah, I'll add some unit tests.

jacderida (approved):

PR #1765, opened 2024-05-23
feat(nat_detection): release nat detection binary

RolandSherwin:
release nat detection binary


Justfile
```diff
@@ -331,6 +342,9 @@ upload-release-assets-to-s3 bin_name:
     faucet)
       bucket="sn-faucet"
       ;;
+    nat-detection)
+      bucket="sn-nat-detection"
```
jacderida: That needs updated to `nat-detection`.

jacderida (approved):

PR #1774, opened 2024-05-27
chore(cli): showing cli final execution result explicitly

maqi:
### Description

Showing the cli final execution result, to address the issue raised at https://forum.autonomi.community/t/thisisnotbetarewardsnet-mini-testnet-26-05-24/39824/140
that when receive a transfer, showing no error in terminal nor log, but no balance credited.

Note: there is printout of `Verifying transfer with the Network...`, but no printout of `Successfully verified transfer.`, 
which indicates a failure during `transfer verification` terminated flow.

### Related Issue

### Type of Change

- [x] Documentation update

### DIFF
--- a/sn_cli/src/bin/main.rs
+++ b/sn_cli/src/bin/main.rs
@@ -140,23 +140,27 @@ async fn main() -> Result<()> {
     // default to verifying storage
     let should_verify_store = !opt.no_verify;
 
-    match opt.cmd {
+    // PowerShell seems having issue to showing the unwrapped error
+    // Hence capture the result and print it out explicity.
+    let cmd_str = format!("{:?}", opt.cmd);
+    let result = match opt.cmd {
         SubCmd::Wallet(cmds) => {
-            wallet_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await?
+            wallet_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await
         }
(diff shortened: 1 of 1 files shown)


sn_cli/src/bin/main.rs
```diff
@@ -140,23 +140,27 @@ async fn main() -> Result<()> {
...
         SubCmd::Folders(cmds) => {
-            folders_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await?
+            folders_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await
         }
         SubCmd::Register(cmds) => {
-            register_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await?
+            register_cmds(cmds, &client, &client_data_dir_path, should_verify_store).await
         }
     };
+    println!("Completed with {result:?} of execute {cmd_str:?}");
```
RolandSherwin, line 163: It would be better if result is at the last, because errors can be large at times.
```suggestion
    println!("Executed {cmd_str:?} with result: {result:?}");
```

RolandSherwin (approved):

sn_cli/src/bin/main.rs
maqi: I think the cmd_str could be long, meanwhile the execution result shall be short in most of time.
Hence put the result in front.

RolandSherwin: Ah I see :+1:

PR #1777, opened 2024-05-28
chore(auditor): refactor spends fetch to be background progressing

joshuef:
This pull request mainly focuses on improving the `sn_auditor` package, particularly the `dag_db.rs` and `main.rs` files. The changes involve adding new libraries, modifying existing methods, and introducing new methods for better DAG collection and handling. The changes are aimed at enhancing the functionality of the `SpendDagDb` struct and improving the overall performance of the software.

The most important changes include:

* Addition of new libraries and modification of existing imports:
  * `tokio = { version = "1.32.0", features = [` in file `sn_auditor/Cargo.toml`: Added `itertools = "0.13.0"` and `futures = "0.3.30"` libraries.
  * `use bls::SecretKey;` in file `sn_auditor/src/dag_db.rs`: Added `use futures::future::join_all;` and modified the `use sn_client::transfers::` import.
  * `use std::time::{SystemTime, UNIX_EPOCH};` in file `sn_auditor/src/dag_db.rs`: Added `use std::time::{Duration, Instant};`.

* Changes to the `SpendDagDb` struct in `sn_auditor/src/dag_db.rs`:
  * Modified the access modifier of `beta_participants` from `pub(crate)` to private and added the `REATTEMPT_INTERVAL` constant.
  * Removed the `update` method and modified the `track_new_beta_participants` and `init_reward_forward_tracking` methods to use `BTreeSet<String>` instead of `Vec<String>`. [[1]](diffhunk://#diff-fc4319f01fee1895543b912d85d9455555b052ecc12a54539497148590b3dc48L183-L232) [[2]](diffhunk://#diff-fc4319f01fee1895543b912d85d9455555b052ecc12a54539497148590b3dc48L266-R222) [[3]](diffhunk://#diff-fc4319f01fee1895543b912d85d9455555b052ecc12a54539497148590b3dc48R252-R438)
  * Introduced new methods `backup_rewards`, `dag_crawling_from_genesis`, and `process_spend`.

* Changes to `sn_auditor/src/main.rs`:
[...]

PR #1785, opened 2024-05-29
Integrate DAG crawl optimisations from #1777

grumbach:
Integrate DAG crawl changes and optimizations from #1777 including:
- `sn_networking` hard coded version
- parsing of discord names provided by url
- same UTXO delay before retry

Other:
- all DAG crawling back in the same place
- Spends can be sent through a channel for custom processing
- switch from `std::sync`'s `RwLock` to `tokio::sync`'s as the lock is used in tokio co-routines and not in threads


sn_networking/src/version.rs
```diff
@@ -80,7 +80,7 @@ fn get_truncate_version_str() -> String {
     let version_str = env!("CARGO_PKG_VERSION");
     let parts = version_str.split('.').collect::<Vec<_>>();
     if parts.len() >= 2 {
-        format!("{}.{}", parts[0], parts[1])
+        format!("{}.{}", parts[0], "16")
```
grumbach: Change the version according to #1777

sn_client/src/audit/dag_crawling.rs
```diff
@@ -38,7 +59,7 @@ impl Client {
     pub async fn spend_dag_build_from(
         &self,
         spend_addr: SpendAddress,
-        max_depth: Option<u32>,
+        spend_processing: Option<Sender<SignedSpend>>,
```
grumbach, line 62: Optional channel Sender through which we send the Spends collected. 
This is useful for real time external processing of those spends for:
- beta rewards
- following progress
- any other applications (could be used for royalties as well)

sn_client/src/audit/dag_crawling.rs
```diff
@@ -21,6 +24,24 @@ enum InternalGetNetworkSpend {
 }
 
 impl Client {
+    pub async fn new_dag_with_genesis_only(&self) -> WalletResult<SpendDag> {
```
grumbach, line 27: Moved this form `sn_auditor` to a `Client` method.

sn_networking/src/version.rs
joshuef: This should have been a blocker on that PR. I'll push a commit removing this for `main` integration. (any new network wont need it)

joshuef (approved):

PR #1789, opened 2024-05-30
feat: launchpad changes

RolandSherwin:
- Revamps the TUI to use the new color pallet
- Adds support for displaying some stats
- Reset the services if the discord id changes

This PR will be followed up with more changes to polish things up.


node-launchpad/src/components/footer.rs
```diff
@@ -6,50 +6,126 @@
...
+            Style::default().fg(GHOST_WHITE)
+        } else {
+            Style::default().fg(LIGHT_PERIWINKLE)
+        };
+
+        let (line1, line2) = match self.current_scene {
             Scene::Home => {
-                "[Ctrl+g] Start nodes, [Ctrl+x] Stop nodes, [O] Set Resources, [D]iscord Username, [Q]uit".into()
+                let line1 = Line::from(vec![
+                    Span::styled(" [Ctrl+G] ", command_style),
```
b-zee, line 86: Is there a reason for the space in here?

node-launchpad/src/components/resource_allocation.rs
```diff
@@ -21,7 +21,8 @@ use crate::{
 use super::{utils::centered_rect_fixed, Component};
 
 pub const GB_PER_NODE: usize = 5;
-pub const GB: usize = 1024 * 1024 * 1024;
+pub const MB: usize = 1000 * 1000;
+pub const GB: usize = MB * 1000;
```
b-zee, line 25: Just a nitpick, but I think we're using `MiB`/`GiB` (mebibyte/gibibyte) mostly in our apps (1024/binary based), lining up with chunks which are 512KiB(?).

b-zee:
TUI looks clean with those colors, nice job!

node-launchpad/src/components/footer.rs
```diff
@@ -6,50 +6,126 @@
...
+                    Span::styled(
+                        "Manage Nodes               ",
+                        Style::default().fg(EUCALYPTUS),
+                    ),
+                    Span::styled("[D] ", Style::default().fg(GHOST_WHITE)),
+                    Span::styled(
+                        "Beta Rewards Programme      ",
+                        Style::default().fg(VERY_LIGHT_AZURE),
+                    ),
+                    Span::styled("[Q] ", Style::default().fg(GHOST_WHITE)),
```
b-zee, line 105: Maybe use lower case `o`, `d` and `q` here. Might be just me being pedantic, but I pressed `Shift` a few times.

Just looked at `dua`, which is quite a popular app using `ratatui`: https://github.com/Byron/dua-cli/blob/ced3b4f5e375278dbee52319eac8750b14eb328a/src/interactive/widgets/help.rs#L43

node-launchpad/src/components/footer.rs
RolandSherwin: Those spaces are to accomodate the borders. I haven't found a clean way to create a border, and then split up the remaining space into a `Layout` yet. 
It is scattered all over the codebase, and I'm not proud of it, but I don't have any alternate route yet.

node-launchpad/src/components/resource_allocation.rs
RolandSherwin: I used `MiB` earlier but here this is for fs storage and I crosschecked the space used by a folder with the ones from macos -> right click -> info and they did not match. So I assume we should be using `MB/GB` for fs?

node-launchpad/src/components/footer.rs
RolandSherwin: The design is based of off a desgin doc shared on Slack, probably Jim is the guy you'd want to discuss this with.

jacderida: Yeah, I've also pointed out the lowercase/uppercase discrepancy a few times. I gravitated towards using shift too.

joshuef:
Screenshot 2024-05-31 at 09 27 38

joshuef:
Can we drop the bg colour? It seems problematic to me

joshuef:
Screenshot 2024-05-31 at 09 28 48

this is via baked in terminal, looks better

PR #1803, opened 2024-05-31
feat: DAG crawling through generic crawling

grumbach:
- Introduces a generic crawling method
- Makes DAG gathering use that generic method
- Adds a feature flag to auditor (`dag-collection`) to allow to choose if we want local DAG gathering or just crawling

### DIFF
--- a/sn_auditor/src/dag_db.rs
+++ b/sn_auditor/src/dag_db.rs
@@ -21,6 +21,7 @@ use std::fmt::Write;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+use tokio::sync::mpsc::Sender;
 use tokio::sync::RwLock;
 
 pub const SPEND_DAG_FILENAME: &str = "spend_dag";
@@ -168,7 +169,6 @@ impl SpendDagDb {
     }
 
     /// Dump DAG to disk
-    #[cfg(feature = "dag-collection")]
     pub async fn dump(&self) -> Result<()> {
         std::fs::create_dir_all(&self.path)?;
         let dag_path = self.path.join(SPEND_DAG_FILENAME);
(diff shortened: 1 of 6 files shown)


sn_auditor/src/dag_db.rs
```diff
@@ -228,39 +229,70 @@ impl SpendDagDb {
...
-            );
+            #[cfg(not(feature = "dag-collection"))]
+            {
+                if let Some(sender) = spend_processing.clone() {
+                    // crawl DAG
+                    let tasks: Vec<_> = addrs_to_get
+                        .iter()
+                        .map(|a| client.spend_dag_crawl_from(*a, sender.clone()))
+                        .collect();
+                    let res = futures::future::join_all(tasks).await;
```
grumbach: Here the sender is cloned for each parallel crawl tasks in `sender.clone()`, meaning the single receiver will receive spends from all these crawling tasks concurrently.

sn_auditor/src/dag_db.rs
```diff
@@ -228,39 +229,70 @@ impl SpendDagDb {
...
-            if let Err(e) = self.dump().await {
-                error!("Failed to dump DAG: {e}");
+                    // update utxos
+                    utxo_addresses.extend(
+                        new_utxos
+                            .into_iter()
+                            .map(|a| (a, Instant::now() + REATTEMPT_INTERVAL)),
+                    );
+                } else {
+                    panic!("There is no point in running the auditor if we are not collecting the DAG or collecting data through crawling. Please enable the `dag-collection` feature or provide beta program related arguments.");
```
grumbach: In the future we can add royalties collection as a background task similar to beta reward tracking

sn_auditor/src/dag_db.rs
```diff
@@ -160,6 +160,7 @@ impl SpendDagDb {
     }
 
     /// Dump DAG to disk
+    #[cfg(feature = "dag-collection")]
```
maqi: this means for beta_tracking audit, the startup cmd shall not bearing `--features=dag-collection` ?
`dag-collection` is currently part of the feature of `svg-dag`

better inform @joshuef and @RolandSherwin for this change, as it may affect the release or setup flow.

sn_auditor/src/dag_db.rs
```diff
@@ -241,39 +242,70 @@ impl SpendDagDb {
...
-            // update utxos
-            let new_utxos = dag.get_utxos();
-            utxo_addresses.extend(
-                new_utxos
-                    .into_iter()
-                    .map(|a| (a, Instant::now() + *UTXO_REATTEMPT_INTERVAL)),
-            );
+            #[cfg(not(feature = "dag-collection"))]
+            {
+                if let Some(sender) = spend_processing.clone() {
```
maqi: will be better put these two branch of code into separate functions.
i.e. 
```code
if cfg!(feature = "dag-collection") {
    crawl_and_generate_local_dag()
} else {
    crawl_only_for_beta_track()
}
```

sn_auditor/src/dag_db.rs
```diff
@@ -241,39 +242,70 @@ impl SpendDagDb {
...
+                    // crawl DAG
+                    let tasks: Vec<_> = addrs_to_get
+                        .iter()
+                        .map(|a| client.spend_dag_crawl_from(*a, sender.clone()))
+                        .collect();
+                    let res = futures::future::join_all(tasks).await;
+                    let mut new_utxos = BTreeSet::new();
+                    for (r, a) in res.into_iter().zip(addrs_to_get) {
```
maqi: `spend_dag_crawl_from` shall return with `WalletResult<(SpendAddress, BTreeSet<SpendAddress>)>` i.e. `(addr_of_spend, utxos_of_spend)`,
instead of using `zip` here

sn_client/src/audit/dag_crawling.rs
```diff
@@ -62,26 +64,93 @@ impl Client {
...
+            info!("Finished verifying SpendDAG from {spend_addr:?} in {elapsed:?}");
+        }
+
+        Ok(dag)
+    }
+
+    /// Crawls the Spend Dag from a given SpendAddress recursively
+    /// following descendants all the way to UTXOs
+    /// Returns the UTXOs reached
+    pub async fn spend_dag_crawl_from(
```
maqi, line 157: shall be feature guarded as well?

sn_client/src/audit/dag_crawling.rs
```diff
@@ -62,26 +64,93 @@ impl Client {
         spend_processing: Option<Sender<SignedSpend>>,
         verify: bool,
     ) -> WalletResult<SpendDag> {
-        info!("Building spend DAG from {spend_addr:?}");
```
maqi, line 65: this function shall now be feature guarded?

sn_client/src/audit/dag_crawling.rs
```diff
@@ -62,26 +64,93 @@ impl Client {
...
-                return Ok(dag);
+                utxos.insert(spend_addr);
+                return Ok(utxos);
             }
             InternalGetNetworkSpend::Error(e) => {
                 return Err(WalletError::FailedToGetSpend(e.to_string()));
             }
         };
-        dag.insert(spend_addr, first_spend.clone());
+        spend_processing
```
maqi, line 185: not sure if we still need that `gen` stuff.

anyway, for beta_track optimisation, we don't need to track down utxos of royalty and forwarded payment
which, seems need to have a separate feature guarded function with this infrastructure.

sn_auditor/src/dag_db.rs
grumbach: Just making sure I understand your comment, beta_tracking audit works with or without DAG collection. This PR just makes collection optional:
- Collection/gathering = spends are stored in a local SpendDag
- Crawling = spends are not kept in a local SpendDag but we still explore the DAG and collect beta rewards

If DAG is not gathered locally there is nothing to dump, this is why I feature flagged the dump fn, but just removed that feature flag as it's not necessary indeed. 
I put `dag-collection` as a dependency to `svg-dag` as it is a prerequisite to it, we cannot make a svg if there is no local DAG.

sn_auditor/src/dag_db.rs
grumbach: Making an external function got rid of this issue.

sn_client/src/audit/dag_crawling.rs
grumbach: This is `sn_client` code, the feature guard is an `sn_auditor` option, the lib shall have no feature guard.

sn_client/src/audit/dag_crawling.rs
grumbach: Same as above, this is `sn_client`

sn_auditor/src/dag_db.rs
maqi: well, I am not sure what `features` to be used during release process for auditor.
maybe just `default` ? which shall be OK.
just a remind that `feature guard` change might affects.

sn_client/src/audit/dag_crawling.rs
grumbach: Agreed, `gen` is a relic from the past, we could get rid of it.

> seems need to have a separate feature guarded function with this infrastructure.

royalties/dag/beta rewards all can go through the same process receiving spends through the spend_processing channel. `sn_client` is lib code so it can stay agnostic IMO.

sn_auditor/src/dag_db.rs
grumbach: Yes agreed, we have to decide what's the default for sn_auditor:
- with local DAG
- without local DAG

As with this PR, the default becomes without, but we have to discuss it together yeah

sn_client/src/audit/dag_crawling.rs
```diff
@@ -62,26 +64,117 @@ impl Client {
...
+                return Err(WalletError::Dag(s));
+            }
+            let elapsed = start.elapsed();
+            info!("Finished verifying SpendDAG from {spend_addr:?} in {elapsed:?}");
+        }
+
+        Ok(dag)
+    }
+
+    /// Crawls the Spend Dag from a set of given SpendAddresses recursively
```
maqi, line 130: As I understand, for the beta_tracking purpose only, this function shall just:
1, tryto get the spends
2, if failed, put that spendaddress as utxo as well
3, if fetched, put the outputs as utxos
4, return all collected utxos as result
which is somehow a `breadth-first search` methodology, and was used by the PR 1777

Meanwhile, according to the comments and the inner function call `spend_dag_crawl_from` it uses, it seems will tryto drain along one branch first, then go to the next.
which is a `depth-first earch` methodology.

I am not sure which one will be better, but my general feeling is that the `wide first` one will be more suitable for the DAG tree we have.

Also, there is `tracking performance` statistics current not being populated in the PR.

I'd strongly suggest to build this auditor separately, then test it against a realy public testnet, to compare the performance with `main`, or even the `PR 1777`, before got it merged.

maqi (changes requested):

sn_auditor/src/dag_db.rs
```diff
@@ -249,41 +250,62 @@ impl SpendDagDb {
                 continue;
             }
 
-            // get a copy of the current DAG
-            let mut dag = { self.dag.clone().read().await.clone() };
+            let new_utxos = if cfg!(feature = "dag-collection") {
+                self.crawl_and_generate_local_dag(
```
maqi, line 254: the `tracking_performance` is only got updated within `beta_background_process_spend`,
which seems now out of sync with the actually background crawling track ?
will need to move the tracking to the correct place so that proper performance comparision can be undertaken.

maqi (changes requested):

maqi (approved):

PR #1816, opened 2024-06-03
feat(launchpad): limit node count to 50

RolandSherwin:
- limits the max node count to be 50

### DIFF
--- a/node-launchpad/src/components/manage_nodes.rs
+++ b/node-launchpad/src/components/manage_nodes.rs
@@ -24,6 +24,7 @@ use super::{utils::centered_rect_fixed, Component};
 pub const GB_PER_NODE: usize = 5;
 pub const MB: usize = 1000 * 1000;
 pub const GB: usize = MB * 1000;
+pub const MAX_NODE_COUNT: usize = 50;
 
 pub struct ManageNodes {
     /// Whether the component is active right now, capturing keystrokes + drawing things.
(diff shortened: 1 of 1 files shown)


node-launchpad/src/components/manage_nodes.rs
```diff
@@ -35,17 +36,18 @@ pub struct ManageNodes {
...
-            nodes_to_start_input: Input::default().with_value(allocated_space.to_string()),
+            nodes_to_start_input: Input::default().with_value(nodes_to_start.to_string()),
             old_value: Default::default(),
         };
         Ok(new)
     }
 
-    fn get_nodes_to_start(&self) -> usize {
+    fn get_nodes_to_start_val(&self) -> usize {
         self.nodes_to_start_input.value().parse().unwrap_or(0)
```
mazzi, line 51: Maybe `unwrap_or(1)`? If it was specified `0` ignore me.

node-launchpad/src/components/manage_nodes.rs
```diff
@@ -82,13 +90,9 @@ impl Component for ManageNodes {
...
                 let nodes_to_start_str = self.nodes_to_start_input.value().to_string();
-                let nodes_to_start = {
-                    let max_space_to_use = std::cmp::min(
-                        self.get_nodes_to_start() * GB_PER_NODE,
-                        self.available_disk_space_gb,
-                    );
-                    max_space_to_use / GB_PER_NODE
-                };
+                let nodes_to_start =
+                    std::cmp::min(self.get_nodes_to_start_val(), self.max_nodes_to_start());
```
mazzi, line 94: Maybe this should be also in `max_nodes_to_start`? Passing `get_nodes_to_start_val()` as a param?

node-launchpad/src/components/manage_nodes.rs
```diff
@@ -283,7 +290,7 @@ impl Component for ManageNodes {
         f.render_widget(info, layer_one[2]);
 
         // ==== help ====
-        let help = Paragraph::new("  Note: Each node will use a small amount of CPU\n  Memory and Network Bandwidth. We recommend\n  starting no more than 5 at a time.")
+        let help = Paragraph::new("  Note: Each node will use a small amount of CPU\n  Memory and Network Bandwidth. We recommend\n  starting no more than 5 at a time (max 50 nodes).")
```
mazzi: Harcoding. Use `MAX_NODE_COUNT` in the string.

node-launchpad/src/components/manage_nodes.rs
RolandSherwin: If we just stored numbers, we'd never unwrap here. 0 just so that we'd know something is up.

node-launchpad/src/components/manage_nodes.rs
RolandSherwin: The `max_nodes_to_start` is also used where we manually type in the `node_count`, i.e. press 0-9.
So, at that place we'd want to have `min(limit_based_on_disk_space, hard_coded_limit)`.
If we considered `get_nodes_to_start_val()` there, then we'd take in the old value there, which is wrong.

mazzi (approved):

PR #1820, opened 2024-06-03
feat(resources): add script to grab foundation cashnote from a server

joshuef:
enable to get the foundation transfer for use by the foundation


grumbach (approved):
LGTM!!

PR #1821, opened 2024-06-03
feat(launchpad): read stats from metrics endpoint

RolandSherwin:
- provide option to start metrics server during `safenode manager add`
- Run the metrics server on `localhost` instead of `0.0.0.0` for security.
- Store and expose the forwarded balance via the metrics endpoint.
    - We already have `wallet_balance` via the metrics endpoint, but it will get to 0 if the reward has been forwarded. So we need to instead track the forwarded rewards separately.
    - Also, to retain the` forwarded_balance` across restarts, we need to cache is locally. Else, people would see a value of 0 after a restart.
- Modify the device status pane in the launchpad to display the stats


joshuef:
It's not clear to me if this alone will result in us _not_ having metrics on new testnets. 

If that's the case I don't want this in until we _can_ have that (via checkbox or what have you at deploy time... default to yes in the workflow trigger)

mazzi (approved):
LGTM 🚀

PR #1823, opened 2024-06-04
chore: small tweak to get foundaiton cash

joshuef:


grumbach (approved):
LGTM! Although 60s for recrawl interval is going to be very heavy at scale

PR #1828, opened 2024-06-04
fix(transfer): mismatched key shall result in decryption error

maqi:
### Description

* add a check_sum pad to ensure mismatched key decryption shall trigger error, instead of a random hash
* add test to check mismatched key shall result in discord_name decryption error

### Related Issue

### Type of Change

### DIFF
--- a/sn_transfers/src/cashnotes/spend_reason.rs
+++ b/sn_transfers/src/cashnotes/spend_reason.rs
@@ -64,7 +64,10 @@ impl SpendReason {
 const MAX_CIPHER_SIZE: usize = std::u8::MAX as usize;
 const DERIVATION_INDEX_SIZE: usize = 32;
 const HASH_SIZE: usize = 32;
-const LIMIT_SIZE: usize = HASH_SIZE + DERIVATION_INDEX_SIZE;
+const CHECK_SUM_SIZE: usize = 8;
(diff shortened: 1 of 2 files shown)


sn_transfers/src/cashnotes/spend_reason.rs
```diff
@@ -89,24 +93,37 @@ impl DiscordName {
...
         let mut nonce_bytes = [0; DERIVATION_INDEX_SIZE];
-        nonce_bytes.copy_from_slice(&bytes[HASH_SIZE..LIMIT_SIZE]);
+        nonce_bytes.copy_from_slice(&bytes[HASH_SIZE..CONTENT_SIZE]);
         let nonce = DerivationIndex(nonce_bytes.to_owned());
-        Self { hash, nonce }
+
+        let mut checksum = [0; CHECK_SUM_SIZE];
+        checksum.copy_from_slice(&bytes[CONTENT_SIZE..LIMIT_SIZE]);
+        if checksum != CHECK_SUM {
+            return Err(TransferError::UserNameDecryptFailed);
```
grumbach: Since there's no decryption here, we can straight away report that the key is invalid?
```suggestion
            return Err(TransferError::InvalidDecryptionKey);
```

grumbach (approved):
LGTM

PR #1856, opened 2024-06-07
feat(networking): allow disabling relay server

b-zee:


sn_node/src/node.rs
```diff
@@ -95,6 +97,7 @@ impl NodeBuilder {
         root_dir: PathBuf,
         owner: Option<String>,
         #[cfg(feature = "upnp")] upnp: bool,
+        relay_server: bool,
```
RolandSherwin, line 100: I think this need not be passed as an arg to `NodeBuilder::new()`. Since this is a builder, we can add a function inside `NodeBuilder` to set the value. And then call it if necessary?

RolandSherwin: The same should've been done for `owner` arg :( can we create a commit for that as well if possible?

joshuef:
So I think f5475dfbe47541339e2be80a0c52d115ae211cd7 might be a better approach? Keep it in, but exclude bootstrap nodes? 

That's make for a more natural "discovery", while also removing the need for many different builds?

---------

If we do still want to disable bootstrapping, perhaps we can instead limit capacity? (And perhaps perhaps we need to blacklist repeat requesters? 🤷 )

Thoughts?

b-zee:
They are both two different ways of solving the problem of overloading the bootstrap nodes. I do think there is value in adding this feature though. Maybe there will be people running very CPU-constrained nodes and turning off the relay server might be a first sensible step. On the other hand, a random node shouldn't be totally overwhelmed by relay requests.

Maybe another thing to look at is being able to configure the amount of relay requests we accept and possibly limit it to only 2-4 peers that can be using us as relay at the same time. (For resource constrained nodes.)

joshuef:
> Maybe another thing to look at is being able to configure the amount of relay requests we accept and possibly limit it to only 2-4 peers that can be using us as relay at the same time. (For resource constrained nodes.)


I think this is the path we likely need to take. This way all nodes can act as relays (otherwise folk will opt-out; later we should enforce this)

Going to close this in favour of that approach for now.

PR #1858, opened 2024-06-07
fix(uploader): fix for mismatched ordering of instructions

RolandSherwin:
There can be a mismatch of ordering between the main loop and the make payment loop because
the instructions are sent via a task(channel.send().await). And there is no guarantee for the
order to come in the same order as they were sent.

We cannot just disobey the instruction inside the child loop, as the mainloop would be expecting
a result back for a particular instruction.

### DIFF
--- a/sn_client/src/uploader/upload.rs
+++ b/sn_client/src/uploader/upload.rs
@@ -824,6 +824,7 @@ impl InnerUploader {
             let mut cost_map = BTreeMap::new();
             let mut current_batch = vec![];
 
+            let mut got_a_previous_force_payment = false;
             while let Some(payment) = make_payment_receiver.recv().await {
                 let make_payments = if let Some((item, quote)) = payment {
                     let xorname = item.xorname();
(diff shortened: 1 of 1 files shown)


sn_client/src/uploader/upload.rs
```diff
@@ -824,22 +824,42 @@ impl InnerUploader {
...
                     let make_payments = !cost_map.is_empty();
-                    trace!("Got a forced forced round of make payment. make_payments: {make_payments:?}");
+                    trace!("Got a forced forced round of make payment.");
+                    // Note: There can be a mismatch of ordering between the main loop and the make payment loop because
+                    // the instructions are sent via a task(channel.send().await). And there is no guarantee for the
+                    // order to come in the same order as they were sent.
+                    //
+                    // We cannot just disobey the instruction inside the child loop, as the mainloop would be expecting
+                    // a result back for a particular instruction.
+                    if !make_payments {
```
maqi, line 846: use `if cost_map.is_empty() {` will be more readable ?

sn_client/src/uploader/upload.rs
```diff
@@ -824,22 +824,42 @@ impl InnerUploader {
...
+                        got_a_previous_force_payment = true;
+                        warn!(
+                            "We were told to force make payment, but cost_map is empty, so we can't do that just yet. Waiting for a task to insert a quote into cost_map"
```
maqi, line 849: the msg actually shall be:
```
We were told cost_map is empty, i.e. no more payment to be maid
```

sn_client/src/uploader/upload.rs
```diff
@@ -824,22 +824,42 @@ impl InnerUploader {
...
-                    trace!("Got a forced forced round of make payment. make_payments: {make_payments:?}");
+                    trace!("Got a forced forced round of make payment.");
+                    // Note: There can be a mismatch of ordering between the main loop and the make payment loop because
+                    // the instructions are sent via a task(channel.send().await). And there is no guarantee for the
+                    // order to come in the same order as they were sent.
+                    //
+                    // We cannot just disobey the instruction inside the child loop, as the mainloop would be expecting
+                    // a result back for a particular instruction.
+                    if !make_payments {
+                        got_a_previous_force_payment = true;
```
maqi, line 847: better rename the var to be `cost_map_was_emptied`

sn_client/src/uploader/upload.rs
```diff
@@ -824,22 +824,42 @@ impl InnerUploader {
...
+                        )
+                    }
+
                     make_payments
                 };
 
                 if make_payments {
+                    // reset force_make_payment
+                    if got_a_previous_force_payment {
+                        info!("A task inserted a quote into cost_map, so we can now make a forced round of payment!");
```
maqi, line 859: better change the comment msg to be:
```
New quote inserted into cost_map, after it was emptied, we shall make a forced round of payment
```

sn_client/src/uploader/upload.rs
```diff
@@ -824,22 +824,42 @@ impl InnerUploader {
...
                 let make_payments = if let Some((item, quote)) = payment {
                     let xorname = item.xorname();
                     trace!("Inserted {xorname:?} into cost_map");
 
                     current_batch.push((xorname, quote.clone()));
                     let _ = cost_map.insert(xorname, (quote.1, quote.2, quote.0.to_bytes()));
-                    cost_map.len() >= batch_size
+                    cost_map.len() >= batch_size || got_a_previous_force_payment
                 } else {
                     // using None to indicate as all paid.
```
maqi, line 837: better update the comment to be : 
```
When the received `payment` is `None`, it notifies no more payment to be made.
There shall be a round of `forced payment` to pay the remaining payments in the cost_map (less than batch_size)
With normal order, there shall be no more payment come in afterwards,
however it could be break due to async channel send.
Hence have to catering such `out-of-order` situation.
```

PR #1860, opened 2024-06-08
feat(auditor): recognize payment forwards with default key

maqi:
### Description

recognize payment forwards with default key

### Related Issue

### Type of Change

### DIFF
--- a/sn_auditor/src/dag_db.rs
+++ b/sn_auditor/src/dag_db.rs
@@ -14,7 +14,9 @@ use color_eyre::eyre::{bail, eyre, Result};
 use graphviz_rust::{cmd::Format, exec, parse, printer::PrinterContext};
 use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


sn_auditor/src/dag_db.rs
```diff
@@ -358,8 +360,24 @@ impl SpendDagDb {
                 .or_default()
                 .insert((addr, amount));
         } else {
+            // check with default key
+            if let Some(default_user_name_hash) =
+                spend.reason().get_sender_hash(&DEFAULT_PAYMENT_FORWARD_SK)
```
grumbach, line 365: We could reduce indentation and remove the return statement below with an `else if let`
```suggestion
        // check with default key
        } else if let Some(default_user_name_hash) =
                spend.reason().get_sender_hash(&DEFAULT_PAYMENT_FORWARD_SK)
```

And put the unknown participant block in a last else

grumbach (approved):
LGTM small style suggestion, nothing blocking

PR #1884, opened 2024-06-14
feat: add scripts for random uploads and wallet generation

joshuef:
Scripts to simplify wallet creation and random data upload

### DIFF
--- a/resources/scripts/make-wallets.sh
+++ b/resources/scripts/make-wallets.sh
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+# Function to print a message in a box
+print_in_box() {
+    local message="$1"
+    local border_length=$((${#message} + 4))
+    local border=$(printf '%*s' "$border_length" | tr ' ' '#')
+
+    echo "$border"
+    echo "# $message #"
+    echo "$border"
+}
+
+# Check for correct number of arguments
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <number_of_wallets> <amount_per_wallet>"
+    exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


resources/scripts/make-wallets.sh
```diff
@@ -0,0 +1,94 @@
...
+    local border=$(printf '%*s' "$border_length" | tr ' ' '#')
+
+    echo "$border"
+    echo "# $message #"
+    echo "$border"
+}
+
+# Check for correct number of arguments
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <number_of_wallets> <amount_per_wallet>"
```
maqi, line 16: if it requires arguments to be passed in,
better give an example of usage in the comment (head of this file, or in the readme correspondently?)

resources/scripts/make-wallets.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# Define directories relative to the current working directory
+APP_DIR=$(pwd)
+CLIENT_DIR="${APP_DIR}/client"
+NEW_CLIENT_DIR_TEMPLATE="${APP_DIR}/client_"
+
+# Initialize an array to store the wallet details
+declare -a WALLETS
+
+# Loop to create and fund wallets
```
maqi, line 32: I think the whole flow can be optimized to:
1, backup the wallet with balance
2, repeated create 5 wallets, collecting their address
3, restore the wallet with balance, transfer fund to new wallets one by one

Not a big issue, as anyway script do the job

maqi (approved):

PR #1889, opened 2024-06-14
Uploader optimisation

maqi:
This pull request primarily focuses on improving the `sn_client` module in the Rust codebase. The changes include refactoring and renaming methods in the `Client` implementation, adding new methods to handle and resend unconfirmed transactions, and improving log messages for better debugging and user experience. Additionally, a new unit test has been added in the `sn_transfers` module to test the payment selection functionality.

### Changes in `sn_client` module:

* `sn_client/src/api.rs`: 
  * The `trace!` log message in the `Client` implementation has been updated to include a pretty printed version of the record key for better debugging.
  * The `RetryStrategy` for `GetRecordCfg` in the `Client` implementation has been removed to avoid unnecessary retries.
  * The `crawl_spend_from_network` method has been renamed to `peek_a_spend` and its `get_quorum` value has been changed to `Quorum::One` to fetch a single copy of the spend.
  * The `is_genesis_spend_present` method now uses the `peek_a_spend` method to check for the presence of the Genesis spend.
* `sn_client/src/uploader/upload.rs`: 
  * The `InnerUploader` implementation now includes a call to `resend_pending_transaction_blocking_loop` method to handle any pending transactions.
* `sn_client/src/wallet.rs`: 
  * A new method `resend_pending_transaction_blocking_loop` has been added to the `WalletClient` implementation. This method continuously resends unconfirmed transactions until they are all uploaded or the user manually terminates the process. It also logs and prints helpful messages for the user to understand the state of unconfirmed transactions.

### Changes in `sn_transfers` module:

* `sn_transfers/src/wallet/api.rs`: 
  * A new unit test `payment_selective` has been added to test the functionality of selecting a payment transaction.

### DIFF
--- a/sn_client/src/wallet.rs
+++ b/sn_client/src/wallet.rs
@@ -669,6 +669,59 @@ impl WalletClient {
         }
     }
 
+    /// This is a blocking loop in cas there is pending transaction.
+    /// It will keeps resending the unconfirmed spend infinitely but explictly.
+    /// Function will only return on success (all unconfirmed spend uploaded),
+    /// or user chose to manualy, but safely, terminate the procedure.
+    pub async fn resend_pending_transaction_blocking_loop(&mut self) -> WalletResult<()> {
+        if !self.wallet.unconfirmed_spend_requests_exist() {
+            return Ok(());
+        }
+        // Wallet shall be all clear to progress forward.
+        while self.wallet.unconfirmed_spend_requests_exist() {
+            info!("Pre-Unconfirmed transactions dected, sending again after 30 seconds...");
+            println!("Pre-Unconfirmed transactions exist, sending again after 30 seconds...");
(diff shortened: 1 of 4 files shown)


sn_client/src/wallet.rs
```diff
@@ -669,6 +669,59 @@ impl WalletClient {
...
+                        );
+                    }
+                }
+            }
+
+            self.resend_pending_transactions(true).await;
+        }
+        info!("Wallet is now all cleared, OK to progress further.");
+        println!("Wallet is now all cleared, OK to progress further.");
+        println!("DO NOTE TERMINATE THE WORK MANUALLY !");
```
joshuef:
```suggestion
        eprintln!("WARNING: Closing the client now could corrupt the wallet.!");
```

joshuef: I i understand correctly, it's not we're out of the loop that it's the danger, so we want to alert for that, right?

joshuef (approved):

PR #1898, opened 2024-06-17
feat(networking): drop SwarmCmds when under load

joshuef:


sn_networking/src/cmd.rs
```diff
@@ -181,6 +181,39 @@ pub enum SwarmCmd {
...
+            SwarmCmd::DialWithOpts { .. } => true,
+            SwarmCmd::GetAllLocalPeers { .. } => false,
+            SwarmCmd::GetKBuckets { .. } => true,
+            SwarmCmd::GetClosestKLocalPeers { .. } => false,
+            SwarmCmd::GetClosestPeersToAddressFromNetwork { .. } => true,
+            SwarmCmd::GetCloseGroupLocalPeers { .. } => false,
+            SwarmCmd::GetSwarmLocalState { .. } => false,
+            SwarmCmd::SendRequest { .. } => true,
+            SwarmCmd::SendResponse { .. } => true,
+            SwarmCmd::RecordStoreHasKey { .. } => true,
```
RolandSherwin, line 198: `SwarmCmd::RecordStoreHasKey, GetNetworkRecord` are used during put_validation (when accpeting payments), we'd want these in probably, as we seem to prioritize PutRecord commands here.

PR #1904, opened 2024-06-18
fix(network): collect all spends and store the first two ordered spends

RolandSherwin:


sn_networking/src/event/kad.rs
```diff
@@ -402,9 +407,40 @@ impl SwarmDriver {
...
-                        .send(Err(GetRecordError::SplitRecord { result_map }))
-                        .map_err(|_| NetworkError::InternalMsgChannelDropped)?;
+                    let mut accumulated_spends = BTreeSet::new();
+                    for (record, _) in result_map.values() {
+                        match get_raw_signed_spends_from_record(record) {
+                            Ok(spends) => {
+                                accumulated_spends.extend(spends);
+                            }
+                            Err(_) => {
+                                break;
```
joshuef: should we break here? Or continue

sn_networking/src/event/kad.rs
```diff
@@ -402,9 +407,40 @@ impl SwarmDriver {
...
+                            }
+                        }
+                    }
+                    if !accumulated_spends.is_empty() {
+                        info!("For record {pretty_key:?} task {query_id:?}, found split record for a spend, accumulated and sending them as a single record");
+                        let accumulated_spends = accumulated_spends
+                            .into_iter()
+                            .take(2)
```
joshuef, line 417: why `2` ? why not more?

sn_networking/src/event/kad.rs
RolandSherwin: Yeah, I think we can just go with continue. Split record will not occur often.

sn_networking/src/event/kad.rs
RolandSherwin: Can we keep a random magic number here for now, say 50? If so, might have to do the same during PUT validation (we store 2 there).

The lower bound of a single signed spend is 825B. Can grow if the inputs/outputs increase.

sn_networking/src/event/kad.rs
```diff
@@ -395,9 +399,40 @@ impl SwarmDriver {
...
+                            value: bytes.to_vec(),
+                            publisher: None,
+                            expires: None,
+                        };
+                        sender
+                            .send(Ok(new_accumualted_record))
+                            .map_err(|_| NetworkError::InternalMsgChannelDropped)?;
+                    } else {
+                        sender
+                            .send(Err(GetRecordError::SplitRecord { result_map }))
```
maqi, line 433: if no entry within the accumulated_record,
it shall be considered as NoHolders, instead of `SplitRecord` ?

sn_networking/src/event/kad.rs
```diff
@@ -395,9 +399,40 @@ impl SwarmDriver {
...
+                        let bytes = try_serialize_record(&accumulated_spends, RecordKind::Spend)?;
+
+                        let new_accumualted_record = Record {
+                            key: peer_record.record.key,
+                            value: bytes.to_vec(),
+                            publisher: None,
+                            expires: None,
+                        };
+                        sender
+                            .send(Ok(new_accumualted_record))
```
maqi: accumualted -> accumulated
same to other places

sn_networking/src/event/kad.rs
RolandSherwin: So we have no idea if the record we're GET'ing is a spend here, we just try to read the header and collect them IF they're spends. If not, we assume it is a chunk/register and just return SplitRecordError

joshuef (approved):

PR #1909, opened 2024-06-19
fix: store multiple double spend attempts

RolandSherwin:


sn_client/src/audit/dag_crawling.rs
```diff
@@ -193,14 +194,27 @@ impl Client {
...
                     .await
                     .map_err(|e| WalletError::SpendProcessing(e.to_string()))?;
-                *s1
+                txs
+            }
+            InternalGetNetworkSpend::DoubleSpend(spends) => {
+                let mut txs = BTreeSet::new();
+                for spend in spends.into_iter() {
+                    txs.insert(spend.spend.spent_tx.clone());
+                    spend_processing
```
RolandSherwin, line 212: Earlier only the first spend was sent for spend_processing, but here we do it for all the spends, is this alright?

sn_client/src/audit/dag_crawling.rs
```diff
@@ -212,13 +226,8 @@ impl Client {
                 return Err(WalletError::FailedToGetSpend(e.to_string()));
             }
         };
-        spend_processing
-            .send(first_spend.clone())
-            .await
-            .map_err(|e| WalletError::SpendProcessing(e.to_string()))?;
 
         // use iteration instead of recursion to avoid stack overflow
```
b-zee, line 230: Comment not relevant anymore?

b-zee (approved):

sn_client/src/audit/dag_crawling.rs
RolandSherwin: I think it is for the one inside the for loop? I'm not sure.

PR #1912, opened 2024-06-20
refactor(network): remove unusued CmdOk type

RolandSherwin:
- The CmdOk was used when we were performing PUTs via `ReqResp` protocol instead of `kad` and thus can be removed.


sn_node/src/put_validation.rs
```diff
@@ -299,7 +302,7 @@ impl Node {
             None => {
                 // Notify replication_fetcher to mark the attempt as completed.
                 self.network.notify_fetch_completed(key.clone());
-                return Ok(CmdOk::DataAlreadyPresent);
+                return Ok(());
```
RolandSherwin, line 305: The trace log to say that the register is already stored, is printed by the `register_validation` function here.

b-zee (approved):

PR #1913, opened 2024-06-20
feat: add script for dag audit node migration

joshuef:
Simple bash script for syncing dag servers to avoid missing users

### DIFF
--- a/resources/scripts/dag-user-comparator.sh
+++ b/resources/scripts/dag-user-comparator.sh
@@ -1,0 +1,121 @@
+#!/bin/bash
+
+# Check if the correct number of arguments is provided
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 <source_server> <target_server>"
+    exit 1
+fi
+
+# Function to check if a command is available
+check_command() {
+    if ! command -v "$1" &> /dev/null; then
+        echo "Error: $1 is not installed."
+        case "$1" in
+            curl)
+                echo "Install curl using:"
+                echo "  sudo apt-get install curl    # Ubuntu/Debian"
+                echo "  sudo dnf install curl        # Fedora"
+                echo "  brew install curl            # MacOS"
+                ;;
+            jq)
+                echo "Install jq using:"
+                echo "  sudo apt-get install jq      # Ubuntu/Debian"
+                echo "  sudo dnf install jq          # Fedora"
+                echo "  brew install jq              # MacOS"
+                ;;
+            rg)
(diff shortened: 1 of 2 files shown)


resources/scripts/dag-user-comparator.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# Fetch the JSON data from the source URL
+echo "Fetching data from $SOURCE_URL"
+source_json=$(curl -s $SOURCE_URL)
+
+# Fetch the JSON data from the target URL
+echo "Fetching data from $TARGET_URL"
+target_json=$(curl -s $TARGET_URL)
+
+# Parse the JSON data to extract names, skipping unknown participants
```
mazzi, line 64: Not sure if we need to remove the `unknown participants`?
At least on `dags-service` I was asked to keep them (incoming change)

resources/scripts/dag-user-sync.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Construct URLs
+SOURCE_URL="http://$SOURCE_SERVER$SOURCE_PATH"
+TARGET_URL="http://$TARGET_SERVER$TARGET_PATH"
+
+# Fetch the JSON data from the source URL
+echo "Fetching data from $SOURCE_URL"
+json_data=$(curl -s $SOURCE_URL)
+
+# Parse the JSON data to extract names, skipping unknown participants
```
mazzi, line 25: Same.

mazzi:
Good one!
Must be so painful to write this on `bash` 😓

mazzi (approved):

resources/scripts/dag-user-sync.sh
joshuef: They'd show up again as soon as found on the new dag, so no need.

It's only the value for decryption that we need, otherwise it defaults to unknown participants (and I feat adding them is probably more prone to errors in this script, with braces, quotes etc. )


I'm not sure. This was supposed to be a quick fix for dag transition, whereas now we'd have the service 🤔 .

mazzi: There was a manual process to fill the missing users in the new DAG.

resources/scripts/dag-user-comparator.sh
joshuef: We cannot really add them. The added users are about decrypting these entries. So it doesn't amke sense to be adding this to the DAG.

PR #1914, opened 2024-06-23
feat: node registry refresh searches for pids

jacderida:
### Description

- 72caf9743 **chore: define network for node manager vm**

  After applying updates on Arch, my Vagrant and libvirt setup doesn't work like it did before. The
  network that Vagrant creates doesn't have internet access. So, the VM needs to be explicitly added
  to the `default` network, rather than the `vagrant-libvirt` network. Shared folders are also a more
  difficult situation now. So one of these were removed.

- 26baf6a4a **feat: node manager logs to stderr**

  A short-lived program like the node manager doesn't need to log to file. The logs can go straight to
  stderr, like most other programs do.

  Two global arguments are added to support this: `--debug` and `--trace`. Setting either of those
  flags will set the logging level to debug or trace, respectively. Normally you can also set an
  environment variable, but I'm leaving this out because there are all sorts of different crate scopes
  you can set with respect to our program.

  BREAKING CHANGE: the previous logging mechanism for logging to file was removed.

  We should also consider actually removing the `Stdout` variant from the `LogOutputDest` enum,
  because it's very non-standard to put logs there.

- 52a587edc **feat: node registry refresh searches for pids**

  The node registry refresh is changed to determine if each service process is running by searching
  for processes that have a binary running at a particular path. Since each node service has its own
  copy of the binary, it is uniquely identified by the location of that binary.

  The refresh is split into partial or full. In the partial case, we will only attempt to find the PID
  by searching for a process by its binary path. If it is found, we set the new PID for the service
  and set its status to `RUNNING`; if it's not found, we set the status to `STOPPED`. This mechanism
  should be robust against user interference, e.g., if they manually start/stop processes by whatever
  means. For a full refresh, we also connect to the node's RPC service to get information about
  connected peers and listeners.

  We will use partial refresh for commands like `start` and `stop`, to help us determine whether we
  should attempt to start/stop services that were already running or not. These commands will
  potentially operate over lots of services so we will avoid the RPC connections here.

  Full refresh will be used when the `status` command runs.

  Finally, I have also reassigned the connected peers data, and added/updated some of the log
  information which I found useful to have during my testing.

### Type of Change

- [ ] Documentation update

### DIFF
--- a/node-launchpad/src/components/home.rs
+++ b/node-launchpad/src/components/home.rs
@@ -95,8 +95,13 @@ impl Home {
         let now = Instant::now();
         debug!("Refreshing node registry states on startup");
         let mut node_registry = NodeRegistry::load(&get_node_registry_path()?)?;
-        sn_node_manager::refresh_node_registry(&mut node_registry, &ServiceController {}, false)
-            .await?;
+        sn_node_manager::refresh_node_registry(
+            &mut node_registry,
+            &ServiceController {},
+            false,
+            true,
+        )
+        .await?;
         node_registry.save()?;
         debug!("Node registry states refreshed in {:?}", now.elapsed());
         home.load_node_registry_and_update_states()?;
--- a/sn_logging/src/layers.rs
+++ b/sn_logging/src/layers.rs
@@ -118,6 +118,12 @@ impl TracingLayers {
                     .event_format(LogFormatter)
                     .boxed()
             }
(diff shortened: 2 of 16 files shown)


joshuef:
> We should also consider actually removing the Stdout variant from the LogOutputDest enum,
because it's very non-standard to put logs there.

Yeh, that would make sense 👍

joshuef (approved):

PR #1921, opened 2024-06-25
Reduce base traffic

maqi:
### Description

This PR contains two works trying to reduce the base traffic usage:
1, less frequent identify info resend
2, avoid replication too often to same target peer

### Related Issue

Fixes #<issue_number> (if applicable).

PR #1922, opened 2024-06-26
chore(node): enlength identify resend interval

joshuef:
Supersedes #1921 

adds in ADR template and identify one for this

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -881,7 +881,7 @@ impl SwarmDriver {
 
         let now = Instant::now();
         self.replication_targets
-            .retain(|_peer_id, timestamp| *timestamp < now);
+            .retain(|_peer_id, timestamp| *timestamp > now);
         // Only carry out replication to peer that not replicated to it recently
         replicate_targets.retain(|peer_id| !self.replication_targets.contains_key(peer_id));
         if replicate_targets.is_empty() {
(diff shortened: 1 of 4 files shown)


sn_networking/src/driver.rs
```diff
@@ -109,6 +109,9 @@ const REQUEST_TIMEOUT_DEFAULT_S: Duration = Duration::from_secs(30);
 // Sets the keep-alive timeout of idle connections.
 const CONNECTION_KEEP_ALIVE_TIMEOUT: Duration = Duration::from_secs(30);
 
+// Inverval of resending identify to connected peers.
+const RESEND_IDENTIFY_INVERVAL: Duration = Duration::from_secs(3600);
```
maqi, line 113: the ADR says reduced to two hour interval ?

sn_networking/src/cmd.rs
```diff
@@ -858,6 +865,17 @@ impl SwarmDriver {
     }
 
     fn try_interval_replication(&mut self) -> Result<()> {
+        let time_since_last_replication = self.last_replication_time.elapsed();
+        if time_since_last_replication < REPLICATION_THROTTLE_INTERVAL {
```
maqi: the `per target` timeout is 45s, meanwhile this `throttle interval` is 60s,
which means `per target` becomes useless, as totally `covered` by this `throttle interval`.

I think this `over all throttle` is an overshot, and we only need `per target` level timeout,
which allows a new node to get updated asap, meanwhile retain the `low frequency of replication`

maqi (changes requested):

maqi (approved):

PR #1926, opened 2024-06-27
test(node): spend simulation test

RolandSherwin:
- attempts to perform spends / double spends / poison old spends and tries to verify all the states at the end.

### DIFF
--- a/.github/workflows/nightly_wan.yml
+++ b/.github/workflows/nightly_wan.yml
@@ -169,8 +169,8 @@ jobs:
         continue-on-error: true
 
       - name: Build testing executable
-        run: cargo test --release -p sn_node --test sequential_transfers --test storage_payments --test double_spend --no-run
-        timeout-minutes: 30
+        run: cargo test --release -p sn_node --features=local-discovery --test sequential_transfers --test storage_payments --test double_spend --test spend_simulation --no-run
+        timeout-minutes: 40
 
       - name: Start a WAN network
         uses: maidsafe/sn-testnet-action@main
@@ -199,7 +199,7 @@ jobs:
           echo "Peer is $SAFE_PEERS"
           echo "Deployment inventory is $SN_INVENTORY"
 
-      - name: execute the spend test
+      - name: execute the sequential transfers test
         run: cargo test --release -p sn_node --test sequential_transfers -- --nocapture --test-threads=1
         env:
           SN_LOG: "all"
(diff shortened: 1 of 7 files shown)


.github/workflows/nightly_wan.yml
```diff
@@ -1,8 +1,9 @@
 name: Nightly -- Full WAN Network Tests
 
 on:
-  schedule:
-    - cron: "0 0 * * *"
+# todo: this is totally broken atm. Fix and re-enable.
+#   schedule:
+#     - cron: "0 0 * * *"
```
grumbach, line 6: Anyone we should tell this about? @jacderida?

sn_node/tests/spend_simulation.rs
```diff
@@ -0,0 +1,697 @@
...
+            random_tx.hash()
+        );
+
+        return Ok(Some(random_tx.clone()));
+    }
+    Ok(None)
+}
+
+/// A spend / transaction is poisonable if all of its outputs are already spent.
+fn find_all_poisonable_spends<'a>(
```
grumbach, line 671: Actually, only one spent output is enough!

sn_node/tests/spend_simulation.rs
```diff
@@ -0,0 +1,697 @@
...
+        received_cash_note: Vec<CashNote>,
+    },
+}
+
+#[derive(Debug)]
+enum SpendStatus {
+    Utxo,
+    Spent,
+    Poisoned,
+    // DoubleSpend,
```
maqi: shall this to be removed?

sn_node/tests/spend_simulation.rs
```diff
@@ -0,0 +1,697 @@
...
+}
+
+#[derive(custom_debug::Debug)]
+struct State {
+    // immutable
+    #[debug(skip)]
+    action_senders: BTreeMap<usize, mpsc::Sender<WalletAction>>,
+    all_wallets: BTreeMap<usize, TempDir>,
+    main_pubkeys: BTreeMap<usize, MainPubkey>,
+    main_pubkeys_inverse: BTreeMap<MainPubkey, usize>,
```
maqi: I think some more detailed comments to explain what those parameters are for will be usefult.
for example, why need this extra `inverse` ?

sn_node/tests/spend_simulation.rs
```diff
@@ -0,0 +1,697 @@
...
+}
+
+#[derive(Debug, Default)]
+struct PendingTasksTracker {
+    pending_send_results: Vec<usize>,
+    pending_receive_results: Vec<usize>,
+}
+
+#[tokio::test]
+async fn cash_note_transfer_double_spend_fail() -> Result<()> {
```
maqi: may need a more detailed comment to explain: what's the detailed scenary of the test, and the expected result .
some general description shall be enough.

sn_node/tests/spend_simulation.rs
RolandSherwin: If an output is unspent, then that output cannot be spent later right. 
I wanted to address `poison`= even after a double spend, the entire branch is valid. i.e, any descendent can be spent.

sn_node/tests/spend_simulation.rs
RolandSherwin: Will do that yeah!

grumbach (approved):
LGTM!

PR #1927, opened 2024-06-28
test(node): consecutive double spends should be tracked

RolandSherwin:
Consider valid spends, A -> B -> C
Now if A is double spent and then B is double spent, then C should not be valid because B is a double spend.
But this does not happen because we reject double spend of B as its parent was invalid.

### DIFF
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -9,7 +9,8 @@
 use crate::{node::Node, quote::verify_quote_for_storecost, Error, Marker, Result};
 use libp2p::kad::{Record, RecordKey};
 use sn_networking::{
-    get_raw_signed_spends_from_record, GetRecordError, NetworkError, MAX_PACKET_SIZE,
+    get_raw_signed_spends_from_record, GetRecordError, NetworkError, SpendVerificationOk,
+    MAX_PACKET_SIZE,
 };
 use sn_protocol::{
     storage::{
@@ -705,6 +706,7 @@ impl Node {
             }
         };
 
+        let mut parent_is_a_double_spend = false;
         // check the received spends and the spends got from the network
         let mut tasks = JoinSet::new();
         for s in signed_spends.into_iter().chain(network_spends.into_iter()) {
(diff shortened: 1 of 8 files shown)


sn_node/src/put_validation.rs
```diff
@@ -735,6 +741,15 @@ impl Node {
             }
         }
 
+        if parent_is_a_double_spend && all_verified_spends.len() == 1 {
+            warn!("Parent is a double spend for {unique_pubkey:?}, ignoring this spend");
+            return Err(Error::InvalidRequest(format!(
+                "Parent is a double spend for {unique_pubkey:?}, ignoring this spend."
+            )));
```
grumbach: Maybe `InvalidRequest` is not exactly the correct error type here?
We have `TransferError::InvalidParentSpend` would that be better?

sn_node/tests/double_spend.rs
```diff
@@ -267,3 +270,134 @@ async fn poisoning_old_spend_should_not_affect_descendant() -> Result<()> {
...
+
+    info!("Verifying the original cashnote of A -> B");
+    let result = client.verify_cashnote(&cash_notes_for_b[0]).await;
+    info!("Got result while verifying the original spend from A -> B: {result:?}");
+    assert!(result.is_err());
+
+    info!("Verifying the original cashnote of B -> C");
+    let result = client.verify_cashnote(&cash_notes_for_c[0]).await;
+    info!("Got result while verifying the original spend from B -> C: {result:?}");
+    assert!(result.is_err());
```
grumbach: Maybe it's better if we specify the exact error type we're expecting so we're sure why we error out?

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -150,31 +151,63 @@ impl SignedSpend {
...
+                .iter()
+                .map(|p| *p.unique_pubkey())
+                .collect::<BTreeSet<_>>();
+            if parent_unique_key.len() > 1 {
+                error!("While verifying parents of {unique_key}, found a parent double spend, but it contained more than one unique_pubkey. This is invalid. Erroring out.");
+                return Err(TransferError::InvalidParentSpend("Invalid parent double spend. More than one unique_pubkey in the parent double spend.".to_string()));
+            }
+
+            // if parent is a double spend, get the actual parent among the parent double spends
+            // We cannot have more than 1 parent double spend with same tx hash.
```
grumbach: > We cannot have more than 1 parent double spend with same tx hash.

That assumption is wrong, one could change a single bit elsewhere (like in reason/purpose or any other metadata) in the spend and sign it and this would create a perfectly identical tx in a double spend pair.

grumbach (approved):
Couple of small comments, but generally LGTM

sn_node/src/put_validation.rs
RolandSherwin: Yeah! That would be perfect.

sn_node/tests/double_spend.rs
RolandSherwin: Right yeah! Will do that.

sn_transfers/src/cashnotes/signed_spend.rs
RolandSherwin: Yeah that makes sense.
But the code should be fine here ig? The is just used to validate the child of a proven parent double spend.
This child is either rejected / stored as double spend. So we're fine here, I suppose. I'll remove the comment.

grumbach: Yes if both double spend parents are to the same descendant then yeah, it should not be a big deal indeed for our verification

PR #1929, opened 2024-07-01
feat(node): check for double payment from client

b-zee:


sn_node/src/put_validation.rs
```diff
@@ -503,10 +503,21 @@ impl Node {
...
+        let (received_fee, mut cash_notes, royalties_cash_notes_r) = self
             .cash_notes_from_transfers(payment.transfers, &wallet, pretty_key.clone())
             .await?;
 
+        // check for spends that we have already received
+        cash_notes.retain(|cash_note| {
+            let spend_addr = SpendAddress::from_unique_pubkey(&cash_note.unique_pubkey());
+            let double_spend = matches!(wallet.get_confirmed_spend(spend_addr), Ok(Some(_spend)));
+            if double_spend {
+                warn!("Double spend detected for record {pretty_key} with unique key {}", cash_note.unique_pubkey());
```
maqi:
```suggestion
                warn!("Double spend {} detected for record payment {pretty_key}", cash_note.unique_pubkey());
```

sn_node/src/put_validation.rs
```diff
@@ -503,10 +503,24 @@ impl Node {
...
             .cash_notes_from_transfers(payment.transfers, &wallet, pretty_key.clone())
             .await?;
 
+        // check for spends that we have already received
+        cash_notes.retain(|cash_note| {
+            let spend_addr = SpendAddress::from_unique_pubkey(&cash_note.unique_pubkey());
+            let double_spend = matches!(wallet.get_confirmed_spend(spend_addr), Ok(Some(_spend)));
+            if double_spend {
+                warn!(
+                    "Double spend detected for record {pretty_key} with unique key {}",
```
maqi:
```suggestion
                    "Double spend {} detected for record payment {pretty_key}"
```

maqi (approved):

sn_node/src/put_validation.rs
```diff
@@ -503,10 +503,24 @@ impl Node {
...
+            if double_spend {
+                warn!(
+                    "Double spend {} detected for record payment {pretty_key}",
+                    cash_note.unique_pubkey()
+                );
+            }
+            // retain the `CashNote` if it's not a double spend
+            !double_spend
+        });
+
```
maqi, line 528: ah, if `cash_notes` becomes `empty` after the filter,
then there is actually no point to undertake the following steps.
so we shall have something like
```rust
if cash_notes.is_empty() {
    info!("All incoming cash_notes are already processed, no need for the further process!");
    return Err(Error::ReusedPayment);
}
```

maqi (changes requested):

maqi (approved):

PR #1942, opened 2024-07-05
fix(network): validate spends even if we get did not get the majority

RolandSherwin:
- We were not considering the records that were returned from the network if the quorum was not fulfilled. But these should be considered during validation.

What the PR buys us is that:
Quickly consolidate double spends attempts and store them in nodes. Clients were already safe, but nodes can possibly be out of sync for a while and would wait for replication to get them to sync. So it just reduces 1 step for the nodes

- Also run spend simulation as its own CI run. The local network was getting overwhelemed.

### DIFF
--- a/sn_networking/src/transfers.rs
+++ b/sn_networking/src/transfers.rs
@@ -33,6 +33,7 @@ impl Network {
         let get_cfg = GetRecordCfg {
             get_quorum: Quorum::Majority,
             retry_strategy: None,
+            // This should not be set.
             target_record: None,
             expected_holders: Default::default(),
         };
--- a/sn_networking/src/transfers.rs
+++ b/sn_networking/src/transfers.rs
@@ -33,7 +33,9 @@ impl Network {
         let get_cfg = GetRecordCfg {
             get_quorum: Quorum::Majority,
             retry_strategy: None,
-            // This should not be set.
+            // This should not be set here. This function is used as a quick check to find the spends around the key during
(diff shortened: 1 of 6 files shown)


sn_node/src/put_validation.rs
```diff
@@ -700,6 +700,23 @@ impl Node {
                 }
                 spends
             }
+            Err(NetworkError::GetRecordError(GetRecordError::NotEnoughCopies {
```
maqi, line 703: I am not sure, but this sounds bit tricky.
a cheaty node can easily provide one copy of a shaky spend, and this means any query attempt to that spend will replied with `NotEnoughCopies`, then be put into further handling.
This actually sounds like an `attack window` opened up ?

maqi (changes requested):

sn_networking/src/transfers.rs
```diff
@@ -33,6 +33,7 @@ impl Network {
         let get_cfg = GetRecordCfg {
             get_quorum: Quorum::Majority,
             retry_strategy: None,
+            // This should not be set.
```
grumbach: Why? Maybe a longer comment would help?

sn_node/src/put_validation.rs
grumbach: Yeah, Roland, when testing locally, do you often see the following log line??

maqi (approved):

PR #1945, opened 2024-07-05
feat: double spend spam protection

grumbach:
### Description

- protects from spend shadowing through spamming attack
- now Network Spend entries can only ever contain 2 entries max, chosen deterministically so all nodes are eventually consistent and there are NO FORKS in the knowledge, and less space is used!
- small cleanup of spend verification API

### The attack

An attack to hide a good spend. Say we have:
A->B->C->D
The attacker wants to get rid of B to poison C and D. They can spam the network with lots of double spends for B until none of the nodes have the original one. The attack is quite intricate and we discussed different alternatives with 2 spends as max for a spend record or unlimited (record size) as max. In both cases the attack was successful.
The solution was to add a descendency check in the case where we have more than 2 spends to choose which to keep:

  -  spend arrives from put
  -  check local kad (assume those are correct)
  -  ask peers for what they have at this addr (verify those)
   - then verify the incoming put ones
   - If we have more than 2 spends in total after all these checks, order by
       - Do they have valid descendants?
       - determistic Hash (order in BTreeSet)
   - And keep 2 of them only
       - If only one has a valid descendency, we successfully saved the good branch
       - If two or more do, we keep two, both become invalid, and the others orphans (invalid too as a result)
       - If none do, the line stops here and that’s what we want

### DIFF
--- a/sn_node/src/put_validation.rs
+++ b/sn_node/src/put_validation.rs
@@ -660,7 +660,6 @@ impl Node {
 
         // trust local spends as we've verified them before
         let local_spends = self.get_local_spends(spend_addr).await?;
-        let mut all_verified_spends = BTreeSet::from_iter(local_spends.into_iter());
 
         // get spends from the network at the address for that unique pubkey
         let network_spends = match self.network().get_raw_spends(spend_addr).await {
@@ -698,8 +697,16 @@ impl Node {
                 vec![]
             }
         };
+        debug!(
+            "For {unique_pubkey:?} got {} local spends, {} from network and {} provided",
+            local_spends.len(),
+            network_spends.len(),
+            signed_spends.len()
+        );
+        debug!("Local spends {local_spends:?}; from network {network_spends:?}; provided {signed_spends:?}");
 
         // only verify spends we don't know of
+        let mut all_verified_spends = BTreeSet::from_iter(local_spends.into_iter());
         let unverified_spends =
             BTreeSet::from_iter(network_spends.into_iter().chain(signed_spends.into_iter()));
         let known_spends = all_verified_spends.clone();
@@ -717,7 +724,7 @@ impl Node {
         }
 
         // gather verified spends
-        let mut double_spent_parent = vec![];
+        let mut double_spent_parent = BTreeSet::new();
         while let Some(res) = tasks.join_next().await {
             match res {
                 Ok((spend, Ok(()))) => {
@@ -726,7 +733,7 @@ impl Node {
                 }
                 Ok((spend, Err(NetworkError::Transfer(TransferError::DoubleSpentParent)))) => {
                     warn!("Parent of {spend:?} was double spent, keeping aside in case we're a double spend as well");
-                    double_spent_parent.push(spend.clone());
+                    let _ = double_spent_parent.insert(spend.clone());
                 }
                 Ok((spend, Err(e))) => {
                     // an error here most probably means the received spend is invalid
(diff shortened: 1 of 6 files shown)


sn_node/src/put_validation.rs
```diff
@@ -28,12 +25,6 @@ use std::collections::BTreeSet;
 use tokio::task::JoinSet;
 use xor_name::XorName;
 
-/// The maximum number of double spend attempts to store that we got from PUTs
-const MAX_DOUBLE_SPEND_ATTEMPTS_TO_KEEP_FROM_PUTS: usize = 15;
-
-/// The maximum number of double spend attempts to store inside a record
-const MAX_DOUBLE_SPEND_ATTEMPTS_TO_KEEP_PER_RECORD: usize = 30;
-
```
grumbach, line 36: The MAX is now 2 as this measure protected us until the attackers reached the MAX double spent attempt to keep!

sn_node/src/put_validation.rs
```diff
@@ -758,33 +741,109 @@ impl Node {
...
+        let mut live_spends = BTreeSet::new();
+        while let Some(res) = tasks.join_next().await {
+            match res {
+                Ok((spend, Ok(_descendant))) => {
+                    let _inserted = live_spends.insert(spend);
+                    if live_spends.len() > 1 {
+                        let array: Vec<_> = live_spends.clone().into_iter().collect();
+                        if let [one, two] = array.as_slice() {
+                            warn!("Got two live spends {one:?} and {two:?}, things are messed up!");
+                            return Ok((one.to_owned().clone(), two.to_owned().clone()));
```
RolandSherwin: If we return early here, will we have an inconsistent state among our nodes if there are more than 2 live branches? 
Having >1 live branch is very unlikely, but should we collect all live branches and then return the first 2, just to be deterministic?

sn_node/src/put_validation.rs
```diff
@@ -758,33 +741,109 @@ impl Node {
...
+                        }
+                    }
+                }
+                Ok((spend, Err(NetworkError::GetRecordError(GetRecordError::RecordNotFound)))) => {
+                    trace!("Spend {spend:?} descendant was not found, continuing...");
+                }
+                Ok((spend, Err(e))) => {
```
RolandSherwin, line 812: Should we handle `GerRecordError::NotEnoughRecords`? An output could've just been spent and maybe did not have enough time to get to the majority nodes?

sn_node/src/put_validation.rs
grumbach: You are absolutely right!

sn_node/src/put_validation.rs
grumbach: Mmmh true, but if the parent is double spent, it should never become enough records as other nodes will reject it. I'd say we can safely ignore this one, what do you think?

RolandSherwin: I was thinking about this situation where we have:
first spend: A -> B -> C utxo
double spend: A -> X utxo

Consider a bad actor spends A and sends it to B, and B being a good wallet sends it to itself (C). But before C got to enough nodes, the bad actor double spends A. 

Now, while checking, we will have 2 spends for A (double spend) and the live branch should be the first one. But if we don't consider the NotEnoughCopies err of B, then we will not consider it as a live branch.

And we might lose the live chain if: the bad actor spams double spend of A, so every node will store the first two `not_live_spends` which might not contain the A-B-C chain.

RolandSherwin: Update: As discussed in slack, we'll be blindly trusting B to be valid even if 1 copy exists. That will open up more attacks. Resolving this.

RolandSherwin (approved):

Issue #1944, opened 2024-07-05
feat(wallet): support for multiple wallets in the client


mickvandijke:
It would be nice to be able to store and switch between multiple wallets using the safe client.

Perhaps the `safe/client/` folder structure should be changed from:

```
..client_files
wallet/
   ..wallet_files
..
```

to:

```
..client_files
wallets/
   WALLET_PUB_KEY/
      ..wallet_files
..
```

After reading https://github.com/maidsafe/safe_network/issues/1752, we could perhaps even go further and change it to also support different networks:

```
..client_files
wallets/
   WALLET_PUB_KEY/
      NETWORK/
         ..wallet_files
..
```

PR #1949, opened 2024-07-08
feat(wallet): wallet secret key file encryption

### LINKED ISSUE #1943: enhancement(wallet): improve wallet private key storage security
Currently, the private key for a user's wallet is stored in `PLATFORM_SPECIFIC_FOLDER/safe/client/wallet/main_secret_key` as plain text, without any encryption. 

This is dangerous in case that someone gets (unwanted) access to a user's computer/data. An attacker could easily steal the user's private key by copying the `main_secret_key` file.

### 1. Store the private key and mnemonic as encrypted with a password and salt

Use something like [pbkdf2](https://docs.rs/pbkdf2/latest/pbkdf2) combined with aead, which uses a salt and password to encrypt/decrypt the private key and only store it as encrypted in the `main_secret_key` file. In the case that a user does not want to set a password for their wallet, the private key would still be encrypted but with a password set to `""`.

--- END Sol. 1 ---

### 2. Derive the private key on-demand from mnemonic using a password

Derive the private key of a wallet on-demand from the mnemonic using a user supplied password (instead of a default one: https://github.com/maidsafe/safe_network/blob/0ae7c0fb712bc0adfee6ba118467be5ee2ff857d/sn_client/src/acc_packet.rs#L16). This means that a password can never be wrong, but it will just lead to a different wallet.

This solution would work better with multi-wallet support. As of now, you can only have one active wallet in `PLATFORM_SPECIFIC_FOLDER/safe/client/wallet`. As different passwords lead to different wallets, there needs to be a way to store wallet files per public key. **Not entirely sure about this, please correct me if I'm wrong.**

Requires:
- https://github.com/maidsafe/safe_network/issues/1944

--- END Sol. 2 ---

mickvandijke:
### Description

Adds an option to save the wallet private key as encrypted with a password to a file called `main_secret_key.encrypted`, inside the wallet directory.

The encrypted secret key file will look like this: 

```json
{
    "encrypted_secret_key": "bdbc29f1e0c3d2211ccfa78c262bd3f42d71235f13b31bab339c682fc9e56d039302870c1f69a004c80c6691b62d51e2",
    "salt": "ee69ccf0008631ac",
    "nonce": "d8a46c7ccef48113f10a6cc1"
}
```

Relevant encryption logic:

```rust
/// Encrypted secret key for storing on disk and decrypting with password
#[derive(Serialize, Deserialize)]
pub(crate) struct EncryptedSecretKey {
    encrypted_secret_key: String,
    pub salt: String,
    pub nonce: String,
}

...

/// Encrypts secret key using pbkdf2 with HMAC<Sha512>.
pub(crate) fn encrypt_secret_key(
    secret_key: &MainSecretKey,
    password: &str,
) -> Result<EncryptedSecretKey> {
    // Generate a random salt
    // Salt is used to ensure unique derived keys even for identical passwords
    let mut salt = [0u8; 8];
    rand::thread_rng().fill(&mut salt);

    // Generate a random nonce
    // Nonce is used to ensure unique encryption outputs even for identical inputs
    let mut nonce = [0u8; 12];
    rand::thread_rng().fill(&mut nonce);

    let mut key = [0; 32];

    // Derive a key from the password using PBKDF2 with HMAC<Sha512>
    // PBKDF2 is used for key derivation to mitigate brute-force attacks by making key derivation computationally expensive
    // HMAC<Sha512> is used as the pseudorandom function for its security properties
    ring::pbkdf2::derive(
        ring::pbkdf2::PBKDF2_HMAC_SHA512,
        NonZeroU32::new(ITERATIONS).unwrap(),
        &salt,
        password.as_bytes(),
        &mut key,
    );

    // Create an unbound key using CHACHA20_POLY1305 algorithm
    // CHACHA20_POLY1305 is a fast and secure AEAD (Authenticated Encryption with Associated Data) algorithm
    let unbound_key = ring::aead::UnboundKey::new(&ring::aead::CHACHA20_POLY1305, &key)
        .map_err(|_| Error::FailedToEncryptKey(String::from("Could not create unbound key.")))?;

    // Create a sealing key with the unbound key and nonce
    let mut sealing_key = ring::aead::SealingKey::new(unbound_key, NonceSeq(nonce));
    let aad = ring::aead::Aad::from(&[]);

    // Convert the secret key to bytes
    let secret_key_bytes = secret_key.to_bytes();
    let mut encrypted_secret_key = secret_key_bytes;

    // seal_in_place_append_tag encrypts the data and appends an authentication tag to ensure data integrity
    sealing_key
        .seal_in_place_append_tag(aad, &mut encrypted_secret_key)
        .map_err(|_| Error::FailedToEncryptKey(String::from("Could not seal sealing key.")))?;

    // Return the encrypted secret key along with salt and nonce encoded as hex strings
    Ok(EncryptedSecretKey {
        encrypted_secret_key: encode(encrypted_secret_key),
        salt: encode(salt),
        nonce: encode(nonce),
    })
}
```

## CLI changes

`wallet create` command now prompts the user if they want to encrypt their new wallet with a password. Also, the following flags have been added:

| Flag | Description |
| - | - |
| --no-password | Skip the set password prompt. |
| --password [PASSWORD] | Provide a password to encrypt the wallet with. |

Added `wallet encrypt` command to encrypt an existing unencrypted wallet with a password.

## Other changes

- CLI commands that use an encrypted wallet, now request a user to provide the password to unlock the wallet.

- The hotwallet stash function now uses a different folder name for saving old wallets. Instead of using the wallet public address: `wallet_PUBLIC_ADDRESS`, it now uses: `wallet_CURRENT_DATETIME`. This change is to support stashing encrypted wallets, without having to decrypt them to get the public address.

### TODO List

- [x] Update CLI flow to optionally encrypt a wallet private key file with a user supplied password
- [x] CLI should prompt the user for a password if the wallet private key file is encrypted

### Related Issue

Fixes #1943. Solution 1.

### Type of Change

- [ ] Documentation update

### Checklist

- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.


mickvandijke:
Should the CLI changes perhaps be added in a separate PR? Else this one might get big.

mickvandijke:
@grumbach @joshuef Should the CLI changes in the TODO list perhaps be added in a separate PR? Else this one might get big.

grumbach:
Yeah include them in! It makes sense in your case!

mickvandijke:
[aa9b547](https://github.com/maidsafe/safe_network/pull/1949/commits/aa9b547a37082f12eaf95765e59be65fd9890657) makes sure that the CLI prompts the user for a password when an encrypted wallet requires it for authentication.

mickvandijke:
PR is ready for review.

sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -84,6 +100,46 @@ impl HotWallet {
...
+        }
+
+        let wallet_key = Self::load_from(root_dir)?.key;
+        let wallet_dir = root_dir.join(WALLET_DIR_NAME);
+
+        // Delete the unencrypted secret key file
+        delete_main_secret_key(&wallet_dir)?;
+
+        // Save the secret key as an encrypted file
+        store_main_secret_key(&wallet_dir, &wallet_key, Some(password.to_owned()))?;
```
grumbach: I would make sure the key is stored successfully before deleting the old one. Else what happens if the `store_main_secret_key` errors out?

sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -200,12 +268,11 @@ impl HotWallet {
...
-    /// to directory root_dir/wallet_ADDRESS
+    /// to directory root_dir/wallet_DATETIME
     pub fn stash(root_dir: &Path) -> Result<PathBuf> {
-        let wallet = HotWallet::load_from(root_dir)?;
         let wallet_dir = root_dir.join(WALLET_DIR_NAME);
-        let addr_hex = &format!("{:?}", wallet.address());
-        let new_name = format!("{WALLET_DIR_NAME}_{addr_hex}");
+        let datetime_str = Local::now().format("%Y-%m-%d_%H-%M-%S").to_string();
+        let new_name = format!("{WALLET_DIR_NAME}_{datetime_str}");
         let moved_dir = root_dir.join(new_name);
```
grumbach: Why switch to date/time?

mickvandijke: Makes it a lot easier to stash encrypted wallets without needing to decrypt them to get the public address.

sn_transfers/src/wallet/authentication.rs
```diff
@@ -0,0 +1,108 @@
...
+pub struct AuthenticationManager {
+    /// Password to decrypt the wallet.
+    /// Wrapped in Secret<> so that it doesn't accidentally get exposed
+    password: Option<Secret<String>>,
+    /// Expiry time of the password.
+    /// Has to be provided by the user again after a certain amount of time
+    password_expires_at: Option<DateTime<Utc>>,
+    /// Path to the root directory of the wallet
+    wallet_dir: PathBuf,
+}
```
grumbach, line 28: I think the wallet ends up in a file dumped on disk and this struct is a field of wallet. 
So if the `password` is `Some`, it ends up dumped to the disk in clear too. 
Even after expiry, an attacker could read the password by `cat`'ing the wallet file. 
Or did I miss something?

grumbach (commented):
Excellent work! Very clean code, love this! 🥇 
Left some comments!

sn_transfers/src/wallet/hot_wallet.rs
mickvandijke: Ah yeah good call, that would be quite a catastrophic failure. I'll swap the order of actions around, so that `store_main_secret_key` is done first.

sn_transfers/src/wallet/hot_wallet.rs
grumbach: > without needing to decrypt them to get the public address.

Not sure I understand, how can you get the public address if it's not in the name anymore?

mickvandijke: The public address can be derived from the secret key. The problem with encrypted wallets is that the secret key is encrypted and needs a password to reveal it.

If we look at the old code:

```rust
let wallet = HotWallet::load_from(root_dir)?; // Would fail if the wallet is encrypted
let addr_hex = &format!("{:?}", wallet.address()); // Needs wallet
```

To fix that, we would have to something like this:

```rust
let wallet = HotWallet::load_encrypted_from_path(root_dir, PASSWORD)?;
let addr_hex = &format!("{:?}", wallet.address());
```

But then the user needs to know and provide the correct password before an encrypted wallet can be stashed. Since this command is only used when creating a new wallet and replacing the old one, I think it makes more sense from a user perspective to not require the password of an encrypted wallet to stash it.

sn_transfers/src/wallet/authentication.rs
mickvandijke: Isn't the wallet file only a watch-only-wallet and not a hotwallet? Also, the `Secret<>` wrapper from the [secrecy crate](https://github.com/iqlusioninc/crates/tree/main/secrecy), prevents just that scenario by not offering a serialize impl. And the debug impl will automatically redact the inner value.

sn_transfers/src/wallet/hot_wallet.rs
mickvandijke: Commit [71267b5](https://github.com/maidsafe/safe_network/pull/1949/commits/71267b5004511fa13eef4dbba705a0a9c6243475) correctly handles this scenario now.

```rust
// Save the secret key as an encrypted file
store_main_secret_key(&wallet_dir, &wallet_key, Some(password.to_owned()))?;

// Delete the unencrypted secret key file
// Cleanup if it fails
if let Err(err) = delete_unencrypted_main_secret_key(&wallet_dir) {
    let _ = delete_encrypted_main_secret_key(&wallet_dir);
    return Err(err);
}
```

grumbach: Just to make sure, `Cleanup if it fails` with `delete_encrypted_main_secret_key` doesn't delete the encrypted one does it?

grumbach: If it does, are we 100% sure that an error above means the `unencrypted_main_secret_key` was NOT deleted?

grumbach: Just trying to make sure we don't lose keys

sn_transfers/src/wallet/hot_wallet.rs
grumbach: Oooh I see! Although sorting by key is better, it's a hack so we don't need to unlock the wallet so do that?
But then on the other hand, filtering by date here is weird, and we will need to cleanup old ones every time...
We will also need to filter by public key when we want to support multi-key in the future so keeping pks is cleaner.
How about keeping the public key in a non-encrypted file along with the wallet?

sn_transfers/src/wallet/authentication.rs
grumbach: > Isn't the wallet file only a watch-only-wallet and not a hotwallet? 

That I'm not sure.

> Also, the Secret<> wrapper from the [secrecy crate](https://github.com/iqlusioninc/crates/tree/main/secrecy), prevents just that scenario

Okay this is nice! We're safe then!

sn_transfers/src/wallet/hot_wallet.rs
mickvandijke: It comes down to https://doc.rust-lang.org/std/fs/fn.remove_file.html. I think that we can reasonably expect that it only fails when it couldn't delete the file (unencrypted `main_secret_key` in this case).

But to be 100% on the safe side, we could not do a cleanup in case of failure of `delete_unencrypted_main_secret_key`. Worst case scenario then, we would have both a `main_secret_key` file and a `main_secret_key.encoded` file in the same wallet directory.

sn_transfers/src/wallet/hot_wallet.rs
mickvandijke: > Oooh I see! Although sorting by key is better, it's a hack so we don't need to unlock the wallet so do that?

Correct!

>  How about keeping the public key in a non-encrypted file along with the wallet?

I thought about using the existing `main_pubkey` file. But the drawback for me was that we can not verify whether that pubkey is actually valid for a locked encrypted wallet/secret key. We'd still have to unlock the wallet to verify if the pubkey actually belongs to that wallet.

> We will also need to filter by public key when we want to support multi-key in the future so keeping pks is cleaner.

When the project starts supporting multiple keys/wallets, I agree that wallets should be saved in folder names that represent their pubkey. I'm not sure what folder structure will be used then (perhaps https://github.com/maidsafe/safe_network/issues/1944 ?). In that case, I see the current `stash` more as a function to backup an old wallet that you may or may not know the password of.

If we do decide on stashing old wallets under a folder name that contains the public address. I'm just a bit stuck on figuring out how we would best deal with stashing encrypted wallets where the user doesn't remember the password of?

mickvandijke: Hey @grumbach ,

I've made the changes we've discussed (using the `main_pubkey` file to get the wallet address): https://github.com/maidsafe/safe_network/pull/1949/commits/ce62fbf7f7da0a9687e61579879c38ba75529c5a

grumbach (approved):
LGTM! Excellent work!

mickvandijke:
Hi @grumbach ,

The test for stashing/unstashing fails only on Windows because of a permission error. I do not have a Windows PC to test myself. Is this a known problem specific to a Windows Docker image?

EDIT: Managed to fix it.

PR #1952, opened 2024-07-10
fix(auditor): create auditor directory if it doesn't exist

RolandSherwin:


sn_auditor/src/dag_db.rs
```diff
@@ -92,15 +92,18 @@ impl SpendDagDb {
         client: Client,
         encryption_sk: Option<SecretKey>,
     ) -> Result<Self> {
+        if !path.exists() {
+            std::fs::create_dir_all(&path)?;
+        }
```
mazzi, line 98: Good one!

What about adding to the documentation of the `new` method that if `path` doesn't exists will be created along with all the parent directories?

Also, let' add a `debug!()` statement there to log that the directory was created.

mazzi (approved):
Nice one!
Small comment but good to go! 🚀

sn_auditor/src/dag_db.rs
RolandSherwin: Thanks, I've added the debug line. The docs for `new` talks about loading a file from the path. So dealing with dir creation is implied there.

PR #1964, opened 2024-07-11
refactor(cli)!: `wallet address` and `wallet create` changes

### LINKED ISSUE #1963: refactor(cli): `wallet address` command should fail when no wallet is available instead of creating a new wallet
The CLI command `wallet address` should simply print the address of the current wallet and nothing more. Currently, it will automatically generate a new wallet when no wallet is found. This behavior makes no sense to me when there is already a `wallet create` command. I propose that the `wallet address` command will be changed to ONLY print the address of the existing wallet or fail with an error: `No wallet found. Please create one using 'wallet create --help'`. And the existing wallet creation logic will be moved to the `wallet create` command.

https://github.com/maidsafe/safe_network/blob/main/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs#L37

https://github.com/maidsafe/safe_network/blob/main/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs#L138

mickvandijke:
### Description

CLI changes.

`wallet address` command no longer creates a new wallet when there is no active wallet.

`wallet create` command now accepts the optional flags `--no-replace`, `--key` and `--derivation`, and eventually also `--password` to optionally encrypt the wallet.

| Flag  | Description |
| - | - |
| --no-replace | Auto answers with no, when asked if user wants to replace an existing wallet. |
| --key | Allows a user to create a wallet from an existing hex encoded private key. |
| --derivation | Allows users to set a derivation passphrase that will be used together with the mnemonic to create a new private key. If not set, a default derivation passphrase will be used. |

The mnemonic will automatically be generated if it doesn't exist already.

`--key` and `--derivation` can't be used together.

### Related Issue

Fixes #1963 

### Type of Change

- [ ] Documentation update

### Checklist

- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.

### DIFF
--- a/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
+++ b/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
@@ -13,6 +13,7 @@ use super::{
 };
 use crate::get_stdin_response;
 
+use autonomi::utils::is_valid_key_hex;
 use bls::SecretKey;
 use clap::Parser;
 use color_eyre::{
@@ -175,6 +176,13 @@ pub(crate) async fn wallet_cmds_without_client(cmds: &WalletCmds, root_dir: &Pat
                 return Err(eyre!(
                     "Only one of `--key` or `--derivation` may be specified"
                 ));
+            }
+            if let Some(key) = key {
+                // Check if key is valid
+                // Doing this early to avoid stashing an existing wallet while the provided key is invalid
+                if !is_valid_key_hex(key) {
+                    return Err(eyre!("Please provide a valid secret key in hex format. It must be 64 characters long."));
+                }
             }
             // Check for existing wallet
             if let Ok(existing_wallet) = WalletApiHelper::load_from(root_dir) {
--- a/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
+++ b/sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
@@ -20,6 +20,7 @@ use color_eyre::{
     Result,
 };
 use dialoguer::Confirm;
+use sn_client::acc_packet::{load_or_create_mnemonic, secret_key_from_mnemonic};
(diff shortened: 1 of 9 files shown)


sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
happybeing, line 1: I suggest the default should be to not replace an existing wallet.

Perhaps do away with the flag and have a `destroy` wallet command, since this is an operation we don't want to happen accidentally.

mickvandijke: Hi @happybeing ,

The `wallet create` command will prompt the user with `Replace existing wallet with new wallet? [y/N]` first when it finds an existing wallet, so that it doesn't get accidentally overwritten.

```rust
// Check for existing wallet
if let Ok(existing_wallet) = WalletApiHelper::load_from(root_dir) {
    let balance = existing_wallet.balance();
    println!("Existing wallet found with balance of {balance}");
    let response = if *no_replace {
        "n".to_string()
    } else {
        get_stdin_response("Replace existing wallet with new wallet? [y/N]")
    };
    if response.trim() != "y" {
        // Do nothing, return ok and prevent any further operations
        println!("Exiting without creating new wallet");
        return Ok(());
    }
    // remove existing wallet
    let new_location = HotWallet::stash(root_dir)?;
    println!("Old wallet stored at {}", new_location.display());
}
```

The old `wallet address` command would create a new wallet if there was none already (several tests relied on that behavior), so `wallet create --no-replace` serves as a simple replacement for that.

happybeing: Ah ok, sounds good.

joshuef (approved):

PR #1960, opened 2024-07-10
feat(auditor): dump fetched spends to local disk

maqi:
This pull request primarily introduces changes to the `SpendDagDb` implementation in the `sn_auditor` package, specifically in the `dag_db.rs` and `main.rs` files. The changes aim to enhance the process of updating the Directed Acyclic Graph (DAG) from the network by storing spends (transactions) on disk for better tracking and auditing. This is achieved by adding a `storage_dir` parameter to the `continuous_background_update` method and modifying how spends are processed.

The most important changes are:

* [`sn_auditor/src/dag_db.rs`](diffhunk://#diff-fc4319f01fee1895543b912d85d9455555b052ecc12a54539497148590b3dc48L207-R207): The `continuous_background_update` method in the `SpendDagDb` implementation now takes an additional `storage_dir: PathBuf` parameter. This directory path is used to store spends on disk.
* [`sn_auditor/src/dag_db.rs`](diffhunk://#diff-fc4319f01fee1895543b912d85d9455555b052ecc12a54539497148590b3dc48L225-R252): The spend processing logic has been updated. Now, each spend is hashed and stored in a `detected_spends` set. If a spend is new (not already in the set), it's written to a file in the `storage_dir`. The file name is a combination of the spend's address and hash.
* [`sn_auditor/src/main.rs`](diffhunk://#diff-41fbce10df15210e82c92ee34bc0329d4eacabd75c03a2bd9f2cf59b3a208828R104-R114): The `main` function now creates a `storage_dir` and passes it to the `initialize_background_spend_dag_collection` function.
* [`sn_auditor/src/main.rs`](diffhunk://#diff-41fbce10df15210e82c92ee34bc0329d4eacabd75c03a2bd9f2cf59b3a208828R194): The `initialize_background_spend_dag_collection` function now takes an additional `storage_dir: PathBuf` parameter. This is passed to the `continuous_background_update` method of the `SpendDagDb` instance. [[1]](diffhunk://#diff-41fbce10df15210e82c92ee34bc0329d4eacabd75c03a2bd9f2cf59b3a208828R194) [[2]](diffhunk://#diff-41fbce10df15210e82c92ee34bc0329d4eacabd75c03a2bd9f2cf59b3a208828L248-R254)

These changes are important as they help in tracking and auditing transactions by storing them on disk. This makes it easier to inspect individual transactions and provides a persistent record of transactions for auditing purposes.

PR #1973, opened 2024-07-16
Auditor supplement features

maqi:
### Description

Replace the PR https://github.com/maidsafe/safe_network/pull/1960

### Related Issue

### DIFF
--- a/sn_auditor/src/dag_db.rs
+++ b/sn_auditor/src/dag_db.rs
@@ -218,7 +218,7 @@ impl SpendDagDb {
     }
 
     /// Update DAG from Network continuously
-    pub async fn continuous_background_update(self) -> Result<()> {
+    pub async fn continuous_background_update(self, storage_dir: PathBuf) -> Result<()> {
         let client = if let Some(client) = &self.client {
             client.clone()
         } else {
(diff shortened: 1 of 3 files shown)


sn_auditor/src/dag_db.rs
```diff
@@ -341,17 +412,87 @@ impl SpendDagDb {
...
+        if royalties.len() > (spend.spend.spent_tx.outputs.len() - 1) / 2 {
+            eprintln!(
+                "Spend: {:?} has incorrect royalty of {}, with amount {} with reason {:?}",
+                spend.spend.unique_pubkey,
+                royalties.len(),
+                spend.spend.amount.as_nano(),
+                spend.spend.reason
+            );
+            eprintln!(
+                "Incorreect royalty spend has {} royalties, {:?} - {:?}",
```
grumbach:
```suggestion
                "Incorrect royalty spend has {} royalties, {:?} - {:?}",
```

sn_auditor/src/dag_db.rs
```diff
@@ -341,17 +412,87 @@ impl SpendDagDb {
...
+                || royalty_pubkeys.contains(&output.unique_pubkey)
+            {
+                let _ = royalties.insert(
+                    SpendAddress::from_unique_pubkey(&output.unique_pubkey),
+                    output.amount.as_nano(),
+                );
+            }
+        }
+
+        if royalties.len() > (spend.spend.spent_tx.outputs.len() - 1) / 2 {
```
grumbach, line 440: What if there is no change output because by chance the amount is perfect?
In this case the `-1` is unnecessary. Although there is no way to know which of the outputs is the change...

grumbach (approved):
LGTM, left small comments, not blocking

sn_auditor/src/dag_db.rs
maqi: this is just for alerting suspicious spend.
with the main nano supplier is currently controlled by us, the case of `no change` is really rare.
so, shall not be a problem here.

sn_auditor/src/dag_db.rs
maqi: this is just for alerting suspicious spend.
with the main nano supplier is currently controlled by us, the case of no change is really rare.
so, shall not be a problem here.

PR #1976, opened 2024-07-17
chore(networking): make ChunkVerification probabalistic

joshuef:


sn_networking/src/event/request_response.rs
```diff
@@ -235,8 +235,10 @@ impl SwarmDriver {
             }
         }
 
-        // Only trigger chunk_proof check when received a periodical replication request.
-        if incoming_keys.len() > 1 {
+        // Only trigger chunk_proof check based every X% of the time
+        let mut rng = thread_rng();
+        // 5% probability
+        if rng.gen_bool(0.05) {
```
maqi: maybe the critieria can be BOTH ? i.e. `if incoming_keys.len() > 1 && rng.gen_bool(0.05)` ?
to ensure the flooding of new data doesn't trigger the chunk_proof check too much as well ?

maqi (approved):

PR #1977, opened 2024-07-17
Reduce replication msg processing

joshuef:
`replication_fetcher.add_keys` will result in the same thing with or without all the preprocessing now

This pull request primarily focuses on refactoring and simplifying the `SwarmDriver` implementation in `sn_networking/src/event/request_response.rs` and the `ReplicationFetcher` implementation in `sn_networking/src/replication_fetcher.rs`. The main changes involve removing unnecessary methods, consolidating code, and improving readability.

Refactoring in `sn_networking/src/event/request_response.rs`:

* Removed the `libp2p::PeerId` import as it was no longer needed.
* Added a `more_than_one_key` variable to improve code readability.
* Simplified the logic for adding keys and triggering chunk_proof checks by using the `more_than_one_key` variable.
* Removed the `select_non_existent_records_for_replications` and `is_in_close_range` methods as they were no longer necessary.

Refactoring in `sn_networking/src/replication_fetcher.rs`:

* Changed the `add_keys` method to filter out locally stored keys from the incoming keys, which simplifies the logic and improves performance.
* Updated the logic for handling new data in the `add_keys` method to use the filtered list of incoming keys. [[1]](diffhunk://#diff-f85f9c20c1dd565a71917b0ac8bc76403964fd06a1575f38fe8702640811a726L104-R118) [[2]](diffhunk://#diff-f85f9c20c1dd565a71917b0ac8bc76403964fd06a1575f38fe8702640811a726L116-R129)
* Updated the logic for filtering out out_of_range keys in the `add_keys` method to use the filtered list of incoming keys.

### DIFF
--- a/sn_networking/src/replication_fetcher.rs
+++ b/sn_networking/src/replication_fetcher.rs
@@ -74,17 +74,30 @@ impl ReplicationFetcher {
     pub(crate) fn add_keys(
         &mut self,
         holder: PeerId,
-        mut incoming_keys: Vec<(NetworkAddress, RecordType)>,
+        incoming_keys: Vec<(NetworkAddress, RecordType)>,
         locally_stored_keys: &HashMap<RecordKey, (NetworkAddress, RecordType)>,
     ) -> Vec<(PeerId, RecordKey)> {
+        // remove locally stored from incoming_keys
+        let mut new_incoming_keys: Vec<_> = incoming_keys
+            .iter()
+            .filter(|(addr, record_type)| {
+                let key = &addr.to_record_key();
+                !locally_stored_keys.contains_key(key)
+                    && !self
+                        .to_be_fetched
+                        .contains_key(&(key.clone(), record_type.clone(), holder))
+            })
+            .cloned()
+            .collect();
+
         self.remove_stored_keys(locally_stored_keys);
         let self_address = NetworkAddress::from_peer(self.self_peer_id);
-        let total_incoming_keys = incoming_keys.len();
+        let total_incoming_keys = new_incoming_keys.len();
 
         // In case of node full, restrict fetch range
         if let Some(farthest_distance) = self.farthest_acceptable_distance {
             let mut out_of_range_keys = vec![];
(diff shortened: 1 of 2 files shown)


sn_networking/src/replication_fetcher.rs
```diff
@@ -74,17 +74,30 @@ impl ReplicationFetcher {
...
         self.remove_stored_keys(locally_stored_keys);
         let self_address = NetworkAddress::from_peer(self.self_peer_id);
-        let total_incoming_keys = incoming_keys.len();
+        let total_incoming_keys = new_incoming_keys.len();
 
         // In case of node full, restrict fetch range
         if let Some(farthest_distance) = self.farthest_acceptable_distance {
             let mut out_of_range_keys = vec![];
-            incoming_keys.retain(|(addr, _)| {
+            new_incoming_keys.retain(|(addr, _)| {
```
maqi, line 100: the `close_range` check within the `select_non_existent_records_for_replications` function happens all the time.
However, the range check here only happens during the `node full`, and using different algorithm.
Which will result in more replication copies to be kept by nodes.

maqi (changes requested):

sn_networking/src/replication_fetcher.rs
joshuef: Yeh so we actually still have that parsing [here](https://github.com/maidsafe/safe_network/pull/1977/files/f35846a36046df9d2f980a25b97f3316b1a4896a#diff-f85f9c20c1dd565a71917b0ac8bc76403964fd06a1575f38fe8702640811a726R148-R153) 

So it's this duplicate that's been removed, and should keep stored records relevant.

maqi (approved):

sn_networking/src/replication_fetcher.rs
joshuef: Sorry for clarity this was pointing to a few lines below the changes where we check the distance and prune based upon that when we have a `distance` set in teh replication fetcher. Which happens on an interval, continuously at the node (so not only when full)

PR #1982, opened 2024-07-18
feat(network): implement blocklist behaviour

RolandSherwin:
- uses `libp2p::allow_block_list` Behavior to deny connection to a bad node.

### DIFF
--- a/sn_networking/src/driver.rs
+++ b/sn_networking/src/driver.rs
@@ -184,20 +184,24 @@ pub enum VerificationKind {
     },
 }
 
-/// NodeBehaviour struct
+/// The behaviors are polled in the order they are defined.
+/// The first struct member is polled until it returns Poll::Pending before moving on to later members.
+/// Prioritize the behaviors related to connection handling.
 #[derive(NetworkBehaviour)]
 #[behaviour(to_swarm = "NodeEvent")]
 pub(super) struct NodeBehaviour {
-    #[cfg(feature = "upnp")]
-    pub(super) upnp: libp2p::swarm::behaviour::toggle::Toggle<libp2p::upnp::tokio::Behaviour>,
-    pub(super) request_response: request_response::cbor::Behaviour<Request, Response>,
-    pub(super) kademlia: kad::Behaviour<UnifiedRecordStore>,
+    pub(super) blocklist:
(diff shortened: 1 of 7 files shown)


sn_networking/src/driver.rs
```diff
@@ -184,20 +184,24 @@ pub enum VerificationKind {
...
-    pub(super) request_response: request_response::cbor::Behaviour<Request, Response>,
-    pub(super) kademlia: kad::Behaviour<UnifiedRecordStore>,
+    pub(super) blocklist:
+        libp2p::allow_block_list::Behaviour<libp2p::allow_block_list::BlockedPeers>,
+    pub(super) identify: libp2p::identify::Behaviour,
     #[cfg(feature = "local-discovery")]
     pub(super) mdns: mdns::tokio::Behaviour,
-    pub(super) identify: libp2p::identify::Behaviour,
-    pub(super) dcutr: libp2p::dcutr::Behaviour,
+    #[cfg(feature = "upnp")]
```
joshuef, line 198: why upnp only?

RolandSherwin: I just reordered the behaviours here, there is no actual change.

sn_networking/src/event/swarm.rs
```diff
@@ -137,7 +137,8 @@ impl SwarmDriver {
                                 our_protocol: IDENTIFY_PROTOCOL_STR.to_string(),
                                 their_protocol: info.protocol_version,
                             });
-                            self.remove_outdated_connections(Some(peer_id));
+                            // Block the peer from any further communication.
+                            self.swarm.behaviour_mut().blocklist.block_peer(peer_id);
```
maqi, line 141: so the already established connection will be disconnected automatically, once the peer is in the blocklist?
i.e. no more explicit `remove_connection` call to that peer is needed ?

sn_networking/src/driver.rs
maqi: yeah, that toggle is only for upnp, and just being moving place for a better ordering

sn_networking/src/event/swarm.rs
RolandSherwin: That is right! It internally calls `ToSwarm::CloseConnection`

maqi (approved):

PR #1988, opened 2024-07-19
refactor(network): reserver trace log level for tracking event stats

RolandSherwin:
- Currently, when using `SN_LOG=all`, the logs of `sn_networking` contains a lot of trace logs related to the event handling statistics. It is not really useful during debugging.
- So this PR aims to move all the useful logs to `debug` level, and now you can use `SN_LOG=v` instead of `SN_LOG=all` to just get the debug level logs.
- Also removes some unused code

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -396,7 +396,7 @@ impl SwarmDriver {
             } => {
                 cmd_string = "PutRecord";
                 let record_key = PrettyPrintRecordKey::from(&record.key).into_owned();
-                trace!(
+                debug!(
                     "Putting record sized: {:?} to network {:?}",
                     record.value.len(),
                     record_key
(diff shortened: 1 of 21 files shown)


sn_networking/src/cmd.rs
```diff
@@ -669,7 +617,7 @@ impl SwarmDriver {
                 // be handled.
                 // `self` then handles the request and sends a response back again to itself.
                 if peer == *self.swarm.local_peer_id() {
-                    trace!("Sending query request to self");
```
joshuef, line 672: Hmmm, i think the point here with trace was that we don't normally want this info?

These logs are more detailed than we'd want by default. (I believe... but maybe that's not the case).

RolandSherwin: Trace contained some useful info alteast for me, but also contained the stats which were clogging things up.

If certain logs are not useful, we can move them to `trace` down the line, but now if we use `SN_LOG=v`, things are much cleaner to read.

joshuef (approved):

PR #1930, opened 2024-07-01
chore!: refactor spend struct

maqi:
BREAKING CHANGE

related to the issue https://github.com/maidsafe/safe_network/issues/1931

### Description

Please provide a brief description of the changes made in this pull request. Highlight the purpose of the changes and the problem they address.

PR #1989, opened 2024-07-19
Massive Spend Refactor

grumbach:
BREAKING CHANGE

Continuing off @maqi's redesign of Spends in refactor PR #1930 
Redesigns the whole transacting flow from `CashNote`s all the way to `Spend` with a simpler design:

### Removing:
- `OfflineTransfer`: all you need to make a transaction, signed
- `TransferInputs` : amassed data to make a transaction's inputs
- `CashNotesAndSecretKey` : this for inputs too 
- `TransactionBuilder`: this makes a transaction from inputs
- `CashNoteBuilder`: this makes cashnotes from transactions
- `UnsignedTransfer`: this is a special case when we don't have a key to sign the transaction
- `Transaction`, `Input`, `Output`: this was in Spends
- the middle spend patch
- lots of unused error variants
- duplicate code
- layers and layers of ~~obfuscation~~ *abstraction*
- more than `500` lines of code
- ...

### Replaced by:
- `UnsignedTransaction`: all you need to make a transaction, not signed
- `SignedTransaction`: all you need to make a transaction, signed

### What's next

- [ ] adapt `spend_simulation` tests that were using the `Transaction`s in `Spends` @RolandSherwin 
- [x] implement `UnsignedTransaction::verify`
- [x] unit test `UnsignedTransaction::new`, `UnsignedTransaction::sign`
- [x] unit test hex serialization for both new types


joshuef:
What's the impact here for OMNI disbursement if any?

grumbach:
> What's the impact here for OMNI disbursement if any?

This reworks the structure of the Spend and refactors the transaction flow, it shouldn't affect OMNI disbursement

sn_client/src/audit/dag_crawling.rs
```diff
@@ -400,47 +408,49 @@ impl Client {
...
-                        .all(|s| s.spend.unique_pubkey == *sn_transfers::GENESIS_SPEND_UNIQUE_KEY)
+                debug!("Depth {depth} - Got {spends_len} spends for parent: {addrs_to_verify:?}");
+                trace!("Spends for {addrs_to_verify:?} - {spends:?}");
+
+                // check if we reached the genesis spend
+                known_ancestors.extend(addrs_to_verify.clone());
+                if spends
+                    .iter()
+                    .all(|s| s.spend.unique_pubkey == *sn_transfers::GENESIS_SPEND_UNIQUE_KEY)
                     && spends.len() == 1
```
grumbach: Got to check if this check is still accurate, wondering if we should not just
- check for genesis spends (not just addr, but content too) and remove them
- keep going for the others

sn_client/src/audit/tests/setup.rs
```diff
@@ -48,23 +46,6 @@ impl MockNetwork {
...
-            .map_err(|e| eyre!("failed to send genesis: {e}"))?;
-        net.genesis_spend = match spent_addrs.as_slice() {
-            [one] => *one,
-            _ => {
-                return Err(eyre!(
-                    "Expected Genesis spend to be unique but got {spent_addrs:?}"
-                ))
-            }
-        };
-
```
grumbach: Why did we remove that from the `MockNetwork::genesis` contructor?

sn_client/src/audit/tests/setup.rs
```diff
@@ -81,7 +62,11 @@ impl MockNetwork {
 
         if balance > 0 {
             let genesis_pk = GENESIS_CASHNOTE.main_pubkey();
-            self.send(genesis_pk, &owner_pk, balance)
+
+            let genesis_sk_main_pubkey = get_genesis_sk().main_pubkey();
+
+            println!("Sending {balance} from genesis {genesis_pk:?} to {owner_pk:?}");
+            self.send(&genesis_sk_main_pubkey, &owner_pk, balance, false)
                 .map_err(|e| eyre!("failed to get money from genesis: {e}"))?;
```
grumbach: What's the change here? Is `genesis_pk` different from `genesis_sk_main_pubkey`?

sn_transfers/src/genesis.rs
```diff
@@ -145,18 +135,12 @@ pub fn get_genesis_sk() -> MainSecretKey {
...
-
 /// Return if provided Spend is genesis spend.
 pub fn is_genesis_spend(spend: &SignedSpend) -> bool {
     let bytes = spend.spend.to_bytes_for_signing();
     spend.spend.unique_pubkey == *GENESIS_SPEND_UNIQUE_KEY
         && GENESIS_SPEND_UNIQUE_KEY.verify(&spend.derived_key_sig, bytes)
-        && is_genesis_parent_tx(&spend.spend.parent_tx)
-        && spend.spend.amount == NanoTokens::from(GENESIS_CASHNOTE_AMOUNT)
+        && spend.spend.amount() == NanoTokens::from(GENESIS_CASHNOTE_AMOUNT)
 }
```
grumbach, line 144: If we know the content of the Genesis Spend, maybe also add those values in this test to make sure it's the one and unique genesis? Do we know those values?
```
&& spend.spend.reason == GENESIS_REASON
&& spend.spend.ancestors == GENESIS_ANCESTORS
&& spend.spend.descendants == GENESIS_DESCENDANTS
```
Assuming GENESIS_SK is destroyed at Network Genesis, this scenario should not happen, but are we ever too safe?

sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
```diff
@@ -372,31 +369,18 @@ async fn send(
...
-        for (i, output) in tx.outputs.iter().enumerate() {
-            println!("\nOutput #{i}:");
-            println!("\tKey: {}", output.unique_pubkey.to_hex());
-            println!("\tAmount: {}", output.amount);
+        for (descendant, amount) in unsigned_tx.output_unique_keys().iter() {
+            println!("\tOutput Key: {}", descendant.to_hex());
+            println!("\tAmount: {amount}");
         }
-    } else {
-        bail!("Transaction is corrupted, no transaction information found.");
```
maqi, line 399: shall we put checks like there shall be non-empty ancestors and descendants ?

sn_transfers/src/cashnotes/cashnote.rs
```diff
@@ -56,14 +57,8 @@ use tiny_keccak::{Hasher, Sha3};
 /// To spend or work with a CashNote, wallet software must obtain the corresponding
 /// MainSecretKey from the user, and then call an API function that accepts a MainSecretKey,
 /// eg: `cashnote.derivation_index(&main_key)`
-#[derive(custom_debug::Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Hash)]
 pub struct CashNote {
-    /// The unique public key of this CashNote. It is unique, and there can never
```
maqi, line 61: Line 23: `A CashNote is like a check.` -> `A CashNote is like a cheque.` 

the above comment of the CashNote (Line 39 - Line 55) may need to be updated to reflect the struct refactor

sn_transfers/src/cashnotes/cashnote.rs
```diff
@@ -56,14 +57,8 @@ use tiny_keccak::{Hasher, Sha3};
...
-    /// be another CashNote with the same public key. It used in SignedSpends.
-    pub unique_pubkey: UniquePubkey,
-    /// The transaction where this CashNote was created.
-    #[debug(skip)]
-    pub parent_tx: Transaction,
     /// The transaction's input's SignedSpends
     pub parent_spends: BTreeSet<SignedSpend>,
```
maqi, line 63: comment of `parent_spends` needs to be updated

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -152,64 +92,49 @@ impl SignedSpend {
...
     /// - Also handles the case of parent double spends.
-    /// - verifies that the parent_spends where spent in our spend's parent_tx.
-    /// - verifies the parent_tx against the parent_spends
-    pub fn verify_parent_spends<'a, T>(&self, parent_spends: T) -> Result<()>
-    where
-        T: IntoIterator<Item = &'a BTreeSet<SignedSpend>> + Clone,
-    {
+    /// - verifies that the parent_spends contains self as an output
+    /// - verifies the sum of total inputs equals to the sum of outputs
+    pub fn verify_parent_spends(&self, parent_spends: &BTreeSet<SignedSpend>) -> Result<()> {
```
maqi, line 102: I am wondering without the middle pay, how the above conditions got all satisfied for the scenario of `multiple_inputs to multiple_outputs` ?

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -152,64 +92,49 @@ impl SignedSpend {
...
-                error!("No parent spend provided for {unique_key}");
-                return Err(TransferError::InvalidParentSpend(
-                    "Parent is empty".to_string(),
-                ));
+        let mut total_inputs: u64 = 0;
+        for (_, spends) in parents_by_key {
+            // check for double spend parents
+            if spends.len() > 1 {
+                error!("While verifying parents of {unique_key}, found a double spend parent: {spends:?}");
+                return Err(TransferError::DoubleSpentParent);
```
maqi, line 120: This actually means the spend itself having duplicated ancestors ?
which is different to the scenario that an ancestor got spent already?
maybe it shall give a different name to differiate these two scenarios?

sn_transfers/src/cashnotes/signed_spend.rs
```diff
@@ -152,64 +92,49 @@ impl SignedSpend {
...
-                None => {
-                    let tx_parent_was_spent_in = parents
-                        .iter()
-                        .map(|p| p.spent_tx_hash())
-                        .collect::<Vec<_>>();
-                    return Err(TransferError::InvalidParentSpend(format!(
-                        "Parent spend was spent in another transaction. Expected: {tx_our_cash_note_was_created_in:?} Got: {tx_parent_was_spent_in:?}"
-                    )));
+            // check that the parent refers to self
+            if let Some(parent) = spends.first() {
```
maqi, line 124: I am not sure why we shall still keep processing in case of `double spend`.
I thought whenever we detect an attempt of double spend, we shall stop the handling and return with error immediately ?

sn_transfers/src/genesis.rs
maqi: > Do we know those values?

Yeah, we do can know these values.

> Assuming GENESIS_SK is destroyed at Network Genesis

shall be OK, as GENESIS_SK is actually not used (and shall not be used) after GENESIS_SPEND generated.

sn_transfers/src/transfers/signed_transaction.rs
```diff
@@ -0,0 +1,81 @@
...
+            cn.verify()?;
+        }
+        for spend in self.spends.iter() {
+            spend.verify()?;
+        }
+        Ok(())
+    }
+
+    /// Create a new `SignedTransaction` from a hex string
+    pub fn from_hex(hex: &str) -> Result<Self> {
```
maqi, line 70: we shall only require `hex` format for a `transfer`, but never for a `transaction` ?

sn_transfers/src/transfers/signed_transaction.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::error::Result;
+use crate::{
+    CashNote, DerivationIndex, MainPubkey, MainSecretKey, NanoTokens, OutputPurpose, SignedSpend,
+    SpendReason, TransferError, UnsignedTransaction,
+};
+use serde::{Deserialize, Serialize};
+
+/// A local transaction that has been signed and is ready to be executed on the Network
+#[derive(custom_debug::Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct SignedTransaction {
```
maqi, line 20: do we really need the entire `trasaction` to be signed ?
I think we only need a bunch of `SignedTrasnfer`s (one generated per output_cashnotes) to be generated ?

sn_transfers/src/transfers/unsigned_transaction.rs
```diff
@@ -0,0 +1,310 @@
...
+        let hex = tx.to_hex()?;
+        let tx2 = UnsignedTransaction::from_hex(&hex)?;
+
+        assert_eq!(tx, tx2);
+        Ok(())
+    }
+
+    #[test]
+    fn test_unsigned_tx_distribution() -> Result<()> {
+        // NB TODO: Implement this test
```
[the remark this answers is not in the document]
maqi: yeah, I think this unit_test is a `must to have` before merge.
the correct distribution especially in the scenario of multiple_inputs to multiple_outputs is critical, and easier to be verified with unit test, but hard to detect issue with large sized network.

sn_transfers/src/transfers/unsigned_transaction.rs
```diff
@@ -0,0 +1,308 @@
...
+                .iter()
+                .map(|s| *s.unique_pubkey())
+                .collect();
+            let mut input_remaining_value = input_value.as_nano();
+            let mut donate_to = BTreeMap::new();
+
+            // take value from input and distribute it to outputs
+            while input_remaining_value > 0 {
+                if let Some((output, amount, purpose)) = outputs_iter.next() {
+                    let amount_to_take = min(input_remaining_value, amount.as_nano());
```
maqi: if the current `output.amount > input_remaining_value`, then here we got `input_remaining_value` sent to that output recipient.
BUT, with the next input, the `outputs_iter.next()` will jump to the next recipient.
Which acutally the previous recipient is short of `output.amount - input_remaining_value` to be paid.

maqi (changes requested):

sn_transfers/src/cashnotes/cashnote.rs
grumbach: Good one! This doc is outdated, check/cheque is not adapted anymore I think, let me fix that!

sn_cli/src/bin/subcommands/wallet/hot_wallet.rs
grumbach: Indeed! That's what I'm doing right now in the `verify` function! Will be in the next PR!

sn_transfers/src/transfers/signed_transaction.rs
grumbach: This is used for the `wo_wallet` which doesn't have a private key to sign the Tx but can nonetheless create a Tx. 
The `hex` format is used so that this wallet can export this unsigned_tx to be signed by a hot_wallet.
I think this was done so we can interoperate with hardware keys and such.

sn_transfers/src/transfers/signed_transaction.rs
grumbach: The transaction itself is not signed, but the `SignedSpend` contained in it are so we can finish the CashNote and as you say send them as `Transfers` which don't need to be signed as they contain the network addresses of the `SignedSpend`s

maqi: Ah, ok.
just double checked the code of `let signed_tx = unsigned_tx.sign(main_key)?;` 
which the `UnSignedTransaction::sign` function only only sign the contains spends individually, instead of signing the entire body.
Though the API is little bit misleading (as the `sign` function normally indicates signing the entire body), the current processing logic is correct.

sn_transfers/src/cashnotes/signed_spend.rs
maqi: looks like it now trying to use `return multiple cash_note, each cn per recipient, i.e. get inputs distributed to outpus explicitly` to resolve this.
So, here, both of the criterias can still be satisfied at the same time.

sn_transfers/src/cashnotes/signed_spend.rs
grumbach: > This actually means the spend itself having duplicated ancestors ?

This means one of the spend's parents has multiple entries and is burnt.

sn_transfers/src/cashnotes/signed_spend.rs
grumbach: Yes all this is covered by the distribution algo in UnsignedTransaction

sn_transfers/src/transfers/unsigned_transaction.rs
grumbach: Excellent find! Now fixed in the last commit!

sn_transfers/src/cashnotes/signed_spend.rs
RolandSherwin: Sorry, I think we return early with an error incase of parent double spend?

sn_transfers/src/transfers/unsigned_transaction.rs
```diff
@@ -93,11 +111,21 @@ impl UnsignedTransaction {
             })
             .collect();
 
+        // order inputs by value, re const after sorting
+        let mut cashnotes_big_to_small = available_cash_notes;
```
maqi, line 115: we shall try to use the small ones first ?

grumbach: Ideally we'd like to minimize the amount of Spends created, using big ones first has that effect for now, but it's not optimal over the long run. 

In the future I imagine we should create a better algorithm that would optimize that and choose the best options with the closest values as to minimize the amount of spends and avoid using huge Cashnotes for small Txs. 

Also, since CashNotes don't have a determined value anymore, we could return the change value to an existing unspent CashNote that would accumulate wealth until it is spent. 

All these nice optimizations will not change the API in this PR so that's also nice!

maqi (approved):

PR #1992, opened 2024-07-22
fix(network): only replicate newly obtained records

RolandSherwin:
- only replicate any newly obtained records (either through PUT/replication), during periodic replication to the same peer.
- if the REPLICATION_GROUP changes and we see a new peer, we replicate ALL records (current behaviour). But if the peer is the same one that we just sent REPLICATE cmd to, then we only try to send new record to him.
- Perform a round of "ALL Key" replication on a much longer interval
- Also request the close group to replicate to us in case we've restarted our node.
- Resend the same set of keys multiple times before updating the timestamp. This is so that, nodes do not miss our updates by chance.
- Also store more replication targets to account for shaky nodes that might join and leave our group.


sn_networking/src/cmd.rs
```diff
@@ -26,20 +28,16 @@ use sn_protocol::{
     NetworkAddress, PrettyPrintRecordKey,
 };
 use sn_transfers::{NanoTokens, PaymentQuote, QuotingMetrics};
-use std::{
-    collections::{BTreeMap, HashMap},
-    fmt::Debug,
-    time::Duration,
-};
+use std::collections::HashMap;
```
maqi: any particular reason to split this out ?

sn_networking/src/cmd.rs
```diff
@@ -26,20 +28,16 @@ use sn_protocol::{
...
 use tokio::sync::oneshot;
 use xor_name::XorName;
 
-use crate::target_arch::Instant;
-
 const MAX_CONTINUOUS_HDD_WRITE_ERROR: usize = 5;
 
-// Shall be synced with `sn_node::PERIODIC_REPLICATION_INTERVAL_MAX_S`
-const REPLICATION_TIMEOUT: Duration = Duration::from_secs(45);
+const REPLICATION_RETRIES_BEFORE_UPDATING_THE_RECORDS_TO_CONSIDER_TIMESTAMP: usize = 3;
```
maqi: the const name is not explain its usage that clear, hence some comment explaination still be necessary?

sn_networking/src/cmd.rs
```diff
@@ -102,7 +100,7 @@ pub enum SwarmCmd {
     },
     /// Get the Addresses of all the Records held locally
     GetAllLocalRecordAddresses {
-        sender: oneshot::Sender<HashMap<NetworkAddress, RecordType>>,
+        sender: oneshot::Sender<Vec<NetworkAddress>>,
```
maqi: still prefer de-duplicated Set/Map if possible, instead of `Vec`, unless there is strong reason of using `Vec`.

sn_networking/src/cmd.rs
```diff
@@ -808,66 +819,210 @@ impl SwarmDriver {
...
 
-        let now = Instant::now();
-        self.replication_targets
-            .retain(|_peer_id, timestamp| *timestamp > now);
-        // Only carry out replication to peer that not replicated to it recently
-        replicate_targets.retain(|peer_id| !self.replication_targets.contains_key(peer_id));
-        if replicate_targets.is_empty() {
-            return Ok(());
+        for peer_id in close_group {
+            let request = Request::Cmd(Cmd::RequestReplication {
```
maqi: composing of the Request can be put outside the for clause ?

sn_networking/src/cmd.rs
```diff
@@ -808,66 +819,210 @@ impl SwarmDriver {
...
+            holder: NetworkAddress::from_peer(self.self_peer_id),
+            keys: records_to_replicate,
+        });
+        let request_id = self
+            .swarm
+            .behaviour_mut()
+            .request_response
+            .send_request(&peer_id, request.clone());
+        debug!("Sending Replicate cmd {request_id:?} with {len} keys to peer {peer_id:?} as RequestReplication was called by that peer",);
+        let _ = self.pending_requests.insert(request_id, None);
```
maqi: we still expect an empty OK response for this request (actually a response) , right?

the reason we have to do that is because the `request_response.send_response(channel, resp)` requires the `channel` to be thrown up, but we have trouble on it, right ?

sn_networking/src/driver.rs
```diff
@@ -650,8 +650,9 @@ pub struct SwarmDriver {
     pub(crate) peers_in_rt: usize,
     pub(crate) bootstrap: ContinuousBootstrap,
     pub(crate) relay_manager: RelayManager,
-    /// The peers that are closer to our PeerId. Includes self.
     pub(crate) replication_fetcher: ReplicationFetcher,
+    /// The peers we are currently replicating to and the (records to consider with > timestamp than this, last replication time, count before updating the records to consider timestamp)
```
maqi: make `records_to_consider` a word?

sn_networking/src/event/request_response.rs
```diff
@@ -49,6 +49,20 @@ impl SwarmDriver {
 
                             self.add_keys_to_replication_fetcher(holder, keys);
                         }
+                        Request::Cmd(sn_protocol::messages::Cmd::RequestReplication {
+                            keys: holding_keys,
+                        }) => {
+                            let response = Response::Cmd(
+                                sn_protocol::messages::CmdResponse::RequestReplication(Ok(())),
+                            );
+                            self.swarm
```
maqi, line 64: yeah, this actuall make one `request-response` messaging to be two `request-OK` messagings.
which actually we can handle it within the swarmdriver directly?
instead have to throw up to the network layer ?
especially when the record_store holds the whole list of stored keys in cache ?

sn_node/src/node.rs
```diff
@@ -81,53 +85,77 @@ const UPTIME_METRICS_UPDATE_INTERVAL: Duration = Duration::from_secs(10);
...
             #[cfg(feature = "upnp")]
-            upnp,
+            upnp: false,
+            was_node_restarted: false,
         }
     }
 
-    #[cfg(feature = "open-metrics")]
+    /// Set to true to enable hole-punching
+    pub fn is_behind_home_network(&mut self, is_behind_home_network: bool) {
```
maqi: the function name indicates it is a `checker`, but actually it is a `setter` ?

sn_node/src/node.rs
```diff
@@ -282,16 +314,29 @@ impl Node {
...
-            debug!("Replication interval set to {replication_interval_time:?}");
+            let mut latest_record_replication_interval =
+                tokio::time::interval(latest_record_replication_interval);
+            let _ = latest_record_replication_interval.tick().await; // first tick completes immediately
 
-            let mut replication_interval = tokio::time::interval(replication_interval_time);
-            let _ = replication_interval.tick().await; // first tick completes immediately
+            let all_record_replication_interval: u64 = rng.gen_range(
+                PERIODIC_ALL_RECORD_REPLICATION_INTERVAL_MAX_S
+                    ..(PERIODIC_ALL_RECORD_REPLICATION_INTERVAL_MAX_S as f64 * 1.05) as u64,
```
maqi: why the `1.05` multiplier ?
it actually makes the `MAX_S` no longer the max ?

sn_networking/src/cmd.rs
RolandSherwin: Nope, probably was done automatically. Will change it back

sn_networking/src/cmd.rs
RolandSherwin: Added comment and updated the var name..

sn_networking/src/cmd.rs
RolandSherwin: Yeah I agree, caller would be confused if they're de-duped. I'll use HashSet.

sn_networking/src/cmd.rs
RolandSherwin: With None, it is a fire and forget thing. We don't wait for their response. 
```
  if let Response::Cmd(CmdResponse::Replicate(Ok(()))) = response {
      // Nothing to do, response was fine
      // This only exists to ensure we dont drop the handle and
      // exit early, potentially logging false connection woes
  } else {
      // responses that are not awaited at the call site must be handled
      // separately
      self.send_event(NetworkEvent::ResponseReceived {
          res: response,
      });
  }
```

sn_networking/src/event/request_response.rs
RolandSherwin: Yeah, that is right. I'll update it 😄

maqi (approved):

RolandSherwin:
Outdated and did not make much impact.

PR #1997, opened 2024-07-23
feat(manager): optionally set the different ports when running local network

RolandSherwin:
- Allow the user to set the metrics, rpc & node port when running a local network.

### DIFF
--- a/sn_node_manager/src/add_services/config.rs
+++ b/sn_node_manager/src/add_services/config.rs
@@ -23,20 +23,46 @@ pub enum PortRange {
     Range(u16, u16),
 }
 
-pub fn parse_port_range(s: &str) -> Result<PortRange> {
-    if let Ok(port) = u16::from_str(s) {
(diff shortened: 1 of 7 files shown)


sn_node_manager/src/add_services/config.rs
```diff
@@ -23,20 +23,46 @@ pub enum PortRange {
...
+                if count != 1 {
+                    error!("The count ({count}) does not match the number of ports (1)");
+                    return Err(eyre!(
+                        "The count ({count}) does not match the number of ports (1)"
+                    ));
+                }
+            }
+            Self::Range(start, end) => {
+                let port_count = end - start + 1;
+                if count != port_count {
```
joshuef, line 57: would this trigger if we specified a greater range than required?

RolandSherwin: Yes, it would. We'd error out.

joshuef (approved):

PR #2000, opened 2024-07-23
chore(node): increase bad node & bootstrap interval

RolandSherwin:
- increases the bad node detection interval to 5mins..10mins. Currently, it is set to 2mins..4mins and that causes a lot of msgs flying around.
- Also slow down the bootstrap process even more depending on the number of peers in RT. Currently, it is set to 5 seconds to every 5 peer in RT, and it has been updated to 10 seconds for every 5 peer in the RT. This is because the bad node detection flow kinda mimics the same thing we're doing here.

### DIFF
--- a/sn_networking/src/bootstrap.rs
+++ b/sn_networking/src/bootstrap.rs
@@ -12,8 +12,9 @@ use tokio::time::Duration;
 
 use crate::target_arch::{interval, Instant, Interval};
 
-/// The interval in which kad.bootstrap is called
-pub(crate) const BOOTSTRAP_INTERVAL: Duration = Duration::from_secs(5);
+/// The default interval at which NetworkDiscovery is triggered. The interval is increased as more peers are added to the
+/// routing table.
+pub(crate) const BOOTSTRAP_INTERVAL: Duration = Duration::from_secs(10);
 
 /// Every BOOTSTRAP_CONNECTED_PEERS_STEP connected peer, we step up the BOOTSTRAP_INTERVAL to slow down bootstrapping
 /// process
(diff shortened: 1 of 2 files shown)


sn_networking/src/bootstrap.rs
```diff
@@ -27,9 +27,12 @@ const LAST_PEER_ADDED_TIME_LIMIT: Duration = Duration::from_secs(180);
 const LAST_BOOTSTRAP_TRIGGERED_TIME_LIMIT: Duration = Duration::from_secs(30);
 
 /// The bootstrap interval to use if we haven't added any new peers in a while.
-const NO_PEER_ADDED_SLOWDOWN_INTERVAL: u64 = 300;
+const NO_PEER_ADDED_SLOWDOWN_INTERVAL_MAX_S: u64 = 600;
```
RolandSherwin, line 31: Not changed. Earlier it was `*2`, now it is `/2`.

sn_node/src/node.rs
```diff
@@ -57,6 +57,14 @@ use sn_protocol::storage::{try_serialize_record, RecordKind, SpendAddress};
 /// This is the max time it should take. Minimum interval at any node will be half this
 pub const PERIODIC_REPLICATION_INTERVAL_MAX_S: u64 = 45;
 
+/// Interval to trigger bad node detection.
+/// This is the max time it should take. Minimum interval at any node will be half this
+const PERIODIC_BAD_NODE_DETECTION_INTERVAL_MAX_S: u64 = 600;
+
+/// Interval to trigger reward forwarding.
+/// This is the max time it should take. Minimum interval at any node will be half this
+const PERIODIC_REWARD_FORWARD_INTERVAL_MAX_S: u64 = 450;
```
RolandSherwin, line 66: Not changed, earlier it was `10*PERIODIC_REPLICATION_INTERVAL_MAX_S`

sn_networking/src/bootstrap.rs
```diff
@@ -12,8 +12,8 @@ use tokio::time::Duration;
 
 use crate::target_arch::{interval, Instant, Interval};
 
-/// The interval in which kad.bootstrap is called
-pub(crate) const BOOTSTRAP_INTERVAL: Duration = Duration::from_secs(5);
+/// The interval in which kad.bootstrap is called.
```
maqi: we no longer using kad.bootstrap anymore, instead using the `network discovery`.
the comment can now be updated.

maqi (approved):

PR #2003, opened 2024-07-24
chore(node): [#1999] panic on upnp event `GatewayNotFound`

### LINKED ISSUE #1999: Node should panic on upnp failure when the `--upnp` flag is specified
Node should panic on upnp failure when the `--upnp` flag is specified. Else users won't directly realize that their nodes might be unreachable.

mickvandijke:
### Description

Makes a node panic on the upnp event `GatewayNotFound`. This event only fires when the node runs with the `--upnp` flag and libp2p couldn't find the upnp gateway to open ports on.

### Related Issue

Fixes #1999 

### Type of Change

- [ ] Documentation update
- [x] Other (please describe): Makes it easier to detect a problem with UPnP.

### Checklist

### DIFF
--- a/sn_networking/src/event/swarm.rs
+++ b/sn_networking/src/event/swarm.rs
@@ -6,6 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::event::TerminateNodeReason;
 use crate::{
     cmd::LocalSwarmCmd,
     event::NodeEvent,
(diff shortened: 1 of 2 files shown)


sn_networking/src/event/swarm.rs
```diff
@@ -80,6 +80,9 @@ impl SwarmDriver {
                 }
                 event_string = "upnp_event";
                 info!(?upnp_event, "UPnP event");
+                if let libp2p::upnp::Event::GatewayNotFound = upnp_event {
+                    panic!("UPnP is not enabled/supported on your gateway. Please rerun without the `--upnp` flag");
```
b-zee: This should be an error that we bubble up. In that case we could also later do graceful shutdowns of our swarm. `sn_networking` is a library crate, so we don't want consumers to get panics.

mickvandijke: We can bubble the error up, but it would be by one level and still within the `sn_networking` lib:

https://github.com/maidsafe/safe_network/blob/main/sn_networking/src/driver.rs#L735-L741

Handling this error gracefully would probably take some refactoring. Not sure if that is worth it. Thoughts?

b-zee (approved):

b-zee (approved):

PR #2011, opened 2024-07-30
feat(networking): initial range based gets implementation

joshuef:


sn_client/src/error.rs
```diff
@@ -64,9 +68,6 @@ pub enum Error {
     #[error(transparent)]
     JoinError(#[from] tokio::task::JoinError),
 
-    /// A general error when verifying a transfer validity in the network.
-    #[error("Failed to verify transfer validity in the network {0}")]
-    CouldNotVerifyTransfer(String),
```
grumbach, line 69: Glad we finally got rid of this one!!

sn_client/src/wallet.rs
```diff
@@ -1155,9 +1162,13 @@ impl Client {
...
-        Err(WalletError::CouldNotVerifyTransfer(
-            "The spends in network were not the same as the ones in the CashNote. The parents of this CashNote are probably double spends.".into(),
-        ))
+        let address = SpendAddress::from_unique_pubkey(&cash_note.unique_pubkey());
+
+        warn!(
+            "BurntSpendAttempt found in CashNote verification at {:?}.",
+            address
+        );
+        Err(WalletError::BurntSpendAttempt)
```
grumbach: Here the error is not burnt spend. If the received spends don't match the spends in the cashnote, we should report it as such. The cashnote itself could be invalid.

sn_networking/src/bootstrap.rs
```diff
@@ -7,45 +7,20 @@
...
-
-/// The bootstrap interval to use if we haven't added any new peers in a while.
-const NO_PEER_ADDED_SLOWDOWN_INTERVAL_MAX_S: u64 = 600;
+pub(crate) const BOOTSTRAP_INTERVAL: Duration = Duration::from_secs(30);
 
 impl SwarmDriver {
     /// This functions triggers network discovery based on when the last peer was added to the RT and the number of
     /// peers in RT. The function also returns a new bootstrap interval that is proportional to the number of
     /// peers in RT, so more peers in RT, the longer the interval.
-    pub(crate) async fn run_bootstrap_continuously(
```
grumbach: Docs outdated this func no longer return anything

sn_networking/src/cmd.rs
```diff
@@ -894,24 +911,67 @@ impl SwarmDriver {
...
-        // Only grab the closest nodes within the REPLICATE_RANGE
-        let mut replicate_targets = closest_k_peers
-            .into_iter()
-            // add some leeway to allow for divergent knowledge
-            .take(REPLICATION_PEERS_COUNT)
+            .get_closest_local_peers(&target_key)
+            .filter_map(|key| {
+                // here we compare _bucket_, not the exact distance.
+                // We want to include peers that are just outside the range
+                // Such that we can and will exceed the range in a search eventually
```
grumbach, line 970: What does get_closest_local_peers return exactly?

sn_networking/src/event/kad.rs
```diff
@@ -369,83 +385,116 @@ impl SwarmDriver {
...
+                info!("For record {pretty_key:?} task {query_id:?}, we have more than one result returned.");
+                // Allow for early bail if we've already seen a split SpendAttempt
+                for (record, _) in result_map.values() {
+                    match get_raw_signed_spends_from_record(record) {
+                        Ok(spends) => {
+                            accumulated_spends.extend(spends);
+                        }
+                        Err(_) => {
+                            continue;
+                        }
```
grumbach: This can happen for Registers too.

sn_networking/src/event/kad.rs
```diff
@@ -369,83 +385,116 @@ impl SwarmDriver {
...
+                    }
+                }
+            }
 
-            debug!("Expecting {expected_answers:?} answers for record {pretty_key:?} task {query_id:?}, received {responded_peers} so far");
+            // we have a Double SpendAttempt and should bail early
```
grumbach: Why bail early? 
I don't think it should be kad code's responsibility to judge the content of the records.

grumbach: I notice this is not new to this PR, I wonder if we could have kad related code ignore the content (chunk/register/spend) and focus on records only?

sn_client/src/wallet.rs
```diff
@@ -1145,8 +1143,17 @@ impl Client {
 
         let mut received_spends = std::collections::BTreeSet::new();
         for result in join_all(tasks).await {
-            let network_valid_spend =
-                result.map_err(|err| WalletError::CouldNotVerifyTransfer(err.to_string()))?;
+            let network_valid_spend = match result {
+                Ok(spend) => Ok(spend),
+                Err(error) => match error {
+                    Error::Network(sn_networking::NetworkError::DoubleSpendAttempt(spends)) => {
```
maqi, line 1157: refactored to `BurntSpend...`

sn_networking/src/cmd.rs
```diff
@@ -531,6 +542,31 @@ impl SwarmDriver {
 
         Ok(())
     }
+
+    /// Return the RecordType
+    pub(crate) fn get_type_from_record(record: &Record) -> Result<RecordType> {
```
maqi, line 555: this seems shall be a helper function to be defined in the `sn_transfer::Spend` ?
or at least shall not be `inside driver` ?

sn_networking/src/driver.rs
```diff
@@ -770,29 +778,105 @@ impl SwarmDriver {
...
+            Ok(sorted_peers) => sorted_peers,
+            Err(err) => {
+                warn!("Problem while sorting peers by address during set_request_range : {err}");
+                return;
+            }
+        };
+
+        let closest_peer = sorted_discovery_peers.first();
+        let farthest_peer = if sorted_discovery_peers.len() >= close_group_majority() {
+            sorted_discovery_peers.get(close_group_majority() - 1)
```
maqi: majority ? not CLOSE_GROUP_SIZE ?
then it shall not be named as `farthest_peer` ?

sn_networking/src/driver.rs
```diff
@@ -770,29 +778,105 @@ impl SwarmDriver {
...
+        } else {
+            sorted_discovery_peers.last()
+        };
+
+        if let Some(closest_peer) = closest_peer {
+            let closest_addr = NetworkAddress::from_peer(**closest_peer);
+            if let Some(farthest_peer) = farthest_peer {
+                let farthest_addr = NetworkAddress::from_peer(**farthest_peer);
+                new_distance_range = farthest_addr
```
maqi: say we are `10000`, and our close group nodes are: `10011`, `10110`, `10111`, `01100`, `01111`
then according to the above calculation, the closest is `10011`, and the farthest (actually the 3rd one) is `10111`, then the calculated new_distance_range will be `00100`, which will actually only include `10011` in its range ?

sn_networking/src/driver.rs
```diff
@@ -770,29 +778,105 @@ impl SwarmDriver {
...
+                trace!("Removed distance range: {:?}", distance.ilog2());
+            }
+        }
+
+        info!(
+            "Adding new distance range: {:?}",
+            new_distance_range.ilog2()
+        );
+
+        self.range_distances.push_back(new_distance_range);
```
maqi: `push` not `insert` ?
will this give `duplicated` distance_range certain chance to distort the final result ?

sn_networking/src/event/kad.rs
```diff
@@ -457,6 +506,64 @@ impl SwarmDriver {
...
+
+        let required_quorum = get_quorum_value(quorum);
+
+        // if we don't even meet quorum, that's it.
+        if !is_last_step && searched_peers_list.len() < required_quorum {
+            return false;
+        }
+
+        // get the farthest distance between peers in the response
+        let mut current_distance_searched = KBucketDistance::default();
```
maqi, line 429: the following block of code seems just set `current_distance_searched` to be the `searched closet distance to the target` ?

sn_networking/src/event/request_response.rs
```diff
@@ -201,16 +208,13 @@ impl SwarmDriver {
...
         );
 
-        // accept replication requests from the K_VALUE peers away,
-        // giving us some margin for replication
-        let closest_k_peers = self.get_closest_k_value_local_peers();
-        if !closest_k_peers.contains(&holder) || holder == self.self_peer_id {
-            debug!("Holder {holder:?} is self or not in replication range.");
+        let more_than_one_key = incoming_keys.len() > 1;
+        // accept replication requests from all peers within our GetRange
+        if !peers.contains(&holder) || holder == our_peer_id {
```
maqi, line 214: `peers` is actually from `self.get_all_local_peers()` ,
which means we accept replication as long as the sender appears in our local RT ?

is it correct? or at least the comment shall be updated to : 
`accept replication requests from all peers`

sn_networking/src/lib.rs
```diff
@@ -146,6 +152,40 @@ pub fn sort_peers_by_key<'a, T>(
 
     Ok(sorted_peers)
 }
+/// Only return peers closer to key than the provided distance
```
maqi, line 175: but the `provided distance` was calculated according to `the node's close range` ?
which could be way smaller than key's actuall close range, when `key` is in a far bucket?

sn_networking/src/lib.rs
```diff
@@ -253,6 +413,31 @@ impl Network {
...
+    pub async fn get_all_local_peers(&self) -> Result<Vec<PeerId>> {
+        let (sender, receiver) = oneshot::channel();
+        self.send_local_swarm_cmd(LocalSwarmCmd::GetAllLocalPeers { sender });
+
+        receiver
+            .await
+            .map_err(|_e| NetworkError::InternalMsgChannelDropped)
+    }
+
+    /// Returns all the PeerId from all the KBuckets from our local Routing Table
```
maqi: but the function name indicates only get `close_range peers` to the record ?

sn_client/src/error.rs
joshuef: Aye, there's still a bunch of `String` placeholders there we should cleanup. I wouldn't dive about there until after the payment refactor work, though.

(We clearly have some issue around organizing errors w/r/t wallets in client. As we pass some errors INTO transfers FROM client which actually then pulls out and converts them FROM transfers)

sn_client/src/wallet.rs
joshuef: Hmmmm, Okay. I thought this was me aligning errors, but clearly I've gone off piste.

That's actually quite helpful in terms of this issue I'm seeing as both the legitimate `BurntSpent`  and this `DivergentParents` (or what have you) error could surface in this test... So I guess that may be the issue I'm seeing here!

sn_networking/src/cmd.rs
joshuef: there's docs on the fn. Just updated them for more clarity:

```Returns peers sorted by closeness to the given key; takes peers from local routing table only.```

File: sn_networking/src/event/kad.rs (hunk 5 above)
Comment (joshuef) [reply sn_networking/src/event/kad.rs]: Ah yeh, but we handle that case with the normal `SplitRecord` error, returning early. We don't attempt to exit early though as there's no benefit to returning with only eg 2/X split records. We'd actually want all of them?

File: sn_networking/src/event/kad.rs (hunk 6 above)
Comment (joshuef) [reply sn_networking/src/event/kad.rs]: >  I don't think it should be kad code's responsibility to judge the content of the records.

Hmm, it's an efficiency thing. If we have a double spend, we could exit earlier instead of waiting for full accumulation. The end result is always the same.

Similarly, we can exit early for Chunks.


It is all very `optimisation`y though. I'll try some runs without this and see where we get to. As you say, maybe we can simplify this out.

File: sn_client/src/wallet.rs (hunk 7 above)
Comment (joshuef) [reply sn_client/src/wallet.rs:1157]: I'm not sure what you're after her @maqi 

Just below:

```rust
 Error::Network(sn_networking::NetworkError::DoubleSpendAttempt(spends)) => {
                        warn!("DoubleSpentAttempt found with {spends:?}");
                        Err(WalletError::BurntSpendAttempt)
                    }
                    ```

File: sn_networking/src/driver.rs (hunk 9 above)
Comment (joshuef) [reply sn_networking/src/driver.rs]: Good point. Will roll back around to this once tests are finalised. I think I need to get into the bug first and this may yet change...

File: sn_networking/src/driver.rs (hunk 10 above)
Comment (joshuef) [reply sn_networking/src/driver.rs]: We `sort_peers_by_address_and_limit` just above here, so they should be properly sorted at this point...

Or am I missing something there?

File: sn_networking/src/driver.rs (hunk 11 above)
Comment (joshuef) [reply sn_networking/src/driver.rs]: We only want to keep the 100 most recent additions. So `push_back` and `pop_front` are used when adding,.

When sorting / findning median, we:

```rust
 let mut sorted_distances = self.range_distances.iter().collect::<Vec<_>>();

        sorted_distances.sort_unstable();

        let median_index = sorted_distances.len() / 2;
 ```
        
I believe that should be fine?

File: sn_networking/src/event/kad.rs (hunk 12 above)
Comment (joshuef) [reply sn_networking/src/event/kad.rs:429]: 🤔  


```rust
   // get the farthest distance between peers in the response
        let mut current_distance_searched = KBucketDistance::default();

        // iterate over peers and see if the distance to the data is greater than the get_range
        for peer_id in searched_peers_list.iter() {
            let peer_address = NetworkAddress::from_peer(*peer_id);
            let distance_to_data = peer_address.distance(data_key_address);
            if current_distance_searched < distance_to_data {
                current_distance_searched = distance_to_data;
            }
        }
```

Should be:

- set `current_distance_searched` = 0
- if the peer's distance to data is further, use that distance.

Which I think isn't only going to be the closest?? (Or i'm missing something).


Thinking about that now though, the distasnce searched may yet be bigger... as the peers could be either side... 

ie, 
- Peer1: Location =1
- Data: Location = 2
- Peer2: Location = 3

So the distance of any peer is 1... But the total searched is 2...

🤔 I'm not sure how best to calculate _that_ in libp2p address terms 🤔

sn_networking/src/event/request_response.rs
joshuef: Ah is only local yes. Updated for clarity

sn_networking/src/lib.rs
joshuef: Hmm, what would you suggest here?

sn_networking/src/cmd.rs
grumbach: So it does not get range?

sn_networking/src/event/kad.rs
grumbach: I think we should wait for full accumulation as if we stop at the first ones we might miss the original one as in the shadowing attack we fixed recently. 
I think low level functions should as much as possible, not take assumptions and do their job in a predictable manner.

sn_networking/src/cmd.rs
joshuef: Cooould be! I'll loop back on this one after some furtther updates

sn_networking/src/event/kad.rs
grumbach: Yeah we'd want all of them, and for spends too also

sn_networking/src/transfers.rs
```diff
@@ -30,8 +28,10 @@ impl Network {
     /// If we get a quorum error, we enable re-try
     pub async fn get_raw_spends(&self, address: SpendAddress) -> Result<Vec<SignedSpend>> {
         let key = NetworkAddress::from_spend_address(address).to_record_key();
+        // use Quorum::One not All here, as we want to collect and return all the spends we can find
+        // Quorum all may be prohibitive to this end
         let get_cfg = GetRecordCfg {
-            get_quorum: Quorum::Majority,
+            get_quorum: Quorum::One,
```
grumbach, line 34: I'm not sure what you mean here? Doesn't Quorum::One stop at the first response??

sn_node/tests/double_spend.rs
```diff
@@ -72,24 +73,31 @@ async fn cash_note_transfer_double_spend_fail() -> Result<()> {
...
+        format!("{should_err1:?}"),
+        format!("Err({:?})", WalletError::BurntSpendAttempt),
+        "Should have been BurntSpendAttempt error, was: {should_err1:?}"
+    );
+
+    assert_eq!(
+        format!("{should_err2:?}"),
+        format!("Err({:?})", WalletError::BurntSpendAttempt),
+        "Should have been BurntSpendAttempt error, was: {should_err2:?}"
+    );
```
grumbach: BurntSpend is a better error name. It's not an attempt, as it's already burnt.

sn_networking/src/event/kad.rs
joshuef: I think we just need to sort to a fixed point. Ideally a distance of 0. But it's all relative, right. So long as we choose _somewhere consistent_. 

I've changed this to derive an address from the genesis spend just now. It would be consistent across the whole network then... 

(Although I'm not actually sure there's any benefit of that from our_peer_id eg... )

joshuef: Actually that's bad and leads to wild distances. I guess we just want the farthest distance to the key we searched for or something similar. Feels simpler

sn_client/src/wallet.rs
maqi: Ah, sry, just mean that: the wording of `DoubleSpend` shall be replaced to `BurntSpend`

sn_networking/src/driver.rs
maqi: yeah they are `sorted` .
but the issue here is that the calculated `range` seems `small`.
and this range will be used later on to decide other `get ranges`, which could then be a problem ?

sn_networking/src/lib.rs
maqi: well, in general, I think trying to use `our close range` for a `far bucket key`, especially based on `local RT peers` will deliver a wrong result with high chance.

for any key that belongs to a `full bucket`, all peers of that bucket shall be contacted for `further closer nodes` . 
but this will then very likely to `re-write the libp2p's interative closeness queries` .

joshuef:
Closing in favour of the merged work in #2033

sn_networking/src/transfers.rs
joshuef: old point no longer relevant with early exit removed.

PR #2013, opened 2024-07-30
fix(manager): dynamic sleep interval during start and upgrade

RolandSherwin:
- Add a timeout to the GET query when fetching the network contacts file. Else it hangs forever.
- By default, we wait for a node to connect to the network before starting the next node.

### DIFF
--- a/sn_node_manager/src/bin/cli/main.rs
+++ b/sn_node_manager/src/bin/cli/main.rs
@@ -284,6 +284,11 @@ pub enum SubCmd {
     /// sudo if you defined system-wide services; otherwise, do not run the command elevated.
     #[clap(name = "start")]
     Start {
+        /// The max time in milliseconds to wait for the RPC connection to the node to be established.
+        ///
+        /// The node is marked as failed if the connection is not established within this time.
+        #[clap(long, default_value = "120000")]
+        connection_timeout: u64,
         /// An interval applied between launching each service.
         ///
         /// Units are milliseconds.
@@ -343,6 +348,11 @@ pub enum SubCmd {
     /// sudo if you defined system-wide services; otherwise, do not run the command elevated.
     #[clap(name = "upgrade")]
     Upgrade {
+        /// The max time in milliseconds to wait for the RPC connection to the node to be established.
+        ///
+        /// The upgrade will fail if the connection is not established within this time.
+        #[clap(long, default_value = "120000")]
+        connection_timeout: u64,
         /// Set this flag to upgrade the nodes without automatically starting them.
         ///
         /// Can be useful for testing scenarios.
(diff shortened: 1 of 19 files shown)


sn_node_manager/src/bin/cli/main.rs
```diff
@@ -284,11 +284,22 @@ pub enum SubCmd {
     /// sudo if you defined system-wide services; otherwise, do not run the command elevated.
     #[clap(name = "start")]
     Start {
+        /// The max time in milliseconds to wait for the RPC connection to the node to be established.
+        ///
+        /// Setting this will automatically scale the interval between each service launch. This argument
+        /// is mutually exclusive with the 'interval' argument.
+        ///
+        /// The upgrade will fail if the connection is not established within this time.
```
jacderida: I don't think we need this line for the `start` command?

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -284,11 +284,22 @@ pub enum SubCmd {
     /// sudo if you defined system-wide services; otherwise, do not run the command elevated.
     #[clap(name = "start")]
     Start {
+        /// The max time in milliseconds to wait for the RPC connection to the node to be established.
+        ///
+        /// Setting this will automatically scale the interval between each service launch. This argument
```
jacderida: I was thinking about this a little bit.

"Setting this will automatically scale the interval between each service launch."

To me, this language implies that the dynamic scaling is only going to apply if you used this argument. However, that's not the case, because there is a default value for the connection timeout. So, the dynamic scaling is really going to apply unless it gets overridden by the `--interval` argument.

I think it could possibly make more sense for the dynamic scaling to be explained on the start command itself, rather than on this connection timeout argument. I'm not really sure there should actually be *any* link between this connection timeout and the dynamic scaling.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -284,11 +284,22 @@ pub enum SubCmd {
...
+        /// is mutually exclusive with the 'interval' argument.
+        ///
+        /// The upgrade will fail if the connection is not established within this time.
+        #[clap(long, default_value = "120000", conflicts_with = "interval")]
+        connection_timeout: u64,
         /// An interval applied between launching each service.
         ///
+        /// Use connection-timeout to scale the interval automatically. This argument is mutually exclusive with the
```
jacderida, line 302: Considering my comment above, I don't think these arguments should really relate to one another, because I don't think the dynamic scaling should be tied to the `connection-timeout` argument.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -278,17 +278,32 @@ pub enum SubCmd {
     },
     /// Start safenode service(s).
     ///
+    /// By default, the interval between starting each service is scaled automatically. The 'connection-timeout'
+    /// controls the maximum time to wait between starting each service. Use the 'interval' argument to set a fixed
+    /// interval and override the dynamic interval.
+    ///
```
b-zee: I haven't dug into the rest of the PR yet, but from the perspective of a node user it's not clear to me what 'dynamic interval' is just from these docs. `connection-timeout` is a limit, but how does the mechanism decide that dynamic interval?

E.g. what is the relation between connections and the dynamic interval?

jacderida: I might be wrong here, but I think the link between the two is that the connection timeout would define the maximum possible value for the dynamic interval. I think that should be communicated on both the start command itself and the argument.

I also agree that we need to explain what the dynamic interval is based on, which I believe is how quickly the previous node establishes connections to a close group. Although, to be honest, just like the previous measurement, which was the length of time it takes to establish an RPC connection, I'm not sure I understand why either of those things would have a bearing on how long we should wait to try and start the _next_ node. I think it would be useful if we explained that in the documentation.

b-zee (approved):

jacderida (approved):

PR #2017, opened 2024-07-31
fix(node): check whether already received an incoming payment

maqi:
### Description

Please provide a brief description of the changes made in this pull request. Highlight the purpose of the changes and the problem they address.

### Related Issue

### Type of Change

### DIFF
--- a/sn_transfers/src/wallet/hot_wallet.rs
+++ b/sn_transfers/src/wallet/hot_wallet.rs
@@ -359,6 +359,13 @@ impl HotWallet {
             .collect()
     }
 
+    /// Checks whether the specified cash_note already presents
+    pub fn cash_note_presents(&mut self, id: &UniquePubkey) -> bool {
+        self.watchonly_wallet
+            .available_cash_notes()
+            .contains_key(id)
+    }
+
     /// Returns all available cash_notes and an exclusive access to the wallet so no concurrent processes can
     /// get available cash_notes while we're modifying the wallet
     /// once the updated wallet is stored to disk it is safe to drop the WalletExclusiveAccess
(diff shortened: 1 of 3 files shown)


sn_transfers/src/wallet/hot_wallet.rs
```diff
@@ -359,6 +359,13 @@ impl HotWallet {
             .collect()
     }
 
+    /// Checks whether the specified cash_note already presents
+    pub fn cash_note_presents(&mut self, id: &UniquePubkey) -> bool {
+        self.watchonly_wallet
+            .available_cash_notes()
+            .contains_key(id)
```
RolandSherwin, line 366: I think we're using a mixture of both `Stateless` methods & `Stateful` methods in our payment forwarding + payment receiving code, and I think that might be prone to errors.

When we first receive the cashnote:
1. We call `cash_note_presents`, which returns the value from `watchonly_wallet` which is from memory inside `KeyLessWallet` (stateful in this context). 
2. We call `get_confirmed_spend` which reads the filesystem are returns the value (stateless).
3. We call `deposit_and_store_to_disk` which writes the cashnote to the filesystem & updates the `KeyLessWallet` wallet (stateful).

When we forward the cashnote:
4. We call `remove_cash_notes_from_disk`, which removes the cashnote from the disk (stateless). Thus, it does not update the state of `KeyLessWallet` that is used during point-1
5. We call `remove_unconfirmed_spend_requests`, which writes the spend to `CONFIRMED_SPENDS_DIR_NAME` (stateless).


Thus, we read from fs directly in 3 places without updating our state, but return a value from memory at 2 place. 


I think we're fine here, but mixing up the different "wallet" structs is a bit confusing and even though in this case we're fine, `remove_cash_notes_from_disk` does not update `KeyLessWallet` when we remove a cashnote, which can be prone to error in the future?

RolandSherwin: Edit: `KeyLessWallet` is reloaded everytime we perform an OP. So the above effect is intended.

RolandSherwin (approved):

PR #2021, opened 2024-08-01
feat(metrics): track the bad node count

RolandSherwin:
- track the number of bad peers that we've detected.

### DIFF
--- a/sn_networking/src/event/mod.rs
+++ b/sn_networking/src/event/mod.rs
@@ -135,7 +135,7 @@ pub enum NetworkEvent {
         bad_behaviour: String,
     },
     /// We have been flagged as a bad node by a peer.
-    FlaggedAsBadNode { flagged_by: NetworkAddress },
+    FlaggedAsBadNode { flagged_by: PeerId },
     /// The records bearing these keys are to be fetched from the holder or the network
     KeysToFetchForReplication(Vec<(PeerId, RecordKey)>),
     /// Started listening on a new address
--- a/sn_networking/src/event/mod.rs
+++ b/sn_networking/src/event/mod.rs
@@ -128,12 +128,14 @@ pub enum NetworkEvent {
         our_protocol: String,
         their_protocol: String,
     },
-    /// The peer is now considered as a bad node, due to the detected bad behaviour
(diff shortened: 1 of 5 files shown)


sn_node/src/log_markers.rs
```diff
@@ -51,6 +51,12 @@ pub enum Marker<'a> {
     /// Valid spend stored
     ValidSpendPutFromClient(&'a PrettyPrintRecordKey<'a>),
 
+    /// The peer has been considered as bad
+    PeerConsideredAsBad(&'a PeerId),
+
+    /// We have been flagged as a bad node by a peer.
+    FlaggedAsBadNode(&'a NetworkAddress),
```
b-zee: Can't we use `PeerId` instead of these `NetworkAddress`es? Maybe that needs a bit of a refactor in `sn_protocol`'s `cmd.rs`, but would seem neater.

If it's not easy or not possible, fine, lets keep it like this.

sn_networking/src/event/mod.rs
```diff
@@ -128,12 +128,14 @@ pub enum NetworkEvent {
...
     },
-    /// The peer is now considered as a bad node, due to the detected bad behaviour
+    /// A peer from our RT is considered as bad due to the included behaviour
     PeerConsideredAsBad {
         detected_by: PeerId,
         bad_peer: PeerId,
         bad_behaviour: String,
     },
+    /// We have been flagged as a bad node by a peer.
+    FlaggedAsBadNode { flagged_by: NetworkAddress },
```
b-zee: Possible to use `PeerId` here too?

sn_node/src/log_markers.rs
RolandSherwin: I think `sn_protocol`'s `cmd.rs` uses `NetworkAddress` everywhere because we don't want structs from other dependencies through the wire? 
Anyway, I tried it and `PeerId` does not implement deserialize, so that will not be possible.

Although I can change things to `PeerId` at other places.

b-zee: Sounds complicated, lets leave it out.

b-zee (approved):

b-zee (approved):

PR #2033, opened 2024-08-09
Spend and range atop main

joshuef:
#1989 and #2011 atop latest main codebase.


joshuef:
closing to replace with just the getrange work atop this

PR #2041, opened 2024-08-12
feat(launchpad): new layout sections and features

mazzi:
### Description

We added a couple of new functionalities to the `launchpad`: drive selection to store nodes data and also to be able to see node logs. Part of this task was to re-arrange the screen layout and hotkeys. This led to several refactors and new features that will follow on this description. As a summary:

* Added navigation bar with `Status`, `Options` and `Help` sections. 
* New sections: `Options` and `Help`.
* We can access the node/s logs from the `Options` section.
* We can choose a different drive where the node data is stored.

#### Configuration
We now store the `mountpoint` and `drive name` of the drive where the node data is stored in our `app_data.json`. We detect from which drive is being executed the `launchpad` to determine this settings on the first run of this version.

#### Refactors
We created two modules for system calls (`system.rs`) and node management (`node_mgmt.rs`)

#### Components/Widgets
We grouped the pop ups into a folder called `popup`.
We created a `footer` and a `header`. Both are `StatefulWidget`s.
We modified `Hyperlink` to be able to include it into `Text`, `Paragraph` or `Block` ratatui elements. More work to be done there.

#### Drive selection
The drives that are entitled to be chosen are the ones that are writable and readable. In some systems and some drives, an alert can appear asking for permissions to write to that drive.

#### `sn_node_manager`
We parametrize the verbosity when resetting nodes. This was creating a glitch when resetting nodes in the `launchpad`.

### Testing

Desktop testing was made. Mostly on macOS. We will need to test more on Windows/Linux.


node-launchpad/src/app.rs
```diff
@@ -46,39 +52,61 @@ impl App {
...
         )
         .await?;
-        let config = Config::new()?;
-        let discord_username_input = BetaProgramme::new(app_data.discord_username.clone());
-        let manage_nodes = ManageNodes::new(app_data.nodes_to_start)?;
-        let footer = Footer::new(app_data.nodes_to_start > 0);
-        let help = HelpPopUp::default();
+        let options = Options::new(
+            app_data.storage_mountpoint.clone().unwrap(),
+            app_data.storage_drive.clone().unwrap(),
```
RolandSherwin: Can we use the `data_dir_path` that is obtained above? eyre will throw a nice error msgs than using unwrap().

Also we could do the same error handling for `app_data.storage_drive`?

node-launchpad/src/config.rs
```diff
@@ -17,6 +18,26 @@ use std::path::PathBuf;
...
 
+// Where to store the Nodes data
+pub fn get_launchpad_nodes_data_dir_path(base_dir: PathBuf) -> Result<PathBuf> {
+    // We get the data dir and we remove the first character which is the / character
+    let mut data_directory = dirs_next::data_dir()
+        .expect("Data directory is obtainable")
+        .to_str()
+        .unwrap()
+        .to_string();
+    data_directory.remove(0);
```
RolandSherwin: `data_dir()` would include the `$HOME` folder, I don't think we should recreate the same strucutre on a different drive. It might be confusing for people and I have not seen that anywhere. I.e., we'd get `F:\Users\Alice\AppData\Roaming`. 

Can we just use `F:\safe\node`?

node-launchpad/src/app.rs
```diff
@@ -46,39 +52,61 @@ impl App {
...
+            app_data.discord_username.clone(),
+        )
+        .await?;
+        let help = Help::new().await?;
+
+        // Popups
         let reset_nodes = ResetNodesPopup::default();
+        let discord_username_input = BetaProgramme::new(app_data.discord_username.clone());
+        let manage_nodes = ManageNodes::new(app_data.nodes_to_start).await?;
+        let change_drive = ChangeDrivePopup::new(app_data.storage_mountpoint.clone().unwrap());
```
RolandSherwin: Same here, can we return an `eyre` error instead of the unwrap?

node-launchpad/src/config.rs
```diff
@@ -66,10 +89,22 @@ impl AppData {
 
         let data = std::fs::read_to_string(config_path)
             .map_err(|_| color_eyre::eyre::eyre!("Failed to read app data file"))?;
+
         let app_data: AppData = serde_json::from_str(&data)
             .map_err(|_| color_eyre::eyre::eyre!("Failed to parse app data"))?;
 
-        Ok(app_data)
+        let mut app_data_cloned = app_data.clone();
```
RolandSherwin: We could just do `let mu app_data` in the above line to avoid this extra line here.

node-launchpad/.config/config.json5
```diff
@@ -1,48 +1,73 @@
...
-      "<Ctrl-r>": {"HomeActions":"TriggerResetNodesPopUp"},
-      "<Ctrl-R>": {"HomeActions":"TriggerResetNodesPopUp"},
-      "<Ctrl-Shift-r>": {"HomeActions":"TriggerResetNodesPopUp"},
+    "Status" : {
+      "<s>": {"MenuActions":"StatusTab"},
+      "<S>": {"MenuActions":"StatusTab"},
+      "<o>": {"MenuActions":"OptionsTab"},
+      "<O>": {"MenuActions":"OptionsTab"},
+      "<h>": {"MenuActions":"HelpTab"},
+      "<H>": {"MenuActions":"HelpTab"},
```
RolandSherwin: Can we not directly call `Action::SwitchScene` here? The `MenuActions` just trigger swtich scene everywhere, so why don't we do it directly here?

RolandSherwin: Similary for `Options` and `Help` screens

node-launchpad/src/components/popup/manage_nodes.rs
```diff
@@ -36,7 +36,7 @@ pub struct ManageNodes {
 }
 
 impl ManageNodes {
-    pub fn new(nodes_to_start: usize) -> Result<Self> {
+    pub async fn new(nodes_to_start: usize) -> Result<Self> {
```
RolandSherwin: We can remove async here?

node-launchpad/src/app.rs
mazzi: `data_dir_path` is the mountpoint + the directory where to store nodes data. I would prefer to store the mountpoint instead of the actual directory as we are storing drive related info (name and mountpoint).

Regarding `eyre`, I can use something like:

```rust
app_data.storage_mountpoint.clone().ok_or_else(|| eyre::eyre!("storage_mountpoint is None"))?,
```

wdyt?

node-launchpad/src/config.rs
mazzi: Yep, we can use just `F:\safe\node`. My only concern is how do we know that is not the default drive the one that we are using so we don't use the `$HOME` path. We can discuss.

node-launchpad/src/config.rs
mazzi: Good one!

node-launchpad/.config/config.json5
mazzi: Did it like this for clarity but you are correct, we save a lot of `match` comparisons.

node-launchpad/src/components/popup/manage_nodes.rs
mazzi: Good catch

node-launchpad/src/config.rs
RolandSherwin: We've agreed that `/` or `C:/` is the default mount point, and they'd use the `data_dir()` structure, whereas the other mounts would use `$MOUNT/safe/node` structure.

node-launchpad/src/app.rs
RolandSherwin: > data_dir_path is the mountpoint + the directory where to store nodes data. 

Yeah, sorry misread it.

And I like the eyre error! We can go with that!

node-launchpad/src/system.rs
```diff
@@ -0,0 +1,148 @@
...
+use sysinfo::Disks;
+
+// Tries to get the default (drive name, mount point) of the current executable
+// to be used as the default drive
+pub fn get_default_mount_point() -> (String, String) {
+    // Create a new System instance
+    let disks = Disks::new_with_refreshed_list();
+
+    // Get the current executable path
+    let exe_path = env::current_exe().expect("Failed to get current executable path");
```
RolandSherwin: Sorry for this nit pick, but we avoid using `expect` and `unwraps` everywhere. In this case yes, the propagated error is not handled and it leads to a panic. Butt if we were to return an error here, then in the future we could handle it if needed.

node-launchpad/src/system.rs
```diff
@@ -0,0 +1,148 @@
...
+    for disk in disks.list() {
+        if exe_path.starts_with(disk.mount_point()) {
+            return (
+                disk.name().to_string_lossy().into(),
+                disk.mount_point().to_string_lossy().into_owned(),
+            );
+        }
+    }
+    // If no matching disk is found, return an empty string or handle the error as needed
+    (String::new(), String::new())
```
RolandSherwin: If empty string is returned here, how are things handled at the other components? Should we be more explicit and return a `None` incase if something doesn't work out instead of an empty string. It would be cleaner to handle.

node-launchpad/src/system.rs
```diff
@@ -0,0 +1,148 @@
...
+pub fn get_mount_point() -> PathBuf {
+    PathBuf::from("/")
+}
+#[cfg(windows)]
+pub fn get_mount_point() -> PathBuf {
+    PathBuf::from("C:\\")
+}
+
+// Gets available disk space in bytes for the given mountpoint
+pub fn get_available_space_b(storage_mountpoint: String) -> Result<usize> {
```
RolandSherwin: Just a tiny nit pick, we can avoid `clones` here tbh. 1) But we can just pass `&str` as argument here OR 2) inside the functions, instead of using `storage_mountpoint.clone()`, we could also just use the references `&storage_mountpoint`. I think the compiler would optimize these `clones` out, but just to be explicit, we could always try using references whenever possible.

node-launchpad/src/config.rs
```diff
@@ -17,6 +19,30 @@ use std::path::PathBuf;
 
 const CONFIG: &str = include_str!("../.config/config.json5");
 
+// Where to store the Nodes data
+// If the base dir is the default mount point, we store in "$HOME/user_data_dir/safe/node"
+// if not we store in "$MOUNTPOINT/safe/node"
+pub fn get_launchpad_nodes_data_dir_path(base_dir: PathBuf) -> Result<PathBuf> {
+    let mut mount_point = PathBuf::new();
+
+    let data_directory = if base_dir == get_mount_point() {
```
RolandSherwin: Nit pick again, but maybe rename `get_mount_point` into `get_primary_mount_point`?

node-launchpad/src/config.rs
```diff
@@ -17,6 +19,30 @@ use std::path::PathBuf;
...
+        dirs_next::data_dir()
+            .expect("Data directory is obtainable")
```
RolandSherwin, line 30: One more panic that we can convert to error.

RolandSherwin: Actually there are a couple more here. We could use `eyre` everwhere.

node-launchpad/src/config.rs
```diff
@@ -17,6 +19,30 @@ use std::path::PathBuf;
...
+            .to_str()
+            .unwrap()
+            .to_string()
+    } else {
+        base_dir.to_str().unwrap().to_string()
+    };
```
RolandSherwin: Instead of `data_directory` being a `string`, can we not just return a `PathBuf` here? It would avoid the unwraps, we really just need a `PathBuf` here.

node-launchpad/src/config.rs
```diff
@@ -52,6 +78,8 @@ pub async fn configure_winsw() -> Result<()> {
 pub struct AppData {
     pub discord_username: String,
     pub nodes_to_start: usize,
+    pub storage_mountpoint: Option<String>,
```
RolandSherwin: Can we make this into `pub storage_mountpoint: Option<PathBuf>`. Less error as people might have weird chars in their Path & converting to string and back to PathBuf can fail.

node-launchpad/src/components/popup/manage_nodes.rs
```diff
@@ -30,19 +30,21 @@ pub struct ManageNodes {
     /// Whether the component is active right now, capturing keystrokes + drawing things.
     active: bool,
     available_disk_space_gb: usize,
+    storage_mountpoint: String,
```
RolandSherwin: `PathBuf` here as well!

node-launchpad/src/action.rs
```diff
@@ -15,13 +15,15 @@ use strum::Display;
...
-    TabActions(TabActions),
+    StatusActions(StatusActions),
+    OptionsActions(OptionsActions),
+
     SwitchScene(Scene),
     SwitchInputMode(InputMode),
 
     StoreDiscordUserName(String),
     StoreNodesToStart(usize),
+    StoreStorageDrive(String, String),
```
RolandSherwin: Can we update this to `StoreStorageDrive {mount: PathBuf, name: String}` to be more explicit and avoid converting between `PathBuf <-> String` as it is failable and can be avoided.

node-launchpad/src/components/options.rs
```diff
@@ -0,0 +1,271 @@
...
+    components::header::Header,
+    mode::{InputMode, Scene},
+    style::{EUCALYPTUS, GHOST_WHITE, LIGHT_PERIWINKLE, VERY_LIGHT_AZURE, VIVID_SKY_BLUE},
+    system,
+};
+use sn_node_manager::config::get_user_safenode_data_dir;
+
+#[derive(Clone)]
+pub struct Options {
+    pub storage_mountpoint: String,
```
RolandSherwin: `PathBuf` here as well! Probably can update it at everyplace tbh.

node-launchpad/src/components/help.rs
```diff
@@ -0,0 +1,256 @@
...
+                    "Download the latest launchpad:",
+                    Style::default().fg(GHOST_WHITE),
+                )])),
+            ]),
+            Row::new(vec![
+                Cell::from(
+                    get_direct_support_link
+                        .to_string()
+                        .into_text()
+                        .unwrap()
```
RolandSherwin, line 115: Can we use `eyre` here as well, and also in the ones below.

node-launchpad/src/widgets/hyperlink.rs
```diff
@@ -26,6 +27,20 @@ impl<'content> Hyperlink<'content> {
     }
 }
 
+// Displays the hyperlink in the terminal using OSC 8.
+// Underline solid \x1b[4m
+// Foreground color 45 \x1b[38;5;45m
+impl fmt::Display for Hyperlink<'_> {
```
RolandSherwin, line 33: For a follow up PR, we should reomve the `HyperLink` struct or try to modify it to fit the use cases.

node-launchpad/src/system.rs
mazzi: Thanks Roland. I've switched to `PathBuf` and used reference to avoid `clone()` 👍

node-launchpad/src/components/help.rs
mazzi: For the links, I believe than an `unwrap()` here will not be an `Error` as we are building the hyperlink in the code and we test.
`.into_text()` on this case, is part of `ansi-to-tui`. Maybe we can handle this when we handle `Hyperlink`.
What do you think?

node-launchpad/src/widgets/hyperlink.rs
mazzi: Indeed, we will need to work on `Hyperlink`. Will come in a follow up PR as you mentioned 👍

RolandSherwin (approved):
Great PR @mazzi! 🚀 Happy to get this in!!

PR #2042, opened 2024-08-14
Docs for currency

grumbach:
Docs for the currency


sn_transfers/README.md
```diff
@@ -1,17 +1,317 @@
...
 
-For error handling, we expose [`Error`](https://github.com/maidsafe/sn_transfers/blob/main/src/error.rs) and [`Result`](https://github.com/maidsafe/sn_transfers/blob/main/src/result.rs) types.
+This `DerivedSecretKey` is used to sign the `Spend` which is then sent to the Network for validation and storage. Once the Network has stored and properly replicated that `Spend`, that `UniquePubkey` is considered to be spent and cannot ever be spent again. If more than one `Spend` entry exist at a given `SpendAddress` on the Network, that key is considered to be burnt which makes any `Spend` refering to it unspendable. 
+
+Without the `DerivationIndex`, there is no way to link a `MainPubkey` to a `UniquePubkey`. Since `UniquePubkey`s are spendable only once, this means every transaction involves new and unique keys which are all unrelated and unlinkable to their original owner's `MainPubkey`.
+
+Under the hood, those types are simply:
```
maqi, line 21: simply -> simple ?

maqi (commented):
A very nice write up.
However, note it is based one the new spend struct, hence may have to be merged of the refactor PR got merged.

maqi (changes requested):
Depends on the PR https://github.com/maidsafe/safe_network/pull/1989

sn_transfers/README.md
```diff
@@ -1,17 +1,317 @@
...
+
+## Spends
+
+When a `UniquePubkey` is spent, the owner creates a `Spend` and signs it with the associated `DerivedSecretKey` before uploading it to the Network. A `Spend` contains the following information:
+
+```rust
+pub struct Spend {
+    pub unique_pubkey: UniquePubkey,
+    pub ancestors: BTreeSet<UniquePubkey>,
+    pub descendants: BTreeMap<UniquePubkey, NanoTokens>,
```
maqi, line 38: still need to keep the `royalties` member field

also, shall the doc cover the `reason` member field as well?

if this doc is just about currency & spend DAG, which these two fields (especially the reason field) are not that related,
at least put a `...` below the line of `descendants`, to avoid give a mis-impression to the readers that `Spend` struct only has that three member fields ?

grumbach: Yes I didn't document that as those fields are more likely to evolve thus the `...`

grumbach: We could add those once stabilized?

maqi (approved):

PR #2043, opened 2024-08-15
feat(node): prune unrelevant records if accumulated too many

maqi:
### Description

There is observed super-high quote due to incorrectly calculated store_cost quote, as an result of not-fully populated RT, but with records accumulated from historical runs.

In one observed scenario, a node accumulated 3k records, normally quoting 10/20 nanos (with 1.1k close_range_records calculated), quoted 4million nanos (with all 3k records calculated as close_range_records) as during the restart when its RT only got 56 peers (network size estimated to be just 76 nodes; where normally shall be 230 peers in RT and network size to be over 30k).

This PR is an attempt to:
* have a background thread checks how many records accumulated so far, with an interval say 1 hour
* once reached 60% of the max_record, trigger clean up
* remove ALL records that not within the close_range that got set

### Type of Change

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -448,6 +448,48 @@ impl NodeRecordStore {
 
         Ok(())
     }
+
+    // When the accumulated record copies exceeds the `expotional pricing point` (max_records * 0.6)
+    // those `out of range` records shall be cleaned up.
+    // This is to avoid `over-quoting` during restart, when RT is not fully populated,
+    // result in mis-calculation of relevant records.
+    pub fn cleanup_unrelevant_records(&mut self) {
+        let accumulated_records = self.records.len();
+        if accumulated_records < 6 * MAX_RECORDS_COUNT / 10 {
+            return;
+        }
+
+        let responsible_range = if let Some(range) = self.responsible_distance_range {
+            range
+        } else {
+            return;
+        };
+
+        let mut removed_keys = Vec::new();
+        self.records.retain(|key, _val| {
+            let kbucket_key = KBucketKey::new(key.to_vec());
+            let is_in_range =
+                responsible_range >= self.local_key.distance(&kbucket_key).ilog2().unwrap_or(0);
+            if !is_in_range {
+                removed_keys.push(key.clone());
+            }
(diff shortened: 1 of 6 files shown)


sn_networking/src/record_store.rs
```diff
@@ -448,6 +448,48 @@ impl NodeRecordStore {
...
+        let mut removed_keys = Vec::new();
+        self.records.retain(|key, _val| {
+            let kbucket_key = KBucketKey::new(key.to_vec());
+            let is_in_range =
+                responsible_range >= self.local_key.distance(&kbucket_key).ilog2().unwrap_or(0);
+            if !is_in_range {
+                removed_keys.push(key.clone());
+            }
+            is_in_range
+        });
```
RolandSherwin, line 477: I think it is better to have the remove functionality at one place. `self.remove(key)` also removes it from `self.record` + cleans up from FS, so we could just do `self.remove(key)` instead of `retain` here.
We might not gain much time saving from `retain` as well as `self.remove()` interacts with fs anyway.

maqi: the reason of remove those keys separately here, is because the `remove` function will try to re-calculate furthest_key, if the to_be_removed key is the current furthest_key.
which, could cause duplicated costly calculation if call remove function individually.

meanwhile, `remove a non-existing entry` is not that costive.

RolandSherwin: Cool, that makes sense!

RolandSherwin (approved):

willief:
Comment from @neo in the forum. @neo is not regitered  on GitHub
Pasted here with his approval.
`
There is one issue I see here, if you remove the chunks rather than just mark them as inactive then you lose the ability to reactivate those chunks when nodes near by disappear and this node becomes responsible for them again. This would affect good nodes too and lose ability from the network.`

maqi:
> you lose the ability to reactivate those chunks when nodes near by disappear and this node becomes responsible for them again

The `relevant range` is far wider than the `close group range`, which means those `un-relevant` chunks shall no longer be useful again, unless there is a disaster shutdown large percentage of nodes.

PR #2044, opened 2024-08-15
feat(network): External Address manager

RolandSherwin:
- If you're a public node, you should blindly add the external address that is reported by others. Instead, you should add it only if the same address has been reported multiple times.
- Also should check for change of IP address.

### DIFF
--- a/sn_networking/src/external_address.rs
+++ b/sn_networking/src/external_address.rs
@@ -1,0 +1,396 @@
+// Copyright 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+#![allow(dead_code)]
+
+use crate::{driver::NodeBehaviour, mulitaddr_get_ip, multiaddr_is_global};
+use itertools::Itertools;
+use libp2p::{multiaddr::Protocol, Multiaddr, PeerId, Swarm};
+use std::{collections::HashMap, net::IpAddr};
+
+const MAX_REPORTS_BEFORE_CONFIRMATION: u8 = 3;
+const MAX_CANDIDATES: usize = 50;
+
+/// Manages the external addresses of a Public node. For a relayed node, the RelayManager should deal with
+/// adding and removing external addresses. We don't manage "local" addresses here.
+// TODO:
+// 1. if the max candidate is reached, kick out the oldest candidate sorted by # of reports
+#[derive(Debug)]
+pub struct ExternalAddressManager {
+    /// All the external addresses of the node
+    address_states: Vec<ExternalAddressState>,
+    current_ip_address: Option<IpAddr>,
+
+    /// The maximum number of reports before an candidate address is confirmed
+    pub max_reports_before_confirmation: u8,
+    /// The maximum number of candidates to store
+    pub max_candidates: usize,
+
+    pub peer_id: PeerId,
+}
+
+impl ExternalAddressManager {
+    pub fn new(peer_id: PeerId) -> Self {
+        Self {
+            address_states: Vec::new(),
+            current_ip_address: None,
+            max_reports_before_confirmation: MAX_REPORTS_BEFORE_CONFIRMATION,
+            max_candidates: MAX_CANDIDATES,
+            peer_id,
+        }
+    }
+
+    pub fn set_reports_before_confirmation(&mut self, max_reports_before_confirmation: u8) {
+        self.max_reports_before_confirmation = max_reports_before_confirmation;
+    }
+
+    pub fn set_max_candidates(&mut self, max_candidates: usize) {
+        self.max_candidates = max_candidates;
+    }
+
+    /// Get the list of candidate addresses
+    pub fn get_candidate_addresses(&self) -> Vec<&Multiaddr> {
+        self.address_states
+            .iter()
+            .filter_map(|state| {
+                if let ExternalAddressState::Candidate { address, .. } = state {
+                    Some(address)
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Get the list of confirmed addresses
+    pub fn get_confirmed_addresses(&self) -> Vec<&Multiaddr> {
+        self.address_states
+            .iter()
+            .filter_map(|state| {
+                if let ExternalAddressState::Confirmed { address, .. } = state {
+                    Some(address)
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Add an external address candidate to the manager.
+    /// If the address has been reported often enough, it is confirmed and added to the swarm.
+    /// If a new IP address has been reported often enough, then we switch to the new IP address and discard the old
+    /// external addresses.
+    pub fn add_external_address_candidate(
+        &mut self,
+        address: Multiaddr,
+        swarm: &mut Swarm<NodeBehaviour>,
+    ) {
+        if !multiaddr_is_global(&address) {
+            debug!("Address is not global, ignoring: {address:?}");
+            return;
+        }
+
+        let Some(address) = self.craft_external_address(&address) else {
+            debug!("Address is ill formed, not added to manager: {address:?}");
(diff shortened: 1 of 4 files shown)


sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+    },
+    Confirmed {
+        address: Multiaddr,
+        num_reports: u8,
+        ip_address: IpAddr,
+    },
+}
+
+impl ExternalAddressState {
+    fn get_multiaddr(&self) -> &Multiaddr {
```
b-zee: Convention in Rust is to not prefix with `get_`. So just `multiaddr()`.

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+            self.get_multiaddr(),
+            self.get_num_reports(),
+        );
+        match self {
+            Self::Candidate { num_reports, .. } => *num_reports = num_reports.saturating_add(1),
+            Self::Confirmed { num_reports, .. } => *num_reports = num_reports.saturating_add(1),
+        }
+    }
+
+    fn get_num_reports(&self) -> u8 {
```
b-zee:
```suggestion
    fn num_reports(&self) -> u8 {
```

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+        match self {
+            Self::Candidate { num_reports, .. } => *num_reports,
+            Self::Confirmed { num_reports, .. } => *num_reports,
+        }
+    }
+
+    fn get_ip_address(&self) -> IpAddr {
```
b-zee:
```suggestion
    fn ip_address(&self) -> IpAddr {
```

sn_networking/src/external_address.rs
b-zee:
```suggestion
    fn multiaddr(&self) -> &Multiaddr {
```

sn_networking/src/lib.rs
```diff
@@ -992,6 +994,19 @@ pub(crate) fn multiaddr_strip_p2p(multiaddr: &Multiaddr) -> Multiaddr {
...
+    for addr in multiaddr.iter() {
+        if let Protocol::Ip4(ip) = addr {
+            return Some(IpAddr::V4(ip));
+        }
+        if let Protocol::Ip6(ip) = addr {
+            return Some(IpAddr::V6(ip));
+        }
+    }
+    None
+}
```
b-zee: Alternative as I saw you used `find_map` before :wink:. (And typo in function name.)

```suggestion
fn multiaddr_get_ip(addr: &Multiaddr) -> Option<IpAddr> {
    addr.iter().find_map(|p| match p {
        Protocol::Ip4(addr) => Some(IpAddr::V4(addr)),
        Protocol::Ip6(addr) => Some(IpAddr::V6(addr)),
        _ => None,
    })
}
```

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+            max_reports_before_confirmation: MAX_REPORTS_BEFORE_CONFIRMATION,
+            max_candidates: MAX_CANDIDATES,
+        }
+    }
+
+    pub fn set_reports_before_confirmation(&mut self, max_reports_before_confirmation: u8) {
+        self.max_reports_before_confirmation = max_reports_before_confirmation;
+    }
+
+    pub fn set_max_candidates(&mut self, max_candidates: usize) {
```
b-zee: Maybe leave out setters we don't need for now?

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+    pub fn new() -> Self {
+        Self {
+            address_states: Vec::new(),
+            current_ip_address: None,
+            max_reports_before_confirmation: MAX_REPORTS_BEFORE_CONFIRMATION,
+            max_candidates: MAX_CANDIDATES,
+        }
+    }
+
+    pub fn set_reports_before_confirmation(&mut self, max_reports_before_confirmation: u8) {
```
b-zee: Leave out setters.

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+        self.max_reports_before_confirmation = max_reports_before_confirmation;
+    }
+
+    pub fn set_max_candidates(&mut self, max_candidates: usize) {
+        self.max_candidates = max_candidates;
+    }
+
+    /// Get the list of candidate addresses
+    pub fn get_candidate_addresses(&self) -> Vec<&Multiaddr> {
```
b-zee: Again, no `get_` (https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)

```suggestion
    pub fn candidates(&self) -> Vec<&Multiaddr> {
```

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+                    Some(address)
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    /// Get the list of confirmed addresses
+    pub fn get_confirmed_addresses(&self) -> Vec<&Multiaddr> {
```
b-zee:
```suggestion
    pub fn confirmed_addresses(&self) -> Vec<&Multiaddr> {
```

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+        let mut new_ip_map = HashMap::new();
+
+        for state in &self.address_states {
+            if let ExternalAddressState::Candidate {
+                ip_address,
+                num_reports,
+                ..
+            } = state
+            {
+                if let Some(current_ip_address) = self.current_ip_address {
```
b-zee: Move this `if` up, so we only do it once, then loop.

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+const MAX_CANDIDATES: usize = 50;
+
+/// Manages the external addresses of a Public node. For a relayed node, the RelayManager should deal with
+/// adding and removing external addresses.
+// TODO:
+// 1. if the max candidate is reached, kick out the oldest candidate sorted by # of reports
+#[derive(Debug)]
+pub struct ExternalAddressManager {
+    /// All the external addresses of the node
+    address_states: Vec<ExternalAddressState>,
```
b-zee, line 26: Maybe it's better to use a `HashMap<Multiaddr, State>` here? This would guarantee the uniqueness per element by multiaddr.

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+    current_ip_address: Option<IpAddr>,
+
+    /// The maximum number of reports before an candidate address is confirmed
+    pub max_reports_before_confirmation: u8,
```
b-zee: Couldn't we turn these into constants? What is the reasoning behind making these runtime variables?

sn_networking/src/external_address.rs
```diff
@@ -0,0 +1,261 @@
...
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
+    pub fn add_external_address(&mut self, address: Multiaddr, swarm: &mut Swarm<NodeBehaviour>) {
+        debug!("Adding external address to manager: {address:?}");
+        // TODO: have a filter to filter out ill formed addresses.
```
b-zee: I think it would be good to have a normalize function, that strips the `p2p`? I think I've seen mostly two forms of multiaddr, one with `p2p` and without.

At least, it is assumed we're not talking about `p2p-circuit` addresses here, right?

b-zee (changes requested):
Left a couple of comments.

Core logic is good I suppose, but would love to see it tested in practice.

sn_networking/src/external_address.rs
RolandSherwin: Yes! Sounds like a good idea!

sn_networking/src/external_address.rs
RolandSherwin: Yeah, I've made the modification to do just that. `craft_external_address`.

And yes, we don't deal with `p2p-circuit` addresses here, those are managed by `relay manager`.

sn_networking/src/external_address.rs
RolandSherwin: Will make the changes after testing this out!

RolandSherwin: Tried making the change, but I prefer the way with just storing the `Vec<States>`, the state contains the multiaddr and I prefer that one instead of storing it at the key OR duplicating it.

b-zee (approved):

PR #2052, opened 2024-08-19
feat(launchpad): list disks with no access

mazzi:
### Description

We show on the drive selection screen, disks that do not have read or write access, replacing the available space column with a legend "No Access", to give feedback to the user.

These drives cannot be selected.

Screenshot 2024-08-19 at 15 18 21


node-launchpad/src/components/popup/change_drive.rs
```diff
@@ -558,6 +563,18 @@ enum DriveStatus {
     Selected,
     #[default]
     NotSelected,
+    NotAvailable,
+}
+
+impl PartialEq for DriveStatus {
```
RolandSherwin: Can we not just derive the trait `#[derive(PartialEq)]` on the enum?

RolandSherwin (approved):

PR #2073, opened 2024-08-29
feat(protocol): add Scratchpad data type

joshuef:


sn_node/src/put_validation.rs
```diff
@@ -12,7 +12,7 @@ use sn_networking::{get_raw_signed_spends_from_record, GetRecordError, NetworkEr
 use sn_protocol::{
     storage::{
         try_deserialize_record, try_serialize_record, Chunk, RecordHeader, RecordKind, RecordType,
-        SpendAddress,
+        Scratchpad, SpendAddress,
```
dirvine, line 15: I am not sure of this name, it feels too loose. Perhaps Account Packet is more clear?

joshuef: I think it's worth there being a separation between the data type on the network and the data being stored.

There's nothing that says this _must_ be an account packet. There could be other usecases for this eg.

So I've been leaning into `AccountPacket` being the data structure + representation of a user's `Account` space (or whatever it ends up being called).

And with that I feel this data type needs a more generic name. 


`Scratchpad` i agree doesn't feel the best, but was the best I could come up with for that. Happy to change to something else to better describe this.

b-zee: How about `FluidPacket`, `MutableData`, `DynamicData` or `FlexiPacket`?

sn_networking/src/record_store.rs
```diff
@@ -1201,6 +1205,147 @@ mod tests {
         assert!(store.get(&r.key).is_none());
     }
 
+    #[tokio::test]
+    async fn can_store_and_retrieve_chunk() {
```
b-zee, line 1159: Is this test relevant to this PR?

sn_protocol/src/storage/scratchpad.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.address.xorname()
+    }
+
+    /// Returns size of contained encrypted_data.
+    pub fn payload_size(&self) -> usize {
+        self.encrypted_data.len()
+    }
+
+    /// Returns size of this scratchpad after serialisation.
+    pub fn serialised_size(&self) -> usize {
```
b-zee: Can we put this in a test to verify this is correct? Also, I don't see this method used, so do we actually need this?

sn_networking/src/record_store.rs
joshuef: No, but was added as I added one for SP and thought it a good addition.

sn_protocol/src/storage/scratchpad.rs
joshuef: removed!

sn_node/src/put_validation.rs
joshuef: After some chat, we agreed on keeping `Scratchpad` as it is reflective of the usecase

b-zee (approved):

PR #2076, opened 2024-08-29
feat(launchpad): connection mode and fixes

mazzi:
### Description

* We create a new option under `Options` screen called `Connection Mode`, were the user can switch between `Automatic`, `Home Network`, `UPnP`, `Custom Ports`.
* We fixed a bug and improve navigability.
* We introduce error popups.
* We unified screen styles.

#### NAT detection
Now we use the latest `nat-detection` crate on `crates.io`. Thanks to a [bug fixed](https://github.com/maidsafe/sn-releases/pull/23) on sn-releases.

#### Error Handling
A new module was created to show error messages on popups. Can be plugged on any component. Documentation on the file.

#### Port Popup
It is a separate popup, in opposition to the established strategy in `Drive Selection` and `Beta Programme`, were follow up popups are self contained on the same file. The reason for this is that we can invoke the `Port Selection` popup directly from the `Options` screen.

#### Node mgmt
A couple of refactors there were we created helper functions to be able to understand better the algorithm.

### Bug fixes
The application will run despite `app_config.json` doesn't have certain keys related to mountpoints.

### Dependencies
Updated `sn-releases` to `0.2.8`.

### Testing
Desktop testing was made. Mostly on macOS. We will need to test more on Windows/Linux.

### Screenshots

Screenshot 2024-08-29 at 16 57 33  Screenshot 2024-08-29 at 16 57 43

Screenshot 2024-08-29 at 16 57 52  Screenshot 2024-08-29 at 16 58 03 |


joshuef (approved):
Tested on mac, code looks good! Nice one @mazzi 

Although I dont understand how to get to the port popup there 🤔

mazzi:
> Tested on mac, code looks good! Nice one @mazzi

Thanks!
 
> Although I dont understand how to get to the port popup there 🤔

When `Custom Ports` is selected, the follow up screen should be `Custom Ports` popup. Anytime with `Custom Ports` selected as `Connection mode`, you can go to `Options` screen, hit `Ctrl+P` and you should be able to edit your custom ports. In the other modes of course, the Port selection popup is disabled.

joshuef (dismissed):
Aha, with the above, I cannot save my custom port range. Hitting enter does nothing for me 🤔

mazzi:
> Aha, with the above, I cannot save my custom port range. Hitting enter does nothing for me 🤔

Most likely your port range is out of a valid one.
We only enable `Enter` when the range is between `PORT_MIN (1024)` and `PORT_MAX (65535)`
Perhaps we will need to provide more feedback to the user...

mazzi:
> Aha, with the above, I cannot save my custom port range. Hitting enter does nothing for me 🤔

I think that your claim is valid as you didn't know this screen beforehand. We added feedback to the user in case that the port range is invalid 👍

joshuef (approved):

PR #2080, opened 2024-09-02
fix(launchpad): storage mountpoint for node data is not set

mazzi:
### Description

Branch from `stable` (it's a hotfix then)

A fix for the error `Storage Mountpoint for node data is not set` was implemented. We modified the code to be able to execute tests injecting the configuration path on the main app.

Tests were written for the following cases:
* Happy case
* Configuration file not found (new installation)
* Configuration exists but with wrong mountpoint defined
* Configuration exists but mountpoint is not set or doesn't exists (case were we upgrade LP version)

Internal tests for configuration parsing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1017,7 +1017,7 @@ dependencies = [
  "js-sys",
  "num-traits",
  "wasm-bindgen",
- "windows-targets 0.52.5",
+ "windows-targets 0.52.6",
 ]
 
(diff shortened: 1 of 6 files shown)


mickvandijke (approved):
Great fix, well done! 🚀

mazzi:
Good catch @jacderida !!
Thanks for this

PR #2082, opened 2024-09-03
feat(networking): initial range based gets implementation

joshuef:
This pull request includes significant updates to the GitHub Actions workflows and several enhancements and fixes in the `sn_client` codebase. The changes improve the reliability and functionality of the workflows and add new methods and error handling to the client code.

### Workflow Improvements:

* [`.github/workflows/merge.yml`](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L768-R768): Updated the `log_file_prefix` and restructured the wallet creation and funding steps to include cleanup and verification processes. [[1]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L768-R768) [[2]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L894-R925)
* [`.github/workflows/merge.yml`](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1R1136-R1137): Added environment variables and a sleep step to ensure proper setup before running subsequent steps. [[1]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1R1136-R1137) [[2]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1R1214-R1227)
* [`.github/workflows/merge.yml`](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L1207-R1254): Improved directory checks and ensured no leftover files before proceeding with tests. [[1]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L1207-R1254) [[2]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L1227-R1267)

### Client Enhancements:

* [`sn_client/src/api.rs`](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L17-R25): Introduced a new method `get_range` and updated existing methods to include verification flags and enhanced error handling. [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L17-R25) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97R309-R313) [[3]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L439-R497)
* [`sn_client/src/api.rs`](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L774-R821): Modified quorum settings and retry strategies for network operations to improve reliability. [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L774-R821) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L808-R855) [[3]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97R895-R921)

### Error Handling Improvements:

* [`sn_client/src/error.rs`](diffhunk://#diff-a49580c567e54fb263fcfe3ab913094367d70f0a12b3681f9c65c197856e987dR16): Added new error types for faucet disbursement and no cash note found scenarios. [[1]](diffhunk://#diff-a49580c567e54fb263fcfe3ab913094367d70f0a12b3681f9c65c197856e987dR16) [[2]](diffhunk://#diff-a49580c567e54fb263fcfe3ab913094367d70f0a12b3681f9c65c197856e987dR27-R28) [[3]](diffhunk://#diff-a49580c567e54fb263fcfe3ab913094367d70f0a12b3681f9c65c197856e987dR51-R53)
* [`sn_client/src/faucet.rs`](diffhunk://#diff-e2aab3de22953c7d7077a1164e3957ef9b0c70d3631e8a44357f5c4e372609f9L41-R55): Implemented retry logic for loading cash notes from the faucet wallet.

### Register Handling:

* [`sn_client/src/register.rs`](diffhunk://#diff-1b86a48cb70322a3ac64029081787ac6c92b95fe1878ae05e65327b224bc787cL772-R774): Updated methods to use the new verification flag when retrieving signed registers from the network. [[1]](diffhunk://#diff-1b86a48cb70322a3ac64029081787ac6c92b95fe1878ae05e65327b224bc787cL772-R774) [[2]](diffhunk://#diff-1b86a48cb70322a3ac64029081787ac6c92b95fe1878ae05e65327b224bc787cL865-R869)

### Test Utilities:

* [`sn_client/src/test_utils.rs`](diffhunk://#diff-0e7d4cd94adcddff8cbd9e20764d29321aadb7695c473e24f560a8d67fbf8fd7L32-R32): Increased the number of retries for loading the faucet wallet in tests.

### Wallet Enhancements:

* [`sn_client/src/wallet.rs`](diffhunk://#diff-d655bcffabb264727633e45f79aff0b7990bb9045b3742352968837f26adca30L1012-R1012): Refined error handling for scenarios where all redeemed cash notes are already spent.

### DIFF
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -14,7 +14,7 @@ use super::{
 use bls::{PublicKey, SecretKey, Signature};
 use libp2p::{
     identity::Keypair,
-    kad::{Quorum, Record},
+    kad::{KBucketDistance, Quorum, Record},
     Multiaddr, PeerId,
 };
 use rand::{thread_rng, Rng};
@@ -303,6 +303,11 @@ impl Client {
     /// ```
     pub fn events_channel(&self) -> ClientEventsReceiver {
         self.events_broadcaster.subscribe()
+    }
+
+    /// Return the underlying network GetRange
+    pub async fn get_range(&self) -> Result<KBucketDistance> {
+        self.network.get_range().await.map_err(Error::from)
     }
 
     /// Sign the given data.
@@ -854,18 +859,26 @@ impl Client {
 
         // When there is retry on Put side, no need to have a retry on Get
         let verification_cfg = GetRecordCfg {
-            get_quorum: Quorum::Majority,
+            get_quorum: Quorum::All,
             retry_strategy: None,
             target_record: record_to_verify,
             expected_holders,
             is_register: false,
         };
+
+        let verification = if verify_store {
+            Some((VerificationKind::Network, verification_cfg))
+        } else {
+            None
+        };
+
         let put_cfg = PutRecordCfg {
-            put_quorum: Quorum::Majority,
+            put_quorum: Quorum::All,
             retry_strategy: Some(RetryStrategy::Persistent),
             use_put_record_to: None,
-            verification: Some((VerificationKind::Network, verification_cfg)),
+            verification,
         };
+
         Ok(self.network.put_record(record, &put_cfg).await?)
     }
 
@@ -902,7 +915,7 @@ impl Client {
         self.try_fetch_spend_from_network(
             address,
             GetRecordCfg {
-                get_quorum: Quorum::Majority,
+                get_quorum: Quorum::All,
                 retry_strategy: Some(RetryStrategy::Balanced),
                 target_record: None,
                 expected_holders: Default::default(),
@@ -935,7 +948,7 @@ impl Client {
         self.try_fetch_spend_from_network(
             address,
             GetRecordCfg {
-                get_quorum: Quorum::Majority,
+                get_quorum: Quorum::All,
                 retry_strategy: None,
                 target_record: None,
                 expected_holders: Default::default(),
@@ -992,9 +1005,7 @@ impl Client {
             }
             Err(err) => {
                 warn!("Invalid signed spend got from network for {address:?}: {err:?}.");
-                Err(Error::CouldNotVerifyTransfer(format!(
-                    "Verification failed for spent at {address:?} with error {err:?}"
-                )))
+                Err(Error::from(err))
             }
         }
     }
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -894,6 +894,7 @@ impl Client {
                 .iter()
                 .cloned()
                 .collect();
+            info!("Expecting holders: {expected_holders:?}");
             (Some(record.clone()), expected_holders)
         } else {
             (None, Default::default())
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -21,7 +21,8 @@ use rand::{thread_rng, Rng};
 use sn_networking::{
     get_signed_spend_from_record, multiaddr_is_global,
     target_arch::{interval, spawn, timeout, Instant},
-    GetRecordCfg, NetworkBuilder, NetworkError, NetworkEvent, PutRecordCfg, VerificationKind,
+    GetRecordCfg, GetRecordError, NetworkBuilder, NetworkError, NetworkEvent, PutRecordCfg,
+    VerificationKind,
 };
 use sn_protocol::{
     error::Error as ProtocolError,
@@ -441,18 +442,59 @@ impl Client {
     /// let xorname = XorName::random(&mut rng);
     /// let address = RegisterAddress::new(xorname, owner);
     /// // Get a signed register
-    /// let signed_register = client.get_signed_register_from_network(address);
+    /// let signed_register = client.get_signed_register_from_network(address, true);
     /// # Ok(())
     /// # }
     /// ```
     pub async fn get_signed_register_from_network(
         &self,
         address: RegisterAddress,
+        is_verifying: bool,
     ) -> Result<SignedRegister> {
         let key = NetworkAddress::from_register_address(address).to_record_key();
+        let get_quorum = if is_verifying {
+            Quorum::All
+        } else {
+            Quorum::Majority
+        };
+        let retry_strategy = if is_verifying {
+            Some(RetryStrategy::Balanced)
+        } else {
+            Some(RetryStrategy::Quick)
+        };
+        let get_cfg = GetRecordCfg {
+            get_quorum,
+            retry_strategy,
+            target_record: None,
+            expected_holders: Default::default(),
+        };
 
-        let maybe_records = self.network.get_register_record_from_network(key).await?;
-        merge_register_records(address, &maybe_records)
+        let maybe_record = self.network.get_record_from_network(key, &get_cfg).await;
+        let record = match &maybe_record {
+            Ok(r) => r,
+            Err(NetworkError::GetRecordError(GetRecordError::SplitRecord { result_map })) => {
+                let mut results_to_merge = HashMap::default();
+
+                for (address, (r, _peers)) in result_map {
+                    results_to_merge.insert(*address, r.clone());
+                }
+
+                return merge_register_records(address, &results_to_merge);
+            }
+            Err(e) => {
+                warn!("Failed to get record at {address:?} from the network: {e:?}");
+                return Err(ProtocolError::RegisterNotFound(Box::new(address)).into());
+            }
+        };
+
+        debug!(
+            "Got record from the network, {:?}",
+            PrettyPrintRecordKey::from(&record.key)
+        );
+
+        let register = get_register_from_record(record)
+            .map_err(|_| ProtocolError::RegisterNotFound(Box::new(address)))?;
+        Ok(register)
     }
 
     /// Retrieve a Register from the network.
@@ -778,7 +820,7 @@ impl Client {
     /// ```
     pub async fn verify_register_stored(&self, address: RegisterAddress) -> Result<SignedRegister> {
         info!("Verifying register: {address:?}");
-        self.get_signed_register_from_network(address).await
+        self.get_signed_register_from_network(address, true).await
     }
 
     /// Quickly checks if a `Register` is stored by expected nodes on the network.
@@ -812,7 +854,7 @@ impl Client {
         address: RegisterAddress,
     ) -> Result<SignedRegister> {
         info!("Quickly checking for existing register : {address:?}");
-        self.get_signed_register_from_network(address).await
+        self.get_signed_register_from_network(address, false).await
     }
 
     /// Send a `SpendCashNote` request to the network. Protected method.
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -467,6 +467,7 @@ impl Client {
             retry_strategy,
             target_record: None,
             expected_holders: Default::default(),
+            is_register: true,
         };
 
         let maybe_record = self.network.get_record_from_network(key, &get_cfg).await;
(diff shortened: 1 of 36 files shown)


sn_client/src/api.rs
```diff
@@ -436,18 +442,59 @@ impl Client {
...
+                return Err(ProtocolError::RegisterNotFound(Box::new(address)).into());
+            }
+        };
+
+        debug!(
+            "Got record from the network, {:?}",
+            PrettyPrintRecordKey::from(&record.key)
+        );
 
-        let maybe_records = self.network.get_register_record_from_network(key).await?;
```
maqi, line 449: I think we shall stay with the current `get_closest` first, then fetch from individual and merge.
swich back to use `kad::get_record` could be problematic, as explained in the slack discussion.

will continue the discussion within the slack discussion thread.

also, this seems not related to this change, and better to be in a separate PR ?

sn_client/src/api.rs
```diff
@@ -849,17 +896,25 @@ impl Client {
 
         // When there is retry on Put side, no need to have a retry on Get
         let verification_cfg = GetRecordCfg {
-            get_quorum: Quorum::Majority,
+            get_quorum: Quorum::All,
```
maqi, line 906: having an ALL with a bit shaky network could slow down or even block the uploader process.
as there will always be one node failes the get verification.
Any particular reason HAVE TO swtich back to ALL ?

same to couple of other places

sn_client/src/wallet.rs
```diff
@@ -1050,23 +1048,40 @@ impl Client {
...
+        info!(
+            "parent spends for cn; {address:?}: {:?}",
+            &cash_note.parent_spends.len()
+        );
+
         for spend in &cash_note.parent_spends {
             let address = SpendAddress::from_unique_pubkey(spend.unique_pubkey());
-            debug!(
-                "Getting spend for pubkey {:?} from network at {address:?}",
+            warn!(
```
maqi: does it need to be as a `warn` ?
`info` shall be enough?

sn_client/src/wallet.rs
```diff
@@ -1050,23 +1048,40 @@ impl Client {
...
 
         let mut received_spends = std::collections::BTreeSet::new();
         for result in join_all(tasks).await {
-            let network_valid_spend =
-                result.map_err(|err| WalletError::CouldNotVerifyTransfer(err.to_string()))?;
+            let network_valid_spend = match result {
+                Ok(spend) => Ok(spend),
+                Err(error) => match error {
+                    Error::Network(sn_networking::NetworkError::DoubleSpendAttempt(spends)) => {
+                        warn!("DoubleSpentAttempt found with {spends:?}");
```
maqi: we shall change to use `BurntSpend` wherever we can ?

sn_networking/src/cmd.rs
```diff
@@ -543,6 +556,31 @@ impl SwarmDriver {
 
         Ok(())
     }
+
+    /// Return the RecordType
+    pub(crate) fn get_type_from_record(record: &Record) -> Result<RecordType> {
```
maqi: this seems shall be a helper function to be defined in the sn_transfer::Spend ?
or at least shall not be inside driver ?

sn_networking/src/cmd.rs
```diff
@@ -946,24 +965,72 @@ impl SwarmDriver {
...
+    }
+
+    /// From all local peers, returns any within current get_range for a given key
+    /// Excludes self
+    pub(crate) fn get_filtered_peers_exceeding_range_or_closest_nodes(
+        &mut self,
+        target_address: &NetworkAddress,
+    ) -> Vec<PeerId> {
+        let filtered_peers = self.get_filtered_peers_exceeding_range(target_address);
+        let cloest_node_buffer_zone = CLOSE_GROUP_SIZE + close_group_majority();
```
maqi: cloest -> closest

sn_networking/src/driver.rs
```diff
@@ -777,28 +789,25 @@ impl SwarmDriver {
                     // logging for handling events happens inside handle_swarm_events
                     // otherwise we're rewriting match statements etc around this anwyay
                     if let Err(err) = self.handle_swarm_events(swarm_event) {
-                        warn!("Error while handling swarm event: {err}");
+                        trace!("Issue while handling swarm event: {err}");
```
maqi: why lower the log level here?
if there is too frequent error message, it might mean there is mis-use ?
we shall mute those too frequent error by `match handle individually`, instead of lower log level for all ?

sn_networking/src/driver.rs
```diff
@@ -777,28 +789,25 @@ impl SwarmDriver {
...
-
-                        if let Some(distance) = self.get_responsbile_range_estimate(&closest_k_peers) {
-                            info!("Set responsible range to {distance}");
-                            // set any new distance to farthest record in the store
-                            self.swarm.behaviour_mut().kademlia.store_mut().set_distance_range(distance);
-                            // the distance range within the replication_fetcher shall be in sync as well
-                            self.replication_fetcher.set_replication_distance_range(distance);
-                        }
+                        let get_range = self.get_request_range();
+                        // set any new distance to farthest record in the store
```
maqi: the range is no longer the `farthest record` ?

sn_networking/src/driver.rs
```diff
@@ -810,32 +819,94 @@ impl SwarmDriver {
...
+        for kbucket in self.swarm.behaviour_mut().kademlia.kbuckets() {
+            for entry in kbucket.iter() {
+                let id = entry.node.key.clone().into_preimage();
+
+                if id != our_peer_id {
+                    all_peers.push(id);
+                }
+            }
+        }
+        all_peers.push(self.self_peer_id);
```
maqi: the function name says `get_all_local_peers_excluding_self` ?

sn_node/src/replication.rs
```diff
@@ -76,12 +79,27 @@ impl Node {
...
-                        }
+                        Err(error) => match error {
+                            sn_networking::NetworkError::DoubleSpendAttempt(spends) => {
+                                debug!("Failed to fetch record {pretty_key:?} from the network, double spend attempt {spends:?}");
+
+                                let bytes = try_serialize_record(&spends, RecordKind::Spend)?;
+
+                                Record {
+                                    key,
+                                    value: bytes.to_vec(),
```
maqi, line 99: this means seralize two spends into one Record?
the `try_serialize_records`, and the `get_spend_from_record` functions are all shall get refactored to accomodate this.

Also, say one reocord is having content [1,2], and the other is with content [1], then a get_record will result with `SplitRecord` with both, then how these two got merged ?

sn_networking/src/event/kad.rs
```diff
@@ -369,92 +382,98 @@ impl SwarmDriver {
...
+            });
+        }
+        Ok(())
+    }
 
-            debug!("Expecting {expected_answers:?} answers for record {pretty_key:?} task {query_id:?}, received {responded_peers} so far");
+    /// Checks passed peers from a request and checks they are sufficiently spaced to
+    /// ensure we have searched enough of the network range as determined by our `get_range`
+    ///
+    /// We expect any conflicting records to have been reported prior ti this check,
```
maqi: prior ti this -> prior this

sn_networking/src/event/kad.rs
```diff
@@ -369,92 +382,98 @@ impl SwarmDriver {
...
+    /// so we assume we're returning unique records only.
+    fn have_we_have_searched_thoroughly_for_quorum(
+        expected_get_range: KBucketDistance,
+        searched_peers_list: &HashSet<PeerId>,
+        data_key_address: &NetworkAddress,
+        quorum: &Quorum,
+    ) -> bool {
+        warn!("Assessing search: range: {:?}, address: {data_key_address:?}, quorum required: {quorum:?}, peers_returned_count: {:?}", expected_get_range.ilog2(), searched_peers_list.len());
```
maqi: info level ?

sn_networking/src/event/kad.rs
```diff
@@ -369,92 +382,98 @@ impl SwarmDriver {
...
-                    debug!("For record {pretty_key:?} task {query_id:?}, fetch completed with non-responded expected holders {:?}", cfg.expected_holders);
-                }
-                let cfg = cfg.clone();
+        // get the farthest distance between peers in the response
+        let mut max_distance_to_data_from_responded_nodes = KBucketDistance::default();
 
-                // Remove the query task and consume the variables.
-                let (_key, senders, result_map, _) = entry.remove();
+        // iterate over peers and see if the distance to the data is greater than the get_range
+        // Fathest peer from the data that has returned it
```
maqi: `Fathest peer from the data that has returned it`
this sentence can now be removed ?

maqi (changes requested):

sn_client/src/api.rs
joshuef: This is required for tests to pass, as the prior version does not work well with this. (We have no retry or majority of responses required... GetRange is naturally more probabilistic and has removed certain premature optimisation from the `get_record` calls which necessitate this I think.

`retry` there is inherently useful as we can hit more holders (now we have a potentially larger range).

So I think this approach is both simpler (`get_record` logic remains the same throughout calls to it), and healthier (going off the CI with/without)

sn_client/src/api.rs
joshuef: > as there will always be one node failes the get verification

Why? 

Right now with `GetRange` all == `CloseGroupSize` as opposed to all returners... so it allows for stricter early exit.

Really though it's not so much about `All` so much as there being a difference in the checking between `verify` or not... So it could be `majority` in both cases and we just remove the `verify` flag on the fn perhaps? (Which I do note is as main... but it seemed handy to have this distinction to me...)

sn_networking/src/cmd.rs
joshuef: It's unrelated to transfers. moved to protocol 👍

sn_node/src/replication.rs
joshuef: > then how these two got merged ?

That's for node layer logic outwith of the get function now. (So the register apis eg, or the spend validation steps which already do this)

joshuef: > this means seralize two spends into one Record?

This is already how this is stored, so there should be no change here. We're just ensuring that we do this on replication now where we did not previously (see `get_record_from_network` handling of double spend)

PR #2090, opened 2024-09-05
test(networking): speed up and simplify address sim

joshuef:
This should open the door to the sim being muuuuch larger. 

goes from 65s on main-> 0.1s on my machine.... 

Could well be I've messed up somethiing in the sim to get that... But it appears to get similar output to pain (still fails a lot with same params, i'd increase the iterations or counts eg...

This pull request introduces significant improvements and refactoring to the `sn_networking/src/record_store.rs` file, particularly within the test module. The changes focus on simplifying the code, improving performance, and enhancing readability. The most important changes include removing unused imports, refactoring the `address_distribution_sim` test, and updating the `pick_cheapest_payee` function.

### Code Simplification and Cleanup:
* Removed unused imports such as `sort_peers_by_key` and `BTreeMap` from the test module (`sn_networking/src/record_store.rs`).

### Refactoring:
* Introduced the `PeerStats` struct to encapsulate peer statistics, replacing the previous `HashMap` approach. This improves code readability and maintainability (`sn_networking/src/record_store.rs`).
* Refactored the `address_distribution_sim` test to use the `PeerStats` struct and `rayon` for parallel iteration, enhancing performance and clarity (`sn_networking/src/record_store.rs`).
* Updated the `pick_cheapest_payee` function to work with the new `PeerStats` struct and use a more efficient approach to find the cheapest payee (`sn_networking/src/record_store.rs`).

### Performance Improvements:
* Utilized `rayon::prelude::*` for parallel iteration in the `address_distribution_sim` test, significantly improving the performance of the simulation (`sn_networking/src/record_store.rs`).
* Optimized the sorting and partitioning logic within the `address_distribution_sim` test to efficiently find and sort peers based on their distance to a target address (`sn_networking/src/record_store.rs`).

These changes collectively enhance the codebase by making it more efficient, easier to understand, and maintain.

### DIFF
--- a/sn_networking/src/record_store.rs
+++ b/sn_networking/src/record_store.rs
@@ -1429,13 +1429,15 @@ mod tests {
     fn address_distribution_sim() {
         use rayon::prelude::*;
 
-        let num_of_peers = 100_000;
-        let num_of_chunks_per_hour = 100_000;
-        let max_hours = 500;
+        // as network saturates, we can see that peers all eventually earn similarly
+        let num_of_peers = 5_000;
+        let num_of_chunks_per_hour = 1_000_000;
+        let max_hours = 50;
 
         //
-        let k = K_VALUE;
-        let close_group_size = k.get() / 3;
+        let k = K_VALUE.get();
+
+        let replication_group_size = k / 3;
 
         // Initialize peers with random addresses
         let mut peers: Vec<PeerStats> = (0..num_of_peers)
@@ -1481,26 +1483,28 @@ mod tests {
                     });
 
                     // Collect close_group_size closest peers
-                    let mut close_group = Vec::with_capacity(close_group_size);
+                    let mut close_group = Vec::with_capacity(replication_group_size);
                     let mut left = partition_point;
                     let mut right = partition_point;
 
-                    while close_group.len() < close_group_size && (left > 0 || right < peers_len) {
+                    while close_group.len() < replication_group_size
+                        && (left > 0 || right < peers_len)
+                    {
                         if left > 0 {
                             left -= 1;
                             close_group.push(left);
                         }
-                        if close_group.len() < close_group_size && right < peers_len {
+                        if close_group.len() < replication_group_size && right < peers_len {
                             close_group.push(right);
                             right += 1;
                         }
                     }
 
                     // Truncate to ensure we have exactly close_group_size peers
-                    close_group.truncate(close_group_size);
+                    close_group.truncate(replication_group_size);
 
                     // Find the cheapest payee among the close group
-                    let Ok(payee_index) = pick_cheapest_payee(&peers, &close_group) else {
+                    let Ok((payee_index, cost)) = pick_cheapest_payee(&peers, &close_group) else {
                         bail!("Failed to find a payee");
                     };
 
@@ -1509,10 +1513,8 @@ mod tests {
                         peer.records_stored.fetch_add(1, Ordering::Relaxed);
 
                         if peer_index == payee_index {
-                            let cost = calculate_cost_for_records(
-                                peer.records_stored.load(Ordering::Relaxed),
-                            );
-                            peer.nanos_earned.fetch_add(cost, Ordering::Relaxed);
+                            peer.nanos_earned
+                                .fetch_add(cost.as_nano(), Ordering::Relaxed);
                             peer.payments_received.fetch_add(1, Ordering::Relaxed);
                         }
                     }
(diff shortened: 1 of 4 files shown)


sn_networking/src/record_store.rs
```diff
@@ -1461,120 +1460,140 @@ mod tests {
...
+        use rayon::prelude::*;
         let num_of_peers = 2000;
         let num_of_chunks_per_itr = 2000;
-
-        for _ in 0..num_of_peers {
-            let peer_id = PeerId::random();
-            let _ = peers.insert(peer_id, (0, 0, 0));
-            peers_vec.push(peer_id);
-        }
+        let close_group_size = close_group_majority().min(num_of_peers);
```
maqi: I think this shall be replication_range to be used.
in practical, this shall be a kbucket_index
in the sim, we used REPLICATION_PEERS_COUNT (7) to simulate, which shall actually be updated to (K_VALUE/2) to be more close to the real situation.
but change to majority of close_group_size (which is just 3) will be too aggressive

sn_networking/src/record_store.rs
```diff
@@ -1461,120 +1460,140 @@ mod tests {
...
+                    let partition_point = peers.partition_point(|peer| {
+                        sorting_target_address.distance(&peer.address) < chunk_distance_to_sorting
+                    });
+
+                    // Collect close_group_size closest peers
+                    let mut close_group = Vec::with_capacity(close_group_size);
+                    let mut left = partition_point;
+                    let mut right = partition_point;
+
+                    while close_group.len() < close_group_size && (left > 0 || right < peers_len) {
```
maqi: the `partition_point` returns the first index that not satisfy the judgement

using this sorted list against a random target, to figure out closest peers against another target,
the current code block has following glitches:
1, the `left` side shall be preferred, as their `distance interval` is smaller than the `right` side peers
2, the distance from different observer will be different, i.e. the peers around the partition_point not guaranteered to always being the closest peers to the new target
3, the while loop will terminate earlier when the partition_point  falls into the beginning ([0,1,2]) or end ([len-2, len-1, len]) of the entire peers list.

for the point 1 and 2, the `glitch` might be acceptable, as the test is just a `sim`

sn_networking/src/record_store.rs
```diff
@@ -1461,120 +1460,140 @@ mod tests {
...
+                        if right < peers_len && close_group.len() < close_group_size {
+                            close_group.push(right);
+                            right += 1;
                         }
                     }
-                    Err(err) => {
-                        panic!("Can't find replicate range of {name:?} with error {err:?}")
+
+                    // Sort the close group by distance to the chunk
+                    close_group.sort_by_key(|&i| chunk_address.distance(&peers[i].address));
```
maqi: there is `while close_group.len() < close_group_size` in the above
which means this sort and truncate is not necessary.

unless you mean the previous loop shall use something like `while close_group.len() < 5 * close_group_size` to reduce the chance of glitch that described above ?

sn_networking/src/record_store.rs
```diff
@@ -1461,120 +1460,140 @@ mod tests {
...
+                    close_group.truncate(close_group_size);
+
+                    let payee_index = pick_cheapest_payee(&peers, &close_group);
+
+                    (close_group, payee_index)
+                })
+                .collect();
+
+            // Update peers stats
+            for (peers_in_replicate_range, payee_index) in chunk_results {
```
maqi: the `pick_cheapest_payee` uses the payment_received & chunks_stored to calculate cost and payee
which means the peer_status shall be updated each time when a chunk updated

hence, I think this `peer stats update block` shall be inside the above while loop, not outside

maqi (changes requested):

sn_networking/src/record_store.rs
joshuef: Okay, moved it into a sequential loop after detecting the desired peers just now 👍

sn_networking/src/record_store.rs
```diff
@@ -1461,199 +1412,205 @@ mod tests {
...
-                panic!("Cannot find stats of {peer:?}");
-            }
-        }
-
-        if let Some(peer_id) = payee {
-            peer_id
-        } else {
-            panic!("Cannot find cheapest payee among {peers_in_close:?}");
-        }
+    fn pick_cheapest_payee(peers: &[PeerStats], close_group: &[usize]) -> usize {
```
maqi: ideally, shall use the same `network::get_fees_from_store_cost_responses` function

maqi (changes requested):

sn_networking/src/record_store.rs
joshuef: hmm, good shout. This algo is wayyy less fair than just choosing lowest cost + no tiebreak atm... 

I guess as we pass in sorted peers (from the test), so the lowest index is effectively random on the chunk store addr. 

But then when we compare addresses, we actually end up always choosing the same... 

Hmmm. If the algo is just "random", that may actually be fine (i mean here, we dont rely on the order passed in, but not doing anything deterministic because we dont need to?)

maqi (approved):

PR #2093, opened 2024-09-05
feat: client API first iteration

b-zee:
Initial implementation of high level API


joshuef (approved):
the test compile needs fixed eg https://github.com/maidsafe/safe_network/actions/runs/10791386235/job/29928681734?pr=2093 

but once thats in gtg!

PR #2096, opened 2024-09-06
fix: replace kad::put_record_to with messaging

maqi:
### Description

the kad::put_record_to will bypass the bridging for home nodes.
which result in home nodes can not receive ChunkWithPayment sent to it.

This PR replaces kad::put_record_to with messaging, which allows bridging. 

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -180,13 +180,6 @@ pub enum NetworkSwarmCmd {
         sender: oneshot::Sender<Result<()>>,
         quorum: Quorum,
     },
-    /// Put record to specific node
-    PutRecordTo {
-        peers: Vec<PeerId>,
-        record: Record,
-        sender: oneshot::Sender<Result<()>>,
-        quorum: Quorum,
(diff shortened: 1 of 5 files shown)


joshuef:
Hmmm, why does kad fail here? Is it because the RT is actually not including these nodes? If so I think we'd want to solve that... 

I'm not sure us sidestepping libp2p is the ideal here...

(marking as donot merge for discussion)

maqi:
> why does kad fail here? Is it because the RT is actually not including these nodes?

Not sure as it really sits deep in the libp2p code.
As it is a client trying to `put_record_to` a private_node close to the target, it is not guaranteered that node shall be in client's RT (no matter the node to be private or public).
As the req/resp seems work fine (as the store_cost query/response can be received by both side.)
Hence I have to doubt it is `libp2p::kad::put_record_to` hit some libp2p hidden issue.

Meanwhile, I even doubt whether the `libp2p::kad::get_record` has the same issue, i.e. the private node not even received the query, even they got the copy via replication.

PR #2100, opened 2024-09-11
chore(cli_aut): feat flag to avoid clippy errors

joshuef:
This pull request includes changes to the `autonomi` project to enhance feature management and conditional compilation. The most important changes involve updating feature dependencies in the `Cargo.toml` file and adding conditional compilation attributes to the `lib.rs` file.

Feature management updates:

* [`autonomi/Cargo.toml`](diffhunk://#diff-ce41b407dcc9b93d4f03678fae5cac35039e04bfd773c3b7b7a263daf6077edcL7-R10): Updated the `default` feature to include "data" and modified the `files` feature to depend on both "transfers" and "data".

Conditional compilation updates:

* [`autonomi/src/lib.rs`](diffhunk://#diff-e0bfa10b2535d8aa691cfe14b14e261dc22f94322602ec07147a5311d4e47a87R50-R55): Added conditional compilation attributes for the `self_encryption` and `wallet` modules based on the "data" and "transfers" features, respectively. Also, introduced a conditional constant `VERIFY_STORE` for the "transfers" feature.

### DIFF
--- a/autonomi/src/lib.rs
+++ b/autonomi/src/lib.rs
@@ -47,7 +47,10 @@ pub use client::{Client, ConnectError, CONNECT_TIMEOUT_SECS};
 
 mod client;
 mod secrets;
+#[cfg(feature = "data")]
 mod self_encryption;
+#[cfg(feature = "transfers")]
 mod wallet;
 
+#[cfg(feature = "transfers")]
 const VERIFY_STORE: bool = true;
--- a/autonomi/Cargo.toml
+++ b/autonomi/Cargo.toml
@@ -4,10 +4,10 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = []
+default = ["data"]
 full = ["data", "files", "fs", "registers", "transfers"]
 data = ["transfers"]
-files = ["transfers"]
+files = ["transfers", "data"]
 fs = []
 local = ["sn_client/local-discovery"]
 registers = ["transfers"]


autonomi/src/wallet/error.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(feature = "transfers")]
```
b-zee: This is not needed as the feature gate is already in `lib.rs` on the `mod wallet`.

autonomi/src/lib.rs
```diff
@@ -47,7 +47,10 @@ pub use client::{Client, ConnectError, CONNECT_TIMEOUT_SECS};
 
 mod client;
 mod secrets;
+#[cfg(feature = "data")]
 mod self_encryption;
+#[cfg(feature = "transfers")]
 mod wallet;
 
+#[cfg(feature = "data")]
```
b-zee: Currently this constant is only used in `transfers`, so maybe change to that.

autonomi/src/client/mod.rs
```diff
@@ -38,6 +38,7 @@ pub const CONNECT_TIMEOUT_SECS: u64 = 20;
 /// ```
 #[derive(Clone)]
 pub struct Client {
+    #[allow(dead_code)]
```
b-zee: If we default to feature `data`, I think it nicely solves this problem here.

b-zee (approved):

Issue #2077, opened 2024-08-29
bug (regression): updating a register can fail on local network with stable-2024.08.2.3 but still works using stable.2024-07-25


happybeing:
I have a bunch of scripts which I use to test my application (awe) on a local network. In brief, these create a local network and then upload a series of websites, some just a single version, but two involve uploading a series of about four versions. 

When uploading multiple versions the following sequence is repeated to load and update the register:
- get the register using Client::get_register()
- sync using ClientRegister::sync()
- write an new value using ClientRegister::write_merging_branches_online()
- sync using ClientRegister::sync()

(The above all happens in awe_website_versions.rs).

The scripts and code to do the above have been run tens if not >100 times without ever seeing the following error, which is happening when I try to update the second multi-version website, but not the first!

When the error occurs (and it occurs at the same point in repeated runs of these scripts with a new local network every time), the `write_merging_branches_online()` function fails with:
```
Failed to add XorName to register: Network(GetRecordError(RecordDoesNotMatch(c7ec9c(754072be1f575e7b94b97f21556067218383dd627d570314e1357d910f9592e9))))
```

If I use stable-2024.08.2.3 the above happens every time. If I use stable.2024-07-25 this has never happened.

Below are the safe_network crate versions I'm building against in each case:
```
# Generated using: awe-dependencies --branch stable-2024.08.2.3
sn_cli = { version = "0.94.1" }
sn_client = { version = "0.109.1" }
sn_peers_acquisition = { version = "0.4.2" }
sn_registers = { version = "0.3.17" }
sn_transfers = { version = "0.18.10" }
sn_protocol = { version = "0.17.7" }

# Generated using: awe-dependencies --branch stable.2024-07-25
# sn_cli = { version = "0.94.0" }
# sn_client = { version = "0.109.0" }
# sn_peers_acquisition = { version = "0.4.1" }
# sn_registers = { version = "0.3.16" }
# sn_transfers = { version = "0.18.9" }
# sn_protocol = { version = "0.17.6" }
```

What is strange to me is:
- it is so far 100% repeatable and 
- always the same website that updates ok, and the same website that doesn't

For all five Registers, including the three other Registers, I create the register and immediately write two values. This always succeeds. There's only an error in one of the two Registers I subsequently try to write a third value too, and it is always the same one.

maqi:
2024.08.2.3 contains a breaking change for client to get Register from network, which is not supported by the current PROD-01.
[...]

PR #2061, opened 2024-08-22
fix(client): fetch register accurately and merge results

maqi:
### Description

This pull request includes several changes to the `sn_client`, `sn_networking`, and `sn_protocol` modules to simplify and improve the handling of register records. The most important changes include removing unnecessary parameters, renaming functions for clarity, and adding new methods for fetching register records from the network.

### Simplification and Code Cleanup:

* [`sn_client/src/api.rs`](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L441-R445): Removed the `is_verifying` parameter from `get_signed_register_from_network` and updated related methods to simplify the code. [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L441-R445) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L806-R771) [[3]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L840-R805) [[4]](diffhunk://#diff-1b86a48cb70322a3ac64029081787ac6c92b95fe1878ae05e65327b224bc787cL773-R773) [[5]](diffhunk://#diff-1b86a48cb70322a3ac64029081787ac6c92b95fe1878ae05e65327b224bc787cL868-R868)
* [`sn_client/src/api.rs`](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L1085-R1067): Renamed `merge_split_register_records` to `merge_register_records` for better clarity and updated related tests. [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L1085-R1067) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L1133-L1140) [[3]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L1191-R1173)

### New Methods and Enhancements:

[...]

PR #2103, opened 2024-09-11
fix(network): register record shall not be verified by entire content

maqi:
### Description

For register record, the get_verification against the entire record value will have a high chance of failure, given the nature of the update_history MerkleReg held by the register.
For register, we can only carry out verification against root value, i.e. the final updated value

### Related Issue

https://github.com/maidsafe/safe_network/issues/2077

### Type of Change

### DIFF
--- a/autonomi/src/client/data.rs
+++ b/autonomi/src/client/data.rs
@@ -88,6 +88,7 @@ impl Client {
             retry_strategy: None,
             target_record: None,
             expected_holders: HashSet::new(),
+            is_register: false,
         };
         let record = self.network.get_record_from_network(key, &get_cfg).await?;
         let header = RecordHeader::from_record(&record)?;
--- a/autonomi/src/client/registers.rs
+++ b/autonomi/src/client/registers.rs
@@ -138,6 +138,7 @@ impl Client {
             retry_strategy: None,
             target_record: None,
             expected_holders: Default::default(),
+            is_register: true,
         };
 
         let record = self.network.get_record_from_network(key, &get_cfg).await?;
--- a/autonomi/src/client/transfers.rs
+++ b/autonomi/src/client/transfers.rs
@@ -318,6 +318,7 @@ async fn store_spend(network: Network, spend: SignedSpend) -> Result<(), Network
         retry_strategy: None,
         target_record: record_to_verify,
         expected_holders,
+        is_register: false,
     };
     let put_cfg = PutRecordCfg {
         put_quorum: Quorum::Majority,
--- a/sn_client/src/api.rs
+++ b/sn_client/src/api.rs
@@ -617,6 +617,7 @@ impl Client {
                 retry_strategy,
                 target_record: None, // Not used since we use ChunkProof
                 expected_holders: Default::default(),
+                is_register: false,
             };
             // The `ChunkWithPayment` is only used to send out via PutRecord.
             // The holders shall only hold the `Chunk` copies.
@@ -702,6 +703,7 @@ impl Client {
             retry_strategy: Some(retry_strategy.unwrap_or(RetryStrategy::Quick)),
             target_record: None,
             expected_holders,
+            is_register: false,
         };
         let record = self.network.get_record_from_network(key, &get_cfg).await?;
         let header = RecordHeader::from_record(&record)?;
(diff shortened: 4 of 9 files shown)


joshuef:
Would be good to get a test to verify this in some fashion. Not sure how best honestly...

And I'm wondering on the impacts here... (not sure how much of this would fall into the scope of this PR, but worth considering.)

During replication... do we fetch registers still with this setup? (I see it doesnt really touch flows there). But say, I have v1, and v2 exists... Would we still fetch + merge?


Is this purely on the client side saying that we return registers if it's the same address basically... but then how do we verify edits/updates? 


And if there _are_ updates / divergence, will the `GET` return all variations or just one at the moment? (Or will it handle `merge` if possible before returning?

maqi:
> get a test to verify this in some fashion

There is `storage_payment_register_creation_succeeds` test, which involves multiple ops mutation, publish and fetch verification.
Which shall cover this already ?
However, yeah, with the limited size of the CI local testnet, to fully simulate register with different update history is really tricky.

> During replication... do we fetch registers still with this setup?

as explained at https://github.com/maidsafe/safe_network/pull/2103/files#diff-c302643eb99427e41bc3ae0fe33cdf9b841a8d55593fc5f20e132674ecfeb603R78
replication fetch doesn't have local target to verify with, hence the `is_register` flag can be set as `false`.

> But say, I have v1, and v2 exists... Would we still fetch + merge?

that's a different question. And the answer is `yes`, we do fetch + merge
For register, during replication, after got it, eventually the [validate_and_store_register](https://github.com/maidsafe/safe_network/blob/main/sn_node/src/put_validation.rs#L442) function will be called.
which internally will get the local register `merged` with the fetched register

>  how do we verify edits/updates?

the `root value`, i.e. the final updated value, will still be verified against the target.
as at https://github.com/maidsafe/safe_network/pull/2103/files#diff-985b0f4f68bd09d2bcb4c2da67ab65c5c68d38d8aac388691e361a295bc8ae4bR169

> And if there are updates / divergence, will the GET return all variations or just one at the moment? (Or will it handle merge if possible before returning?

After the merge of [PR 2061](https://github.com/maidsafe/safe_network/pull/2061), the fetched register will be the `merged` result of all copies from the closest nodes to the target.

happybeing:
Supporting developers using these APIs is a good way to test this and much else. So long as the plan remains to replace the native token with a blockchain token, I will not help with that but am planning to keep `awe` working and people using that will in effect be testing things, at least until the native token is ditched. 

You could also use `awe` yourselves, and may find the 'inspect-register' command that is part of the `awe` CLI is helpful in debugging any issues.

joshuef:
> Supporting developers using these APIs is a good way to test this and much else.

Yeh that'll be the true final test, but just wanting to make sure we have an automated CI step that gets a minimum case in here.

joshuef:
> the root value, i.e. the final updated value, will still be verified against the target.
as at [#2103 (files)](https://github.com/maidsafe/safe_network/pull/2103/files#diff-985b0f4f68bd09d2bcb4c2da67ab65c5c68d38d8aac388691e361a295bc8ae4bR169)

Aha, okay so we're comparing the final computed value only. 

To me this is not the `root` value... or rather, the `root` to me implies the _first_ entry. Whereas here we mean the cumulative final hash after all merging...

If that's the case, i'd suggest renaming the idea of the `base` register to be `merged_register` or something more clear?

maqi:
> To me this is not the `root` value... or rather, the `root` to me implies the _first_ entry. Whereas here we mean the cumulative final hash after all merging...

The [crdts::MerkleReg document](https://docs.rs/crdts/latest/crdts/merkle_reg/struct.MerkleReg.html) defines the `roots` as:
```
The MerkleReg is a Register CRDT that uses the Merkle DAG structure to track the current value(s) held by this register. The roots of the Merkle DAG are the current concurrent values.
```
Here just try to follow that definition.
 
> If that's the case, i'd suggest renaming the idea of the `base` register to be `merged_register` or something more clear?

`base_register` is a member field of the `SignedRegister`.
it has been used like this before this PR, and I think is not related to the work ?

joshuef:
Okay fair enough! @maqi , thanks for clarifying 👍

joshuef (approved):

PR #2109, opened 2024-09-13
feat: enable e2e wan nightly tests

ermineJose:
### Description

Enable first test case e2e tests of WAN nightly builds in repo.

### Related Issue

Fixes daily nightly WAN tests

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### Checklist

- [X ] New and existing unit tests pass locally with my changes.

### DIFF
--- a/.github/workflows/nightly_wan.yml
+++ b/.github/workflows/nightly_wan.yml
@@ -133,6 +133,7 @@ jobs:
             ~/.local/share/safe/client/logs/*/*.log*
 
       - name: destroy network
+        if: always()
         uses: maidsafe/sn-testnet-control-action/destroy-network@main
         with:
           network-name: ${{ env.NETWORK_NAME }}
--- a/.github/workflows/nightly_wan.yml
+++ b/.github/workflows/nightly_wan.yml
@@ -1,11 +1,11 @@
 name: Nightly -- Full WAN Network Tests
 
 on:
-  # schedule:
-  #   - cron: "0 0 * * *"
-# enable below for testing purpose.
-  pull_request:
-    branches: ["*"]
+  schedule:
+    - cron: "0 0 * * *"
+# enable as below for testing purpose.
+  # pull_request:
+  #   branches: ["*"]
   workflow_dispatch:
 
 env:
--- a/.github/workflows/nightly_wan.yml
+++ b/.github/workflows/nightly_wan.yml
@@ -62,7 +62,12 @@ jobs:
         run: |
            echo "Peer is $SAFE_PEERS"
            echo "Deployment inventory is $SN_INVENTORY"
-      
+
+      - name: start faucet
+        uses: maidsafe/sn-testnet-control-action/start-faucet@main
+        with:
+            network-name: ${{ env.NETWORK_NAME }}
+               
       - name: Obtain the funds from the faucet
         run: |
           set -e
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly_wan.yml
```diff
@@ -1,13 +1,17 @@
...
+    - cron: "0 0 * * *"
+# enable as below for testing purpose.
+  # pull_request:
+  #   branches: ["*"]
   workflow_dispatch:
 
 env:
   CARGO_INCREMENTAL: 0 # bookkeeping for incremental builds has overhead, not useful in CI.
+  DEFAULT_PAYMENT_FORWARD_SK: 49113d2083f57a976076adbe85decb75115820de1e6e74b47e0429338cef124a
+  NETWORK_NAME: DEV-01
```
RolandSherwin, line 14: We might have to keep a unique `NETWORK_NAME` per `job`.
All the jobs inside the file will run in parallel and if you were to re-use the same testnet-name, we'd have a mess.

.github/workflows/nightly_wan.yml
```diff
@@ -29,93 +33,94 @@ jobs:
...
+      - name: launch ${{ env.NETWORK_NAME }}
+        uses: maidsafe/sn-testnet-control-action/launch-network@main
+        with:
+          ansible-forks: ${{ env.ANSIBLE_FORKS }}
+          beta-encryption-key: ${{ env.DEFAULT_PAYMENT_FORWARD_SK }}
+          environment-type: development
+          faucet-version: ${{ env.FAUCET_VERSION }}
+          log-format: json
+          network-name: ${{ env.NETWORK_NAME }}
+          network-contacts-file-name: ${{ env.NETWORK_CONTACTS_FILE_NAME }}
```
RolandSherwin: This env is not set anywhere? Also this var is optional and so can be ignored. https://github.com/maidsafe/sn-testnet-control-action/blob/057aa565d5fa3f5871f12de6ec49bf3257e1f0ee/launch-network/action.yml#L24

We can remove this variable.

.github/workflows/nightly_wan.yml
```diff
@@ -29,93 +33,94 @@ jobs:
...
-          subnet-id: subnet-018f2ab26755df7f9
-          node-count: 20
-          vm-count: 1
+
+      - name: launch ${{ env.NETWORK_NAME }}
+        uses: maidsafe/sn-testnet-control-action/launch-network@main
+        with:
+          ansible-forks: ${{ env.ANSIBLE_FORKS }}
+          beta-encryption-key: ${{ env.DEFAULT_PAYMENT_FORWARD_SK }}
+          environment-type: development
```
RolandSherwin, line 50: Instead of providing the `environment-type: development` (which would set a pre configured VM count & node-count per VM), can you set the following:
```
node-vm-count: 10
node-count: 20,
uploader-vm-count: 0
bootstrap-node-vm-count: 0 (I think 0 should work, if not, set 1)
```

We don't want to spin up so many VMs tbh.

.github/workflows/nightly_wan.yml
```diff
@@ -29,93 +33,94 @@ jobs:
...
-          security-group-id: sg-0d47df5b3f0d01e2a
-          subnet-id: subnet-018f2ab26755df7f9
-          node-count: 20
-          vm-count: 1
+
+      - name: launch ${{ env.NETWORK_NAME }}
+        uses: maidsafe/sn-testnet-control-action/launch-network@main
+        with:
+          ansible-forks: ${{ env.ANSIBLE_FORKS }}
+          beta-encryption-key: ${{ env.DEFAULT_PAYMENT_FORWARD_SK }}
```
RolandSherwin: I think it should work without this key? The default key will be used if not provided. So we can remove this variable.

.github/workflows/nightly_wan.yml
```diff
@@ -29,93 +33,94 @@ jobs:
...
+          environment-type: development
+          faucet-version: ${{ env.FAUCET_VERSION }}
```
RolandSherwin: We don't want to provide any binary versions here, since `safe-network-branch` and `safe-network-user` are used. I.e,. we are compiling the binary we need from the provided branch instead of using pre-compiled binary.

Also `env.FAUCET_VERSION` is not defined anywhere, so this might not be doing anything. We can remove this variable.

.github/workflows/nightly_wan.yml
```diff
@@ -29,93 +33,94 @@ jobs:
...
+          cargo run --bin safe --release -- --log-output-dest=data-dir register edit -n baobao wood
         env:
           SN_LOG: "all"
         timeout-minutes: 2
-
+      
       - name: Fetch network logs
-        if: always()
-        uses: maidsafe/sn-testnet-action@main
+        uses: maidsafe/sn-testnet-control-action/fetch-logs@main
```
RolandSherwin, line 118: Are these logs being written to `~/.local/share/safe/node/`?

PR #2119, opened 2024-09-17
feat(node-manager): :sparkles: adds maintain cli command

anthonyra:
### Description

Please provide a brief description of the changes made in this pull request. Highlight the purpose of the changes and the problem they address.

### Related Issue

None

### Type of Change

- [ ] Documentation update
- [x] Other (please describe): Noticed that there were some #[expect] macros failing the build. Updated them to #[allow]. Currently builds. I also noticed there were some issues with the `--path` flag and aligning the arguments with the functions.

### Checklist

### DIFF
--- a/node-launchpad/src/components.rs
+++ b/node-launchpad/src/components.rs
@@ -38,7 +38,7 @@ pub trait Component {
     /// # Returns
     ///
     /// * `Result<()>` - An Ok result or an error.
-    #[expect(unused_variables)]
+    #[allow(unused_variables)]
     fn register_action_handler(&mut self, tx: UnboundedSender<Action>) -> Result<()> {
         Ok(())
     }
@@ -51,7 +51,7 @@ pub trait Component {
     /// # Returns
     ///
     /// * `Result<()>` - An Ok result or an error.
-    #[expect(unused_variables)]
+    #[allow(unused_variables)]
     fn register_config_handler(&mut self, config: Config) -> Result<()> {
         Ok(())
     }
@@ -93,7 +93,7 @@ pub trait Component {
     /// # Returns
     ///
     /// * `Result<Option<Action>>` - An action to be processed or none.
-    #[expect(unused_variables)]
+    #[allow(unused_variables)]
     fn handle_key_events(&mut self, key: KeyEvent) -> Result<Vec<Action>> {
         Ok(vec![])
     }
@@ -106,7 +106,7 @@ pub trait Component {
     /// # Returns
     ///
     /// * `Result<Option<Action>>` - An action to be processed or none.
-    #[expect(unused_variables)]
+    #[allow(unused_variables)]
     fn handle_mouse_events(&mut self, mouse: MouseEvent) -> Result<Vec<Action>> {
         Ok(vec![])
     }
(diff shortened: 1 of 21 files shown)


anthonyra:
Left this as draft since I'm not a huge rust developer and not 100% familiar with the project layout. The intention is to add the ability to the `node-manager` CLI to maintain a specific number of nodes running. This would be great for the DeEEP Network simplifying deployments. I didn't touch any tests and wasn't able to build the artifacts (I'm on Apple silicon). Let me know how I can help

jacderida:
Hi Anthony,

Thanks for the contribution! Could you elaborate more please on how you envision this command being used?

Cheers,

Chris

anthonyra:
@jacderida We want to host nodes for individuals (my parent company is NerdNode but building for DeEEP Network also). As such, with the current `safenode-manager` CLI we would need to manage the number of nodes an individual wants and scale up or down as necessary. With maintain we can simply call `safenode-manager maintain --count 6` to scale up or `safenode-manager maintain --count 1` to scale down.

jacderida:
I see, thanks! Let me take a look at this in a little more detail. I might need to wait until the weekend, but, it looks like a good feature.

node-launchpad/src/components.rs
```diff
@@ -38,7 +38,7 @@ pub trait Component {
     /// # Returns
     ///
     /// * `Result<()>` - An Ok result or an error.
-    #[expect(unused_variables)]
+    #[allow(unused_variables)]
```
joshuef, line 41: can we leave these as `expect` please, otherwise we can drop the lint

joshuef:
thanks for this @anthonyra . QQ: In the current setup what's missing is `count` is that right?

Just wondering if another command is required here vs a flag to auto start the correct node counts.

node-launchpad/src/components.rs
anthonyra: I'm not too familiar with rust, but I couldn't `cargo build` with those remaining as `expect`. Unless there's some cargo config settings that need to be set?

anthonyra:
> QQ: In the current setup what's missing is `count` is that right?

In the current setup, the CLI doesn't have the `maintain_n_running_nodes` exposed. My goal was to simply boil this up to the CLI. As such, the old function had `max_nodes_to_run` which I changed to `count` to align with the `add` sub-command. But while doing so noticed that there were other issues. Most notable one is that the path's flag "seemed" to be broken and some of the other arguments "shuffled". I attempted to install `safenode-manager` with the path flag set and got an os 13 error also after seeing the discrepancy. 

```
1 service(s) to be added
Error: 
   0: Permission denied (os error 13)
```

> Just wondering if another command is required here vs a flag to auto start the correct node counts.

I thought about this also but it's a hybrid of the `add` and `start` sub-commands from what I could tell requiring a mix of each to work. I could be wrong here just figured this would be the "quickest" lift from my understanding of the project.

node-launchpad/src/components.rs
RolandSherwin: Hey, I think it might be because it was a new feature that was stablizied in the latest rust version. Could you run `rustup update` && `rustup default stable` and try compiling it?

anthonyra: It appears that it does allow me to build but results in a lot more orange and red in my IDE.. Also is there a way to force a rustc version so that it warns others before attempting to build? -- Looks like maybe

```
[package]
rust-version = "1.81.0"
```

jacderida:
Hi @anthonyra 

I really do appreciate the PR here, but, the node manager is going to have a significant feature removed, and so the code in this PR might diverge away too much.

For now, we are going to close it, but, we may reopen it in the future.

I'm sorry for how long we took to get back to you on this. We've been very busy trying to hit our release schedule.

PR #2123, opened 2024-09-18
feat(protocol): derive serialisations for RecordKind

joshuef:


maqi:
the RecordKind need to be refactored for the future plan of introducing payment-data assosciation.
which may need to retain with the current custom serialisation/deserialisation, or swtich to derive.
To avoid future confliction, closing this PR now.

PR #2124, opened 2024-09-18
fix(autonomi): verify register on fetch

b-zee:
Registers are fetched and contain CRDT operations. These operations
should be valid, so we check once we fetched it. Then later we assert
the register is valid.

### DIFF
--- a/autonomi/src/client/registers.rs
+++ b/autonomi/src/client/registers.rs
@@ -28,6 +28,8 @@ pub enum RegisterError {
     Network(#[from] NetworkError),
     #[error("Serialization error")]
     Serialization,
+    #[error("Register could not be verified (corrupt)")]
+    FailedVerification,
     #[error("Payment failure occurred during register creation.")]
     Pay(#[from] PayError),
     #[error("Failed to retrieve wallet payment")]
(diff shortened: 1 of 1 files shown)


joshuef:
there was a bug that led you here right @b-zee ? Can we get a test case of that bug?

(also what's the implication of the change from `base_register()` to `register()` ?

b-zee:
>(also what's the implication of the change from base_register() to register() ?

This was the source of the bug. I assumed when we freshly fetched a register, the 'base register' would contain all the values, but this is only after applying the CRDT operations. Getting that base register plus operations was a fallible method, which was why I didn't like calling it.

maqi:
> when we freshly fetched a register, the 'base register' would contain all the values, but this is only after applying the CRDT operations.

Yeah, that's true.
I think this might being the reason why it is called `base_register` within `SignedRegister`.
I do agree the current naming/flow is kind of `confusing` (as I am also just started touch this part due to recent community member's issue raised related to it)

However, I think it is kind of `mis-used`, as `ops` shall always just being `local mutation history`, and published `register record` shall only retain the `MerkleReg` for the `last updated value + DAG tree for the update history` ?

Following is what I think shall happen to `ops` and `base_register` member fields of `SignedRegister` : 
* the `ops` are all `local` ones to record `mutate instructions issued by client`
* client locally write `mutate instructions` into `SignedRegister.ops`
* during `publish`, all `ops` will be applied to the `base_register` and then uploaded to network (and only `base_register` got uploaded as a `register record`)
* When fetch from network, a signed_register using the fetched register record as its base_register will be created, with its `ops` to be empty

maqi (approved):

PR #2125, opened 2024-09-18
feat(client): pick a random payee from candidates within range of lowest

maqi:
### Description

To avoid certain node that is in over-lapped responsible range got biased to be picked as a payee 
(i.e. it has to give high quotes due to received extra records, compared with its group neighbours)

refactor the `payee selection scheme` to be: 
all nodes holding relevant_records within certain range of `lowest` of them become candidates, 
then randomly select one from those candidates as a `payee`.

### DIFF
--- a/sn_networking/src/lib.rs
+++ b/sn_networking/src/lib.rs
@@ -58,6 +58,7 @@ use libp2p::{
     request_response::OutboundFailure,
     Multiaddr, PeerId,
 };
+use rand::prelude::SliceRandom;
 use rand::Rng;
 use sn_protocol::{
     error::Error as ProtocolError,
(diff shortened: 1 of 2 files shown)


sn_networking/src/lib.rs
```diff
@@ -958,40 +959,55 @@ impl Network {
...
-            },
-        )
+        .filter(|(_, _, quote)| {
+            let count = quote.quoting_metrics.close_records_stored as u64;
+            count <= upper_bound
+        })
+        .collect();
+
+    if filtered_costs.is_empty() {
+        return Err(NetworkError::NoStoreCostResponses);
```
joshuef, line 991: This should be a separate and clear error

sn_networking/src/lib.rs
```diff
@@ -1129,14 +1143,10 @@ mod tests {
...
-        let expected_price = costs[0].2.cost.as_nano();
+        let lowest_price = costs[0].2.cost.as_nano();
         let (_peer_id, _key, price) = get_fees_from_store_cost_responses(costs)?;
 
-        assert_eq!(
-            price.cost.as_nano(),
-            expected_price,
-            "price should be {expected_price}"
-        );
+        assert!(price.cost.as_nano() >= lowest_price);
```
joshuef, line 1149: hmm, this assert doesn't really tell us anything now.

sn_networking/src/record_store.rs
```diff
@@ -1745,20 +1745,19 @@ mod tests {
                     "store cost is not 'balanced', expected ratio max/min to be < 100, but was {}",
                     max_store_cost / min_store_cost
                 );
-                assert!(
-                    (max_earned / min_earned) < 1500,
-                    "earning distribution is not balanced, expected to be < 1500, but was {}",
-                    max_earned / min_earned
-                );
+                if min_earned != 0 {
```
joshuef, line 1748: i'm not sure this should be applied. effectively removes the check for balance if we are unbalanced... 🤔

sn_networking/src/record_store.rs
```diff
@@ -1729,7 +1729,7 @@ mod tests {
 
             // Check termination condition
             if hour == max_hours {
-                let acceptable_percentage = 0.01; //%
+                let acceptable_percentage = 0.12; //%
```
joshuef, line 1732: so with our change we end up with loads more empty nodes? 🤔 

hmm. This number feels a bit magic and a bit high.

joshuef (approved):
I'm unsure on this. 

I think it might be part one of a two part solution here. 

This is in the world of "good client by default", in which case this seems valid. We would have to assume "cheap clients" exist too. So the issue would then persist... (at a potential network health cost).


The other side is the "is my node healthy" check... Which I guess is a node-manager conceit... But basically "if im not being paid... restart the node + move"... as I think it would be folly to ignore that `cheap clients` will exist. 


not sure what you think on that @maqi ? 



Not sure

joshuef (changes requested):
as above

maqi:
> We would have to assume "cheap clients" exist too

Right, I got what you mean.
Considering that, I think the client side `selection scheme` has to stay with the current `cheapest`.
Then I'd agree with have to `if im not being paid... restart the node + move` from `node-manager` level

PR #2131, opened 2024-09-19
ci: nightly releases

jacderida:
- 7639624c6 **chore: standardise versioning for binaries**

  The RFC for the release process stipulated there should be four versioning arguments:
  * --version
  * --crate-version
  * --package-version
  * --network-version

  Here, `--crate-version` is the semantic version of the crate, `--package-version` is the release
  cycle version, e.g., `2024.09.1.1`, and `--network-version` prints the compatible network protocol.
  The `--version` argument will then print all of this information. The `--package-version` argument
  does not apply to the nightly release.

  The approach for printing the version information is to provide our own `version` flag, by using the
  `disable_version_flag` attribute with `clap`. If you want to use `clap`, all the information needs
  to be completely static and available at compile time. This is convoluted, especially for things
  like the network protocol version and the crate version, and it would have led to a lot of
  repetition. We can avoid the difficulty by providing the information dynamically, and we can use the
  `sn_build_info` crate to define the version string once. On binaries that used subcommands, for the
  versioning to work correctly, the subcommand needs to be made optional.

  The `get_bin_version` helper used by the node manager was updated parse the new versioning
  information from both stable and nightly releases. Now, the function can handle different
  `--version` output formats by extracting the version number for stable releases prefixed with a 'v'
  and the date for nightly releases.

  Since we are going to introduce a nightly build, this commit also introduces a nightly feature,
  which will control what versioning information is used. For the nightly, the binaries will be
  versioned with the date on which they are built. The `--package-version` argument does not apply to
  the nightly release; it is not necessary if all the binaries have the same version, which is the
  date.

- 73194e3b2 **ci: nightly releases**

  The nightly release will build the binary set from the `main` branch every night at midnight UTC.

  The workflow:

  * Builds the binaries with the `nightly` feature, which versions them using the current date, rather
    than a Semantic Version.
  * Uploads packaged binaries to S3, again with the date for the version.
  * Also uploads packaged binaries to S3 with 'nightly' as the version, which will always be the
    latest. The previous 'nightly' version will be deleted from the bucket.
  * Creates a Github Release and uploads archives with the binaries packaged by architecture, as per
    the stable release. To prevent our releases page being flooded, the previous nightly release will
    be deleted before the current one is created.

### DIFF
--- a/nat-detection/src/main.rs
+++ b/nat-detection/src/main.rs
@@ -35,7 +35,7 @@ const RETRY_INTERVAL: Duration = Duration::from_secs(10);
 /// - 11: Public under UPnP
 /// - 12: Private or Unknown NAT
 #[derive(Debug, Parser)]
-#[clap(version, author, verbatim_doc_comment)]
+#[clap(disable_version_flag = true)]
 struct Opt {
     /// Port to listen on.
     ///
@@ -60,14 +60,49 @@ struct Opt {
 
     #[command(flatten)]
     verbose: clap_verbosity_flag::Verbosity,
+
+    /// Print the crate version
+    #[clap(long)]
+    crate_version: bool,
+
+    /// Print the package version
+    #[clap(long)]
+    #[cfg(not(feature = "nightly"))]
+    package_version: bool,
+
+    /// Print version information.
+    #[clap(long)]
+    version: bool,
 }
 
 #[tokio::main]
 async fn main() -> Result<()> {
     color_eyre::install()?;
 
-    // Process command line arguments.
     let opt = Opt::parse();
+
+    if opt.version {
+        println!(
+            "{}",
+            sn_build_info::version_string(
+                "Autonomi NAT Detection",
+                env!("CARGO_PKG_VERSION"),
+                None
+            )
+        );
+        return Ok(());
+    }
+
+    if opt.crate_version {
+        println!("Crate version: {}", env!("CARGO_PKG_VERSION"));
+        return Ok(());
+    }
+
+    #[cfg(not(feature = "nightly"))]
+    if opt.package_version {
+        println!("Package version: {}", sn_build_info::package_version());
+        return Ok(());
+    }
 
     let registry = tracing_subscriber::registry().with(tracing_subscriber::fmt::layer());
     // Use `RUST_LOG` if set, else use the verbosity flag (where `-vvvv` is trace level).
--- a/node-launchpad/src/bin/tui/main.rs
+++ b/node-launchpad/src/bin/tui/main.rs
@@ -16,7 +16,7 @@ use color_eyre::eyre::Result;
 use node_launchpad::{
     app::App,
     config::configure_winsw,
-    utils::{initialize_logging, initialize_panic_handler, version},
+    utils::{initialize_logging, initialize_panic_handler},
 };
 #[cfg(target_os = "windows")]
 use sn_node_manager::config::is_running_as_root;
@@ -25,7 +25,7 @@ use std::{env, path::PathBuf};
 use tokio::task::LocalSet;
 
 #[derive(Parser, Debug)]
-#[command(author, version = version(), about)]
+#[command(disable_version_flag = true)]
 pub struct Cli {
     #[arg(
         short,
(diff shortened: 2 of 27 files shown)


node-launchpad/src/bin/tui/main.rs
```diff
@@ -21,11 +21,12 @@ use node_launchpad::{
 #[cfg(target_os = "windows")]
 use sn_node_manager::config::is_running_as_root;
 use sn_peers_acquisition::PeersArgs;
+use sn_protocol::version::IDENTIFY_PROTOCOL_STR;
 use std::{env, path::PathBuf};
 use tokio::task::LocalSet;
 
 #[derive(Parser, Debug)]
-#[command(author, version = version(), about)]
+#[command(author, version = version(), about, name = "Autonomi Node Launchpad")]
```
RolandSherwin: The `fn version()` is from the init of the node launchpad crate and it displays the author + data dir path, which is incorrect. We can modify it to dispaly the correct versions.

But I'm not sure `protocol version` applies to launchpad as it can use any safenode version with any protocol version within. So, maybe we can display just the `CARGO_PKG_VERSION` here?

node-launchpad/src/bin/tui/main.rs
```diff
@@ -53,12 +54,41 @@ pub struct Cli {
...
     #[command(flatten)]
     pub(crate) peers: PeersArgs,
+
+    /// Print the crate version.
+    #[clap(long)]
+    crate_version: bool,
+
+    /// Print the network protocol version.
+    #[clap(long)]
+    protocol_version: bool,
```
RolandSherwin: As mentioned above, launchpad is not restricted to a certain protocol version, it can use any safenode with any protocol version. So should we remove this arg?

nat-detection/src/main.rs
```diff
@@ -60,15 +61,43 @@ struct Opt {
...
 
+    if opt.crate_version {
+        println!("Crate version: {}", env!("CARGO_PKG_VERSION"));
+        return Ok(());
+    }
+
+    if opt.protocol_version {
+        println!("Network version: {}", IDENTIFY_PROTOCOL_STR.to_string());
+        return Ok(());
+    }
```
RolandSherwin: Nat detection binary is a standalone one and does not use any protocol versions. This might potentially cause confusion?

node-launchpad/src/utils.rs
```diff
@@ -135,12 +127,12 @@ macro_rules! trace_dbg {
 
 pub fn version() -> String {
```
RolandSherwin: Looks like you have modified this function, but can we also:
- get rid of `author` from here
- remove protocol version st
- remove data_dir_path
?

sn_build_info/src/lib.rs
```diff
@@ -33,3 +34,58 @@ pub const fn git_branch() -> &'static str {
...
+            git_sha(),
+        )
+    } else {
+        full_version_info(crate_version, protocol_version)
+    }
+}
+
+pub fn log_version_info(crate_version: &str, protocol_version: &str) {
+    if cfg!(feature = "nightly") {
+        debug!("nightly build info: {}", nightly_git_info());
```
RolandSherwin, line 102: Should nightly display the `protocol_version` as well? I think it will be useful.

sn_cli/src/bin/main.rs
```diff
@@ -39,6 +40,22 @@ const CLIENT_KEY: &str = "clientkey";
...
+        println!("{}", env!("CARGO_PKG_VERSION"));
+        return Ok(());
+    }
+
+    if opt.protocol_version {
+        println!("{}", IDENTIFY_PROTOCOL_STR.to_string());
+        return Ok(());
+    }
+
+    if opt.package_version {
```
RolandSherwin, line 67: Should we not restrict `package_version` to only print when not nightly? i.e., `#[cfg(not(feature = "nightly"))]`. If so, this applies to the `Opt` struct `pub package_version: bool,` field as well.

sn_node_manager/src/bin/cli/main.rs
```diff
@@ -16,28 +16,42 @@ use sn_node_manager::{
...
     debug: bool,
 
+    /// Print the package version.
+    #[cfg(not(feature = "nightly"))]
+    #[clap(long)]
+    pub package_version: bool,
+
+    /// Print the network protocol version.
+    #[clap(long)]
+    pub protocol_version: bool,
```
RolandSherwin: Similar to the comment on launchpad, I think `protocol version` will not apply to safenode manager as well.

sn_node_manager/src/bin/daemon/main.rs
```diff
@@ -22,21 +22,32 @@ use sn_service_management::{
...
+    /// Print the package version.
+    #[cfg(not(feature = "nightly"))]
+    #[clap(long)]
+    pub package_version: bool,
+    /// Specify a port for the daemon to listen for RPCs. It defaults to 12500 if not set.
+    #[clap(long, default_value_t = DAEMON_DEFAULT_PORT)]
+    port: u16,
+    /// Print the network protocol version.
+    #[clap(long)]
+    pub protocol_version: bool,
```
RolandSherwin: Same here, protocol_version is not applicable here.

sn_node_rpc_client/src/main.rs
```diff
@@ -9,26 +9,36 @@
...
     #[clap(subcommand)]
     cmd: Cmd,
+
+    /// Print the crate version.
+    #[clap(long)]
+    crate_version: bool,
+
+    /// Print the network protocol version.
+    #[clap(long)]
+    protocol_version: bool,
```
RolandSherwin: The `rpc_client` is not restricted by the protocol version too. It can communicate to any `safenode` if the RPC protocol is the same. I think we should remove this one.

sn_build_info/src/lib.rs
RolandSherwin: Can we also add the following here?
```
  debug!("version: {crate_version}");
  debug!("network version: {protocol_version}");
```

node-launchpad/src/bin/tui/main.rs
RolandSherwin: This can be removed.

RolandSherwin (approved):

sn_build_info/src/lib.rs
jacderida: The version (the date) is in the nightly build info. However, I added the network version line.

PR #2135, opened 2024-09-24
feat(launchpad): new stats with layout and interaction

mazzi:
### Description

New info on the status screen.
Start/Stop/Running now has spinners and not a popup. Only the popup is being shown when the NAT detection process is happening.

Follow up PR: start/stop of individual nodes.

Screenshot 2024-09-24 at 14 26 42


RolandSherwin:
Hey! I was testing out the build, and I've encountered this following flow which I think might not be the intended one?
- Reset nodes
- setup mode to upnp
- start nodes => this runs NAT detection, which should not happen since we've set the mode manually right?

node-launchpad/src/components/status.rs
```diff
@@ -417,6 +426,7 @@ impl Component for Status {
                     }
 
                     if self.lock_registry.is_some() {
+                        // FIXME: What do we show to the user?
```
RolandSherwin: When `lock_registry.is_some()`, we've got the pop up saying that we're starting/stopping nodes. Do you mean that we should display something else instead?

RolandSherwin:
Tested out the build and I think `Mbps` displays the cumulative "inbound / outbound bytes" instead of the `rate of flow`. 
I think `new_value previous_value / time` should work? Where `time = time between each metric fetch` which can be the same constant value (5 sec I think between each metric fetch?) or can be calculated, to be more accurate?

mazzi:
> Hey! I was testing out the build, and I've encountered this following flow which I think might not be the intended one?
> 
>     * Reset nodes
> 
>     * setup mode to upnp
> 
>     * start nodes => this runs NAT detection, which should not happen since we've set the mode manually right?

Good testing! Yes, after reset we do always NAT. Not sure if it the specs where those, but make sense that we don't run NAT except in Auto mode. Do you agree with the last statement? I will modify but most likely in another PR because this doesn't belong here.

node-launchpad/src/components/status.rs
mazzi: You are right. My mistake. This prevents multiple execution of this part of the code. I'll remove the `FIXME`.

mazzi:
> Tested out the build and I think `Mbps` displays the cumulative "inbound / outbound bytes`instead of the`rate of flow`. I think `new_value previous_value / time`should work? Where`time = time between each metric fetch` which can be the same constant value (5 sec I think between each metric fetch?) or can be calculated to be more accurate?

The intended metric is to show what's incoming whats' outgoing. Not sure if we need to display the rate, because that will as you correctly pointed out, to handle a bit of historic data and we cannot do it so easily 🤷

RolandSherwin:
>  but make sense that we don't run NAT except in Auto mode. Do you agree with the last statement? 

Yes, agree with that.

> The intended metric is to show what's incoming whats' outgoing. Not sure if we need to display the rate, because that will as you correctly pointed out, to handle a bit of historic data and we cannot do it so easily 🤷

I assumed that we wanted the rate. The rate would slow down after a while and the cumulative value would change very slowly. OR if the safenode has been running for a while, the cumulative value might get really high and would become meaningless at one point?

`rate` would just require us to store the value from the previous metric fetch & the time.

mazzi:
@RolandSherwin just pushed the implementation. Seems to work fine.

RolandSherwin (approved):
Great work!

PR #2136, opened 2024-09-24
feat(register)!: network only store ops list

maqi:
BREAKING CHANGE!

crdt MerkleReg only stored in client locally
node side only store ops list (signed by owner)

### DIFF
--- a/autonomi/src/client/registers.rs
+++ b/autonomi/src/client/registers.rs
@@ -10,27 +10,22 @@
 pub use bls::SecretKey as RegisterSecretKey;
 use sn_evm::Amount;
 use sn_evm::AttoTokens;
-use sn_networking::GetRecordError;
 use sn_networking::VerificationKind;
 use sn_protocol::storage::RetryStrategy;
 pub use sn_registers::{Permissions as RegisterPermissions, RegisterAddress};
-use tracing::warn;
 
 use crate::client::data::PayError;
 use crate::client::Client;
 use bytes::Bytes;
 use evmlib::wallet::Wallet;
 use libp2p::kad::{Quorum, Record};
-use sn_networking::GetRecordCfg;
-use sn_networking::NetworkError;
-use sn_networking::PutRecordCfg;
+use sn_networking::{GetRecordCfg, GetRecordError, NetworkError, PutRecordCfg};
 use sn_protocol::storage::try_deserialize_record;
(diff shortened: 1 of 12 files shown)


autonomi/src/client/registers.rs
```diff
@@ -42,27 +43,68 @@ pub enum RegisterError {
...
+        };
+
+        if let Some(value) = initial_value {
+            register.write_atop(&value, &owner)?;
+        }
+
+        Ok(register)
+    }
+
+    pub fn write_atop(&mut self, entry: &[u8], owner: &SecretKey) -> Result<(), RegisterError> {
```
b-zee: Lets not make these `pub`. I want to keep the public interface simple.

autonomi/src/client/registers.rs
```diff
@@ -42,27 +43,68 @@ pub enum RegisterError {
...
-            .register()
-            .expect("register to be valid")
+        self.crdt_reg
             .read()
             .into_iter()
             .map(|(_hash, value)| value.into())
             .collect()
     }
+
+    pub fn new(
```
b-zee:
```suggestion
    fn new(
```

sn_registers/src/reg_crdt.rs
```diff
@@ -9,17 +9,21 @@
...
+    collections::{BTreeSet, HashSet},
     fmt::{self, Debug, Display, Formatter},
     hash::Hash,
 };
+use xor_name::XorName;
 
 /// Register data type as a CRDT with Access Control
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd)]
-pub(crate) struct RegisterCrdt {
+pub struct RegisterCrdt {
```
b-zee, line 26: Is there a specific motivation for making these items `pub`?

happybeing: @b-zee I think that's needed so that the API can expose the MerkleReg. That's necessary for access to the history, and without it usefulness of Registers becomes very limited. Not useful really!

You can look at/use the `awe inspect-register` command and implementation for some of the things that are needed.

b-zee: Yes, we do want to expose the underlying register. But that would be the `MerkleReg` type. Our `RegisterCrdt` is not something we want to promise anything about for the future.

b-zee (approved):

PR #2157, opened 2024-09-30
feat(metrics): shunned count across time frames

RolandSherwin:
- adds a new `/metrics_extended` endpoint. It contains all the same metrics from `/metrics` with a few more detailed ones.
- A new endpoint was added because we do not want to overwhelm our monitoring setup. If a metric is useful to us, we can move it to `/metrics`
- Adds a shunned count metrics that is split across time frames. This is useful to tell if we were recently marked as bad by others.

### DIFF
--- a/sn_networking/src/metrics/bad_node.rs
+++ b/sn_networking/src/metrics/bad_node.rs
@@ -1,0 +1,293 @@
+// Copyright 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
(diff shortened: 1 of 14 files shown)


sn_networking/src/metrics/bad_node.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn update_should_move_to_next_state() -> eyre::Result<()> {
```
jacderida, line 158: My only suggestion here would be that these could potentially be several different test cases? However, I could be wrong--I don't know enough about the context.

jacderida (approved):

sn_networking/src/metrics/bad_node.rs
RolandSherwin: Since the logic is inside a for loop, I think having just a single test would be sufficient 😄

PR #2159, opened 2024-09-30
Feat change connection confirmation

mazzi:
### Description

Screenshot 2024-09-30 at 18 38 03

We add a confirmation popup when changing connection mode, mentioning that the nodes will be reset when taking that action. 

We fixed a bit part of the flow, like cancelling changing to custom mode, was not maintaining the old value.

### DIFF
--- a/node-launchpad/src/components/popup/connection_mode.rs
+++ b/node-launchpad/src/components/popup/connection_mode.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::default::Default;
+use std::{default::Default, rc::Rc};
 
 use super::super::utils::centered_rect_fixed;
 
@@ -16,7 +16,9 @@ use ratatui::{
     layout::{Alignment, Constraint, Direction, Layout, Rect},
     style::{Style, Stylize},
     text::{Line, Span},
-    widgets::{Block, Borders, HighlightSpacing, List, ListItem, ListState, Padding, Paragraph},
+    widgets::{
+        Block, Borders, HighlightSpacing, List, ListItem, ListState, Padding, Paragraph, Wrap,
+    },
 };
 use strum::IntoEnumIterator;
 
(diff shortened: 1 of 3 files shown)


node-launchpad/src/components/popup/connection_mode.rs
```diff
@@ -305,7 +211,248 @@ impl Component for ChangeConnectionModePopUp {
...
+                .border_style(Style::new().fg(VIVID_SKY_BLUE)),
+        );
+        clear_area(f, layer_zero);
+
+        // split into 3 parts, paragraph, dash, buttons
+        let layer_two = Layout::new(
+            Direction::Vertical,
+            [
+                // for the text
+                Constraint::Length(9),
```
RolandSherwin, line 240: Is it possible move this code into a common function for both `port_range.rs` and `connection_mode.rs`. It think there are a lot of moving components here and keeping them in sync might be hard. For e.g., port_range contains `Constraint::Length(8)` instead of `Constraint::Length(9)`.

RolandSherwin:
Just something that I've noticed while trying out the build, but when starting the nodes, we don't have the pop anymore saying that we're starting nodes anymore, was this intentional? 
If so, maybe we should show some indication? Currently, there is none. Maybe we can add a new safenode to the table with "starting" status?

Anyway, this is not related to the PR! 😄

mazzi:
> Just something that I've noticed while trying out the build, but when starting the nodes, we don't have the pop anymore saying that we're starting nodes anymore, was this intentional? If so, maybe we should show some indication? Currently, there is none. Maybe we can add a new safenode to the table with "starting" status?
> 
> Anyway, this is not related to the PR! 😄

We discarded the "Starting Nodes..." popup in favor of spinners and the status "Starting" when starting nodes. I will make more testing. We are showing the popup only when NAT detection is needed. Thanks for the feedback!

Screenshot 2024-10-01 at 09 10 12

node-launchpad/src/components/popup/connection_mode.rs
mazzi: Indeed. I've tried with a helper module, but in the other hand we wanted self contained scenes/screens. I understand that this confirmation dialog is the same in the flows for the different types of connections, but they reside in different flow steps.

You would suggest a helper module with this function then?

RolandSherwin: Fair enough, we can skip this one!

RolandSherwin (approved):

PR #2163, opened 2024-10-01
refactor(autonomi): get rid of the native api

mickvandijke:
Removes the native implementation from `autonomi`.


grumbach (approved):
LGTM!

PR #2167, opened 2024-10-01
feat(metrics):  add close group shunned metrics

RolandSherwin:
- adds two new metrics to `/metrics_extended` endpoint based on the shunned_count:
   - `shunned_by_close_group` - The number of close group peers that have shunned our node
   - `shunned_by_old_close_group` - The number of close group peers that have shunned our node. This contains the peers that were once in our close group but have since been evicted.

### DIFF
--- a/sn_networking/src/event/mod.rs
+++ b/sn_networking/src/event/mod.rs
@@ -23,11 +23,11 @@ use libp2p::{
 
 use sn_protocol::{
     messages::{Query, Request, Response},
-    NetworkAddress, PrettyPrintRecordKey,
+    NetworkAddress, PrettyPrintRecordKey, CLOSE_GROUP_SIZE,
 };
 use sn_transfers::PaymentQuote;
 use std::{
-    collections::BTreeSet,
+    collections::{BTreeSet, HashSet},
     fmt::{Debug, Formatter},
 };
 use tokio::sync::oneshot;
@@ -216,6 +216,28 @@ impl Debug for NetworkEvent {
 }
 
 impl SwarmDriver {
+    /// Check for changes in our close group
+    #[cfg(feature = "open-metrics")]
+    pub(crate) fn check_for_change_in_our_close_group(&mut self) {
+        // this includes self
+        let closest_k_peers = self.get_closest_k_value_local_peers();
+
+        let new_closest_peers: Vec<_> =
+            closest_k_peers.into_iter().take(CLOSE_GROUP_SIZE).collect();
+
+        let old = self.close_group.iter().cloned().collect::<HashSet<_>>();
+        let new_members: Vec<_> = new_closest_peers
+            .iter()
+            .filter(|p| !old.contains(p))
+            .collect();
+        if !new_members.is_empty() {
+            debug!("The close group has been updated. The new members are {new_members:?}");
+            debug!("New close group: {new_closest_peers:?}");
+            self.close_group = new_closest_peers.clone();
+            self.record_change_in_close_group(new_closest_peers);
+        }
+    }
+
     /// Update state on addition of a peer to the routing table.
     pub(crate) fn update_on_peer_addition(&mut self, added_peer: PeerId) {
         self.peers_in_rt = self.peers_in_rt.saturating_add(1);
@@ -225,6 +247,11 @@ impl SwarmDriver {
         );
         self.log_kbuckets(&added_peer);
         self.send_event(NetworkEvent::PeerAdded(added_peer, self.peers_in_rt));
+
+        #[cfg(feature = "open-metrics")]
+        if self.metrics_recorder.is_some() {
+            self.check_for_change_in_our_close_group();
+        }
 
         #[cfg(feature = "open-metrics")]
         if let Some(metrics_recorder) = &self.metrics_recorder {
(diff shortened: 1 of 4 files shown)


sn_networking/src/event/mod.rs
```diff
@@ -244,6 +271,11 @@ impl SwarmDriver {
         self.log_kbuckets(&removed_peer);
         self.send_event(NetworkEvent::PeerRemoved(removed_peer, self.peers_in_rt));
 
+        #[cfg(feature = "open-metrics")]
+        if self.metrics_recorder.is_some() {
+            self.check_for_change_in_our_close_group();
```
maqi, line 276: seems there is only `log` within this function, and the change not sent to `metrics_recorder` .
is it ok ?

sn_networking/src/metrics/bad_node.rs
```diff
@@ -77,38 +111,158 @@ impl TimeFrameType {
...
+    pub(crate) fn record_shunned_metric(&mut self, peer: PeerId) {
+        // increment the metric if the peer is in the close group (new or old) and hasn't shunned us before
+        if self.close_group_peers.contains(&peer) {
+            if !self.close_group_peers_that_have_shunned_us.contains(&peer) {
+                self.metric_current_group.inc();
+                self.close_group_peers_that_have_shunned_us.insert(peer);
+            }
+        } else if self.old_close_group_peers.iter().any(|(p, _)| p == &peer)
+            && !self
+                .old_close_group_peers_that_have_shunned_us
```
maqi: looks like `close_group_peers_that_have_shunned_us` and `old_close_group_peers_that_have_shunned_us` can be merged into one set of `peers_shunned_us`

Then there are couple of places that can have code simplified to avoid `shifting between these two sets` .

sn_networking/src/metrics/bad_node.rs
```diff
@@ -77,38 +111,158 @@ impl TimeFrameType {
...
 
-    pub fn record_shunned_metric(&mut self) {
+        if !new_members.is_empty() {
+            debug!("The close group has been updated. The new members are {new_members:?}. The evicted members are {evicted_members:?}");
+            self.close_group_peers = new_closest_peers;
+
+            if self.old_close_group_peers.len() > MAX_EVICTED_CLOSE_GROUP_PEERS {
+                // clean the oldest Instant ones
+                self.old_close_group_peers
+                    .sort_by_key(|(_, instant)| std::cmp::Reverse(*instant));
```
maqi: a VecDeque then call `push_back` and `pop_front` shall achieve the same?
then don't need the timestamp ?

maqi (changes requested):

sn_networking/src/metrics/bad_node.rs
RolandSherwin: Hey @maqi, we already have a `shunned_count` metric available to us, but it does not paint the whole picture as we care more about if our close group still considers us as bad.
These metrics are primarily focused on LP to show the node's health.

We need these 2 metrics as separate ones as our close group will keep changing once we are marked as bad by them (as they block communication to us, but not instantly as we still might keep them in our RT for a while), so we need both of them.

sn_networking/src/metrics/bad_node.rs
RolandSherwin: Aye, that will work! Thanks

sn_networking/src/event/mod.rs
RolandSherwin: There is a `self.record_change_in_close_group(new_closest_peers);` within the function that will send it to `metrics_recorder`

sn_networking/src/metrics/bad_node.rs
maqi: i think the separated metric you want is : `self.metric_current_group;` and `self.metric_old_group` ?
I'm not saying merge these two.

Say if there is merged `peers_shunned_us`, then the code block here will be:
```rust
if !self.peers_shunned_us.contains(&peer){
    if self.close_group_peers.contains(&peer) {
        self.metric_current_group.inc();
    } else if self.old_close_group_peers.iter().any(|(p, _)| p == &peer) {
        self.metric_old_group.inc();
    }    
}
let _ = self.peers_shunned_us.insert(peer);
```

RolandSherwin: Right rigt I see.. Lemme try it out! That will simplify the code a lot.

maqi: And you can then have third metric: `non-close-group-peers that shunned us`
which is: `self.peers_shunned_us.len() - self.metric_current_group - self.metric_old_group`

sn_networking/src/metrics/bad_node.rs
```diff
@@ -168,18 +166,14 @@ impl BadNodeMetrics {
 impl ShunnedByCloseGroup {
     pub(crate) fn record_shunned_metric(&mut self, peer: PeerId) {
         // increment the metric if the peer is in the close group (new or old) and hasn't shunned us before
-        if self.close_group_peers.contains(&peer) {
-            if !self.close_group_peers_that_have_shunned_us.contains(&peer) {
+        if !self.old_new_group_shunned_list.contains(&peer) {
+            if self.close_group_peers.contains(&peer) {
                 self.metric_current_group.inc();
-                self.close_group_peers_that_have_shunned_us.insert(peer);
+                self.old_new_group_shunned_list.insert(peer);
```
maqi, line 172: this can be outside of the inner if clause but within the outer if clause
so that only `one code line of insertion` will be enough.

of you can even do `if self.old_new_group_shunned_list.insert(peer).is_none() { ... }` to avoid `separated check then insert`

maqi (approved):

sn_networking/src/metrics/bad_node.rs
RolandSherwin: >this can be outside of the inner if clause but within the outer if clause
so that only one code line of insertion will be enough.

Do you mean within this clause? ` !self.old_new_group_shunned_list.contains(&peer)`
But that would contain ALL peers that are observed by us. I only want the ones that are in the close group (either old / new).

PR #2179, opened 2024-10-02
feat: use EVM_NETWROK=local var to fetch local network deta

RolandSherwin:
- Write local evm node details to a CSV 
- This would allow us to get the details from other applications without parsing/hardcoding the values (e.g,. the GH action to run local networks)
- Use `EVM_NETWROK=local` to get the values from the CSV
- `safenode-manager` now supports `evm-local` flag to read from CSV

### DIFF
--- a/evmlib/src/testnet.rs
+++ b/evmlib/src/testnet.rs
@@ -60,6 +60,7 @@ pub fn start_node() -> AnvilInstance {
     // Spin up a local Anvil node.
     // Requires you to have Foundry installed: https://book.getfoundry.sh/getting-started/installation
     Anvil::new()
+        .port(4343_u16)
         .try_spawn()
         .expect("Could not spawn Anvil node")
 }
--- a/sn_node_manager/src/bin/cli/subcommands/evm_network.rs
+++ b/sn_node_manager/src/bin/cli/subcommands/evm_network.rs
@@ -1,7 +1,17 @@
+// Copyright (C) 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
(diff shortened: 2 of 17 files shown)


evmlib/src/testnet.rs
```diff
@@ -58,6 +58,7 @@ pub fn start_node() -> AnvilInstance {
     // Spin up a local Anvil node.
     // Requires you to have Foundry installed: https://book.getfoundry.sh/getting-started/installation
     Anvil::new()
+        .port(4343_u16)
```
mickvandijke, line 63: I think the behaviour should be here that we can optionally specify a port or else it should be random. If we use a constant port and have tests running on the same machine, they might yield "address already in use"

sn_node_manager/src/bin/cli/subcommands/evm_network.rs
```diff
@@ -20,22 +30,29 @@ pub enum EvmNetworkCommand {
...
+    /// Use the local EVM testnet, loaded from a CSV file.
+    EvmLocal,
 }
 
-#[allow(clippy::from_over_into)]
-impl Into<EvmNetwork> for EvmNetworkCommand {
-    fn into(self) -> EvmNetwork {
+impl EvmNetworkCommand {
+    /// Convert the command into an EvmNetwork
+    pub fn into_network(self) -> Result<EvmNetwork> {
```
mickvandijke: Since it returns a result now, a better name might be `try_into_network`.

evmlib/src/testnet.rs
b-zee: Shouldn't we run a single anvil node usually? I don't think we will run two local testnets at the same time? I think this is the same with other ports we have used in the past like `8000` for the faucet?

mickvandijke: I think that ideally every test should start a new EVM node, just to have a clean starting state and to avoid side effects.

b-zee: In that case, we could still run it on the same port as we have only a single evm node running for each test?

mickvandijke: Tests run in parallel if the host machine is multi-threaded. And also, the OS can sometimes take a while to free a port after closing an application. That's what I'm a bit worried about.

We can always just try this out though and fix later if we run into problems. Having to manually specify all the EVM params is annoying.

mickvandijke (approved):

PR #2183, opened 2024-10-02
feat(auditor): support lowercase only query

maqi:
### Description

Make auditor supporting `lowercase only` queries


maqi:
Prefer to resolve the issue on discord side

maqi:
Closed as in favour of https://github.com/maidsafe/dags-service/pull/7

PR #2185, opened 2024-10-03
feat(evmlib): Chunk->Data

joshuef:


sn_evm/CHANGELOG.md
```diff
@@ -266,7 +266,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.14.35](https://github.com/maidsafe/safe_network/compare/sn_transfers-v0.14.34...sn_transfers-v0.14.35) - 2024-01-09
 
 ### Added
-- *(client)* extra sleep between chunk verification
+- *(client)* extra sleep between data verification
```
grumbach, line 270: Not sure we should change this in the Changelog file too!!

joshuef: ups! FNR gone TOO FAR

evmlib/artifacts/DataPayments.json
mickvandijke, line 1: I need to update the smart contract first to reflect these name changes in here. 

Can this be left out in this PR? I'll open up a follow-up PR with the updated contract and new contract address.

joshuef: im not sure what parts need left out (but yeh, whatever makes sense)

joshuef: (is the contract not in here? I thought it was contained in the evm crate)

PR #2188, opened 2024-10-03
fix(node_manager): lowercase owner for discord

mazzi:
### Description

We use owner in lowercase to be able to compute scores accordingly on Discord.

Related https://github.com/maidsafe/dags-service/pull/7
Related https://github.com/maidsafe/safe_network/pull/2180

### DIFF
--- a/sn_node_manager/src/add_services/mod.rs
+++ b/sn_node_manager/src/add_services/mod.rs
@@ -79,7 +79,12 @@ pub async fn add_node(
     }
 
     let owner = match &options.owner {
-        Some(owner) => Some(owner.to_lowercase()),
+        Some(owner) => {
+            if owner.chars().any(|c| c.is_uppercase()) {
+                warn!("Owner name ({owner}) contains uppercase characters and will be converted to lowercase");
+            }
+            Some(owner.to_lowercase())
+        }
         None => None,
     };
 
--- a/sn_node_manager/src/add_services/mod.rs
+++ b/sn_node_manager/src/add_services/mod.rs
@@ -78,6 +78,11 @@ pub async fn add_node(
         check_port_availability(port_option, &node_registry.nodes)?;
     }
 
+    let owner = match &options.owner {
+        Some(owner) => Some(owner.to_lowercase()),
+        None => None,
+    };
+
     let safenode_file_name = options
         .safenode_src_path
         .file_name()
@@ -221,7 +226,7 @@ pub async fn add_node(
             name: service_name.clone(),
             node_ip: options.node_ip,
             node_port,
-            owner: options.owner.clone(),
+            owner: owner.clone(),
             rpc_socket_addr,
             safenode_path: service_safenode_path.clone(),
             service_user: options.user.clone(),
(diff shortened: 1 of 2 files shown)


sn_node_manager/src/add_services/mod.rs
```diff
@@ -78,6 +78,11 @@ pub async fn add_node(
         check_port_availability(port_option, &node_registry.nodes)?;
     }
 
+    let owner = match &options.owner {
+        Some(owner) => Some(owner.to_lowercase()),
```
maqi: shall we raise a warning here in case an `UpperCase` input detected ?

at least a `warning log` ?

maqi (approved):

sn_node_manager/src/add_services/mod.rs
mazzi: Good idea.

jacderida:
Hey mazzi,

The node manager has a test suite. It would be nice if we could add a test case for this. The tests can be a bit unwieldy though because they use a lot of mocking and stuff, so if it's too much hassle, don't worry about it.

mazzi:
> Hey mazzi,
> 
> The node manager has a test suite. It would be nice if we could add a test case for this. The tests can be a bit unwieldy though because they use a lot of mocking and stuff, so if it's too much hassle, don't worry about it.

I've modified an existing test. I didn't wanted to copy past the one that I've modified.
Did you see it? It's alright?
Thanks!

jacderida:
> > Hey mazzi,
> > The node manager has a test suite. It would be nice if we could add a test case for this. The tests can be a bit unwieldy though because they use a lot of mocking and stuff, so if it's too much hassle, don't worry about it.
> 
> I've modified an existing test. I didn't wanted to copy past the one that I've modified. Did you see it? It's alright? Thanks!

Right, thanks. However, wouldn't it need an assertion that the assigned value is in lowercase?

mazzi:
> > > Hey mazzi,
> > > The node manager has a test suite. It would be nice if we could add a test case for this. The tests can be a bit unwieldy though because they use a lot of mocking and stuff, so if it's too much hassle, don't worry about it.
> > 
> > 
> > I've modified an existing test. I didn't wanted to copy past the one that I've modified. Did you see it? It's alright? Thanks!
> 
> Right, thanks. However, wouldn't it need an assertion that the assigned value is in lowercase?

The assertion was already there. I just modified the input. Please check the whole test.

jacderida (approved):

jacderida:
> > > > Hey mazzi,
> > > > The node manager has a test suite. It would be nice if we could add a test case for this. The tests can be a bit unwieldy though because they use a lot of mocking and stuff, so if it's too much hassle, don't worry about it.
> > > 
> > > 
> > > I've modified an existing test. I didn't wanted to copy past the one that I've modified. Did you see it? It's alright? Thanks!
> > 
> > 
> > Right, thanks. However, wouldn't it need an assertion that the assigned value is in lowercase?
> 
> The assertion was already there. I just modified the input. Please check the whole test.

Ah ok cool, sorry. Approved.

PR #2198, opened 2024-10-04
feat(node): check if we have enough space before starting

RolandSherwin:
- node - Check if we have enough space before starting the node
- nodemanager - check if the `(avaialable space - space reserved by other nodes)` is enough to start the `count` nodes.


sn_node_manager/src/cmd/node.rs
```diff
@@ -849,3 +860,110 @@ fn summarise_any_failed_ops(
     }
     Ok(())
 }
+
+/// Check if the mount point has enough space to add the required number of nodes. It also factors in the space used
+/// by other nodes running on the same mount point.
+fn check_available_space(
```
jacderida, line 866: Hey, I would have a suggestion here.

Perhaps you could split this function out possibly into two, and have one that passes in the nodes to operate on, without passing in the registry?

Then you could add some unit tests for the function that does the storage calculation?

I think a few test cases would probably help to understand the different cases and failures here.

jacderida (approved):

sn_node/src/bin/safenode/main.rs
```diff
@@ -360,6 +364,41 @@ You can check your reward balance by running:
     }
 }
 
+fn check_node_disk_space(root_dir: &Path) -> Result<()> {
+    let disks = Disks::new_with_refreshed_list();
+    let disks = disks.list();
+    let required_space_bytes = (MAX_RECORDS_COUNT * *MAX_CHUNK_SIZE) as u64;
```
maqi, line 370: as mentioned at https://github.com/maidsafe/safe_network/commit/85c81e6c26c6ffcd50bcc5a82470ec87f9cafd21
the node_size shall be calculated as : `MAX_RECORDS_COUNT * (MAX_CHUNK_SIZE / 2)`

This calculation also appears duplicated in other place, hence better make it a helper facility

maqi (changes requested):

RolandSherwin:
Not required for now.

RolandSherwin:
This is currently blocked. Even our testnet machines are not spacious enough to meet the requirements. Also this can be easily bypassed, closing for now.

PR #2206, opened 2024-10-07
feat(auditor): restart webservice on client request

maqi:
### Description

auditor restart webservice on client request


maqi:
replaced by https://github.com/maidsafe/safe_network/pull/2207
as in favour of merge into stable branch

PR #2209, opened 2024-10-08
feat: simpler evm connection and env

grumbach:


RolandSherwin (approved):

evm_testnet/src/main.rs
```diff
@@ -125,6 +125,21 @@ impl TestnetData {
         if let Some((tokens, gas)) = self.tokens_and_gas {
             println!("Genesis wallet balance (atto): (tokens: {tokens}, gas: {gas})");
         }
+
+        println!();
+        println!("--------------");
+        println!("Run the CLI or Node with the following env vars set to use this network:");
```
RolandSherwin, line 131: We now store these to a csv and load them if `EVM_NETWORK=local`, so maybe we could just mention that here.

PR #2212, opened 2024-10-08
feat(autonomi): expose public writeable registers

RolandSherwin:
- expose `RegisterPermissions` from the new client
- enable public register e2e test


grumbach (approved):
LGTM

PR #2213, opened 2024-10-09
Evm range

joshuef:
range based gets ahead of evm

replaces #2082

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -744,6 +744,10 @@ jobs:
               echo "EVM_NETWORK has been set to $EVM_NETWORK"
           fi
 
+      - name: Wait from network to stabilise
+        shell: bash
+        run: sleep 30
+
       - name: Verify the routing tables of the nodes
         run: cargo test --release -p sn_node --features="local" --test verify_routing_table -- --nocapture
         env:
@@ -769,7 +773,7 @@ jobs:
         uses: maidsafe/sn-local-testnet-action@evm-dev
         with:
           action: stop
-          log_file_prefix: safe_test_logs_data_location
+          log_file_prefix: safe_test_logs_data_location_routing_table
           platform: ${{ matrix.os }}
 
       - name: Verify restart of nodes using rg
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -531,15 +531,19 @@ jobs:
   #         platform: ${{ matrix.os }}
   #         build: true
 
-  #     - name: Check SAFE_PEERS was set
-  #       shell: bash
-  #       run: |
-  #         if [[ -z "$SAFE_PEERS" ]]; then
-  #           echo "The SAFE_PEERS variable has not been set"
-  #           exit 1
-  #         else
-  #           echo "SAFE_PEERS has been set to $SAFE_PEERS"
-  #         fi
+  # # incase the faucet is not ready yet
+  # - name: 30s sleep for faucet completion
+  #   run: sleep 30
+
+  # - name: Check SAFE_PEERS was set
+  #   shell: bash
+  #   run: |
+  #     if [[ -z "$SAFE_PEERS" ]]; then
+  #       echo "The SAFE_PEERS variable has not been set"
+  #       exit 1
+  #     else
+  #       echo "SAFE_PEERS has been set to $SAFE_PEERS"
+  #     fi
 
   #     - name: execute token_distribution tests
   #       run: cargo test --release --features=local,distribution token_distribution -- --nocapture --test-threads=1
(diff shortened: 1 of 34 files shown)


.github/workflows/merge.yml
```diff
@@ -279,70 +279,70 @@ jobs:
...
-  #     - name: Start a client to edit a register writable by anyone (new client is not the owner)
-  #       if: matrix.os == 'windows-latest'
-  #       run: ./target/release/safe --log-output-dest=data-dir register edit ${{ steps.register-address-windows.outputs.REGISTER_ADDRESS }} water
-  #       env:
-  #         SN_LOG: "all"
-  #       timeout-minutes: 10
+      #     #
+      #     # Next two steps are same with a slight difference in the way they write to the output file (GITHUB_OUTPUT vs ENV:GITHUB_OUTPUT)
+      #     #
+      #     - name: Start a client to create a register writable by anyone
```
grumbach: Is this an editor auto indent artifact?

second.txt
```diff
@@ -0,0 +1 @@
+This is an error message
```
grumbach: what's this file?

sn_networking/src/cmd.rs
```diff
@@ -983,24 +974,72 @@ impl SwarmDriver {
...
+    /// From all local peers, returns any within current get_range for a given key
+    /// Excludes self
+    pub(crate) fn get_filtered_peers_exceeding_range_or_closest_nodes(
+        &mut self,
+        target_address: &NetworkAddress,
+    ) -> Vec<PeerId> {
+        let filtered_peers = self.get_filtered_peers_exceeding_range(target_address);
+        let closest_node_buffer_zone = CLOSE_GROUP_SIZE + close_group_majority();
+        if filtered_peers.len() >= closest_node_buffer_zone {
+            filtered_peers
```
grumbach, line 1016: Aren't we limiting the range-based search to a number instead of a range here?
Wouldn't that just make the address mining attack possible again?

sn_networking/src/event/kad.rs
```diff
@@ -333,26 +355,32 @@ impl SwarmDriver {
...
     /// - if multiple content hashes are found, we return a SplitRecord Error
     ///   And then we stop the kad query as we are done here.
+    ///   We do not need to wait for GetRange to be exceeded here and should return early.
     fn accumulate_get_record_found(
         &mut self,
         query_id: QueryId,
         peer_record: PeerRecord,
         _stats: QueryStats,
-        step: ProgressStep,
+        _step: ProgressStep,
```
grumbach: Why keep it if we don't use it?

sn_networking/src/transfers.rs
```diff
@@ -51,38 +49,14 @@ impl Network {
     /// If we get a quorum error, we increase the RetryStrategy
     pub async fn get_spend(&self, address: SpendAddress) -> Result<SignedSpend> {
         let key = NetworkAddress::from_spend_address(address).to_record_key();
-        let mut get_cfg = GetRecordCfg {
+        let get_cfg = GetRecordCfg {
             get_quorum: Quorum::All,
```
grumbach, line 53: Just wondering, what does Quorum::All mean with get range? All the peers that hold the data in the address range? Or all the ones in the close group?

grumbach (commented):
Good work! Couple of questions. 
I understand this was very well testet on non-EVM net, just to make sure, did we test that locally on a small testnet with EVM?

.github/workflows/merge.yml
joshuef: must be the autoformatter aye

second.txt
joshuef: must be from a local run testing output at some juncture 🤔 removing anyway

sn_networking/src/cmd.rs
joshuef: with `get_filtered_peers_exceeding_range` we get all the local peers we have within the range, so we're returning that just now, unless there's not enough, then we just grab closest here.

sn_networking/src/transfers.rs
joshuef: You can see [in kad.rs](https://github.com/maidsafe/safe_network/pull/2213/files#diff-4c7753b4198da926db9ea80d1e10daac655d3472dbc84edc034f41a54ffd0022R434-R462) we have `have_we_have_searched_thoroughly_for_quorum` which basically relaxes checks when not `Quorum::All`.

`Spends` were quorum all (as that's where the added security really benefits)

joshuef:
>  I understand this was very well testet on non-EVM net, just to make sure, did we test that locally on a small testnet with EVM?

I wanted to see what CI was doing, first. Tested locally just now and can up/download a whole folder via the cli 👍

PR #2221, opened 2024-10-10
feat: WASM bindings

b-zee:
- **feat(autonomi): JS bindings**
- **fix(autonomi): use proper dep**
- **refactor(autonomi): error return; camel case**
- **docs(autonomi): add example doc to logInit**
- **refactor/fix(autonomi): wasm32 changes for API**


joshuef (approved):
approved ready for if/when we get that last dep with fix released

PR #2222, opened 2024-10-10
fix(test): use separate wallets to prevent payment failures

RolandSherwin:
- the churn test was failing because we were using the same wallet concurrently. This caused one tx to go out before the other one got confirmed.

### DIFF
--- a/sn_node/tests/common/client.rs
+++ b/sn_node/tests/common/client.rs
@@ -7,10 +7,14 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use autonomi::Client;
+use evmlib::wallet::Wallet;
 use eyre::Result;
+use sn_evm::Amount;
 use sn_protocol::safenode_proto::{NodeInfoRequest, RestartRequest};
 use sn_service_management::{get_local_node_registry_path, NodeRegistry};
+use std::str::FromStr;
 use std::{net::SocketAddr, path::Path};
+use test_utils::evm::get_new_wallet;
 use test_utils::testnet::DeploymentInventory;
 use test_utils::{evm::get_funded_wallet, peers_from_env};
 use tokio::sync::Mutex;
@@ -35,7 +39,7 @@ const LOAD_FAUCET_WALLET_RETRIES: usize = 6;
 // mutex to restrict access to faucet wallet from concurrent tests
 static FAUCET_WALLET_MUTEX: Mutex<()> = Mutex::const_new(());
 
-pub async fn get_client_and_funded_wallet() -> (Client, evmlib::wallet::Wallet) {
+pub async fn get_client_and_funded_wallet() -> (Client, Wallet) {
     match DeploymentInventory::load() {
         Ok(_inventory) => {
             todo!("Not implemented yet for WanNetwork");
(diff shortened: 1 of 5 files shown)


evmlib/src/utils.rs
```diff
@@ -1,3 +1,13 @@
+// Copyright 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+#![allow(dead_code)]
```
RolandSherwin: Even though this is a lib crate, clippy says that the `evm_network_from_env` is never used. But it is indeed used by everyone that is calling `evm_network_from_env` 🤔

sn_node/tests/common/client.rs
```diff
@@ -107,20 +111,17 @@ pub fn get_all_rpc_addresses(_skip_genesis_for_droplet: bool) -> Result<Vec<Sock
...
-//         }
-//         Err(_) => NonDroplet::get_funded_wallet(client, to_wallet_dir, false).await,
-//     }
-// }
+/// Transfer tokens from the provided wallet to a newly created wallet
+/// Returns the newly created wallet
+pub async fn transfer_to_new_wallet(from: &Wallet, amount: usize) -> Result<Wallet> {
+    match DeploymentInventory::load() {
+        Ok(_inventory) => {
+            todo!("Not implemented yet for WanNetwork");
```
maqi, line 119: This is only for testing purpose ?
if so, better not use the word of `Not implemented yet for WanNetwork` , which gives wrong impression that `transfer to another receiver is not supported yet`.
there is `transfer_token` function provided to achieve this I think ?
or just say this is not used for test?

sn_node/tests/data_with_churn.rs
```diff
@@ -315,21 +320,40 @@ fn store_chunks_task(
...
-                .data_put(random_data, &wallet)
-                .await
-                .inspect_err(|err| {
-                    println!("Error to put chunk: {err:?}");
-                    error!("Error to put chunk: {err:?}")
-                })?;
-
-            println!("Stored Chunk/s at {data_map:?} after a delay of: {delay:?}");
-            info!("Stored Chunk/s at {data_map:?} after a delay of: {delay:?}");
+            // FIXME: The client does not have the retry repay to different payee feature yet.
```
maqi, line 323: In that case, then a retry is no use?

maqi (approved):

sn_node/tests/common/client.rs
RolandSherwin: Yes, these are just for testing as it is inside `sn_node/tests`. For the WAN network, we are yet to re-enable the tooling that is required. 
These should be temporary.

sn_node/tests/data_with_churn.rs
RolandSherwin: This retry is just to make sure that we pick a different payee if the original one goes offline. This will do the job for us now.

PR #2226, opened 2024-10-11
feat: API rework for fs less WASM compatibility

grumbach:
### Description

```
         Archive          this is a bunch of files with their paths       `archive` API layer, allows for directories (multiple files)
        /       \
   file Data   file Data  this is a Datamap of chunks of a file           `data` API layer, allows for files (multiple chunks)
     / | |      / | | \ 
    c  c c     c  c c  c  those are Chunks                                `network layer`, allows for 1-chunk of data bytes
```

- rework of the API according to [discussion here](https://maidsafe.discourse.team/t/autonomi-files-api-for-wasm/1010)
- removed `files` feature flag, now only `fs` and non fs (which is WASM compatible)
- remove `evmlib` dependency from `autonomi` as we should use `sn_evm` instead
- stop exposing low level `XorName` in the APIs
- some cleanup in the crate


autonomi/src/client/error.rs
```diff
@@ -0,0 +1,87 @@
+// Copyright 2024 MaidSafe.net limited.
```
b-zee: I don't think we should put all errors in one module. The idea is to be very specific, so we might end up with way more error types. We've already become lazy recently by just piling error variants up into these structs.

If you look at `std`, you'll see all errors also with their respective functions or modules. It keeps it much easier to edit and find the related error types.

autonomi/src/client/mod.rs
```diff
@@ -7,16 +7,22 @@
...
 pub mod address;
+pub mod error;
 
+#[cfg(feature = "data")]
+pub mod archive;
 #[cfg(feature = "data")]
 pub mod data;
-#[cfg(feature = "files")]
+#[cfg(feature = "fs")]
 pub mod files;
```
b-zee: Maybe we could rename the `files` module to `fs`? As everything in there is related to the local file system.

autonomi/tests/file.rs
```diff
@@ -6,38 +6,77 @@
...
+use std::fs::File;
+use std::io::{BufReader, Read};
 use std::time::Duration;
 use test_utils::{evm::get_funded_wallet, peers_from_env};
 use tokio::time::sleep;
+use walkdir::WalkDir;
 
+// With a local evm network, and local network, run:
+// EVM_NETWORK=local cargo test --features="fs,local" --package autonomi --test file
+#[cfg(feature = "fs")]
```
b-zee: This attribute can be removed as we have the top level guard already.

b-zee (approved):

autonomi/src/client/error.rs
grumbach: Oh okay, fair point

autonomi/src/client/mod.rs
grumbach: Make sense

autonomi/tests/file.rs
grumbach: Do we?

b-zee: See line 9.

PR #2230, opened 2024-10-14
feat(autonomi): enable files benchmark

RolandSherwin:


test_utils/src/evm.rs
```diff
@@ -23,7 +23,7 @@ pub fn get_funded_wallet() -> evmlib::wallet::Wallet {
     const DEFAULT_WALLET_PRIVATE_KEY: &str =
         "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
 
-    let private_key = env::var("EVM_PRIVATE_KEY").unwrap_or(DEFAULT_WALLET_PRIVATE_KEY.to_string());
+    let private_key = env::var("SECRET_KEY").unwrap_or(DEFAULT_WALLET_PRIVATE_KEY.to_string());
```
b-zee, line 26: What is the reason for this change? Shouldn't we keep this a bit more explicit?

b-zee (approved):

test_utils/src/evm.rs
RolandSherwin: It is to be consistent with the other places we use this var. Maybe yeah, I could change it to `EVM_PRIVATE_KEY` or `EVM_SECRET_KEY` everywhere.

PR #2231, opened 2024-10-14
chore(launchpad): refactor status screen

mazzi:
### Description

To unify patterns and to be prepared for the next feature, we use now `NodeItem` and a `StatefulTable` for the status screen as we use on the drive selection and connection popups.

With this architecture is easier to understand how to update the real time node status and changing rendering modes per node. We update stats and spinners on `Tick`.

Next feature: Call To Actions on individual nodes.


node-launchpad/src/components/status.rs
```diff
@@ -151,6 +166,94 @@ impl Status {
         Ok(status)
     }
 
+    fn update_node_items(&mut self) -> Result<()> {
+        // Iterate over existing node services and update their corresponding NodeItem
+        if let Some(ref mut items) = self.items {
+            for (node_item, item) in self.node_services.iter().zip(&mut items.items) {
```
RolandSherwin: Would this cause an issue if `node_services.len() > items.len()`? The `items` field is updated only during `// If items is None, create a new list (fallback)` part of the code. This would mean that if we were to add a new node, just the `node_services` would be updated and not `items`. Then the `zip()` would ignore these new nodes.

RolandSherwin: The solution is to keep `node_services` and `items` in sync.

mazzi: We will not delete nodes just now. That hopefully will come in the next PR when we manage nodes individually.
Thanks for your feedback @RolandSherwin !

RolandSherwin (approved):

PR #2237, opened 2024-10-15
feat(node): kill nodes running on host with little spare CPU

joshuef:
This pull request includes a small change to the `sn_node/Cargo.toml` file. The change adds a new dependency on the `sysinfo` crate with version `0.30.8` and disables the default features.

* [`sn_node/Cargo.toml`](diffhunk://#diff-e255381d82815cad4f0b0eaab942daafe80d19b247510a636789fdaa83310b52R63): Added a new dependency on `sysinfo` with version `0.30.8` and disabled default features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8256,6 +8256,7 @@ dependencies = [
  "sn_service_management",
  "sn_transfers",
  "strum",
+ "sysinfo",
  "tempfile",
  "test_utils",
  "thiserror",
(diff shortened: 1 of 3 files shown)


RolandSherwin (approved):

maqi (changes requested):
I have the impression that even for a normal node, during the procedure of replicating certain amount of records, the CPU usage could increased to 100% for a short while.

Also, the CPU metric collected is a `system wide CPU usage`, which could be affected by other factors as well.
say `running many nodes on same machine`, or `playing a CPU savage game while running a node as a spare` ?

I do suggest have this PR tested properly (with a small sized inner devnet) before merge.

jacderida:
I'm going to close this PR and replace it with a PR from my branch.

PR #2241, opened 2024-10-15
feat(autonomi): implement client events

RolandSherwin:
- emit `UploadSummary` as an event during each `data_put`
- print stats to the cli stdout


grumbach (approved):
LGTM!

PR #2242, opened 2024-10-15
API 3rd party transaction signer integration

mickvandijke:
The goal of this PR is to allow 3rd party wallet apps to sign (and send) transactions that would otherwise be automatically signed and send within the API with the private key of a wallet.

3rd party wallet apps do not expose the private key of a wallet, so we can't pass that to the API. Alternative API endpoints have to be provided that allow us to skip automatic data payments and submit proof of payments directly instead.

The transaction input data will still be formed in the API however, to make the signing and sending of transactions from the 3rd party wallet side as easy as possible.

### DIFF
--- a/evmlib/src/external_signer.rs
+++ b/evmlib/src/external_signer.rs
@@ -18,11 +18,10 @@ use std::collections::HashMap;
 pub enum Error {
     #[error("Network token contract error: {0}")]
     NetworkTokenContract(#[from] network_token::Error),
-    #[error("Chunk payments contract error: {0}")]
-    ChunkPaymentsContract(#[from] data_payments::error::Error),
+    #[error("Data payments contract error: {0}")]
+    DataPaymentsContract(#[from] data_payments::error::Error),
 }
 
-#[allow(dead_code)]
 /// Approve an address / smart contract to spend this wallet's payment tokens.
 ///
 /// Returns the transaction calldata (input, to).
@@ -36,7 +35,6 @@ pub fn approve_to_spend_tokens_calldata(
     network_token.approve_calldata(spender, value)
 }
 
-#[allow(dead_code)]
 /// Transfer payment tokens from the supplied wallet to an address.
 ///
 /// Returns the transaction calldata (input, to).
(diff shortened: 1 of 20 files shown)


evmlib/src/external_signer.rs
```diff
@@ -0,0 +1,98 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Network token contract error: {0}")]
+    NetworkTokenContract(#[from] network_token::Error),
+    #[error("Chunk payments contract error: {0}")]
+    ChunkPaymentsContract(#[from] data_payments::error::Error),
+}
+
+#[allow(dead_code)]
```
b-zee: Why are these needed for `pub`?

evmlib/src/external_signer.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub to: Address,
+    pub approve_spender: Address,
+    pub approve_amount: Amount,
+}
+
+#[allow(dead_code)]
+/// Use this wallet to pay for chunks in batched transfer transactions.
+/// If the amount of transfers is more than one transaction can contain, the transfers will be split up over multiple transactions.
+///
+/// Returns PayForQuotesCalldataReturnType, containing calldata of the transaction batches along with the approval details for the spender.
```
b-zee, line 61: What exactly is 'call data'?

evmlib/src/external_signer.rs
mickvandijke: The Error struct? It needs the same visibility as the `pay_for_quotes_calldata` function that potentially returns it

evmlib/src/external_signer.rs
mickvandijke: I have used calldata as a general term to refer to the data that is needed to form a transaction, including transaction input and the to address. `Calldata` the type is an alias to `Bytes` and should be set as the input of a transaction.

Example: https://github.com/mickvandijke/safe_network/blob/metamask-integration/autonomi/tests/external_signer.rs#L45

evmlib/src/external_signer.rs
b-zee: I meant all `#[allow(dead_code)]` attributes in those file, sorry, don't know why GitHub showed the lines before that.

b-zee (approved):

grumbach (approved):
Amazing work!

evmlib/src/external_signer.rs
mickvandijke: Good catch, removed them all now

PR #2251, opened 2024-10-16
feat: improved archives with metadata

grumbach:


autonomi/src/client/archive.rs
```diff
@@ -6,7 +6,11 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    time::SystemTime,
```
b-zee: I think we need to use `SystemTime` from `wasmtimer` here to prevent wasm problems.

```rust
#[cfg(not(target_arch = "wasm32"))]
pub use std::time::SystemTime;
#[cfg(target_arch = "wasm32")]
pub use wasmtimer::std::SystemTime;
```

or actually rather use our `sn_networking/src/target_arch.rs` mechanism here.

autonomi/src/client/archive.rs
```diff
@@ -43,6 +76,58 @@ impl Archive {
 
         Ok(root_serialized)
     }
+
+    /// Rename a file in an archive
+    /// Note that this does not upload the archive to the network
+    pub fn rename_file(&mut self, old_path: &Path, new_path: &Path) -> Result<(), ArchiveError> {
+        let data_addr = self
```
b-zee: This would not be just the `data_addr`, maybe use generic `key`?

autonomi/src/client/archive.rs
```diff
@@ -21,11 +25,40 @@ use xor_name::XorName;
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Archive {
-    pub map: HashMap<PathBuf, DataAddr>,
+    map: HashMap<PathBuf, (DataAddr, Metadata)>,
+}
+
+/// Metadata for a file in an archive
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Metadata {
+    pub created: SystemTime,
```
b-zee: Should we also add `modified` here? And update that on rename for example?

autonomi/src/client/archive.rs
```diff
@@ -21,11 +25,40 @@ use xor_name::XorName;
 /// The address of an archive on the network. Points to an [`Archive`].
 pub type ArchiveAddr = XorName;
 
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum ArchiveError {
```
b-zee:
```suggestion
pub enum RenameError {
```

If we also add a `remove` method later, we might rename this into something else. But let's not use `ArchiveError`, as it's generic and we'll probably make the mistake of piling all our error variants in here.

autonomi/src/client/archive.rs
grumbach: To assure compatibility between JS apps and rust apps, we should use the same type. 
I'll use u64 as time since UNIX_EPOCH.

autonomi/src/client/archive.rs
grumbach: Indeed, renamed to value as more accurate!

b-zee (approved):

autonomi/src/client/archive.rs
```diff
@@ -6,7 +6,15 @@
...
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
+
+#[cfg(not(target_arch = "wasm32"))]
+pub use std::time::{Duration, SystemTime, UNIX_EPOCH};
+#[cfg(target_arch = "wasm32")]
+pub use wasmtimer::std::{Duration, SystemTime, UNIX_EPOCH};
```
b-zee: Could we extend `target_arch` to export these?

PR #2258, opened 2024-10-16
feat(autonomi): allow the uploader to work with the new apis

RolandSherwin:
The uploader module provides an interface to upload data to the network, with the ability to retry failed uploads,
make repayments for failed payments, and verify the data after it has been uploaded.

### Example
```rust
let mut uploader = Uploader::new(client, wallet);
uploader.insert_chunks(chunks);
let summary = uploader.start_upload().await?;
```

### Configuration
The `Uploader` can be configured using the `UploadCfg` struct. The most notable options are the `batch_size` and
`payment_batch_size` which determine the number of data that are processed in parallel and the number of payments
that are made in a single evm transaction, respectively.
Also the `max_repayments_for_failed_data` option determines the maximum number of repayments to make if the
initial payment fails.


autonomi/src/client/utils.rs
```diff
@@ -110,35 +109,39 @@ impl Client {
...
-                target_record: None,
-                expected_holders: Default::default(),
-                is_register: false,
-            };
+        let put_cfg = if let Some(cfg) = cfg {
+            cfg
+        } else {
+            let verification = {
+                let verification_cfg = GetRecordCfg {
+                    get_quorum: Quorum::N(NonZero::new(2).expect("2 is non-zero")),
```
maqi: better use `CLOSE_GROUP_SIZE / 2` instead of a majec value ?

autonomi/src/uploader/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// Returns a receiver for UploadEvent.
+    /// This method is optional and the upload process can be performed without it.
+    pub fn get_event_receiver(&mut self) -> mpsc::Receiver<UploadEvent> {
+        self.inner
+            .as_mut()
+            .expect("Uploader::new makes sure inner is present")
+            .get_event_receiver()
+    }
+
+    /// Insert a list of chunk paths to upload to upload.
```
maqi: => Insert a list of chunk paths into uploader to upload

autonomi/src/uploader/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+    #[cfg(feature = "fs")]
+    pub fn insert_chunk_paths(&mut self, chunks: impl IntoIterator<Item = (XorName, PathBuf)>) {
+        self.inner
+            .as_mut()
+            .expect("Uploader::new makes sure inner is present")
+            .insert_chunk_paths(chunks);
+    }
+
+    /// Insert a list of chunks to upload to upload.
```
maqi: => Insert a list of chunks into uploader to upload

autonomi/src/uploader/upload.rs
```diff
@@ -0,0 +1,1235 @@
...
+                    Some((*xorname, GetStoreCostStrategy::Cheapest))
+                } else {
+                    None
+                }
+            })
+            .collect();
+    }
+
+    // registers have to be verified + merged with remote replica, so we have to fetch it first.
+    #[cfg(feature = "registers")]
```
maqi, line 103: so `data` and `registers` features can not be enabled at the same time?
that's bit strange, and is not that user-friendly.
I thought autonomi-cli shall provide different API for upload data or register, but on autonomi level, they shall use the same (or different, if the flow has to be different) functions to complete the task.

autonomi/src/uploader/upload.rs
```diff
@@ -0,0 +1,1235 @@
...
+                uploader.upload_final_balance = uploader
+                    .wallet
+                    .balance_of_tokens()
+                    .await
+                    .inspect_err(|err| {
+                        error!("Failed to get wallet balance: {err:?}");
+                    })?;
+            }
+
+            #[cfg(test)]
```
maqi: the below log is also valid for `non test` scenario ?

autonomi/src/uploader/upload.rs
```diff
@@ -0,0 +1,1235 @@
...
+                        reg_addr,
+                        task_result_sender.clone(),
+                    );
+                }
+            }
+        }
+
+        // try to push register if we have enough buffer.
+        // No other checks for the same reason as the above step.
+        #[cfg(feature = "registers")]
```
maqi: the following block of code can be inside the same above block, under one feature guard ?

maqi (changes requested):
In general, I don't think `Register` need to use `uploader`, as the content of `Register` shall be small.

Looking at the PR, to make `uploader` also supports `Register`, there are many extra feature_guard/code_blocks have to be introduced.
Which makes code more complexed and bit hard to follow.

I'd suggest make `Register` related API a separate set of functions provided via autonomi general API, instead of from `uploader`

maqi (approved):

autonomi/src/uploader/upload.rs
RolandSherwin: As discussed in slack, we should be feature gating things here. But maybe we should split this up into smaller components for better segregation in the future.

PR #2262, opened 2024-10-18
fix(networking): limit conncurrent data verifications

joshuef:
Limit concurrent verification now we do more of them. We do not need to verify 30 chunks at one time per node. Better to spread the load, do less but more nodes.

### DIFF
--- a/sn_networking/src/event/request_response.rs
+++ b/sn_networking/src/event/request_response.rs
@@ -15,13 +15,15 @@ use libp2p::{
     request_response::{self, Message},
     PeerId,
 };
-use rand::{rngs::OsRng, Rng};
+use rand::{rngs::OsRng, seq::SliceRandom, Rng};
 use sn_protocol::{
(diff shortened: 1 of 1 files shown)


sn_networking/src/event/request_response.rs
```diff
@@ -242,8 +244,17 @@ impl SwarmDriver {
 
         let event_sender = self.event_sender.clone();
         let _handle = tokio::spawn(async move {
-            let keys_to_verify =
-                Self::select_verification_data_candidates(&peers, &all_keys, &sender);
+            let keys_to_verify: Vec<_> = {
+                let mut vec = Self::select_verification_data_candidates(&peers, &all_keys, &sender)
```
maqi, line 248: The `select_verification_data_candidates` function itself only return 1 chunk candidate to be verified.
Hence, this PR has no effect on reducing the workload.

maqi (changes requested):

PR #2263, opened 2024-10-18
feat: user data stored in vault

grumbach:
- UserData can now be stored on the Network in a Vault so no need to keep addresses and register keys around
- JS bindings to get_user_data_from_vault and put_user_data_to_vault allowing the webapp to make use of this
- Autonomi API still offers lower level APIs to create custom vaults. To differenciate those a u64 hash of the app name is used in Scratchpad
- Chose to allow UserData to include names for data we keep track of: namely Registers and Archives. Although it is optional. 
- Improved the API of vault which returned clunky options instead of managing underlying errors
- exposed Bytes used throughout sn_protocol's API as higher level libs should were using the original lib leading to possible future problems if the dependencies versions mismatch

### DIFF
--- a/sn_protocol/src/storage/scratchpad.rs
+++ b/sn_protocol/src/storage/scratchpad.rs
@@ -23,8 +23,8 @@ pub struct Scratchpad {
     /// Network address. Omitted when serialising and
     /// calculated from the `encrypted_data` when deserialising.
     address: ScratchpadAddress,
-    /// Data version
-    version: u64,
+    /// Data encoding: custom apps using scratchpad should use this so they can identify the type of data they are storing
+    data_encoding: u64,
     /// Contained data. This should be encrypted
     #[debug(skip)]
     encrypted_data: Bytes,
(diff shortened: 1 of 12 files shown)


sn_protocol/src/storage/scratchpad.rs
```diff
@@ -35,10 +37,11 @@ pub struct Scratchpad {
 
 impl Scratchpad {
     /// Creates a new instance of `Scratchpad`.
-    pub fn new(owner: PublicKey) -> Self {
+    pub fn new(owner: PublicKey, version: u64) -> Self {
         Self {
             address: ScratchpadAddress::new(owner),
             encrypted_data: Bytes::new(),
+            version,
```
joshuef: I wonder if this `version` is confusing w/ the counter there. 

Probably one name should change. 

maybe this is `data_encoding` or something? 🤔

joshuef (approved):

sn_protocol/src/storage/scratchpad.rs
grumbach: Make sense, data_encoding is excellent

PR #2224, opened 2024-10-11
Range based gets

joshuef:
This pull request includes several changes to the `.github/workflows/merge.yml` file and the `sn_client` module to improve functionality and error handling. The most important changes include updating job steps in the GitHub Actions workflow, adding new methods and parameters in the `sn_client` API, and refining error handling in the `sn_client` module.

### GitHub Actions Workflow Improvements:
* Renamed a job step to "Start a client to upload cost estimate" and adjusted the corresponding command in `.github/workflows/merge.yml`.
* Updated the log file prefix for stopping nodes to `safe_test_logs_data_location_routing_table` in `.github/workflows/merge.yml`.
* Added commented-out steps for sleeping, checking faucet funding, and verifying directory contents in `.github/workflows/merge.yml` [[1]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1R1222-R1235) [[2]](diffhunk://#diff-bdb37b2ad65e049a0fc043e88813fdcf7ca0118abc11d99bbda676adfcbbb5a1L1211-R1262).

### `sn_client` API Enhancements:
* Added `KBucketDistance` to the imports and introduced the `get_range` method in `sn_client/src/api.rs` [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L17-R25) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97R309-R313).
* Modified the `get_signed_register_from_network` method to include `is_verifying` parameter and adjusted related methods to use this parameter for quorum and retry strategy [[1]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L408-R467) [[2]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L745-R793) [[3]](diffhunk://#diff-6aa8955102cfa98bc83700a7a1860ae21ace8bc5545df1ef0e4d6388defaab97L779-R827).

### Error Handling Improvements:
[...]

PR #2265, opened 2024-10-18
chore(node): restore ChunkVerification frequency to be reduced

maqi:
### Description

Restored the ChunkVerification frequency to be reduced, which was got removed during the merge range work :
https://github.com/maidsafe/safe_network/pull/2224/files#diff-f9a909155bd3d3e5ad85ae2f220de1d7d30cea5330fc236c480dbed832f27d7cL246


maqi:
not required as @joshuef suggested

PR #2270, opened 2024-10-19
chore: accumulate register at the network level during split

RolandSherwin:
- If we don't handle a split register at the lower level, then we keep retrying until the get retires run out
- Also it is upto the caller to make sure to merge the split record, which is hard to perform at every caller. e.g., replication fetch, during PUT verification etc.

I wanted to keep the network layer not exposed to the "data types", but we've already got Spends being accumulated there, and since we don't have any tests for spends right now, I did not attempt to change the current design. Maybe we should rearrange things later (when tests for spends are enabled) to make sure that the network layer stays clean.

### DIFF
--- a/autonomi/src/client/registers.rs
+++ b/autonomi/src/client/registers.rs
@@ -147,23 +147,11 @@ impl Client {
                     try_deserialize_record(&record).map_err(|_| RegisterError::Serialization)?;
                 signed_reg
             }
-            // manage forked register case
             Err(NetworkError::GetRecordError(GetRecordError::SplitRecord { result_map })) => {
-                debug!("Forked register detected for {address:?} merging forks");
-                let mut registers: Vec<SignedRegister> = vec![];
-                for (_, (record, _)) in result_map {
-                    registers.push(
-                        try_deserialize_record(&record)
-                            .map_err(|_| RegisterError::Serialization)?,
-                    );
-                }
-                let register = registers.iter().fold(registers[0].clone(), |mut acc, x| {
-                    if let Err(e) = acc.merge(x) {
-                        warn!("Ignoring forked register as we failed to merge conflicting registers at {}: {e}", x.address());
-                    }
-                    acc
-                });
-                register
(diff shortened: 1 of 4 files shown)


maqi (changes requested):
the PR seems moved split handling from autonomi to network ?

anyway, there are some places of using 'spend' incorrectly in the comment (and even the commit title ?)
shall they be corrected to `register` or `record` correspondently ?

RolandSherwin:
@maqi , as mentioned in slack, this is moved to sn_networking for the following reason:
- the get_record will be used at node & client, hence a need for a common `split register handling fn`. This is assuming we know that the record is a register.
- But if we have no clue about the register, say inside replication, we would only surface the error after all the retires are exhausted. Where we would then handle it, but ideally we should have dealt with the error during the first GET instead of retrying and exhausting the retries.

autonomi/src/client/registers.rs
```diff
@@ -147,23 +147,11 @@ impl Client {
...
-                let register = registers.iter().fold(registers[0].clone(), |mut acc, x| {
-                    if let Err(e) = acc.merge(x) {
-                        warn!("Ignoring forked register as we failed to merge conflicting registers at {}: {e}", x.address());
-                    }
-                    acc
-                });
-                register
+                error!("Got split record error for register at address: {address}. This should've been handled at the network layer");
+                Err(RegisterError::Network(NetworkError::GetRecordError(
+                    GetRecordError::SplitRecord { result_map },
```
maqi, line 153: Register is always `merged` if split detected.
And, due to the fact that register-record holds list of ops, it might be possible the get_record will report with split.

I don't think here a merge can be skipped , neither shall the error to be raised up ?

also, as in this PR, the `merge` is moved into `network` level, this means on client level here, we shall never receive an error of split ?
In that case, we shall raise/log something more serious ?

RolandSherwin: Yeah, we'd not get this error on the client side. I have just noted this down for future reference.

maqi (approved):

PR #2298, opened 2024-10-23
fix: do not retry for non-bad peer connections

joshuef:
Do not always redial failed peers at networking level, leave to higher levels. (otherwise more state needed here to prevent loops)

### DIFF
--- a/sn_networking/src/event/swarm.rs
+++ b/sn_networking/src/event/swarm.rs
@@ -7,8 +7,8 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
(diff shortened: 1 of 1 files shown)


mazzi (approved):
LGTM. Maybe is worth to analyze (if has not been done yet) the different cases on
```
L400: let should_clean_peer = match error
```
and why we are cleaning up for `DialError::Aborted` `DialError::DialPeerConditionFalse` only?

joshuef:
> and why we are cleaning up for DialError::Aborted DialError::DialPeerConditionFalse only?

We're _not_ cleaning up for those. As those are issues on _our_ end, essentially

joshuef:
(This went in via another PR just FYI)

PR #2308, opened 2024-10-24
fix(networking): replication factor of 5 to reduce CPU overhead

joshuef:


joshuef:
Closing as CPU here is a fair whack higher than joshuef:MinViableChangesReduced

PR #2310, opened 2024-10-24
fix(ci): make the nightly tests work with evm

RolandSherwin:
- also increases "large file" download time from 2 mins to 3 mins inisde `merge.yml`

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,7 +15,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+            safe_path: /home/runner/.local/share/safe
+          - os: windows-latest
+            safe_path: C:\\Users\\runneradmin\\AppData\\Roaming\\safe
+          - os: macos-latest
+            safe_path: /Users/runner/Library/Application\ Support/safe
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -26,77 +32,181 @@ jobs:
...
           # happens if we build the node manager via testnet action
           CARGO_TARGET_DIR: ${{ matrix.os == 'windows-latest' && './test-target' || '.' }}
         timeout-minutes: 15
 
-      - name: Create and fund a wallet to pay for files storage
+
+      # FIXME: do this in a generic way for localtestnets
+      - name: export default secret key
+        if: matrix.os != 'windows-latest'
+        run: echo "SECRET_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" >> $GITHUB_ENV
```
mazzi, line 75: This secret keys should be a repo secret maybe?

.github/workflows/nightly.yml
```diff
@@ -26,77 +32,181 @@ jobs:
...
+        shell: bash
+      - name: Set secret key for Windows
+        if: matrix.os == 'windows-latest'
+        run: echo "SECRET_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
mazzi, line 79: Same

.github/workflows/nightly.yml
```diff
@@ -167,13 +263,12 @@ jobs:
       - name: Run logging tests
         timeout-minutes: 25
         run: cargo test --release --package sn_logging
-
+  
       - name: Run register tests
-        shell: bash
         timeout-minutes: 50
+        run: cargo test --release -package sn_registers
```
mazzi:
```suggestion
        run: cargo test --release --package sn_registers
```

.github/workflows/nightly.yml
RolandSherwin: Hey @mazzi, this is just the local evm SECRET KEY, we don't mind exposing it.
The solution would be probably to let the CLI use it without configuring things.

RolandSherwin: It is already exposed through readme and stuff 😄

maqi (changes requested):
* comment out tests (spend, token) can actually be removed ?
* can the `nightly` tests got executed somehow? or it can only be executed (to confirm they work as expected) `AFTER` the merge and wait for a day ?

maqi (approved):

PR #2311, opened 2024-10-24
chore: remove restart from `safenode` process

jacderida:
We now have code that shuts down `safenode` processes if system-wide CPU usage has been over 50% for 5 consecutive minutes. We don't want the nodes to be restarted automatically; node operators can do that when they see fit.

I'm not actually sure why this behaviour was ever desirable. It may have been before we were regularly deploying using services, where the OS service infrastructure can provide restarts on failures.


jacderida:
This seems to be required for the RPC restart command. Going to close the PR.

PR #2314, opened 2024-10-25
feat: private data, private archives, vault support and CLI integration

grumbach:
- private data
- private archives
- private file/dir
- vault integration for private data
- CLI integration of all the above
- CLI uploads use private as default and --public as an option

Note that unlike public data, private data can only be downloaded by the owner! Either from the same client (needs local user data) OR from another client with a synced vault! Here's the flow:
- client pushes private data
- private data access is kept in local user data
- client syncs it to vault
- from another device/app (with the same EVM key), client loads vault
- this new device now had access to the private data too

### DIFF
--- a/autonomi/src/client/data_private.rs
+++ b/autonomi/src/client/data_private.rs
@@ -1,0 +1,138 @@
+// Copyright 2024 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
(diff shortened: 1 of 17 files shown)


autonomi/src/client/data_private.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    pub fn from_hex(hex: &str) -> Result<Self, hex::FromHexError> {
+        let data = hex::decode(hex)?;
+        Ok(Self(Chunk::new(Bytes::from(data))))
+    }
+
+    /// Get a private address for [`PrivateDataAccess`]. Note that this is not a network address, it is only used for refering to private data client side.
+    pub fn address(&self) -> String {
+        hash_to_short_string(&self.to_hex())
+    }
```
grumbach, line 38: This is kind of an important design decision. Wanted to outline this here. Public data refers to a Network address which points to a DataMap. `PrivateDataAccess` is not an address it IS THE DATAMAP. As it it can be very big (up to max chunk size), we need a way to refer to it in the client side, thus using its hash.

mickvandijke (approved):
LGTM 🔥

PR #2318, opened 2024-10-25
fix(launchpad): update status after resetting

mazzi:
When resetting nodes, we clear the status screen.

We add more logging.

### DIFF
--- a/node-launchpad/src/components/status.rs
+++ b/node-launchpad/src/components/status.rs
@@ -46,6 +46,7 @@ use std::{
     time::{Duration, Instant},
     vec,
 };
+use strum::Display;
 use tokio::sync::mpsc::UnboundedSender;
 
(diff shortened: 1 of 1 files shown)


node-launchpad/src/components/status.rs
```diff
@@ -273,6 +274,11 @@ impl Status<'_> {
         Ok(())
     }
 
+    fn clear_node_items(&mut self) {
+        debug!("Cleaning items on Status page");
+        self.items.as_mut().unwrap().items.clear();
```
RolandSherwin:
```suggestion
        if let Some(items) = self.items.as_mut() {
        	items.clear();
        	debug!("Cleared the items on status page");
        }
```
Better to avoid unwraps..

RolandSherwin (approved):

PR #2327, opened 2024-10-25
refactor: wasm & metamask needed changes

mickvandijke:


autonomi/src/uploader/upload.rs
```diff
@@ -75,13 +77,26 @@ pub(super) async fn start_upload(
...
+                task_result_sender.clone(),
+                uploader.cfg.payment_batch_size,
+            )?;
+        }
+        PaymentOption::Receipt(receipt) => {
+            let _ = task_result_sender
+                .send(TaskResult::MakePaymentsOk {
+                    payment_proofs: receipt.clone(),
+                })
+                .await;
```
RolandSherwin: I think we should probably not support `Receipt` insidde the uploader? The uploader would perform get_store cost even if we provide the receipt here, also I'm not really sure how this would work with the whole flow. It was not designedd with pre-paid data in mind.

I think we should make the uploader moduler if we were to support receipts. In the meantime, we can error out if we don't want to support receipts?

mickvandijke: We need prepaid uploads for the webapp that releases Tuesday, so I had to quick and dirty get support in there. 

Do you have a better idea of how we might do it? Ideally, It must be very quick to do, as I need to get started with coupling this API to the webapp.

RolandSherwin: I took a look at the code again and I think this commit should cover the edge cases.. https://github.com/maidsafe/safe_network/commit/09b0a72dd29a85206812b6c837759389a9a1fd7f

mickvandijke: Brilliant, thank you!!

grumbach (approved):
Fantastic work!! LGTM!

PR #2332, opened 2024-10-28
repl perf improvements

joshuef:
Will be replaced by PR https://github.com/maidsafe/safe_network/pull/2336

### DIFF
--- a/sn_networking/src/cmd.rs
+++ b/sn_networking/src/cmd.rs
@@ -29,6 +29,7 @@ use sn_protocol::{
     NetworkAddress, PrettyPrintRecordKey,
 };
 use std::{
+    cmp::Ordering,
     collections::{BTreeMap, HashMap},
     fmt::Debug,
     time::Duration,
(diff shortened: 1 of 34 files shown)


sn_networking/src/cmd.rs
```diff
@@ -980,26 +984,30 @@ impl SwarmDriver {
...
+            .collect();
 
-                // Map KBucketKey<PeerId> to PeerId.
-                let peer_id = key.into_preimage();
-                Some(peer_id)
+        // Binary search to find the index where we exceed the acceptable range
+        let split_index = sorted_peers
+            .binary_search_by(|key| {
+                let distance = target_key.distance(key);
+                if distance >= acceptable_distance_range {
```
maqi, line 998: note this is a change from `kbucket index range check` to `exact range check`,
which could be too stict that not leaving any margin for edge cases.

sn_networking/src/record_store.rs
```diff
@@ -70,14 +70,14 @@ const MIN_STORE_COST: u64 = 1;
...
-    local_key: KBucketKey<PeerId>,
     /// The address of the peer owning the store
     local_address: NetworkAddress,
     /// The configuration of the store.
     config: NodeRecordStoreConfig,
-    /// A set of keys, each corresponding to a data `Record` stored on disk.
+    /// Main records store remains unchanged for compatibility
     records: HashMap<Key, (NetworkAddress, RecordType)>,
+    /// Additional index organizing records by distance bucket
+    records_by_bucket: HashMap<u32, HashSet<Key>>,
```
maqi, line 80: I think we don't need separate fields of `records_by_bucket` and `records`, they can be merged into:
`records: HashMap<u32, HashMap<Key, (NetworkAddress, RecordType)>>`
during `key` query, the extra effor of calculating kbucket index will be compenstated by break a large map into smaller ones

and above all, it avoids hassale of `have to sync two maps` (which could be vulerable to bugs)

jacderida:
This was merged in #2336

PR #2333, opened 2024-10-28
feat: improved fs uploads performance

grumbach:
- Parallel file uploading with capped concurrency to avoid memory burst
- Parallel chunk uploading with capped concurrency to avoid memory burst
- Local testing reveals this PR improves upload performance of CLI directories upload by **x50** FIFTY!!
- Configurable batching for files and chunks through env vars

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -110,6 +110,10 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
+      - name: Run autonomi tests
+        timeout-minutes: 25
+        run: cargo test --release --package autonomi --lib --features="full,fs"
+
       - name: Run node tests
         timeout-minutes: 25
         run: cargo test --release --package sn_node --lib
--- a/autonomi/src/client/fs.rs
+++ b/autonomi/src/client/fs.rs
@@ -21,11 +21,19 @@ use super::archive::{Archive, ArchiveAddr};
 use super::data::{DataAddr, GetError, PutError};
 
 /// Number of files to upload in parallel.
+/// Can be overridden by the `FILE_UPLOAD_BATCH_SIZE` environment variable.
 pub static FILE_UPLOAD_BATCH_SIZE: LazyLock<usize> = LazyLock::new(|| {
-    std::thread::available_parallelism()
-        .map(|n| n.get())
-        .unwrap_or(1)
-        * 8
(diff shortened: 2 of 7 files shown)


autonomi/src/client/fs.rs
```diff
@@ -8,15 +8,19 @@
...
 use bytes::Bytes;
 use sn_evm::EvmWallet;
 use sn_networking::target_arch::{Duration, SystemTime};
 use std::path::PathBuf;
+use tokio::task::JoinError;
 
 use super::archive::{Archive, ArchiveAddr};
 use super::data::{DataAddr, GetError, PutError};
 
+pub(crate) const FILE_UPLOAD_BATCH_SIZE: usize = 128;
```
b-zee: What's the rationale behind 128? Do we need batching at all?

grumbach: I tried with unlimited and my memory useage spiked to the moon. 
So it seems you were right about batching, and we probably need some. Local testing revealed that 128 was a sweet spot (at least on my machine), not too high for mem, not too slow.

mazzi:
Good one! What about a couple of tests? To check file integrity and probably we can profile in the future.
What do you think?

autonomi/src/client/fs.rs
b-zee: I do worry a little about the relative performance (e.g. PC vs Raspberry Pi). Can we relate this value to CPU core count?

https://doc.rust-lang.org/std/thread/fn.available_parallelism.html

grumbach: Oooh excellent! Yes!

.github/workflows/merge.yml
```diff
@@ -110,6 +110,10 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
+      - name: Run autonomi tests
+        timeout-minutes: 25
+        run: cargo test --release --package autonomi --lib --features="full,fs"
+
       - name: Run autonomi tests
```
ermineJose: i think the same workflow is repeated twice! **Run autonomi tests**
Maybe some rebase conflicts?

grumbach: Good catch! I'm actually sneaking this back in as it was removed as a side effect of: 
https://github.com/maidsafe/safe_network/pull/2334

ermineJose: oh cool!

PR #2335, opened 2024-10-28
Reduce ChunkVerificationIntensity per node

joshuef:
(Last commit is new)

We were looping over all chunks (one at a time), a) printing that to logs! (a lot of reads), and b) causing probably longer lasting feedback loops here based upon replication... 

here we more simply verify only one key at once (chosen at random), and could then just increase the amount of ndoes we verify against... instead of scrutinizing one for potentially thousands of keys!

================================================================================

Will be replaced by the PR https://github.com/maidsafe/safe_network/pull/2336


jacderida:
This was merged in #2336

PR #2338, opened 2024-10-28
metrics: cap node wallet balance to i64::MAX

RolandSherwin:


mazzi (approved):
LGTM. This will fix the overflow but no the metric gathering.
Watch out with the PR title

PR #2347, opened 2024-10-29
incrase chunk verifications and responsbile range

joshuef:


maqi:
work has been partically covered by other PRs

PR #2351, opened 2024-10-29
feat(autonomi): download chunks in parallel

RolandSherwin:


grumbach (approved):
Good one! Maybe a bigger batch size for download would make sense?

autonomi/src/client/utils.rs
```diff
@@ -35,16 +35,25 @@ impl Client {
...
+                        })
+                        .map(|chunk| EncryptedChunk {
+                            index: info.index,
+                            content: chunk.value,
+                        })
+                }
+            })
+            .buffered(*CHUNK_DOWNLOAD_BATCH_SIZE);
+
+        while let Some(encrypted_chunk) = stream.next().await {
```
maqi: this while loop will terminate on the first `failed to fetch` as well,
which will give incorrect input to the `decrypt`
there shall be some error out for retry or print out to give more explicit error info the user ?

RolandSherwin: The return type is a result. We get a `None` if the iter runs out. We do short circuit on error below.
I've renamed it to `encrypted_chunk_result` to be more accurate.

PR #2358, opened 2024-10-30
fix: allow split records during crdt record puts verifications

grumbach:


b-zee (approved):

sn_networking/src/lib.rs
```diff
@@ -1017,6 +1017,11 @@ impl Network {
                             NetworkAddress::from_record_key(&record_key),
                         ));
                     }
+                    Err(NetworkError::GetRecordError(GetRecordError::SplitRecord { .. }))
```
maqi, line 1020: `non Wasm` build has a handling code `handle_split_record_error` trying to merge `Registers`, i.e. here get Register won't emit a SplitRecord.

not sure about Vault, but if it is a crdt type, then a merge shall be undertaken as well.

meanwhile, the `Wasm` build donesn't try to carry out `merge` yet, which could be an issue.

Swallowing an error here is also bit concerning, hence I'd suggest give it a proper handling here (if emits) or still keeps raise the error up.

maqi (changes requested):

maqi (approved):

PR #2364, opened 2024-10-30
feat(autonomi): support prepaid put operations

mickvandijke:
Allows doing `put` operations with a `Receipt` as proof of payment instead of having to pass a `Wallet` and doing payments on the fly.

This is needed for web3 wallet support and also opens the doors to retrying put operations without having to pay again if we store the receipts somewhere.

You can run an integration test that puts a private file in a vault using receipts with this command (make sure to run a local network):

```
EVM_NETWORK=local cargo test --package=autonomi --features=local,data,vault,external-signer --test external_signer
```


grumbach (approved):
LGTM!

PR #2365, opened 2024-10-30
fix: restart node only on restart rpc command

RolandSherwin:
- If a node wants to be terminated due to high cpu usage, `run_node` would return `Ok(None)` and it will be shutdown
- If `run_node` returns `Ok(Some(_))` we restart the node.


jacderida:
Closing this because it was cherry picked into the RC branch. It couldn't be merged in because it was branched from `main`. It will be merged back in to `main` when the RC branch is merged. If not, it can be re-opened.

PR #2368, opened 2024-10-31
chore(node): create a RecordCache struct

maqi:
### Description

create a RecordCache struct

PR #2367, opened 2024-10-31
Attempt to fix CPU issues, need more work to fully remove RBS

grumbach:
Shall get replaced by PR https://github.com/maidsafe/safe_network/pull/2369

PR #2357, opened 2024-10-29
fix(node): ignore timestamp elapsed call failure

maqi:
### Description

ignore timestamp elapsed call failure

### Type of Change

Please mark the types of changes made in this pull request.

- [x] Bug fix (non-breaking change which fixes an issue)

PR #2369, opened 2024-10-31
Revert "Revert "Revert "Merge pull request #2224 from joshuef/RangeBasedGets"""

maqi:
This reverts commit 49115fd211e3417f4d613c18a73cda898adb682d.

CLOSE https://github.com/maidsafe/safe_network/pull/2357
CLOSE https://github.com/maidsafe/safe_network/pull/2360
CLOSE https://github.com/maidsafe/safe_network/pull/2367
CLOSE https://github.com/maidsafe/safe_network/pull/2368

### Description

Take PR 2224 out of the production code

### DIFF
--- a/sn_evm/src/data_payments.rs
+++ b/sn_evm/src/data_payments.rs
@@ -244,17 +244,29 @@ impl PaymentQuote {
             return false;
         }
 
+        // TODO: Double check if this applies, as this will prevent a node restart with same ID
+        if new_quote.quoting_metrics.received_payment_count
+            < old_quote.quoting_metrics.received_payment_count
+        {
+            info!("claimed received_payment_count out of sequence");
+            return false;
+        }
+
         let old_elapsed = if let Ok(elapsed) = old_quote.timestamp.elapsed() {
             elapsed
         } else {
-            info!("timestamp failure");
-            return false;
+            // The elapsed call could fail due to system clock change
+            // hence consider the verification succeeded.
+            info!("old_quote timestamp elapsed call failure");
+            return true;
         };
         let new_elapsed = if let Ok(elapsed) = new_quote.timestamp.elapsed() {
             elapsed
         } else {
-            info!("timestamp failure");
-            return false;
+            // The elapsed call could fail due to system clock change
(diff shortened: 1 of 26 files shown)


sn_evm/src/data_payments.rs
```diff
@@ -244,17 +244,29 @@ impl PaymentQuote {
...
         };
         let new_elapsed = if let Ok(elapsed) = new_quote.timestamp.elapsed() {
             elapsed
         } else {
-            info!("timestamp failure");
-            return false;
+            // The elapsed call could fail due to system clock change
+            // hence consider the verification succeeded.
+            info!("new_quote timestamp elapsed call failure");
+            return true;
```
grumbach, line 269: Not 100% sure about this, @mickvandijke could you confirm this makes sense?

grumbach (commented):
Amazing work @maqi !

sn_evm/src/data_payments.rs
mickvandijke: I don't really understand why these checks are done at all. Maybe I'm missing something. But wouldn't we only need to verify a quote's signature to check if it is made by this peer? Why do we need to store historical quotes and compare with them?

maqi: This was to confirm a quote given by other node are not faked.

mickvandijke: The quote signature being valid already rules out that the quote could have been faked by another node or client. So I think we can get rid of this.

This might be out of scope for this PR though, we can do it in a follow-up PR.

mickvandijke:
Great work! 💪

b-zee (approved):

sn_evm/src/data_payments.rs
maqi: > This was to confirm a quote given by other node are not faked.

Sry, I don't mean `faked by other`, I mean `faked by the quoter itself` .
i.e. a node gives a quote with out_of_sequence data to gaim higher chance to be picked as a payee, or got a overpay by the client

PR #2377, opened 2024-11-04
feat: remove external address on too many connection error

RolandSherwin:
- Remove listerner as a external address addr if the swarm reports an expired listerner.
- Remove an external address on high error rate.

### DIFF
--- a/sn_networking/src/event/swarm.rs
+++ b/sn_networking/src/event/swarm.rs
@@ -63,8 +63,10 @@ impl SwarmDriver {
                     relay_peer_id, ..
                 } = *event
                 {
-                    self.relay_manager
-                        .on_successful_reservation_by_client(&relay_peer_id, &mut self.swarm);
+                    if let Some(relay_manager) = self.relay_manager.as_mut() {
+                        relay_manager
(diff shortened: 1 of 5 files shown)


sn_networking/src/event/swarm.rs
```diff
@@ -300,13 +304,15 @@ impl SwarmDriver {
...
             }
-
             SwarmEvent::NewListenAddr {
                 mut address,
                 listener_id,
             } => {
                 event_string = "new listen addr";
 
+                info!("Local node is listening {listener_id:?} on {address:?}");
+                println!("Local node is listening on {address:?}"); // TODO: make it print only once
```
maqi: is this printout really necessary ?
could pollute cli terminal a lot ?

maqi (approved):

sn_networking/src/event/swarm.rs
RolandSherwin: We were already doing that, I just moved the line to the top. Maybe I'll remove it, I searched if we were using that stdout anywhere, but can't find any results.

PR #2397, opened 2024-11-06
feat(autonomi): run self encryption non-blocking

b-zee:
Previously, the `encrypt` method might block other tasks from
progressing in the same runtime (thread).

In JS this would cause an issue as it's single threaded, meaning the UI
got blocked during encryption

Has to be compiled using nightly compiler arguments, e.g. to run tests:
```
RUSTFLAGS='-C target-feature=+atomics,+bulk-memory,+mutable-globals' CARGO_TARGET_DIR=/home/zamaan/src/safe_network/target/wasm wasm-pack test --release --firefox autonomi -Z build-std=std,panic_abort --test wasm
```


mickvandijke (approved):

mickvandijke:
Unfortunately I did not test this properly before approving. When I tried to actually use it in the webapp it caused quite some problems.

First I needed a way for Nuxt/Vite to enable the `SharedArrayBuffer` in a local dev environment. I tried many proposed solutions, but none of them worked. Eventually got a working setup like this in the `nuxt.config.ts` file:

```
export default defineNuxtConfig({
  ...
  nitro: {
    routeRules: {
      '/**': {
        headers: {
          'Cross-Origin-Embedder-Policy': 'require-corp',
          'Cross-Origin-Opener-Policy': 'same-origin'
        },
      },
    },
  },
  ...
})
```

We might have to do something similar on the platform we deploy the static site at.

Then I ran into this error on `Chrome`:

```
RuntimeError: Atomics.wait cannot be called in this context
```

And this error on `Firefox`:

```
TypeError: waiting is not allowed on this thread
```

I don't think the `Chrome` error is fixable. The `Firefox` error we might be able to fix, but I haven't found a solution yet.

So I propose that we rollback this PR and potentially deal with the `sync` encryption in the API consumer side instead of within the API.

PR #2400, opened 2024-11-07
feat(launchpad): node selection and log view

mazzi:
### Description

Node selection enabled. Hitting `L` will show logs for the node.

Screenshot 2024-11-07 at 12 05 38

### DIFF
--- a/node-launchpad/src/components/status.rs
+++ b/node-launchpad/src/components/status.rs
@@ -14,6 +14,7 @@ use super::{
 };
 use crate::action::OptionsActions;
 use crate::components::popup::port_range::PORT_ALLOCATION;
+use crate::components::utils::open_logs;
 use crate::config::get_launchpad_nodes_data_dir_path;
 use crate::connection_mode::ConnectionMode;
 use crate::error::ErrorPopup;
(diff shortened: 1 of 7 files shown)


node-launchpad/src/components/status.rs
```diff
@@ -1080,6 +1092,7 @@ impl<T> StatefulTable<T> {
             None => self.last_selected.unwrap_or(0),
         };
         self.state.select(Some(i));
+        self.last_selected = Some(i);
```
RolandSherwin, line 1095: I think the `TableState` now has a `self.state.select_next(), select_previous(), etc`  fn, I remembered seeing that on a reddit release post last week! Checked it out and it indeed it has one. This can simplify a lot of code here

RolandSherwin (approved):

mazzi:
Closed in favor of https://github.com/maidsafe/safe_network/pull/2405

Follow up PR required

node-launchpad/src/components/status.rs
mazzi: @RolandSherwin I've just checked `self.state.select_next()` and `select_previous()`. I think that the behavior is not the one that we expect. Goes out of bounds after reaching the last element instead of selecting the first one. Same the other way around. I will wait another release to see the evolution. I'll keep an eye on it.

https://ratatui.rs/highlights/v029/#table-support-selecting-columns-and-cells-%EF%B8%8F

PR #2409, opened 2024-11-07
feat: make LP multithreaded and limit local set to node mgmnt

RolandSherwin:
- This would allow the LP to take in keystrokes, tick etc, while we are starting nodes, upgrading etc.


RolandSherwin:
Closed  in favor of #2410

PR #2410, opened 2024-11-07
feat: make LP multithreaded and limit local set to node mgmnt

RolandSherwin:
#### Fixed
- Launchpad now runs on multiple threads and this would allow operations to be performed while nodes are being started, upgraded etc.


mazzi (approved):
💥 🚀

PR #2411, opened 2024-11-07
fix(launchpad): change copy on popup estimated time

mazzi:
### Description

We display an estimated time for the upgrade.

Screenshot 2024-11-07 at 18 34 35

### DIFF
--- a/node-launchpad/src/app.rs
+++ b/node-launchpad/src/app.rs
@@ -120,7 +120,7 @@ impl App {
         let change_connection_mode = ChangeConnectionModePopUp::new(connection_mode)?;
         let port_range = PortRangePopUp::new(connection_mode, port_from, port_to);
         let rewards_address = RewardsAddress::new(app_data.discord_username.clone());
-        let upgrade_nodes = UpgradeNodesPopUp::default();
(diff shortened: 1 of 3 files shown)


node-launchpad/src/app.rs
```diff
@@ -120,7 +120,7 @@ impl App {
         let change_connection_mode = ChangeConnectionModePopUp::new(connection_mode)?;
         let port_range = PortRangePopUp::new(connection_mode, port_from, port_to);
         let rewards_address = RewardsAddress::new(app_data.discord_username.clone());
-        let upgrade_nodes = UpgradeNodesPopUp::default();
+        let upgrade_nodes = UpgradeNodesPopUp::new(app_data.nodes_to_start);
```
RolandSherwin, line 123: If I'm not wrong, this is set during the init of `UpgradeNodesPopUp`, what if user has added 2 more nodes? This would not be updated right?

mazzi: That is true. Also `fixed_inteval` maybe can be passed as an argument. I will review the first one.
Good catch!

jacderida (approved):

PR #2414, opened 2024-11-07
feat(network): accumulate split scratchpads

RolandSherwin:


sn_networking/src/lib.rs
```diff
@@ -656,6 +657,34 @@ impl Network {
                         }
                     }
                 }
+                // Accumulate scratchpad
+                else if kind == &RecordKind::Scratchpad {
```
maqi: would prefer this to be changed to `match` pattern, so that any future ones won't got `skipped through` accidently

sn_networking/src/lib.rs
```diff
@@ -656,6 +657,34 @@ impl Network {
...
+                    info!("For record {pretty_key:?}, we have a split record for a scratchpad. Accumulating scratchpads");
```
maqi: logs need to be updated

sn_networking/src/lib.rs
```diff
@@ -656,6 +657,34 @@ impl Network {
...
+                    let Ok(scratchpad) = try_deserialize_record::<Scratchpad>(record) else {
+                        error!(
+                            "Failed to deserialize scratchpad {pretty_key}. Skipping accumulation"
+                        );
+                        continue;
+                    };
+
+                    if !scratchpad.is_valid() {
+                        warn!("Rejecting Scratchpad PUT with invalid signature");
```
maqi: put an address so that the log will be more explicit for tracing?

maqi (approved):

PR #2426, opened 2024-11-11
feat(autonomi): expose inner MerkleReg type

b-zee:
This inner `MerkleReg` CRDT allows access to the history (DAG) of the register.


happybeing:
Thanks @b-zee, may I suggest a couple of extras and that you mention the history:

```
   // Owner's PublicKey is used to verify API calls which are signed using the owner's RegisterSecretKey
    pub fn owner(&self) -> PublicKey {
        self.signed_reg.base_register().address().owner()
    }

    // RegisterPermissions determine who can mutate the register
    pub fn permissions(&self) -> &RegisterPermissions {
        self.signed_reg.base_register().permissions()
    }

   /// Get the low-level CRDT of the register
   /// This is needed to allow access to register history
    pub fn inner_merkle_reg(&self) -> &MerkleReg<Entry> {
        // Unstable method according to documentation.
        &self.crdt_reg.merkle_reg()
    }
```

b-zee:
Excellent suggestions @happybeing!

maqi (changes requested):
there are somee rustc compiling failure reports got included by mistake ?

PR #2431, opened 2024-11-11
test(CI): execute large_file_upload test with websocket

maqi:
### Description

execute large_file_upload test with websocket

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -5,7 +5,7 @@ on:
   # on main, we want to know that all commits are passing at a glance, any deviation should help bisecting errors
   # the merge run checks should show on master and enable this clear test/passing history
   merge_group:
-    branches: [main, evm-dev]
+    branches: [main, alpha*, beta*, rc*]
   pull_request:
     branches: ["*"]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/merge.yml
```diff
@@ -1265,6 +1270,24 @@ jobs:
...
+        timeout-minutes: 1
+        run: |
+          incoming_connection_errors=$(rg "IncomingConnectionError" "${{ matrix.node_data_path }}" -c --stats | \
+            rg "(\d+) matches" | rg "\d+" -o) || { echo "Failed to find IncomingConnectionError error"; exit 0; }
+          if [ -z "$incoming_connection_errors" ]; then
+            echo "Doesn't find any IncomingConnectionError error !"
+          else
+            echo "Found $incoming_connection_errors IncomingConnectionError errors."
+          fi
+          if ! rg "UnexpectedEof" "${{ matrix.node_data_path }}" -c --stats; then
```
RolandSherwin, line 1284: We encounter different error types other than `UnexpectedEof`,  that are related to a connection being closed just because a previous connection to the same peer succeeded. Such as:

`UnexpectedEof`
`Custom { kind: Other, error: Right(Left(Right(Select(Failed)))) }`
`error: Io(Os { code: 104, kind: ConnectionReset, message: \"Connection reset by peer\" }) })))))) } })`

Maybe it'd be hard to get the count for all the types, we could just skip it and just find the count for IncommingConnectionError

maqi: the test `without ws` will have no such specific Eof error detected, hence I'd expect the same after the patch, when `ws` is enabled.

We'd expect a drop on the IncomingConnectionError as well.
which, I'd suggest to put a threshold in for the `with ws` test, to avoid any future regression

RolandSherwin: Right makes sense!

RolandSherwin (approved):

maqi:
In favour of https://github.com/maidsafe/safe_network/pull/2433

PR #2432, opened 2024-11-11
fix(network): do not log connection error for established connections

RolandSherwin:


sn_networking/src/driver.rs
```diff
@@ -819,6 +820,10 @@ pub struct SwarmDriver {
     // Peers that having live connection to. Any peer got contacted during kad network query
     // will have live connection established. And they may not appear in the RT.
     pub(crate) live_connected_peers: BTreeMap<ConnectionId, (PeerId, Instant)>,
+    /// The peers that we connected to recently to.
```
maqi: The peers that we recently connected to.

sn_networking/src/event/swarm.rs
```diff
@@ -659,6 +674,30 @@ impl SwarmDriver {
...
+    /// Older entries are removed if the list exceeds 50.
+    fn insert_latest_connected_peers(&mut self, addr: Multiaddr) {
+        while self.latest_connected_peers.len() >= 50 {
+            // remove the oldest entry
+            let Some(oldest) = self
+                .latest_connected_peers
+                .iter()
+                .min_by_key(|(_, time)| *time)
+                .map(|(addr, _)| addr.clone())
+            else {
```
maqi, line 694: this `else` branch is bit weird, on what condition, there will `break` happens ?

RolandSherwin: That is the `let else` pattern. Basically translates the following
```
let x = if let Some(x) = item  {
 x
else {
 return;
};
```
to:
```
if Some(x) = item else {
 return;
}
```
Avoids some boilerplate code

maqi (approved):

PR #2441, opened 2024-11-13
fix: do not use tcp/ws for relay connections

RolandSherwin:
- trying to request a WS connection to be our Relay causes too many `CircuitReqDenied`. This PR would make sure that we do not use WS connection for relay purposes.


maqi (approved):

RolandSherwin:
Needs further investigation. In main, the tcp/ws is implicitly ignored inside that function.

PR #2444, opened 2024-11-14
feat(launchpad): node actions

mazzi:
#### Internal crates
* Changes on `node-manager` to be less verbose (`Verbosity::Minimal` on calls)

### Main feature
* Now we can start/stop, add, remove individual nodes. For this we implemented a locking mechanism (semaphore) per node (`locked` on `NodeItem`). It's set to `true` before starting a `NodeManagementTask` and `false` when succeeds or returns with error. There are some helper functions for that effect.
 
#### Validation
* Validations when adding nodes (`[+]`) regarding the maximum amount of nodes and available disk space. Some helpers were created for the error messages.
* Validation of some edge cases like operations without nodes, when the application starts from scratch.
* New popup screen when trying to remove (`[-]`) a node. Gives a warning.

### Styling
* Changed footer style and behavior depending on the application state. Now is sort of flexible when shrinking the terminal (minimal resolution 80x24).

### Extras
* Some minor error message formatting.
* Changed the legend when upgrading nodes as is difficult to pass the amount of nodes between scenes. Instead of calculating total time of upgrade, we just mention how much takes per node.

### Bug fix 🐞 
* We added `CONNECTION_TIMEOUT_START` as a constant. We were using a hardcoded value as timeout when starting nodes that was making few RPC connection attempts and failing too early. 

🛑  Do not merge straight away so we squash and merge.

### DIFF
--- a/node-launchpad/.config/config.json5
+++ b/node-launchpad/.config/config.json5
@@ -8,8 +8,8 @@
       "<h>": {"SwitchScene":"Help"},
       "<H>": {"SwitchScene":"Help"},
 
-      "<Ctrl-s>": {"StatusActions":"StartNodes"},
-      "<Ctrl-S>": {"StatusActions":"StartNodes"},
+      "<Ctrl-s>": {"StatusActions":"StartStopNode"},
+      "<Ctrl-S>": {"StatusActions":"StartStopNode"},
       "<Ctrl-Shift-s>": {"StatusActions":"StartNodes"},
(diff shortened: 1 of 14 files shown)


RolandSherwin:
I'm working on merging this one. c.c @mazzi

RolandSherwin:
Superseded by https://github.com/maidsafe/autonomi/pull/2807

PR #2446, opened 2024-11-15
chore!: improve ChunkProofVerification

maqi:
### Description

🚨 Breaking Changes

improve ChunkProofVerification

### DIFF
--- a/sn_node/src/node.rs
+++ b/sn_node/src/node.rs
@@ -40,7 +40,10 @@ use std::{
     },
     time::Duration,
 };
-use tokio::{sync::mpsc::Receiver, task::{spawn, JoinSet}};
+use tokio::{
+    sync::mpsc::Receiver,
+    task::{spawn, JoinSet},
+};
 
 use sn_evm::EvmNetwork;
 
@@ -63,10 +66,10 @@ const HIGHEST_SCORE: usize = 100;
 
 /// Any nodes bearing a score below this shall be considered as bad.
 /// Max is to be 100 * 100
-const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 2000;
+const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 5000;
 
-/// in ms, expecting average StorageChallenge complete time to be around 500ms.
-const TIME_STEP: usize = 100;
+/// in ms, expecting average StorageChallenge complete time to be around 250ms.
+const TIME_STEP: usize = 20;
 
 /// Helper to build and run a Node
 pub struct NodeBuilder {
@@ -268,7 +271,7 @@ impl Node {
                 tokio::time::interval(UNRELEVANT_RECORDS_CLEANUP_INTERVAL);
             let _ = irrelevant_records_cleanup_interval.tick().await; // first tick completes immediately
 
-            // use a random neighbour storege challenge ticker to ensure
+            // use a random neighbour storage challenge ticker to ensure
             // neighbour do not carryout challenges at the same time
             let storage_challenge_interval: u64 =
                 rng.gen_range(STORE_CHALLENGE_INTERVAL_MAX_S / 2..STORE_CHALLENGE_INTERVAL_MAX_S);
@@ -333,7 +336,7 @@ impl Node {
 
                         let _handle = spawn(async move {
                             Self::storage_challenge(network).await;
-                            trace!("Periodic storege challenge took {:?}", start.elapsed());
+                            trace!("Periodic storage challenge took {:?}", start.elapsed());
                         });
                     }
                 }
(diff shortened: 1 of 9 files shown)


sn_node/src/node.rs
```diff
@@ -277,6 +278,17 @@ impl Node {
                 tokio::time::interval(UNRELEVANT_RECORDS_CLEANUP_INTERVAL);
             let _ = irrelevant_records_cleanup_interval.tick().await; // first tick completes immediately
 
+            // use a random neighbour storege challenge ticker to ensure
+            // neighbour do not carryout challenges at the same time
+            let storege_challenge_interval: u64 =
```
grumbach: typo: storage

sn_node/src/node.rs
```diff
@@ -491,28 +514,14 @@ impl Node {
...
+                        return;
                     }
-                    // Now ALL attempts failed, hence report the issue.
-                    // Note this won't immediately trigger the node to be considered as BAD.
-                    // Only the same peer accumulated three same issue
-                    // within 5 mins will be considered as BAD.
-                    // As the chunk_proof_check will be triggered every periodical replication,
-                    // a low performed or cheaty peer will raise multiple issue alerts during it.
+                    info!("Peer {peer_id:?} failed storage existence challenge.");
+                    // TODO: shall the challenge failure immediately triggers the node to be removed?
```
grumbach: I'd say reduce connection score when we introduce the feature. If score falls too low, sever connection.

sn_node/src/node.rs
```diff
@@ -712,6 +717,126 @@ impl Node {
...
+            results.push((key.clone(), result));
+        } else {
+            let all_local_records = network.get_all_local_record_addresses().await;
+
+            if let Ok(all_local_records) = all_local_records {
+                // Only `ChunkRecord`s can be consistantly verified
+                let mut all_chunk_addrs: Vec<_> = all_local_records
+                    .iter()
+                    .filter_map(|(addr, record_type)| {
+                        if *record_type == RecordType::Chunk {
```
grumbach: Why only chunks? I think all records should be checked.

sn_node/src/node.rs
```diff
@@ -770,58 +895,51 @@ impl Node {
...
-    } else {
-        error!(
-                 "To verify peer {peer_id:?} Could not get ChunkProof for {key:?} as we don't have the record locally."
-            );
-        true
-    };
+    let request = Request::Query(Query::GetChunkExistenceProof {
+        key: key.clone(),
+        nonce,
+        expected_answers: CLOSE_GROUP_SIZE,
```
grumbach: Why close group size? Not sure I understand What expected_answers represents? It feels like 1 is an exception for client reqs, maybe this var needs to be more explicit?

grumbach (commented):
Good work in general

sn_node/src/node.rs
maqi: register/spend has nature of could have different versions under the same name,
which, it's not guaranteed to be always consistence across the network.
they are also `small` in generally cases, which `cheaper` to be cached, and include them could pollute the scoring ?

sn_node/src/node.rs
maqi: >Why close group size? 

just picked to utilize an existing const.

as mentioned in comments/docs, this factor affects performance a lot.
Ideally, client set this factor based on its expectaction of security and acceptance of slowness.
Meanwhile node ste this factor based on its dynamic performance.

and, yeah, regarding the naming, any better suggestion ?
how about `difficulty` ?

sn_node/src/node.rs
```diff
@@ -341,6 +365,17 @@ impl Node {
...
                     }
+                    // runs every storage_challenge_interval time
+                    _ = storage_challenge_interval.tick() => {
+                        let start = Instant::now();
+                        debug!("Periodic storage challenge triggered");
+                        let network = self.network().clone();
+
+                        let _handle = spawn(async move {
+                            Self::storage_challenge(network).await;
+                            trace!("Periodic storege challenge took {:?}", start.elapsed());
```
grumbach: storage typo

sn_node/src/node.rs
```diff
@@ -277,6 +290,17 @@ impl Node {
                 tokio::time::interval(UNRELEVANT_RECORDS_CLEANUP_INTERVAL);
             let _ = irrelevant_records_cleanup_interval.tick().await; // first tick completes immediately
 
+            // use a random neighbour storege challenge ticker to ensure
```
grumbach: Storage

sn_node/src/node.rs
grumbach: They can be self verifiable, so even if different, the sig is what matters

grumbach: If we only verify storage of chunks what's my incentive as a node to keep other records? 
We should check all records regardless of what type.

grumbach (commented):
Good work! I think we should include all records regardless of the type in the check. During verification, differing records can be dealt with by the verifier (fetch unknown reg, spend, scratch). All data is self verifiable so detecting bad data is easy.

sn_node/src/node.rs
maqi: > They can be self verifiable, so even if different, the sig is what matters

Unfortunately, they are not.
To verify the signature, the `owner's pub key` must be known.
But nodes doesn't have that info, unless being contacted by the client.

suggested to have client check all types of records

grumbach (approved):
Excellent work Qi!

PR #2459, opened 2024-11-21
feat(networking): add bootstrap cache for peer discovery

dirvine:
# Bootstrap Cache Implementation

## Summary
This PR introduces a bootstrap cache mechanism to improve network bootstrapping efficiency. The cache persists information about previously known peers, enabling faster network rejoining and reducing cold-start latency.

## Motivation
When nodes join the network, they need to discover and connect to existing peers. Without a persistent cache, nodes must rediscover the network from scratch each time they restart, which is inefficient and time-consuming. This implementation addresses that limitation.

## Technical Details
- Implements a persistent cache storing known peer information
- Cache is maintained in the [bootstrap_cache](cci:7://file:///Users/davidirvine/Desktop/safe_network/bootstrap_cache:0:0-0:0) directory
- Entries include peer IDs and their last known addresses
- Cache is automatically updated as peers are discovered
- Stale entries are pruned periodically

## Impact
- Faster network rejoining for nodes
- Reduced network bootstrap time
- More resilient peer discovery process
- Lower network overhead during node startup

## Testing
The implementation includes:
- Unit tests for cache operations
- Integration tests for peer discovery
- Persistence verification tests

## Future Considerations
- Cache entry expiration policy could be tuned based on network metrics
- Additional peer metadata could be stored for better peer selection

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,6 @@ sn_node_manager/.vagrant
 uv.lock
 *.so
 *.pyc
-
-*.pyc
 *.swp
 
+/vendor/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "autonomi",
     "autonomi-cli",
+    "bootstrap_cache",
     "evmlib",
     "evm_testnet",
     "sn_build_info",
--- a/bootstrap_cache/Cargo.toml
+++ b/bootstrap_cache/Cargo.toml
@@ -18,8 +18,20 @@ tempfile = "3.8.1"
 thiserror = "1.0"
 tokio = { version = "1.0", features = ["full", "sync"] }
 tracing = "0.1"
+url = "2.4.0"
 
 [dev-dependencies]
 wiremock = "0.5"
 tokio = { version = "1.0", features = ["full", "test-util"] }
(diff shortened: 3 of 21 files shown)


rreive:
How is security addressed in this model?

dirvine:
> How is security addressed in this model?

This is not a security related issue per se' It's a mechanism to reconnect to the network easily. It is cross any client/node on the machine. 

The security aspect would be who you connect to. If somebody broke into your machine and replaced the file with all bad nodes you could end up on the wrong network, but in that case I imagine you would quickly know. 

In terms of connecting to a wrong network clients would not find their data and does woudl likely not be paid, but it's unlikely to cause you to reveal any secret information unless you were using a badly written app, but that app woudl still be bad on a good nework.

loziniak:
Looks like worth an [RFC](https://github.com/maidsafe/rfcs)...?

loziniak:
Could this cache be used also by clients?

RolandSherwin:
Yes clients can use this @loziniak , refer to https://github.com/maidsafe/autonomi/pull/2487 for the work on top of this!

RolandSherwin:
Superseded by https://github.com/maidsafe/autonomi/pull/2487

PR #2486, opened 2024-12-03
chore: use `ant` as the `antctl` service user

jacderida:
We want the service user to be aligned with the user we use for the deployment, and we've already changed that to the `ant` user.


ant-node-manager/src/bin/cli/main.rs
```diff
@@ -1411,14 +1411,14 @@ async fn configure_winsw(verbosity: VerbosityLevel) -> Result<()> {
...
+    // `C:\Users\<username>\autonomi\winsw.exe`, sitting it alongside the other safe-related binaries.
     //
     // However, if the node manager has been obtained by other means, we can put winsw.exe
     // alongside the directory where the services are defined. This prevents creation of what would
-    // seem like a random `safe` directory in the user's home directory.
+    // seem like a random `autonomi` directory in the user's home directory.
     let safeup_winsw_path = dirs_next::home_dir()
         .ok_or_else(|| eyre!("Could not obtain user home directory"))?
-        .join("safe")
+        .join("autnomi")
```
RolandSherwin:
```suggestion
        .join("autonomi")
```

RolandSherwin (approved):

PR #2487, opened 2024-12-03
feat: impl bootstrap cache

RolandSherwin:
- Introduces `ant-bootstrap` crate which replaces the `ant-peers-acquisition` crate.
- Every node now by default writes to a single `bootstrap_cache_<network_id>.json` file in the data dir. It contains a map of PeerId to their addresses. We track the success/failure of each address.
- These files are synced on periodic interval. Nodes don't overwrite, but instead `sync` the values.
- `network-contacts` feature flag is now removed, and now we use a `--testnet` flag to denote if we should connect to the mainnet or to a testnet.


ant-peers-acquisition/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/maidsafe/autonomi"
 version = "0.5.7"
 
 [features]
+default = ["network-contacts"]
```
b-zee: Maybe we should integrate `ant-peers-acquisition` into `ant-bootstrap`? So we have just one crate related to bootstrap contacts.

Justfile
```diff
@@ -68,15 +68,15 @@ build-release-artifacts arch nightly="false":
     cargo binstall --no-confirm cross
     cross build --release --target $arch --bin nat-detection $nightly_feature
     cross build --release --target $arch --bin node-launchpad $nightly_feature
-    cross build --release --features network-contacts,websockets --target $arch --bin autonomi $nightly_feature
+    cross build --release --features network-contacts,websockets --target $arch --bin ant $nightly_feature
```
b-zee: Could we merge these changes in a separate PR?

ant-peers-acquisition/Cargo.toml
RolandSherwin: Yes! That is the plan, we'd use the `ant-bootstrap::PeersArgs` instead of the other one.

Justfile
RolandSherwin: Sorry I think that was a previous git commit that somehow snuck in here, it is updated now.

ant-peers-acquisition/Cargo.toml
b-zee: Cool!

.gitignore
```diff
@@ -36,8 +36,7 @@ sn_node_manager/.vagrant
 .venv/
 uv.lock
 *.so
-*.pyc
-
 *.pyc
 *.swp
 
+/vendor/
```
b-zee, line 42: What is this for?

ant-peers-acquisition/Cargo.toml
jacderida: Could we just lose this as a 'feature'? I think that was very excessive anyway.

RolandSherwin: Yes! We'd remove this.
The `network-contacts` just indicates if we want to fetch the `mainnet` contacts or not.

This can be replaced by a simple `--testnet` flag.

.gitignore
RolandSherwin: No idea, it was from David's commit.

ant-networking/src/driver.rs
```diff
@@ -1005,6 +1025,39 @@ impl SwarmDriver {
...
                 },
+                Some(()) = Self::conditional_interval(&mut bootstrap_cache_save_interval) => {
+                    let Some(bootstrap_cache) = self.bootstrap_cache.as_mut() else {
+                        continue;
+                    };
+                    let Some(current_interval) = bootstrap_cache_save_interval.as_mut() else {
+                        continue;
+                    };
+
+                    if let Err(err) = bootstrap_cache.sync_and_save_to_disk(true).await {
```
maqi: this will be risky as will slow down the `swarm_driver` handling, and may even result in `dead-lock`, if running multiple nodes on one machine.
`swarm_driver` shall only handle small/instant tasks. any `async` functions (potentially could be long or blocked by others) shall be passed to a spwaned task to handle, or even fired as an event to higher layer (network) to be handled.

maqi (changes requested):

maqi:
the memcheck CI run has a step of `Start a node instance to be restarted` , which currently using : 
```code
./target/release/antnode --root-dir $RESTART_TEST_NODE_DATA_PATH --log-output-dest $RESTART_TEST_NODE_DATA_PATH --local --rewards-address "0x03B770D9cD32077cC0bF330c13C114a87643B124" &
```
and relies on the `ANT_PEERS` env to be set.

It will be the best to refactor it to use `bootstrap cache` shared_file to startup with, so that the work can be verified by CI directly.

ant-networking/src/driver.rs
```diff
@@ -1005,6 +1025,58 @@ impl SwarmDriver {
...
+                    }
+
+                    // add a variance of 1% to the max interval to avoid all nodes writing to disk at the same time.
+                    let max_cache_save_duration =
+                        Self::duration_with_variance(bootstrap_cache.config().max_cache_save_duration, 1);
+
+                    // scale up the interval until we reach the max
+                    let new_duration = Duration::from_secs(
+                        std::cmp::min(
+                         current_interval.period().as_secs() * bootstrap_cache.config().cache_save_scaling_factor,
```
maqi: just to ensure that an overflow multiple won't cause a panic

ant-networking/src/driver.rs
```diff
@@ -1005,6 +1025,58 @@ impl SwarmDriver {
...
+                    };
+                    let Some(current_interval) = bootstrap_cache_save_interval.as_mut() else {
+                        continue;
+                    };
+                    let start = Instant::now();
+
+                    let config = bootstrap_cache.config().clone();
+                    let mut old_cache = bootstrap_cache.clone();
+
+                    let new = match BootstrapCacheStore::new(config) {
```
maqi, line 1040: doesn't need to have a new one ?
just cotinue with the existing one shall be ok?

anyway, this depends on how the BootstrapCache handles `merge`.

ant-networking/src/event/swarm.rs
```diff
@@ -506,6 +515,15 @@ impl SwarmDriver {
                     }
                 };
 
+                // Just track failures during outgoing connection with `failed_peer_id` inside the bootstrap cache.
+                // OutgoingConnectionError without peer_id can happen when dialing multiple addresses of a peer.
+                // And similarly IncomingConnectionError can happen when a peer has multiple transports/listen addrs.
+                if let (Some((_, failed_addr, _)), Some(bootstrap_cache)) =
```
maqi: I think this block of code can be moved into the below block of `if should_clean_peer {` ?
where a peer is for sure hits a serious issue and got removed from the RT ?

ant-node/src/node.rs
```diff
@@ -160,11 +177,25 @@ impl NodeBuilder {
             None
         };
 
+        if !self.initial_peers.is_empty() && self.bootstrap_cache.is_some() {
```
maqi: this block (180 - 191) of code can be down in : 
```rust
let initial_peers = match (self.initial_peers.is_empty(), self.bootstrap_cache) {
    (true, Some(_)) => return Err(Error::InitialPeersAndBootstrapCacheSet),
    (true, None) => self.initial_peers.clone(),
    (false, Some(cache)) => cache.get_sorted_addrs().cloned().collect(),
    (false, None) => vec![],
};
```
Which will be more readable ?

also, shall `initial_peers` always be preferred, i.e. it shall be `(true, _) => self.initial_peers.clone(),` ?

maqi (approved):

ant-bootstrap/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn sync(&mut self, other: &Self) {
+        trace!("Syncing our state {self:?} with and other: {other:?}.");
+        if self.last_seen == other.last_seen {
+            return;
+        }
+
+        self.success_count = self.success_count.saturating_add(other.success_count);
+        self.failure_count = self.failure_count.saturating_add(other.failure_count);
+
+        // if at max value, reset to 0
```
maqi, line 166: shall be `capped`, not `reset` ?
i.e. no need to take this reset step

maqi (approved):

ant-networking/src/driver.rs
RolandSherwin: The existing one will contain the tracked peers and will only be cleared after calling `flush`. That is spawned as a separate task, so we cannot clone it as well. So need a new one here.

ant-node/src/node.rs
RolandSherwin: Ahh.. we cannot get the initial bootstrap peers from the cache store anylonger as it won't be initialized anymore. 
The loading of cache happens at a higher layer and it is fed into `initial_peers`. I will modify according to this, thanks for the catch!

ant-bootstrap/src/lib.rs
RolandSherwin: This won't be an issue in the real world, but for some reason if we were to reached the max for `success`, then only the `failures` will be tracked until it catches up with the `success`, making this peer useless as it will not be reliable (success = failure = max)

So just wrap around to 0.

PR #2495, opened 2024-12-05
Quoting upgrade nodeside

grumbach:


ant-networking/src/record_store.rs
```diff
@@ -1736,255 +1628,4 @@ mod tests {
...
-        rewards_addr: RewardsAddress,
-        records_stored: AtomicUsize,
-        nanos_earned: AtomicU64,
-        payments_received: AtomicUsize,
-    }
-
-    // takes a long time to run
-    #[ignore]
-    #[test]
-    fn address_distribution_sim() {
```
maqi, line 1751: I'd suggest move the above `calculate_cost_for_records` function into the test mod, and retain these tests.
This serves as `simulating tests of what pricing curve to be`.

ant-node/src/put_validation.rs
```diff
@@ -646,22 +645,22 @@ impl Node {
...
                 .current_reward_wallet_balance
                 .set(new_value);
         }
         self.events_channel()
-            .broadcast(crate::NodeEvent::RewardReceived(storecost, address.clone()));
+            .broadcast(crate::NodeEvent::RewardReceived(AttoTokens::from(reward_amount), address.clone()));
 
         // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
-        info!("Total payment of {storecost:?} atto tokens accepted for record {pretty_key}");
+        info!("Total payment of {reward_amount:?} atto tokens accepted for record {pretty_key}");
```
maqi, line 657: this and couple of following logs/prints could be misleading if switched to say `pay median` or `gradual emission` scheme.
So, need to make sure the returned `reward_amount` is really `the amount paid to this node directly`, or `the cost calculated according to the quoting metrics and pricing curve`, or something else.

if `self.evm_network().verify_data_payment()` cann't return with something like `(transaction_total_amount, paid_to_node, calculated_from_metrics)`, then we shall be careful on the wording here.

ant-networking/src/lib.rs
```diff
@@ -406,54 +406,51 @@ impl Network {
...
                     storage_proofs,
-                }) => {
+                })) => {
                     if !storage_proofs.is_empty() {
-                        debug!("Storage proofing during GetStoreCost to be implemented.");
+                        debug!("Storage proofing during GetStoreQuote to be implemented.");
                     }
-                    all_costs.push((peer_address, payment_address, PaymentQuote::zero()));
+                    info!("Address {record_address:?} was already paid for according to {peer_address:?}, ending quote request");
+                    return Ok(vec![]);
```
maqi, line 442: this actually creates a hole that if any malicious node replies with RecordExists , it will blocks the upload.
I'd suggest leave it and continue with collect of other responses.
Then decided whether to pay or not by: if find majority of RecordExists responses, or verified can fetch from network

grumbach:
Continued in #2497

PR #2496, opened 2024-12-05
fix(metrics): use the newer crate name for all the metrics

RolandSherwin:


ant-node/src/metrics.rs
```diff
@@ -59,7 +59,7 @@ enum RecordType {
 
 impl NodeMetricsRecorder {
     pub(crate) fn new(registries: &mut MetricsRegistries) -> Self {
-        let node_metadata_sub_registry = registries.metadata.sub_registry_with_prefix("sn_node");
+        let node_metadata_sub_registry = registries.metadata.sub_registry_with_prefix("ant_node");
         node_metadata_sub_registry.register(
             "safenode_version",
```
jacderida: Hey Roland, I think we need to rename this one too? `safenode_version` -> `antnode_version`.

jacderida (approved):

PR #2497, opened 2024-12-05
Upgradable smart contracts and updated quotation flow

mickvandijke:
Client changes are a WIP.


evmlib/src/contract/payment_vault/implementation.rs
```diff
@@ -0,0 +1,30 @@
...
...8116603f0116810167ffffffffffffffff81118282101715610cc357610cc3610c1f565b604052818152828201602001871015610cda575f5ffd5b816020840160208301375f602083830101528093505050509250929050565b5f5f828403610120811215610d0c575f5ffd5b610d168585610bdb565b9250606060bf1982011215610d29575f5ffd5b5060c0830190509250929050565b602081525f82518060208401528060208501604085015e5f604082850101526040601f19601f83011684010191505092915050565b5f5f60208385031215610d7d575f5ffd5b823567ffffffffffffffff811115610d93575f5ffd5b8301601f81018513610da3575f5ffd5b803567ffffffffffffffff811115610db9575f5ffd5b856020606083028401011115610dcd575f5ffd5b6020919091019590945092505050565b5f5f60408385031215610dee575f5ffd5b8235610df981610c0b565b946020939093013593505050565b5f60208284031215610e17575f5ffd5b8135610b9481610c0b565b634e487b7160e01b5f52603260045260245ffd5b8135610e4181610c0b565b81546001600160a01b0319166001600160a01b039190911617815560208201356001820155604090910135600290910155565b5f60208284031215610e84575f5ffd5b5051919050565b5f82518060208501845e5f92019182525091905056fe360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbca26469706673582212203894ca52be6e6323aa3d296efd566c7f21d1723d4c66c56aed8a5f75a96b579d64736f6c634300081c0033";
```
mickvandijke: This absurd long byte code is only temporary

ant-node/src/put_validation.rs
```diff
@@ -646,22 +645,22 @@ impl Node {
...
                 .current_reward_wallet_balance
                 .set(new_value);
         }
         self.events_channel()
-            .broadcast(crate::NodeEvent::RewardReceived(storecost, address.clone()));
+            .broadcast(crate::NodeEvent::RewardReceived(AttoTokens::from(reward_amount), address.clone()));
 
         // vdash metric (if modified please notify at https://github.com/happybeing/vdash/issues):
-        info!("Total payment of {storecost:?} atto tokens accepted for record {pretty_key}");
+        info!("Total payment of {reward_amount:?} atto tokens accepted for record {pretty_key}");
```
maqi, line 666: this and couple of following logs/prints could be misleading if switched to say pay median or gradual emission scheme.
So, need to make sure the returned reward_amount is really the amount paid to this node directly, or the cost calculated according to the quoting metrics and pricing curve, or something else.

if self.evm_network().verify_data_payment() cann't return with something like (transaction_total_amount, paid_to_node, calculated_from_metrics), then we shall be careful on the wording here.

ant-networking/src/lib.rs
```diff
@@ -406,54 +406,51 @@ impl Network {
...
                     storage_proofs,
-                }) => {
+                })) => {
                     if !storage_proofs.is_empty() {
-                        debug!("Storage proofing during GetStoreCost to be implemented.");
+                        debug!("Storage proofing during GetStoreQuote to be implemented.");
                     }
-                    all_costs.push((peer_address, payment_address, PaymentQuote::zero()));
+                    info!("Address {record_address:?} was already paid for according to {peer_address:?}, ending quote request");
+                    return Ok(vec![]);
```
maqi: this actually creates a hole that if any malicious node replies with RecordExists , it will blocks the upload.
I'd suggest leave it and continue with collect of other responses.
Then decided whether to pay or not by: if find majority of RecordExists responses, or verified can fetch from network

evmlib/src/contract/payment_vault/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+    let mut amount = Amount::ZERO;
+
+    // TODO: @mick change this for loop to a batch when the smart contract changes
+    for (quote_hash, quoting_metrics, rewards_address) in payment {
+        let payment_verification: PaymentVerification = payment_vault
+            .verify_payment(quoting_metrics, (quote_hash, rewards_address, Amount::ZERO))
+            .await
+            .map(|is_valid| PaymentVerification {
+                quote_hash,
+                amount_paid: Amount::from(1), // TODO: update placeholder amount when the smart contract changes
```
maqi: this will be used by node instance, hence `can not` be a placeholder.
i.e. the smart_contract API must return the value before RC.
though, the smart_contract itself can use such placeholder for future impl.

evmlib/src/contract/payment_vault/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+pub async fn get_market_price(
+    network: &Network,
+    quoting_metrics: QuotingMetrics,
+) -> Result<Amount, error::Error> {
+    let provider = http_provider(network.rpc_url().clone());
+    let payment_vault = PaymentVaultHandler::new(*network.data_payments_address(), provider);
+    payment_vault.get_quote(quoting_metrics).await
+}
+
+/// Helper function to verify whether a data payment is valid
```
maqi: better put a comment to explain the return value is `the amount just paid to self` or `the amount paid to all payees`

evmlib/src/contract/payment_vault/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+                is_valid,
+            })?;
+
+        // CODE REVIEW: should we fail on a single invalid payment?
+        if !payment_verification.is_valid {
+            return Err(error::Error::PaymentInvalid);
+        }
+
+        if owned_quote_hashes.contains(&quote_hash) {
+            amount += payment_verification.amount_paid;
```
maqi, line 53: seems the returned value is for `amount paid to all payees` .
which is different to other places of logs and metrics.
need to sync of that.

maqi (changes requested):

maqi (approved):

grumbach:
replaced by #2519 (rebased)

PR #2498, opened 2024-12-06
WIP: progress in quoting land

grumbach:


ant-networking/src/lib.rs
```diff
@@ -406,6 +408,10 @@ impl Network {
             .send_and_get_responses(&close_nodes, &request, true)
             .await;
 
+        // consider data to be already paid for if 1/3 of the close nodes already have it
+        let mut peer_already_have_it = 0;
+        let enough_peers_already_have_it = close_nodes.len() / 3;
```
maqi, line 413: to avoid malicious attacke, better at least have `2` holders.
for `close_nodes.len() = 5`, `enough_peers_already_have_it` will have a value of `1`, which is somehow vulunerable.
maybe make it `std::cmp::max(2, close_nodes.len()/3)` or `close_nodes.len() / 2` instead.

ant-node/src/put_validation.rs
```diff
@@ -614,24 +614,30 @@ impl Node {
...
-        let quote_expiration_time_in_secs = quote_expiration_time
+        let _quote_expiration_time_in_secs = quote_expiration_time
             .duration_since(UNIX_EPOCH)
             .map_err(|e| {
                 Error::InvalidRequest(format!(
                     "Payment quote timestamp is invalid for record {pretty_key}: {e}"
                 ))
             })?
             .as_secs();
+        // NB TODO @mick: can we check if the quote has expired with block time in evmlib? Or should nodes do it manually here? Else keep the block below
```
maqi, line 625: the `quote` was made by the node, not related to the block time, hence have to do it manually ?
unless node using block time on quote, which requires an extra query to the block?

mickvandijke: We can't really couple a transaction to a quote anymore since we no longer use the transaction's event logs to verify payments. So we'd have to do it manually.

ant-networking/src/lib.rs
grumbach: Make sense let's go with: `close_nodes.len() / 2`

PR #2504, opened 2024-12-08
feat: provide new versioning mechanism

jacderida:
Implements the versioning scheme that was discussed and agreed.

It only uses the version from the `ant-protocol` crate; no keys or other external sources are provided. Only the `MAJOR` and `MINOR` parts of the version are considered. The patch version can advance without being considered.

We have discussed changing the version of other crates such that they are synchronised with this protocol version.

### DIFF
--- a/ant-protocol/src/version.rs
+++ b/ant-protocol/src/version.rs
@@ -9,88 +9,97 @@
 use lazy_static::lazy_static;
 
 lazy_static! {
-    /// The node version used during Identify Behaviour.
-    pub static ref IDENTIFY_NODE_VERSION_STR: String =
+    /// The version used to identify the node
+    pub static ref NODE_VERSION: String =
         format!(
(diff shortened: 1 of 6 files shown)


ant-protocol/src/version.rs
```diff
@@ -9,88 +9,97 @@
...
 
-    /// The req/response protocol version
-    pub static ref REQ_RESPONSE_VERSION_STR: String =
+    /// The req/response version
+    pub static ref REQ_RESPONSE_VERSION: String =
         format!(
-            "/safe/node/{}/{}",
-            get_truncate_version_str(),
-            get_key_version_str(),
+            "ant/node/{}",
```
maqi: shall be `/ant/node/{}`

maqi (approved):

jacderida:
This will be implemented in another PR.

PR #2508, opened 2024-12-09
feat!: use a simple network id to differentiate between network

RolandSherwin:


ant-protocol/src/version.rs
```diff
@@ -7,39 +7,83 @@
...
 lazy_static! {
+    /// The network_id is used to differentiate between different networks.
+    /// The default is set to 1 and it represents the mainnet.
+    pub static ref NETWORK_ID: RwLock<u8> = RwLock::new(1);
+
     /// The node version used during Identify Behaviour.
-    pub static ref IDENTIFY_NODE_VERSION_STR: String =
-        format!(
-            "safe/node/{}/{}",
+    pub static ref IDENTIFY_NODE_VERSION_STR: RwLock<String> =
```
maqi, line 18: Why need the extra `RwLock` ?
same to others

RolandSherwin: Because the `network_id` is now passed as a runtime argument, using `--network-id` and not an env var that we can fetch here..

maqi (approved):

PR #2511, opened 2024-12-09
Feat upgradable smart contracts and updated quotation flow

mickvandijke:
`autonomi` compiles!

Still some tests to fix most likely though and need to have the final smart contract implementation to fix local deployment on Anvil.

Look for `@anselme` and `// CODE REVIEW` for parts where I was a bit unsure!


ant-evm/src/data_payments.rs
```diff
@@ -77,6 +77,21 @@ impl ProofOfPayment {
             .any(|(_, quote)| quote.has_expired())
     }
 
+    /// Returns all quotes by given peer id
+    pub fn quotes_by_peer(&self, peer_id: &PeerId) -> Vec<&PaymentQuote> {
```
mickvandijke, line 81: Should this also check that the quote is valid for the same peer id?

grumbach: I think this is caller responsibility? The name doesn't suggest a check is there

autonomi/src/client/registers.rs
```diff
@@ -333,10 +341,11 @@ impl Client {
             expected_holders: Default::default(),
             is_register: true,
         };
+
         let put_cfg = PutRecordCfg {
             put_quorum: Quorum::All,
             retry_strategy: None,
-            use_put_record_to: Some(vec![payee]),
+            use_put_record_to: Some(payees), // CODE REVIEW: should we put to all or just one here?
```
grumbach, line 348: All sounds good. Maybe in the future, to strain the network less we could put to 1 by 1 until the data is stored

grumbach: This can come as an optimization

autonomi/src/client/utils.rs
```diff
@@ -146,19 +150,21 @@ impl Client {
         let put_cfg = PutRecordCfg {
             put_quorum: Quorum::One,
             retry_strategy: Some(RetryStrategy::Balanced),
-            use_put_record_to: Some(vec![storing_node]),
+            use_put_record_to: Some(storing_nodes), // CODE REVIEW: do we put to all payees or just one?
```
grumbach, line 153: Same here, this would be a good optimization down the line but maybe in the put_record code layer?

evmlib/src/contract/payment_vault/mod.rs
```diff
@@ -20,3 +21,38 @@ pub async fn get_market_price(
...
+            .map(|is_valid| PaymentVerification {
+                quote_hash,
+                amount_paid: Amount::from(1), // TODO: update placeholder amount when the smart contract changes
+                is_valid,
+            })?;
+
+        // CODE REVIEW: should we fail on a single invalid payment?
+        if !payment_verification.is_valid {
+            return Err(error::Error::PaymentInvalid);
+        }
```
grumbach, line 50: For now yeah, I guess this will be a question again when we batch this

PR #2514, opened 2024-12-10
feat(networking): add TransactionWithPayment

b-zee:


ant-node/src/put_validation.rs
```diff
@@ -186,6 +186,55 @@ impl Node {
...
+                        "Record's key {pretty_key:?} does not match with the value's TransactionAddress, ignoring PUT."
+                    );
+                    return Err(Error::RecordKeyMismatch);
+                }
+
+                let already_exists = self.validate_key_and_existence(&net_addr, &key).await?;
+
+                // The transaction may already exist during the replication.
+                // The payment shall get deposit to self even the transaction already presents.
+                // However, if the transaction already presents, the incoming one maybe for edit only.
```
maqi: The comment (originally for Register ?) shall be updated
As, unlike Register, Transaction is not allowed to be edited.

grumbach: If a new one comes in both should be kept!

grumbach: So no edits as in removal. But if there are multiple incoming for an address, BOTH/ALL copies should be kept

b-zee: Shall I just remove the comment?

maqi: change
```
However, if the transaction already presents, the incoming one maybe for edit only.
```
to
```
However, if the transaction already presents, the incoming one shall be appended with the existing one, if content is different.
```
shall be enough

maqi (approved):

PR #2517, opened 2024-12-10
Feat upgradable smart contracts and updated quotation flow

mickvandijke:
- Updates the payment vault interface and allows us to deploy the payment vault locally on Anvil.
- Fixes receipts only containing paid quotes instead of the 5 quotes needed for verification on the node side.
- Adds a (quick and dirty) rate limit to the contract's `getQuote` call.


evmlib/src/contract/payment_vault/mod.rs
```diff
@@ -33,24 +33,21 @@ pub async fn verify_data_payment(
...
-            })?;
+    let payment_verifications: Vec<_> = payment
+        .into_iter()
+        .map(interface::IPaymentVault::PaymentVerification::from)
+        .collect();
 
+    let payment_verification_results = payment_vault.verify_payment(payment_verifications).await?;
+
+    for payment_verification_result in payment_verification_results {
         // CODE REVIEW: should we fail on a single invalid payment?
```
maqi, line 44: I think we should.

maqi (approved):

PR #2520, opened 2024-12-11
chore: avoid access Distance private field of U256

maqi:
### Description

avoid access Distance private field of U256
also rolls back the libp2p dependency to the previous 0.54.1

### DIFF
--- a/ant-networking/Cargo.toml
+++ b/ant-networking/Cargo.toml
@@ -20,6 +20,7 @@ upnp = ["libp2p/upnp"]
 
 [dependencies]
 aes-gcm-siv = "0.11.1"
+alloy = { version = "0.5.3", default-features = false, features = ["std", "reqwest-rustls-tls", "provider-anvil-node", "sol-types", "json", "signers", "contract", "signer-local", "network"] }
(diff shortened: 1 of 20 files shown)


jacderida (approved):

ant-networking/Cargo.toml
```diff
@@ -20,6 +20,7 @@ upnp = ["libp2p/upnp"]
 
 [dependencies]
 aes-gcm-siv = "0.11.1"
+alloy = { version = "0.5.3", default-features = false, features = ["std", "reqwest-rustls-tls", "provider-anvil-node", "sol-types", "json", "signers", "contract", "signer-local", "network"] }
```
mickvandijke, line 23: Was this only added for the `U256` type? Cause the type is in `ant_evm::common::U256` already 🤔 

EDIT: Hmm actually, it is in evmlib::common::U256, but `ant_evm` doesn't re-export it yet. I think it would be a cleaner solution to re-export it through `ant_evm` than adding `Alloy` as dependency.

mickvandijke: ^ I'll do that in my merge commit for PR # 2519

PR #2526, opened 2024-12-12
feat: create wallet file from env sk

ermineJose:
### Description

create a wallet file with sk taken from the env.


ant-cli/src/wallet/fs.rs
```diff
@@ -137,7 +137,15 @@ pub(crate) fn select_wallet_address() -> Result<String, Error> {
         0 => {
```
mickvandijke: Do we only want to use the ENV wallet when we have 0 local wallets? Shouldn't we always default to the ENV wallet if it exists?

mickvandijke (approved):
LGTM!

PR #2529, opened 2024-12-13
fixes to address connection errors and rise on adding peers into RT & extra disk writes

maqi:
### Description
Contains following fixes:
* not to carry out connection pruning, this is no longer required due to the libp2p upgrades
* bootstrap node replacement only to be carried out once, as due to the switch to the bootstrap cache, workload is already distributed and could have peers `replacing each other`.

### DIFF
--- a/ant-networking/src/driver.rs
+++ b/ant-networking/src/driver.rs
@@ -759,7 +759,6 @@ impl NetworkBuilder {
             quotes_history: Default::default(),
             replication_targets: Default::default(),
             last_replication: None,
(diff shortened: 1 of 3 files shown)


maqi:
The `not to prune connection` fix has a bad result from staging run, due to the `open connections` to be kept rise to over 350, compared with previous 72.
Result in a 500% rise on CPU usage as well.
This is due to the libp2p and have to abandon the fix.

The other fix has been moved into PR https://github.com/maidsafe/autonomi/pull/2531 to be merged separately.

PR #2538, opened 2024-12-16
chore: make metrics logging to work with local networks only

RolandSherwin:
- This would mean that production builds will not log the system metrics to the log file.
- We keep it just for the tests running using `local` flag.


jacderida:
This was merged as part of #2539

PR #2546, opened 2024-12-17
Payment batch size decrease

mickvandijke:
Sets max transfers per transaction to `256`.


maqi (changes requested):
seems the entire `data_payments` folder can be removed ?

mickvandijke:
> seems the entire `data_payments` folder can be removed ?

Yes, we are not using it anymore

maqi:
seems the decreasement is no longer required, due to the smart_contract change.

PR #2566, opened 2024-12-23
chore: always specify the quoting metric explicitly

RolandSherwin:


ant-networking/src/cmd.rs
```diff
@@ -103,7 +103,7 @@ pub enum LocalSwarmCmd {
     /// Returns the quoting metrics and whether the record at `key` is already stored locally
     GetLocalQuotingMetrics {
         key: RecordKey,
-        sender: oneshot::Sender<(QuotingMetrics, bool)>,
+        sender: oneshot::Sender<Option<(QuotingMetrics, bool)>>,
```
grumbach: Why return an option here?

ant-networking/src/lib.rs
```diff
@@ -775,7 +775,7 @@ impl Network {
     pub async fn get_local_quoting_metrics(
         &self,
         key: RecordKey,
-    ) -> Result<(QuotingMetrics, bool)> {
+    ) -> Result<Option<(QuotingMetrics, bool)>> {
```
grumbach: What does None mean in this case? I think `Result<Option<_>>` is a confusing pattern

ant-networking/src/record_store_api.rs
```diff
@@ -117,13 +117,13 @@ impl UnifiedRecordStore {
         &self,
         key: &RecordKey,
         network_size: Option<u64>,
-    ) -> (QuotingMetrics, bool) {
+    ) -> Option<(QuotingMetrics, bool)> {
```
grumbach: Why do we use `Option` here?

ant-node/src/node.rs
```diff
@@ -618,6 +618,14 @@ impl Node {
...
                         }
                     }
+                    Ok(None) => {
+                        error!("Quoting metrics not found for {key:?}. This might be because we are using a ClientRecordStore??. This should not happen");
+                        QueryResponse::GetStoreQuote {
+                            quote: Err(ProtocolError::GetStoreQuoteFailed),
+                            peer_address: NetworkAddress::from_peer(self_id),
+                            storage_proofs,
+                        }
+                    }
```
grumbach: Would getting rid of this `Option` type solve the issue of the `"This should not happen"` error here?

grumbach (changes requested):
Nice tweaks, but I'm not sure about the `Option` addition in the quoting metrics

ant-networking/src/cmd.rs
RolandSherwin: Lemme just try to return an error instead of None type there. That makes more sense and it is cleaner

ant-networking/src/record_store_api.rs
```diff
@@ -119,51 +133,50 @@ impl UnifiedRecordStore {
...
-                warn!("Calling set_distance_range at Client. This should not happen");
+            Self::Client(_) => {
+                error!("Calling set_distance_range at Client. This should not happen");
             }
             Self::Node(store) => store.set_responsible_distance_range(distance),
         }
     }
 
-    pub(crate) fn get_farthest(&self) -> Option<RecordKey> {
+    pub(crate) fn get_farthest(&self) -> Result<Option<RecordKey>> {
```
grumbach, line 178: In what case do we have a `Ok(None)`?
Any chance we can simplify the return type to `Result<RecordKey>` or does the `Option` have any special meaning here?

ant-networking/src/record_store_api.rs
```diff
@@ -119,51 +133,50 @@ impl UnifiedRecordStore {
...
             Self::Client(_) => {
-                warn!("Calling payment_received at Client. This should not happen");
+                error!("Calling payment_received at Client. This should not happen");
             }
             Self::Node(store) => store.payment_received(),
         }
     }
 
-    pub(crate) fn get_farthest_replication_distance(&self) -> Option<U256> {
+    pub(crate) fn get_farthest_replication_distance(&self) -> Result<Option<U256>> {
```
grumbach, line 157: Same here

grumbach (approved):
LGTM, two questions about `Result<Option<...>>` return types

ant-networking/src/record_store_api.rs
RolandSherwin: It is None when the record store is empty / when the farthest record is not calculated. It is from this field.
```
    /// Farthest record to self
    farthest_record: Option<(Key, Distance)>,
```

PR #2574, opened 2024-12-24
Data further refactor

dirvine:
wip - this PR is on top of the data_refactor PR


ant-bootstrap/tests/cli_integration_tests.rs
```diff
@@ -176,3 +182,45 @@ async fn test_test_network_peers() -> Result<(), Box<dyn std::error::Error>> {
...
+    let cache_path = temp_dir.path().join("cache.txt");
+
+    let peer_addr = format!(
+        "/ip4/{}/udp/8080/quic-v1/p2p/12D3KooWRBhwfeP2Y4TCx1SM6s9rUoHhR5STiGwxBhgFRcw3UERE",
+        LOCAL_IP
+    );
+
+    // ... rest of the test ...
+    Ok(())
+}
```
grumbach: Agreed that this would be a nice feature! Maybe add a TODO comment in the tests here so we don't forget them as is?

ant-evm/src/amount.rs
```diff
@@ -67,7 +67,7 @@ impl AttoTokens {
 
     /// Converts the Nanos into bytes
     pub fn to_bytes(&self) -> Vec<u8> {
```
grumbach: Here the comment still refers to Nanos

ant-evm/src/amount.rs
```diff
@@ -67,7 +67,7 @@ impl AttoTokens {
...
-        self.0.as_le_bytes().to_vec()
+        self.0.to_be_bytes::<32>().to_vec()
```
grumbach, line 70: Any specific reasons to move to big endian and 32 here?

ant-networking/Cargo.toml
```diff
@@ -58,7 +59,6 @@ prometheus-client = { version = "0.22", optional = true }
 rand = { version = "~0.8.5", features = ["small_rng"] }
 rayon = "1.8.0"
 rmp-serde = "1.1.1"
-self_encryption = "~0.30.0"
```
grumbach: Glad we were able to trim dependencies here!

ant-networking/src/driver.rs
```diff
@@ -300,6 +300,11 @@ impl NetworkBuilder {
         }
     }
 
+    pub fn local(mut self, local: bool) -> Self {
+        self.local = local;
+        self
+    }
+
```
grumbach, line 307: This is a GREAT pattern!

ant-evm/src/amount.rs
dirvine: Really network type endianness (most significant bit first) The 32 states we want exactly 32bytes (256 bits)
Otherwise we risk returning a variable length vec, well we would if using different types at least. So here we force the 32bytes. 

In short we guarantee a 32byte network endian number here really.

ant-evm/src/amount.rs
dirvine: Yea I missed that one. This file was failing tests mind you, that's the only reason I ended up making changes to it.

autonomi/README_WASM.md
```diff
@@ -1,95 +0,0 @@
-# Autonomi JS API
-
-Note: the JS API is experimental and will be subject to change.
```
grumbach, line 3: Are we sure we want to delete these WASM docs now?

dirvine: yes we gave up on wasm for now. I hope we introduce it at a later stage, but we need to ensure we do not hamper network or disk access when we do

autonomi/src/client/data/public.rs
```diff
@@ -44,100 +49,72 @@ impl Client {
...
-
-            let summary = UploadSummary {
-                record_count,
-                tokens_spent,
-            };
-            if let Err(err) = channel.send(ClientEvent::UploadComplete(summary)).await {
-                error!("Failed to send client event: {err:?}");
+        match &self.mode {
+            ClientMode::ReadWrite(_) => {
+                let now = ant_networking::target_arch::Instant::now();
```
grumbach, line 46: Why introduce a mode here?

autonomi/src/client/error.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Errors that can occur during data storage operations
+#[derive(Debug, thiserror::Error)]
+pub enum PutError {
+    /// No wallet available for write operations
+    #[error("Write operations require a wallet. Use upgrade_to_read_write to add a wallet.")]
+    NoWallet,
+    /// Network-related error
+    #[error("Network error: {0}")]
+    Network(#[from] NetworkError),
```
grumbach, line 22: There was a discussion about putting errors were they happen before, instead of all in a single file were they become un-linked to their code. 
What's your take on this?

autonomi/src/client/mod.rs
```diff
@@ -52,33 +60,70 @@ const CLIENT_EVENT_CHANNEL_SIZE: usize = 100;
...
-/// let client = Client::init().await?;
-/// # Ok(())
-/// # }
-/// ```
 #[derive(Clone)]
 pub struct Client {
     pub(crate) network: Network,
     pub(crate) client_event_sender: Arc<Option<mpsc::Sender<ClientEvent>>>,
     pub(crate) evm_network: EvmNetwork,
+    pub(crate) mode: ClientMode,
```
grumbach, line 119: What's the reason behind mode here?

dirvine: The mode is for local network mode (mens) or not.

autonomi/src/client/pointer.rs
```diff
@@ -35,20 +39,17 @@ pub enum PointerError {
...
-        
+
         match record {
             Some(record) => {
                 let (_, pointer): (Vec<u8>, Pointer) = rmp_serde::from_slice(&record.value)
                     .map_err(|_| PointerError::Serialization)?;
                 Ok(pointer)
             }
-            None => Err(PointerError::Corrupt),
+            None => Err(PointerError::FailedVerification),
```
grumbach, line 45: Does `FailedVerification` cover all possible errors behind `get_local_record`?
I think having at least a String with some information on the actual underlying error would be extremely valuable

autonomi/src/client/utils.rs
```diff
@@ -17,82 +19,71 @@ use bytes::Bytes;
...
+            GetError::Decryption(crate::self_encryption::Error::SelfEncryption(e.into()))
+        })?;
+        let temp_path = temp_file.path().to_owned();
+
+        // Create a closure to fetch chunks
+        let client = self.clone();
+        let get_chunks = move |xor_names: &[XorName]| -> Result<Vec<Bytes>, SelfEncryptionError> {
+            let mut chunks = Vec::with_capacity(xor_names.len());
+            for xor_name in xor_names {
+                match futures::executor::block_on(client.chunk_get(*xor_name)) {
```
grumbach, line 45: Why not just await here instead of `block_on`?

autonomi/src/network/local.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+impl LocalNode {
+    pub async fn start() -> Result<Self> {
+        Self::start_internal(true).await
+    }
+
+    pub async fn start_secondary() -> Result<Self> {
+        Self::start_internal(false).await
+    }
```
grumbach, line 23: I'm not sure what `start` and `start_secondary` involves here, could this be precised?

autonomi/src/client/data/public.rs
dirvine: The idea is we have a read only client (no wallet etc.) and a full client that can put so needs a wallet

autonomi/tests/register.rs
```diff
@@ -22,7 +22,7 @@ use tokio::time::sleep;
 async fn register() -> Result<()> {
     let _log_appender_guard = LogBuilder::init_single_threaded_tokio_test("register", false);
 
-    let client = Client::init_local().await?;
+    let client = Client::init_local(true).await?;
```
grumbach, line 25: It's not 100% clear what the `true` here means, if an option on/off is needed maybe another constructor or the following pattern:
```
Client::init_local().with_option_AAA().await?;
```

grumbach (commented):
Lots of great additions! Having docs in here is a real game changer!
Couple of notes and questions but overall excellent work!

dirvine:
closing as it's too big and the local flag is leading to confusion here

PR #2577, opened 2024-12-29
fix: use smaller timeouts for circuit relay connections

jacderida:
This was originally PR #2568.

It was prematurely merged into an RC branch and that RC has since been abandoned.

We can release it in a new RC if it gets through a comparison.


jacderida:
In an environment comparison this code was shown to have unacceptable increases in resource usage.

PR #2587, opened 2024-12-31
feat: rename linked list to graph

grumbach:


dirvine (approved):
Nice and simple PR for review, agree with the name change. I was unsure of Entry, but it is an entry really as the whole thing is a graph and we don't hold that as a type actually as it's never ending :-) 

Nice one though

PR #2589, opened 2025-01-01
fix(launchpad): use default path for nodes data dir when root(unix)

harismuzaffer:
### Description

when launchpad is run as root(on unix) and base directory is same as primary mount point, use default path instead. Otherwise currently default is to use user's data directory which in case of root, is not accessible to user ant(against which antnode binary runs)

Here is the github issue: https://github.com/maidsafe/autonomi/issues/2584


RolandSherwin (approved):
Thanks for the PR! Looks good to me 😄

harismuzaffer:
> Thanks for the PR! Looks good to me 😄

Thanks alot!

PR #2590, opened 2025-01-02
chore(node): more logs when carry out payment verification

maqi:
### Description

more logs when carry out payment verification


grumbach (approved):
LGTM

PR #2591, opened 2025-01-02
fix(record-store): ensure record sync before marking as stored

dirvine:
# Record Store Sync Improvements

## Problem
The `can_store_after_restart` test was failing because records were not being properly synced to disk before being marked as stored. This could lead to potential data loss in case of system crashes or unexpected restarts.

## Solution
- Added proper file sync operation after writing record to disk using `sync_all()`
- Combined file write and sync operations into a single `Result` chain using `and_then`
- Improved error handling for both write and sync operations
- Only send `AddLocalRecordAsStored` command after both operations succeed

## Testing
- All tests now pass, including the previously failing `can_store_after_restart` test
- Added proper error handling and logging for write/sync failures
- Verified that records are properly persisted after system restarts

## Impact
This change improves data reliability by ensuring that records are properly synced to disk before being marked as stored. This prevents potential data loss in scenarios where the system crashes or restarts unexpectedly after a write but before the data is fully synced to disk.

### DIFF
--- a/ant-networking/src/record_store.rs
+++ b/ant-networking/src/record_store.rs
@@ -687,17 +687,18 @@ impl NodeRecordStore {
         spawn(async move {
             let key = r.key.clone();
             if let Some(bytes) = Self::prepare_record_bytes(r, encryption_details) {
-                let cmd = match fs::write(&file_path, bytes) {
+                let cmd = match fs::write(&file_path, bytes)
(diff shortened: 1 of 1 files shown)


grumbach (approved):
LGTM

maqi:
The `can_store_after_restart` test is being executed stand alone and pass.

PR #2593, opened 2025-01-03
feat: remove feat flags

grumbach:


dirvine (approved):
LTM

mickvandijke:
Interesting. The test that fails here seems to have been enabled by this PR. Apparently it wasn't somehow executing before because of the feat flags used/unused in the workflow?

Anyway. The issue here is that two tests run in parallel and both post transactions to the same EVM network at the same time. The ideal fix is to run an EVM network for every test, as tests shouldn't indirectly influence each other's results. However as a quick fix, I will just make a PR that forces these tests to run in serie instead of in parallel.

PR #2594, opened 2025-01-03
chore: publish docs on merge

jacderida:
We don't need to publish the documentation as part of the PR; it can be done after we merge.

The `GH_TOKEN` variable is also set to a personal access token that should have permission to push to the repository.

There is a small change in the documentation that removes a typo. It was used to test the process.


grumbach (approved):
Great!

PR #2595, opened 2025-01-03
WIP: local as runtime opt

grumbach:


ant-networking/src/driver.rs
```diff
@@ -254,7 +253,7 @@ pub(super) struct NodeBehaviour {
     pub(super) blocklist:
         libp2p::allow_block_list::Behaviour<libp2p::allow_block_list::BlockedPeers>,
     pub(super) identify: libp2p::identify::Behaviour,
-    #[cfg(feature = "local")]
+    /// mDNS behaviour to use in local mode
```
b-zee, line 256: Should we not make this a `Toggle`? Then disable it completely with non-local mode to prevent home network noise.

grumbach: If you look at the implementation I check if we are in local mode before executing anything. But not sure if it does anything else? Not too familiar with how Behaviors work.

b-zee: If there is no `Toggle` mDNS behavior will be activated, thus broadcasting the address on the home network.

b-zee: So if a public node is running on a home network it will still broadcast its address to a local test network's nodes. Shouldn't be a problem because of different network IDs/identifiers (plus the fact, like you mentioned, that you don't handle the mDNS events), but would still add a lot of noisy logs for those local nodes I presume.

ant-networking/src/driver.rs
```diff
@@ -684,7 +681,6 @@ impl NetworkBuilder {
             request_response,
             kademlia,
             identify,
-            #[cfg(feature = "local")]
             mdns,
```
b-zee, line 690: This is where the mDNS behavior is executed both with and withouth the `--local` flag.

ant-networking/src/event/swarm.rs
```diff
@@ -287,26 +286,28 @@ impl SwarmDriver {
...
-
-                                info!(%addr, "mDNS node discovered and dialing");
-
-                                if let Err(err) = self.dial(addr.clone()) {
-                                    warn!(%addr, "mDNS node dial error: {err:?}");
+                // mDNS is only relevant in local mode
+                if self.local {
+                    match *mdns_event {
+                        mdns::Event::Discovered(list) => {
+                            if self.local {
```
b-zee: This line is now duplicate because of the above added line.

ant-networking/src/driver.rs
grumbach: Mmmh I see, that's not ideal. @dirvine mentioned getting rid of mdns all together eventually, maybe it's not worth the effort now if we plan to remove it altogether?

ant-networking/src/event/swarm.rs
grumbach: Oh indeed!

ant-networking/src/driver.rs
b-zee: Wouldn't mind getting it out, but `Toggle` is fine with me too. You can use the UPnP behavior as an example which is a bunch of lines below mDNS in the `build` method.

b-zee (approved):

PR #2602, opened 2025-01-06
fix: shall only have Chunk and NonChunk for RecordType

maqi:
### Description

RecordType is only used within node instance to help with identify record content having different versions.
Hence only need to have entries of `Chunk` or `NonChunk`.


grumbach (approved):
LGTM

PR #2610, opened 2025-01-07
feat(ant-cli): report already paid for records

b-zee:


autonomi/src/client/data/public.rs
```diff
@@ -87,7 +87,7 @@ impl Client {
             return Err(last_chunk_fail.1);
         }
 
-        let record_count = chunks.len() + 1;
+        let record_count = (chunks.len() + 1) - skipped_payments;
```
grumbach, line 90: Im not sure record_count should be changed. Or it should be renamed to record_paid

autonomi/src/client/graph.rs
```diff
@@ -121,7 +121,8 @@ impl Client {
         // send client event
         if let Some(channel) = self.client_event_sender.as_ref() {
             let summary = UploadSummary {
-                record_count: 1,
+                record_count: 1usize.saturating_sub(skipped_payments),
```
grumbach: Same here

grumbach (approved):
Looks good, with a small comment on record_count

autonomi/src/client/data/public.rs
b-zee: The file uploader has [this](
https://github.com/b-zee/safe_network/blob/c995d16b235f4eae482b62d154223738172d5399/ant-cli/src/commands/file.rs#L79) piece of code:
```rust
    if summary.record_count == 0 {
        println!("All chunks already exist on the network.");
    } else {
```

I will rename to `record_paid`, good name suggestion!

PR #2611, opened 2025-01-07
chore: refactor RecordKind, creating DataTypes

maqi:
### Description

refactor RecordKind, which is now based on a new enum of DataTypes.
DataTypes becomes the sole place to reflects all data types that natively supported by the autonomi network.


grumbach (approved):
Nice one!

PR #2613, opened 2025-01-09
feat(client): try get evm network from env by default on client::init…

mickvandijke:
Adjusts the default for `ClientConfig` to first try getting the EVM network from ENV before falling back to the default network option:

```rust
impl Default for ClientConfig {
    fn default() -> Self {
        Self {
            ...
            evm_network: EvmNetwork::try_from_env().unwrap_or_default(),
        }
    }
}
```
This makes it easier to create a client without manually having to specify the EVM network. This also fixes a problem in our tests where `Client::init_local` would select the default EVM network instead of the one specified in the ENV: "EVM_NETWORK=local".


grumbach (approved):
LGTM

PR #2620, opened 2025-01-10
feat: fix docs and bad namings in API

grumbach:


ant-protocol/src/storage/graph.rs
```diff
@@ -22,7 +22,7 @@ pub struct GraphEntry {
     pub owner: PublicKey,
     pub parents: Vec<PublicKey>,
     pub content: GraphContent,
-    pub outputs: Option<Vec<(PublicKey, GraphContent)>>,
+    pub outputs: Vec<(PublicKey, GraphContent)>,
     /// signs the above 4 fields with the owners key
```
dirvine, line 26: The reason for option here is that some patterns we do not have outputs, only Parent. i.e. when used as a file change history the GraphEntry can only go backwards. So the content is a direct or whatever and changes with each entry. There are no outputs in that pattern though.

autonomi/nodejs/dist/linkedList.d.ts
```diff
@@ -1,9 +1,9 @@
-import { LinkedListOptions, PaymentOption } from './types';
```
dirvine, line 1: nice catch

dirvine (commented):
Nice @grumbach The docs were old so good to see those changes there too. I left a couple of comments on the GraphEntry type outputs. Looks good to me apart from those comments though

autonomi/nodejs/dist/linkedList.d.ts
dirvine: we should also rename this file

autonomi/nodejs/dist/linkedList.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.LinkedList = void 0;
-class LinkedList {
+exports.GarphEntry = void 0;
```
b-zee: Typo

autonomi/nodejs/dist/linkedList.js
```diff
@@ -22,4 +22,4 @@ class LinkedList {
         throw new Error('Not implemented');
     }
 }
-exports.LinkedList = LinkedList;
+exports.GarphEntry = GarphEntry;
```
b-zee: Typo

autonomi/nodejs/dist/linkedList.js
b-zee: Oh I see it in more places. Better just replace all `garph`.

ant-protocol/src/storage/graph.rs
b-zee: An empty vec can be considered as a pattern. But an `Option` makes it more obvious, though might be a little more complex handling in code.

grumbach: Empty vec indeed sounds like the right balance between usability and simplicity, what do you think?

b-zee (approved):

PR #2624, opened 2025-01-13
fix(metrics): avoid intialize un-necessary system resources

maqi:
### Description

It turned out that new_all function opens FD for un-necessary system resources, which result in accumulated memory-usage issue.
Hence initialize with nothing and then refresh only related.

This PR also contains the CI test to confirm there is no extra and acculuated FDs to be held by the node process.

### Type of Change

### DIFF
--- a/.github/workflows/memcheck.yml
+++ b/.github/workflows/memcheck.yml
@@ -164,6 +164,22 @@ jobs:
         run: pgrep antnode | wc -l
         if: always()
 
+      - name: confirm opened FDs
+        shell: bash
+        timeout-minutes: 1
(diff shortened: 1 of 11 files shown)


rid-dim:
just a note; most certainly you already know ... but cpu usage is significantly increased (in intervals) when applying this change and especially when running many nodes (~200) this leads to high load-averages because a lot of operations seem to pile up in intervals

oh - stupid me ... is the force-push there the successor of the first version? :-O

maqi:
In favour of PR https://github.com/maidsafe/autonomi/pull/2639

PR #2627, opened 2025-01-14
feat(node): add easy way to launch node

b-zee:


b-zee:
Closed in favor of #2642

PR #2628, opened 2025-01-14
fix(network): consider MultiAddrNotSupported as a serioud issu..

RolandSherwin:
- Libp2p throws Outgoing conn errors for all the listen addr (including private) of the remote peer even though we try to dial just the global/public addr. This would mean that we get `MultiaddrNotSupported` error for the private addr of the peer every single time.
- Thus, just a single `MultiaddrNotSupported` error here should not be a huge concern and we should not remove the peer based on this.
- We now remove the peer ONLY IF ALL the addrs have `MultiaddrNotSupported`

### DIFF
--- a/ant-networking/src/event/swarm.rs
+++ b/ant-networking/src/event/swarm.rs
@@ -434,21 +434,27 @@ impl SwarmDriver {
                         // unless there are _specific_ errors (connection refused eg)
                         error!("Dial errors len : {:?}", errors.len());
                         let mut there_is_a_serious_issue = false;
+                        // Libp2p throws errors for all the listen addr (including private) of the remote peer even
+                        // though we try to dial just the global/public addr. This would mean that we get
+                        // MultiaddrNotSupported error for the private addr of the peer.
+                        //
+                        // Just a single MultiaddrNotSupported error is not a critical issue, but if all the listen
+                        // addrs of the peer are private, then it is a critical issue.
+                        let mut all_multiaddr_not_supported = true;
                         for (_addr, err) in errors {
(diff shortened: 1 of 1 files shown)


ant-networking/src/event/swarm.rs
```diff
@@ -434,21 +434,27 @@ impl SwarmDriver {
...
-                                    warn!("Multiaddr not supported : {addr:?}");
+                                    warn!("OutgoingConnectionError: Transport::MultiaddrNotSupported {addr:?}. This can be ignored if the peer has atleast one global address.");
                                     #[cfg(feature = "loud")]
                                     {
-                                        println!("Multiaddr not supported : {addr:?}");
+                                        warn!("OutgoingConnectionError: Transport::MultiaddrNotSupported {addr:?}. This can be ignored if the peer has atleast one global address.");
                                         println!("If this was your bootstrap peer, restart your node with a supported multiaddr");
                                     }
-                                    // if we can't dial a peer on a given address, we should remove it from the routing table
-                                    there_is_a_serious_issue = true
```
maqi, line 449: as the default value of `all_multiaddr_not_supported` is to be true, here actually doing the same? then what's the point to  change ?

RolandSherwin: We set it to `all_multiaddr_not_supported = false` below. 
Only if the entire `Vec<Error>` is `MultiaddrNotSupported`, we really go after this error, else this error is harmless.

maqi (approved):

b-zee (approved):

PR #2634, opened 2025-01-15
feat(network): introduce relay reservation score

RolandSherwin:
This PR will now also merge #2617, #2618 and #2628 from the hotfix branch that was not released.

### DIFF
--- a/ant-networking/src/event/identify.rs
+++ b/ant-networking/src/event/identify.rs
@@ -6,180 +6,27 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::collections::HashSet;
-use std::time::Instant;
-
+use crate::relay_manager::is_a_relayed_peer;
+use crate::{multiaddr_is_global, multiaddr_strip_p2p, NetworkEvent, SwarmDriver};
 use ant_protocol::version::{IDENTIFY_NODE_VERSION_STR, IDENTIFY_PROTOCOL_STR};
+use libp2p::identify::Info;
 use libp2p::kad::K_VALUE;
 use libp2p::multiaddr::Protocol;
 use libp2p::swarm::dial_opts::{DialOpts, PeerCondition};
 use libp2p::Multiaddr;
-
-use crate::error::Result;
-use crate::relay_manager::is_a_relayed_peer;
-use crate::{multiaddr_is_global, multiaddr_strip_p2p, NetworkEvent, SwarmDriver};
+use std::collections::HashSet;
+use std::time::Instant;
 
 impl SwarmDriver {
-    pub(super) fn handle_identify_event(
-        &mut self,
-        identify_event: libp2p::identify::Event,
-    ) -> Result<()> {
-        let start = Instant::now();
-
+    pub(super) fn handle_identify_event(&mut self, identify_event: libp2p::identify::Event) {
         match identify_event {
             libp2p::identify::Event::Received {
                 peer_id,
                 info,
                 connection_id,
             } => {
-                debug!(conn_id=%connection_id, %peer_id, ?info, "identify: received info");
-
-                let our_identify_protocol = IDENTIFY_PROTOCOL_STR.read().expect("IDENTIFY_PROTOCOL_STR has been locked to write. A call to set_network_id performed. This should not happen.").to_string();
-
-                if info.protocol_version != our_identify_protocol {
-                    warn!(?info.protocol_version, "identify: {peer_id:?} does not have the same protocol. Our IDENTIFY_PROTOCOL_STR: {our_identify_protocol:?}");
-
-                    self.send_event(NetworkEvent::PeerWithUnsupportedProtocol {
-                        our_protocol: our_identify_protocol,
-                        their_protocol: info.protocol_version,
-                    });
-                    // Block the peer from any further communication.
-                    self.swarm.behaviour_mut().blocklist.block_peer(peer_id);
-                    if let Some(dead_peer) =
-                        self.swarm.behaviour_mut().kademlia.remove_peer(&peer_id)
-                    {
-                        error!("Clearing out a protocol mistmatch peer from RT. Something went wrong, we should not have added this peer to RT: {peer_id:?}");
-                        self.update_on_peer_removal(*dead_peer.node.key.preimage());
-                    }
-
-                    return Ok(());
-                }
-
-                let our_agent_version = IDENTIFY_NODE_VERSION_STR.read().expect("IDENTIFY_NODE_VERSION_STR has been locked to write. A call to set_network_id performed. This should not happen.").to_string();
-                // if client, return.
-                if info.agent_version != our_agent_version {
-                    return Ok(());
-                }
-
-                let has_dialed = self.dialed_peers.contains(&peer_id);
-
-                // If we're not in local mode, only add globally reachable addresses.
-                // Strip the `/p2p/...` part of the multiaddresses.
-                // Collect into a HashSet directly to avoid multiple allocations and handle deduplication.
-                let mut addrs: HashSet<Multiaddr> = match self.local {
-                    true => info
-                        .listen_addrs
-                        .into_iter()
-                        .map(|addr| multiaddr_strip_p2p(&addr))
-                        .collect(),
-                    false => info
-                        .listen_addrs
-                        .into_iter()
-                        .filter(multiaddr_is_global)
-                        .map(|addr| multiaddr_strip_p2p(&addr))
-                        .collect(),
-                };
-
-                let has_relayed = is_a_relayed_peer(&addrs);
-
-                let is_bootstrap_peer = self
-                    .bootstrap_peers
-                    .iter()
-                    .any(|(_ilog2, peers)| peers.contains(&peer_id));
-
-                // Do not use an `already relayed` peer as `potential relay candidate`.
-                if !has_relayed && !is_bootstrap_peer {
-                    if let Some(relay_manager) = self.relay_manager.as_mut() {
-                        debug!(
-                            "Adding candidate relay server {peer_id:?}, it's not a bootstrap node"
-                        );
-                        relay_manager.add_potential_candidates(&peer_id, &addrs, &info.protocols);
-                    }
-                }
-
-                // When received an identify from un-dialed peer, try to dial it
-                // The dial shall trigger the same identify to be sent again and confirm
-                // peer is external accessible, hence safe to be added into RT.
-                if !self.local && !has_dialed {
-                    // Only need to dial back for not fulfilled kbucket
-                    let (kbucket_full, already_present_in_rt, ilog2) = if let Some(kbucket) =
-                        self.swarm.behaviour_mut().kademlia.kbucket(peer_id)
-                    {
-                        let ilog2 = kbucket.range().0.ilog2();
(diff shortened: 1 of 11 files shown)


ant-networking/src/metrics/mod.rs
```diff
@@ -289,12 +289,6 @@ impl NetworkMetricsRecorder {
                 });
             }
             Marker::QuotingMetrics { quoting_metrics } => {
-                let _ = self.relevant_records.set(
-                    quoting_metrics
-                        .close_records_stored
-                        .try_into()
-                        .unwrap_or(i64::MAX),
-                );
```
RolandSherwin, line 297: Set twice.

ant-networking/src/relay_manager.rs
```diff
@@ -9,47 +9,105 @@
...
+use std::{
+    collections::{btree_map::Entry, BTreeMap, HashMap, HashSet, VecDeque},
+    time::SystemTime,
+};
 
 const MAX_CONCURRENT_RELAY_CONNECTIONS: usize = 4;
 const MAX_POTENTIAL_CANDIDATES: usize = 1000;
 
+/// We could get multiple incoming connections from the same peer through multiple relay servers, and only one of them
+/// would succeed. So we wait and collect all such connections peer 'from' peer, instead of just recording the
```
maqi: > So we wait and collect all such connections peer 'from' peer, 

got an extra `peer` ?

ant-networking/src/relay_manager.rs
```diff
@@ -9,47 +9,105 @@
...
+///  success/failure for eachconnection.
```
maqi: eachconnection -> each connection

ant-networking/src/relay_manager.rs
```diff
@@ -9,47 +9,105 @@
...
+    /// relay.
+    reservation_health: RelayReservationHealth,
+}
+
+#[derive(Debug, Default)]
+struct RelayReservationHealth {
+    /// We could have multiple incoming connections from the same peer through multiple relay servers. But we could
+    /// just have a single one as well.
+    /// This is a map of the 'from peer' to the multiple relay servers that the incoming connections are coming through.
+    incoming_connections_from_remote_peer: BTreeMap<PeerId, ConnectionsFromPeer>,
```
maqi, line 74: `ConnectionsFromPeer` is defined as `(PeerId, ConnectionId, SystemTime, Option<bool>)`, which has `PeerId` by itself
it seems it shall be defined as `<ConnectionId, SystemTime, Option<bool>)` instead ?

ant-networking/src/relay_manager.rs
```diff
@@ -9,47 +9,105 @@
...
 const MAX_POTENTIAL_CANDIDATES: usize = 1000;
 
+/// We could get multiple incoming connections from the same peer through multiple relay servers, and only one of them
+/// would succeed. So we wait and collect all such connections peer 'from' peer, instead of just recording the
+///  success/failure for eachconnection.
+const MAX_DURATION_TO_TRACK_INCOMING_CONNECTIONS_PER_PEER: std::time::Duration =
+    std::time::Duration::from_secs(20);
+
+/// The connections from a single peer through multiple relay servers.
+type ConnectionsFromPeer = Vec<(PeerId, ConnectionId, SystemTime, Option<bool>)>;
```
maqi, line 42: better have comments to explain which each field stands for or purpose to be used.
for example, what's the bool flag for and what if it is None ?

ant-networking/src/relay_manager.rs
```diff
@@ -9,47 +9,105 @@
...
+struct RelayReservationHealth {
+    /// We could have multiple incoming connections from the same peer through multiple relay servers. But we could
+    /// just have a single one as well.
+    /// This is a map of the 'from peer' to the multiple relay servers that the incoming connections are coming through.
+    incoming_connections_from_remote_peer: BTreeMap<PeerId, ConnectionsFromPeer>,
+    reservation_score: BTreeMap<PeerId, ReservationStat>,
+}
+
+#[derive(Debug, Default, Clone)]
+struct ReservationStat {
```
maqi, line 83: shall this to be the score of the `recent X number attempts` , so that it can reflect the realtime of the health ?

ant-networking/src/relay_manager.rs
```diff
@@ -239,3 +378,227 @@ impl RelayManager {
...
+        for from_peer in to_remove {
+            debug!("Removing {from_peer:?} from the incoming_connections_from_remote_peer");
+            self.incoming_connections_from_remote_peer
+                .remove(&from_peer);
+        }
+
+        self.log_reservation_score();
+    }
+
+    /// Clean up the stats for relay servers that we are no longer connected to.
```
maqi: will this make us `re-connect` to a `faulty-server` and experience another round of tracking then drop ?

ant-networking/src/relay_manager.rs
```diff
@@ -239,3 +378,227 @@ impl RelayManager {
...
+                debug!("Could not obtain elapsed time.");
+                continue;
+            };
+
+            if elapsed < MAX_DURATION_TO_TRACK_INCOMING_CONNECTIONS_PER_PEER {
+                debug!("There is still an active incoming connection from {from_peer:?} that is waiting to be established. Skipping.");
+                continue;
+            }
+
+            // if atleast one connection has been established, we can update the stats.
```
maqi: atleast -> at least

ant-networking/src/relay_manager.rs
```diff
@@ -239,3 +378,227 @@ impl RelayManager {
...
+                debug!("The incoming connections from {from_peer:?} are empty. Skipping.");
+                continue;
+            };
+
+            let Ok(elapsed) = SystemTime::now().duration_since(*latest_time) else {
+                debug!("Could not obtain elapsed time.");
+                continue;
+            };
+
+            if elapsed < MAX_DURATION_TO_TRACK_INCOMING_CONNECTIONS_PER_PEER {
```
maqi, line 491: not quite sure, MAX_DURATION_TO_TRACK_INCOMING_CONNECTIONS_PER_PEER is a fixed duration.
However, according to the log msg, `waiting to be established` is an event driven status change, not `time-driven`
Don't get why `not timed out` means it is still waiting to be established?

ant-networking/src/relay_manager.rs
```diff
@@ -239,3 +378,227 @@ impl RelayManager {
...
+                        }
+                    }
+                }
+            }
+
+            to_remove.push(*from_peer);
+        }
+
+        for from_peer in to_remove {
+            debug!("Removing {from_peer:?} from the incoming_connections_from_remote_peer");
```
maqi: `incoming_connections_from_remote_peer` only got pruned when got `updated within above block of code`
what if entries remained in the container due to some bug/error, which makes it growing infinitely ?

ant-networking/src/relay_manager.rs
RolandSherwin: It is the `relay server peer`. I'll update the docs as suggested below

ant-networking/src/relay_manager.rs
RolandSherwin: Yeah, we could add a block list here.

ant-networking/src/relay_manager.rs
RolandSherwin: A random peer could make `multiple` incoming connections through multiple of our relays, so we can't just act on the events, instead we have to give ourselves some time to collect all these incoming connections.

Sometimes we could get 1 incoming connection, sometimes 4, we can't be sure.

ant-networking/src/relay_manager.rs
RolandSherwin: The only scenario where this can happen is  from the below code
```
 let Some(latest_time) = connections.iter().map(|(_, _, time, _)| time).max() else {
                debug!("The incoming connections from {from_peer:?} are empty. Skipping.");
                continue;
            };

            let Ok(elapsed) = SystemTime::now().duration_since(*latest_time) else {
                debug!("Could not obtain elapsed time.");
                continue;
            };
```

This can only happen if there is a bug in our code. The cleanup does not depend on the events, it depends on just the `MAX_DURATION_TO_TRACK_INCOMING_CONNECTIONS_PER_PEER`.

Just depending on a timer is fine because a connection cannot take more than 15s before a `ConnectionEstablished` or a `IncomingConnectionError` is emitted tbh.

maqi: ok. fine then.

just `This can only happen if there is a bug in our code`  could happen

ant-networking/src/event/identify.rs
```diff
@@ -6,187 +6,236 @@
...
-                    if let Some(relay_manager) = self.relay_manager.as_mut() {
-                        debug!(
-                            "Adding candidate relay server {peer_id:?}, it's not a bootstrap node"
-                        );
-                        relay_manager.add_potential_candidates(&peer_id, &addrs, &info.protocols);
-                    }
-                }
+                    .any(|entry| entry.node.key.preimage() == &peer_id);
+
+                // // If the bucket contains any of a bootstrap node,
```
maqi: this commented out block can now be removed?

ant-networking/src/event/identify.rs
```diff
@@ -6,187 +6,236 @@
...
+
+        // When received an identify from un-dialed peer, try to dial it
+        // The dial shall trigger the same identify to be sent again and confirm
+        // peer is external accessible, hence safe to be added into RT.
+        if !self.local && !has_dialed {
+            // Only need to dial back for not fulfilled kbucket
+            if kbucket_full {
+                debug!("received identify for a full bucket {ilog2:?}, not dialing {peer_id:?} on {addrs:?}");
+                return;
+            } else if already_present_in_rt {
```
maqi: this `already_present_in_rt` has been checked above

ant-networking/src/event/identify.rs
```diff
@@ -6,187 +6,236 @@
...
+                    .addresses(addrs.iter().cloned().collect())
+                    .build(),
+            ) {
+                warn!(%peer_id, ?addrs, "dialing error: {err:?}");
+            }
+
+            return;
+        }
+
+        // If we are not local, we care only for peers that we dialed and thus are reachable.
```
maqi: comment doesn't match the condition check (i.e. `self.local || has_dialed` ) ?
also `self.local || has_dialed` is just the `else` of `!self.local && !has_dialed` , why make it separate `if clause` ?

ant-networking/src/event/identify.rs
```diff
@@ -6,187 +6,236 @@
...
+            if let Err(err) = self.swarm.dial(
+                DialOpts::peer_id(peer_id)
+                    .condition(PeerCondition::NotDialing)
+                    .addresses(addrs.iter().cloned().collect())
+                    .build(),
+            ) {
+                warn!(%peer_id, ?addrs, "dialing error: {err:?}");
+            }
+
+            return;
```
maqi: may not be required

ant-networking/src/event/identify.rs
RolandSherwin: I haven't modified anything there. The comment has to be updated there. I think we should add the peers directly to the RT if local.

RolandSherwin: Also added an else there.

RolandSherwin:
superseded by https://github.com/maidsafe/autonomi/pull/2691

PR #2638, opened 2025-01-15
feat(pricing): make QuotingMetrics supports data_type and data_size pricing

maqi:
### Description

Future support for the pricing on data_type and data_size


grumbach (approved):
LGTM

PR #2640, opened 2025-01-17
fix(external_address): limit some of the incoming error from...

RolandSherwin:
- fix(external_address): limit some of the incoming error from affecting the score

We should not consider all the incoming connection error while calculating the external address score. Some of these error are harmless, as we might have a connection established right before/after the error.

This does not prevent all such harmless error, but it is better than just considering every error as an issue.


maqi (approved):

RolandSherwin:
superseded by https://github.com/maidsafe/autonomi/pull/2692

PR #2641, opened 2025-01-17
feat(node): prevent free upload via replication

maqi:
### Description

Use healthy score to prevent free upload via replication.

This is achieved by:
 * only trust replication source with enough healthy status
 * when not having enough healthy knowledge, only trust majority

### DIFF
--- a/ant-networking/src/cmd.rs
+++ b/ant-networking/src/cmd.rs
@@ -153,6 +153,10 @@ pub enum LocalSwarmCmd {
     AddNetworkDensitySample {
         distance: Distance,
     },
+    /// Send peer scores (collected from storage challenge) to replication_fetcher
+    NotifyPeerScores {
+        peer_scores: Vec<(PeerId, bool)>,
+    },
 }
 
 /// Commands to send to the Swarm
@@ -311,6 +315,9 @@ impl Debug for LocalSwarmCmd {
             }
             LocalSwarmCmd::AddNetworkDensitySample { distance } => {
                 write!(f, "LocalSwarmCmd::AddNetworkDensitySample({distance:?})")
+            }
+            LocalSwarmCmd::NotifyPeerScores { peer_scores } => {
+                write!(f, "LocalSwarmCmd::NotifyPeerScores({peer_scores:?})")
             }
         }
     }
@@ -936,6 +943,10 @@ impl SwarmDriver {
             LocalSwarmCmd::AddNetworkDensitySample { distance } => {
                 cmd_string = "AddNetworkDensitySample";
                 self.network_density_samples.add(distance);
+            }
+            LocalSwarmCmd::NotifyPeerScores { peer_scores } => {
+                cmd_string = "NotifyPeerScores";
+                self.replication_fetcher.add_peer_scores(peer_scores);
             }
         }
 
--- a/ant-networking/src/cmd.rs
+++ b/ant-networking/src/cmd.rs
@@ -23,7 +23,7 @@ use ant_protocol::{
 use libp2p::{
     kad::{
         store::{Error as StoreError, RecordStore},
-        KBucketDistance as Distance, Quorum, Record, RecordKey,
+        KBucketDistance as Distance, Quorum, Record, RecordKey, K_VALUE,
     },
     Multiaddr, PeerId,
 };
@@ -961,7 +961,7 @@ impl SwarmDriver {
             }
             LocalSwarmCmd::AddFreshReplicateRecords { holder, keys } => {
                 cmd_string = "AddFreshReplicateRecords";
-                self.add_keys_to_replication_fetcher(holder, keys, true);
+                let _ = self.add_keys_to_replication_fetcher(holder, keys, true);
             }
         }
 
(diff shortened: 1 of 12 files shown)


ant-networking/src/cmd.rs
```diff
@@ -153,6 +153,15 @@ pub enum LocalSwarmCmd {
     AddNetworkDensitySample {
         distance: Distance,
     },
+    /// Send peer scores (collected from storage challenge) to replication_fetcher
+    NotifyPeerScores {
+        peer_scores: Vec<(PeerId, bool)>,
```
grumbach, line 158: Does that send our local peer scores to other peers? I'm not sure this should be shared as attackers can easily make these up.

ant-networking/src/replication_fetcher.rs
```diff
@@ -50,6 +50,13 @@ pub(crate) struct ReplicationFetcher {
     /// used when the node is full, but we still have "close" data coming in
     /// that is _not_ closer than our farthest max record
     farthest_acceptable_distance: Option<Distance>,
+    /// Scoring of peers collected from storage_challenge.
+    /// To be a trustworthy replication source, the peer must has two latest scoring both healthy.
+    peers_scores: HashMap<PeerId, (VecDeque<bool>, Instant)>,
```
grumbach, line 55: Should this be in here? I wonder if replication is the only thing that this affects? Shouldn't this be in the Node struct?

ant-node/src/node.rs
```diff
@@ -922,20 +926,26 @@ impl Node {
...
 
+        let mut peer_scores = vec![];
         while let Some(res) = tasks.join_next().await {
             match res {
                 Ok((peer_id, score)) => {
-                    if score < MIN_ACCEPTABLE_HEALTHY_SCORE {
+                    let is_healthy = score < MIN_ACCEPTABLE_HEALTHY_SCORE;
+                    if !is_healthy {
                         info!("Peer {peer_id:?} failed storage challenge with low score {score}/{MIN_ACCEPTABLE_HEALTHY_SCORE}.");
                         // TODO: shall the challenge failure immediately triggers the node to be removed?
```
grumbach, line 936: This sounds like a good idea. Maybe in a next PR so we can test and compare in a testnet to make sure we got the correct threshold?

ant-node/src/replication.rs
```diff
@@ -174,4 +176,61 @@ impl Node {
...
+                        Ok(false) => {
+                            info!(
+                                "Received a fresh update against a non-existing record of {addr:?}"
+                            );
+                            continue;
+                        }
+                        Err(err) => {
+                            info!("Failed to verify the local existence of {addr:?} with error {err:?}");
+                            continue;
+                        }
```
grumbach, line 218: Do we deal with split record case here? 
Say I have a different GraphEntry or ScratchPad at the same address here, what happens?

ant-networking/src/replication_fetcher.rs
```diff
@@ -323,6 +286,228 @@ impl ReplicationFetcher {
...
+        if self.peers_scores.len() > 20 {
+            let mut oldest_peer = PeerId::random();
+            let mut oldest_timestamp = Instant::now();
+            for (peer_id, (_peer_scores, last_seen)) in self.peers_scores.iter() {
+                if *last_seen < oldest_timestamp {
+                    oldest_timestamp = *last_seen;
+                    oldest_peer = *peer_id;
+                }
+            }
+            let _ = self.peers_scores.remove(&oldest_peer);
```
grumbach, line 318: Feels here like this could be exploited by spamming us with newer peer, which would make us forget old pals. 
I'd keep good old peers a bit longer. What do you think?

grumbach (commented):
Excellent work Qi! Couple of questions as a first pass!

ant-networking/src/cmd.rs
maqi: no. `replication_fetcher` is just a container held by the smart_driver instance inside the node.
`peer_scores` is only held within a node, never shared out

ant-networking/src/replication_fetcher.rs
maqi: `Node` struct needs to be `clonable` and for `async` ops.
hence get it held here.

If other places need to utilize it as well in future, some infrastructure refactor might be required.

ant-node/src/node.rs
maqi: yes

ant-node/src/replication.rs
maqi: here don't need to.
it's when the copy fetched, then to handle the split case.

ant-networking/src/replication_fetcher.rs
maqi: those `newer peers` then can only `replicate in` with majority copies.
replacing old ones with new ones is the most natural way to do.
we could improve it in future anyway, as a non-breaking change to improve the security.

grumbach: Ok make sense

grumbach (approved):
LGTM!

PR #2642, opened 2025-01-17
feat: add `NodeSpawner` and `NetworkSpawner`

mickvandijke:
REQUIRES mDNS REMOVAL PR: https://github.com/maidsafe/autonomi/pull/2652

Adds the option to spawn nodes and networks from code using `NodeSpawner` and `NetworkSpawner`.

TODO:

- [x] Fn to shutdown a node.
- [x] Fn to shutdown a network.
- [x] Test that verifies nodes in a network know each other.


ant-node/src/lib.rs
```diff
@@ -41,20 +45,22 @@ pub use self::{
...
 };
-
-use ant_evm::RewardsAddress;
+use tokio::sync::watch;
 
 /// Once a node is started and running, the user obtains
 /// a `NodeRunning` object which can be used to interact with it.
 #[derive(Clone)]
 pub struct RunningNode {
+    shutdown_sender: watch::Sender<bool>,
```
RolandSherwin, line 63: We send `TerminateNode(String)` via the `NodeEventsChannel` below, can this be re-used here?

mickvandijke: It could be reused for the node (`NodeEvents`) loop, but for the `SwarmDriver` loop it wouldn't make sense, as the broadcast channel uses `ant_node::event::NodeEvent`, which is inaccessible from the `ant-networking` crate in which the `SwarmDriver` loop resides.

ant-node/examples/spawn_local_network.rs
```diff
@@ -0,0 +1,33 @@
+use ant_node::spawn::network::NetworkSpawner;
```
maqi: missing copyrigt doc ?

ant-node/src/spawn/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod network;
```
maqi: missing copy right doc

ant-node/src/spawn/network.rs
```diff
@@ -0,0 +1,234 @@
+use crate::spawn::node::NodeSpawner;
```
maqi: missing copy right doc

ant-node/src/spawn/node.rs
```diff
@@ -0,0 +1,228 @@
+use crate::utils::get_root_dir_and_keypair;
```
maqi: missing copy right doc

ant-node/src/utils.rs
```diff
@@ -0,0 +1,80 @@
+use ant_protocol::node::get_antnode_root_dir;
```
maqi, line 9: missing copy right doc

maqi (changes requested):
for the files under the `spawn`, as the file names of `network.rs` and `node.rs` is very similar to the existing `Node` `Network, can they be renamed to `network_spawner.rs` `node_spawner.rs` to match the struct names ?

Also, seems the functions within the NetworkSpawner and NodeSpawner are very much similar to the existing ones, for example `ant_node\src\bin\antnode\main.rs::create_secret_key_file`
I am wondering why there is new duplicated functions as well?

It seems `NetworkSpawner` and `NodeSpawner` are two `utility wrapping struct`, which seems the PR title is bit misleading.
Shall the PR Title change to: Add NodeSpawner and NetworkSpawner ?

mickvandijke:
> for the files under the `spawn`, as the file names of `network.rs` and `node.rs` is very similar to the existing `Node` `Network, can they be renamed to `network_spawner.rs` `node_spawner.rs` to match the struct names ?

Ah yes, great suggestion!

> Also, seems the functions within the NetworkSpawner and NodeSpawner are very much similar to the existing ones, for example `ant_node\src\bin\antnode\main.rs::create_secret_key_file` I am wondering why there is new duplicated functions as well?

Functions like `create_secret_key_file` etc.. have been moved to `ant_node/src/utils.rs`. They shouldn't be duplicate.

> It seems `NetworkSpawner` and `NodeSpawner` are two `utility wrapping struct`, which seems the PR title is bit misleading. Shall the PR Title change to: Add NodeSpawner and NetworkSpawner ?

Sure, makes sense!

maqi (approved):

PR #2645, opened 2025-01-20
feat: remove deprecated registers

grumbach:
- remove registers
- remove outdated references to spends, auditor and transactions


b-zee (approved):

happybeing:
@b-zee Does this mean you're happy the new data types and they're working on the live network?

I've not seen updates or announcements in their status for developers so been busy on other areas.

Thanks.

grumbach:
Hey @happybeing thanks for coming in, I appreciate your work on the Network and will try to keep you informed better about the new data type coming in to replace Registers. I'm currently working on `GraphEntry` a new data type (on main) which is the replacement for, and at the crossroads between Registers, Spends and Transactions but without their respective weaknesses and in some cases, design flaws. I think @dirvine has mentioned them but maybe we didn't communicate enough, that's our bad, sorry about that! 
The `GraphEntry` should allow for more powerful use of BLS keys and no update limit (unlike registers had), while staying generic and light for the Network (very little node work).

happybeing:
@grumbach thanks for getting back to me. 

I have kept up with the plan for GraphEntry but assumed the new data types would be in place before removing Registers which I'm using until Pointer and GraphEntry are ready.

So seeing this I thought maybe that was the case. The next release (removing Registers) will create a barrier for development until the replacements are included I'm afraid.

Although problems with uploads have also made me hold back from a release. 

I've held `awe` back since October, and now have its replacement `dweb` almost ready so it's a shame not to be able to show these as the marketing has now begun and TGE is imminent. 

I do appreciate you are all doing your best at a particularly busy time.

PR #2647, opened 2025-01-21
feat: massive cleanup of autonomi API repo structure

grumbach:
- The 4 datatypes are now easy to find
- Error types are where they should be
- The high level data types are in they separate module


b-zee (approved):

b-zee:
Perhaps rename `/datatypes/` to `/data_types/`.

PR #2648, opened 2025-01-21
fix(client): retry donwload if it fails

RolandSherwin:


jacderida:
I will raise this as a new PR from the correct hotfix branch.

PR #2655, opened 2025-01-22
fix(client): non-blocking sleep during quoting re-attempts

maqi:
### Description

non-blocking sleep during quoting re-attempts

### Type of Change

### DIFF
--- a/autonomi/src/client/quote.rs
+++ b/autonomi/src/client/quote.rs
@@ -217,7 +217,7 @@ async fn fetch_store_quote_with_retries(
         }
         // Shall have a sleep between retries to avoid choking the network.
         // This shall be rare to happen though.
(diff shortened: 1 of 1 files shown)


autonomi/src/client/quote.rs
```diff
@@ -217,7 +217,7 @@ async fn fetch_store_quote_with_retries(
         }
         // Shall have a sleep between retries to avoid choking the network.
         // This shall be rare to happen though.
-        std::thread::sleep(std::time::Duration::from_secs(5));
+        let _ = tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
mickvandijke:
```suggestion
        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```

It shouldn't return anything

mickvandijke (approved):

PR #2657, opened 2025-01-22
feat(client): implement ClientOperationCfg to control any client oper…

RolandSherwin:


b-zee (approved):

maqi (approved):

ant-cli/src/commands/file.rs
```diff
@@ -105,9 +134,20 @@ pub async fn upload(file: &str, public: bool, peers: NetworkPeers) -> Result<()>
...
+pub async fn download(
+    addr: &str,
+    dest_path: &str,
+    peers: NetworkPeers,
+    read_quorum: Option<ResponseQuorum>,
+    json: bool,
+) -> Result<()> {
+    let mut client_operation_config = ClientOperationConfig::default();
+    if let Some(read_quorum) = read_quorum {
+        client_operation_config.chunk_read_quorum(read_quorum);
```
grumbach, line 148: Not sure I understand why we would need a quorum for chunk? Isn't the default of 1 sufficient?

autonomi/src/client/config.rs
```diff
@@ -0,0 +1,315 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct ClientOperationConfig {
+    /// Configuration for chunk operations.
+    pub(crate) chunk_operation_config: ChunkOperationConfig,
+    /// Configuration for graph operations.
+    pub(crate) graph_operation_config: GraphOperationConfig,
+    /// Configuration for pointer operations.
+    pub(crate) pointer_operation_config: PointerOperationConfig,
+    /// Configuration for scratchpad operations.
+    pub(crate) scratchpad_operation_config: ScratchpadOperationConfig,
```
grumbach, line 49: Do these four really need to be 4 different types?? They look the same to me

autonomi/src/client/data_types/chunk.rs
```diff
@@ -110,13 +104,34 @@ fn hash_to_short_string(input: &str) -> String {
...
     pub async fn chunk_get(&self, addr: ChunkAddress) -> Result<Chunk, GetError> {
+        let result = self.chunk_get_inner(addr).await;
+        // Reporting
+        if let Some(channel) = self.client_event_sender.as_ref() {
+            let xor_name = addr.xorname();
+            let event = if result.is_ok() {
+                ClientEvent::DownloadSucceeded(*xor_name)
+            } else {
+                ClientEvent::DownloadFailed(*xor_name)
+            };
```
grumbach: Here the error is not shown so we will have no idea why get failed. I think it would be helpful to keep the error messages.

autonomi/src/client/data_types/graph.rs
```diff
@@ -54,7 +54,38 @@ impl Client {
...
+                ClientEvent::DownloadSucceeded(*xor_name)
+            } else {
+                ClientEvent::DownloadFailed(*xor_name)
+            };
+
+            if let Err(err) = channel.send(event).await {
+                error!("Failed to send client event: {err:?}");
+            }
+        }
+        result
```
grumbach: This pattern is for every type. Can't we make this a function so we reduce this to one line? This way we don't need the new tedious inner pattern for every type?

autonomi/src/client/data_types/pointer.rs
```diff
@@ -52,11 +52,32 @@ pub enum PointerError {
...
-            get_quorum: Quorum::Majority,
-            retry_strategy: Some(RetryStrategy::Balanced),
+            get_quorum: self.operation_config.pointer_operation_config.read_quorum,
+            retry_strategy: self
+                .operation_config
+                .pointer_operation_config
+                .read_retry_strategy,
             target_record: None,
             expected_holders: Default::default(),
         };
```
grumbach, line 72: I don't see how this is improving our code. 
By reading this code I have no idea what quorum we are using here.

grumbach (changes requested):
Good work! Couple of areas where I have doubts and questions.

ant-cli/src/commands/file.rs
RolandSherwin: Yes that is right, 1 is sufficient for general use case. But we would like to test out replication performance of the nodes and this would require us to tweak the read quorum.

autonomi/src/client/config.rs
RolandSherwin: Yeah, I was thinking about the same and decided on 4 types for these reasons:
- You have to deal with matching, if you have a generic `NetworkOperationConfig` and we also avoid creating a flag/enum to denote each type of data.
  - We could ofc have a generic struct inside each of these `*OperationConfig` to reduce some duplication of the definitions, but since this is a simple config struct, it would be nice if things remained flat?
- Each data type could get it's own set of unique configs later down the road, and this makes things easier.

autonomi/src/client/data_types/chunk.rs
RolandSherwin: Do you mean pass the error type through the `Event` channel? I don't think `Error` implement `Copy` for us to do that :/ I'm not sure if there is a better alternative.

autonomi/src/client/data_types/graph.rs
RolandSherwin: Sounds good! Will try it out!

autonomi/src/client/data_types/pointer.rs
RolandSherwin: The configs are set inside a single file for all the data types `autonomi/src/client/config.rs`. 

I think both of these patterns have their own tradeoffs.. :/ For instance, with the old method we got to see the configs without indirections when we are editing in that area. But this also meant that if people are not looking into that particular area of code, sometimes incorrect values can creep in. I think prior to a couple rebases, I found places with 0 retries for verification.

RolandSherwin: The other alternative is to set `Option` for the inner configs and override the default values with those `Some<config>` values. I tried it earlier, but I like the ergonomics of `config.rs`..

grumbach (approved):

ant-node/src/put_validation.rs
```diff
@@ -693,16 +693,6 @@ impl Node {
...
-        // loud mode: print a celebratory message to console
-        #[cfg(feature = "loud")]
-        {
-            println!("🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟   RECEIVED REWARD   🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟");
-            println!(
-                "Total payment of {reward_amount:?} atto tokens accepted for record {pretty_key}"
-            );
-            println!("🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟🌟");
-        }
-
```
grumbach, line 711: Wait this is useful, please keep it

grumbach (changes requested):

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
+        failed_files: result
+            .as_ref()
+            .map_or_else(|(_, failed_files)| failed_files.clone(), |_| vec![]),
+    };
+
+    if json {
+        download_output.print_json();
+    } else {
+        download_output.print();
+    }
```
grumbach, line 81: Can't this be part of client config so we don't need to carry around a json bool?

autonomi/Cargo.toml
```diff
@@ -23,7 +23,6 @@ name = "put_and_dir_upload"
 default = []
 external-signer = ["ant-evm/external-signer"]
 extension-module = ["pyo3/extension-module"]
-loud = []
```
grumbach, line 26: This feature is very useful for local testing for us, I think we should keep it around. Most users will never us it but that's fine, it's off by default.

autonomi/src/client/high_level/files/archive_public.rs
```diff
@@ -161,12 +161,6 @@ impl Client {
             .to_bytes()
             .map_err(|e| PutError::Serialization(format!("Failed to serialize archive: {e:?}")))?;
 
-        #[cfg(feature = "loud")]
-        println!(
-            "Uploading public archive referencing {} files",
-            archive.map().len()
-        );
-
```
grumbach, line 169: Although I agree it's not the prettiest way to do it, it's an excellent UX for CLI. Until we find another way I think we need to keep it

autonomi/src/client/high_level/files/fs_private.rs
```diff
@@ -102,8 +102,6 @@ impl Client {
             }
         }
 
-        #[cfg(feature = "loud")]
-        println!("Upload completed in {:?}", start.elapsed());
```
grumbach, line 106: Same here

autonomi/src/client/high_level/files/fs_private.rs
```diff
@@ -128,8 +126,6 @@ impl Client {
         wallet: &EvmWallet,
     ) -> Result<DataMapChunk, UploadError> {
         info!("Uploading file: {path:?}");
-        #[cfg(feature = "loud")]
-        println!("Uploading file: {path:?}");
```
grumbach, line 132: Same here

autonomi/src/client/mod.rs
```diff
@@ -266,6 +252,10 @@ impl Client {
 
         client_event_receiver
     }
+
+    pub fn get_evm_network(&self) -> &EvmNetwork {
+        &self.evm_network
+    }
```
grumbach: This is confusing, there is another fn named `get_evm_network` that actually gets it. I suggest we rename this to just `evm_network()` so we know it's not "get"ing it from anywhere

autonomi/src/client/mod.rs
```diff
@@ -320,23 +310,23 @@ async fn handle_event_receiver(
...
                         sender
                             .send(Err(ConnectError::TimedOut))
-                            .expect("receiver should not close");
+                            .expect("Could not send TimedOut, receiver should not close");
                     }
                 }
             }
             event = event_receiver.recv() => {
-                let event = event.expect("receiver should not close");
+                let event = event.expect("Could not receiver Network event, receiver should not close");
```
grumbach, line 337: Although this originates not in this PR, I suggest we BAN `expect` altogether as they do fail sometimes!

ant-cli/Cargo.toml
```diff
@@ -27,7 +27,7 @@ ant-bootstrap = { path = "../ant-bootstrap", version = "0.1.4" }
 ant-build-info = { path = "../ant-build-info", version = "0.1.23" }
 ant-logging = { path = "../ant-logging", version = "0.2.45" }
 ant-protocol = { path = "../ant-protocol", version = "0.3.3" }
-autonomi = { path = "../autonomi", version = "0.3.5", features = [ "loud" ] }
```
grumbach, line 30: This feature although I agree not the most elegant way to do it, made the UX great. Removing it will result in a frustrating user experience while waiting for uploads.

ant-networking/src/event/mod.rs
```diff
@@ -366,8 +363,6 @@ impl SwarmDriver {
         }
 
         info!("kBucketTable has {index:?} kbuckets {total_peers:?} peers, {kbucket_table_stats:?}, estimated network size: {estimated_network_size:?}");
-        #[cfg(feature = "loud")]
-        println!("Estimated network size: {estimated_network_size:?}");
```
grumbach, line 371: This too is very useful for local tests

ant-networking/src/event/mod.rs
```diff
@@ -242,9 +242,6 @@ impl SwarmDriver {
         let n_peers = self.peers_in_rt;
         info!("New peer added to routing table: {added_peer:?}, now we have #{n_peers} connected peers");
 
-        #[cfg(feature = "loud")]
-        println!("New peer added to routing table: {added_peer:?}, now we have #{n_peers} connected peers");
-
```
grumbach, line 248: This too

ant-networking/src/event/swarm.rs
```diff
@@ -435,11 +435,6 @@ impl SwarmDriver {
                             match err {
                                 TransportError::MultiaddrNotSupported(addr) => {
                                     warn!("Multiaddr not supported : {addr:?}");
-                                    #[cfg(feature = "loud")]
-                                    {
-                                        println!("Multiaddr not supported : {addr:?}");
-                                        println!("If this was your bootstrap peer, restart your node with a supported multiaddr");
-                                    }
```
grumbach: This as well

ant-cli/Cargo.toml
b-zee: Also not a proponent of removing this. Users on the forum are already showing a lack of understanding what's happening when using the CLI.

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
     client: &mut Client,
-) -> Result<()> {
+) -> Result<(), (DownloadError, Vec<(String, String)>)> {
     let archive = client
-        .archive_get(private_address)
+        .archive_get(private_address.clone())
         .await
-        .wrap_err("Failed to fetch data from address")?;
+        .map_err(|err| {
+            error!("Failed to fetch archive from address: {private_address:?}");
```
b-zee, line 98: Use `inspect_err` to report errors.

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
         };
 
         let path = PathBuf::from(dest_path).join(path);
         let here = PathBuf::from(".");
         let parent = path.parent().unwrap_or_else(|| &here);
-        std::fs::create_dir_all(parent)?;
-        std::fs::write(path, bytes)?;
-        progress_bar.clone().inc(1);
+        std::fs::create_dir_all(parent).map_err(|err| {
+            error!("Failed to create parent directories for {path:?}: {err}");
```
b-zee, line 124: Use `inspect_err`

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
+            (err.into(), all_errs.clone())
+        })?;
+        std::fs::write(&path, bytes).map_err(|err| {
+            error!("Failed to write file {path:?}: {err}");
```
b-zee, line 128: Use `inspect_err`

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
     dest_path: &str,
     client: &mut Client,
-) -> Result<()> {
-    let archive = client
-        .archive_get_public(address)
-        .await
-        .wrap_err("Failed to fetch data from address")?;
+) -> Result<(), (DownloadError, Vec<(String, String)>)> {
+    let archive = client.archive_get_public(address).await.map_err(|err| {
+        error!("Failed to fetch archive from address: {address:?}");
```
b-zee, line 152: Use `inspect_err`.

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
         };
 
         let path = PathBuf::from(dest_path).join(path);
         let here = PathBuf::from(".");
         let parent = path.parent().unwrap_or_else(|| &here);
-        std::fs::create_dir_all(parent)?;
-        std::fs::write(path, bytes)?;
-        progress_bar.clone().inc(1);
+        std::fs::create_dir_all(parent).map_err(|err| {
+            error!("Failed to create parent directories for {path:?}: {err}");
```
b-zee, line 178: Use `inspect_err`.

ant-cli/src/actions/download.rs
```diff
@@ -7,123 +7,192 @@
...
+            (err.into(), all_errs.clone())
+        })?;
+        std::fs::write(&path, bytes).map_err(|err| {
+            error!("Failed to write file {path:?}: {err}");
```
b-zee, line 182: Use inspect_err.

b-zee (changes requested):

autonomi/src/client/mod.rs
b-zee: Agreed, `get_` is unconventional for getters in Rust too.

RolandSherwin: Modified it!

ant-cli/src/actions/download.rs
RolandSherwin: We actually map the error to a different type here, so it should be fine for the logging to go in here I think.

ant-cli/src/actions/download.rs
RolandSherwin: Same as above, we actually map the error to a different type here, so it should be fine for the logging to go in here.

ant-cli/src/actions/download.rs
RolandSherwin: Same as above.

RolandSherwin:
superseded by https://github.com/maidsafe/autonomi/pull/2708

PR #2658, opened 2025-01-23
feat: actually working pointers, key derivation, various fixes

grumbach:
- actually working pointers (node and client side)
- new simpler API for pointers as an alternative to the lower level put and get
- e2e tests for all that to prove it works
- various API fixes
- bls key derivation tools for autonomi

### DIFF
--- a/ant-node/src/put_validation.rs
+++ b/ant-node/src/put_validation.rs
@@ -16,7 +16,7 @@ use ant_protocol::storage::GraphEntry;
 use ant_protocol::{
     storage::{
         try_deserialize_record, try_serialize_record, Chunk, DataTypes, GraphEntryAddress, Pointer,
-        RecordHeader, RecordKind, Scratchpad, ValidationType,
+        PointerAddress, RecordHeader, RecordKind, Scratchpad, ValidationType,
     },
     NetworkAddress, PrettyPrintRecordKey,
 };
@@ -241,28 +241,44 @@ impl Node {
                 res
             }
             RecordKind::DataOnly(DataTypes::Pointer) => {
-                // Pointers should always be paid for
-                error!("Pointer should not be validated at this point");
-                Err(Error::InvalidPutWithoutPayment(
-                    PrettyPrintRecordKey::from(&record.key).into_owned(),
-                ))
+                let pointer = try_deserialize_record::<Pointer>(&record)?;
+                let net_addr = NetworkAddress::from_pointer_address(pointer.network_address());
+                let pretty_key = PrettyPrintRecordKey::from(&record.key);
+                let already_exists = self
+                    .validate_key_and_existence(&net_addr, &record.key)
+                    .await?;
+
+                if !already_exists {
+                    warn!("Pointer at address: {:?}, key: {:?} does not exist locally, rejecting PUT without payment", pointer.network_address(), pretty_key);
+                    return Err(Error::InvalidPutWithoutPayment(
+                        PrettyPrintRecordKey::from(&record.key).into_owned(),
+                    ));
+                }
+
+                let res = self
+                    .validate_and_store_pointer_record(pointer, record.key.clone(), true, None)
+                    .await;
+                if res.is_ok() {
+                    let content_hash = XorName::from_content(&record.value);
+                    Marker::ValidPointerPutFromClient(&pretty_key).log();
+
+                    // Notify replication_fetcher to mark the attempt as completed.
+                    self.network().notify_fetch_completed(
+                        record.key.clone(),
+                        ValidationType::NonChunk(content_hash),
+                    );
+                }
+                res
             }
             RecordKind::DataWithPayment(DataTypes::Pointer) => {
                 let (payment, pointer) =
                     try_deserialize_record::<(ProofOfPayment, Pointer)>(&record)?;
 
-                // check if the deserialized value's PointerAddress matches the record's key
                 let net_addr = NetworkAddress::from_pointer_address(pointer.network_address());
-                let key = net_addr.to_record_key();
-                let pretty_key = PrettyPrintRecordKey::from(&key);
-                if record.key != key {
-                    warn!(
-                        "Record's key {pretty_key:?} does not match with the value's PointerAddress, ignoring PUT."
-                    );
-                    return Err(Error::RecordKeyMismatch);
-                }
-
-                let already_exists = self.validate_key_and_existence(&net_addr, &key).await?;
+                let pretty_key = PrettyPrintRecordKey::from(&record.key);
+                let already_exists = self
+                    .validate_key_and_existence(&net_addr, &record.key)
+                    .await?;
 
                 // The pointer may already exist during the replication.
                 // The payment shall get deposit to self even if the pointer already exists.
@@ -278,11 +294,17 @@ impl Node {
                     }
                 }
 
-                let res = self.validate_and_store_pointer_record(pointer, key, true, Some(payment));
+                let res = self
+                    .validate_and_store_pointer_record(
+                        pointer,
+                        record.key.clone(),
+                        true,
+                        Some(payment),
+                    )
+                    .await;
                 if res.is_ok() {
                     let content_hash = XorName::from_content(&record.value);
-                    Marker::ValidPointerPutFromClient(&PrettyPrintRecordKey::from(&record.key))
-                        .log();
+                    Marker::ValidPointerPutFromClient(&pretty_key).log();
 
                     // Notify replication_fetcher to mark the attempt as completed.
                     self.network().notify_fetch_completed(
@@ -340,6 +362,7 @@ impl Node {
                 let pointer = try_deserialize_record::<Pointer>(&record)?;
                 let key = record.key.clone();
                 self.validate_and_store_pointer_record(pointer, key, false, None)
+                    .await
             }
         }
     }
@@ -675,8 +698,50 @@ impl Node {
         Ok(local_entries)
     }
 
+    /// Get the local Pointer for the provided `PointerAddress`
+    /// This only fetches the Pointer from the local store and does not perform any network operations.
+    /// If the local Pointer is not present or corrupted, returns `None`.
+    async fn get_local_pointer(&self, addr: PointerAddress) -> Option<Pointer> {
+        // get the local Pointer
+        let record_key = NetworkAddress::from_pointer_address(addr).to_record_key();
+        debug!("Checking for local Pointer with key: {record_key:?}");
+        let local_record = match self.network().get_local_record(&record_key).await {
+            Ok(Some(r)) => r,
+            Ok(None) => {
+                debug!("Pointer is not present locally: {record_key:?}");
+                return None;
+            }
+            Err(e) => {
+                error!("Failed to get Pointer record at {addr:?}: {e}");
+                return None;
+            }
+        };
+
+        // deserialize the record and get the Pointer
+        let local_header = match RecordHeader::from_record(&local_record) {
+            Ok(h) => h,
+            Err(_) => {
+                error!("Failed to deserialize Pointer record at {addr:?}");
+                return None;
+            }
+        };
+        let record_kind = local_header.kind;
+        if !matches!(record_kind, RecordKind::DataOnly(DataTypes::Pointer)) {
+            error!("Found a {record_kind} when expecting to find Pointer at {addr:?}");
+            return None;
+        }
+        let local_pointer: Pointer = match try_deserialize_record(&local_record) {
+            Ok(p) => p,
+            Err(_) => {
+                error!("Failed to deserialize Pointer record at {addr:?}");
+                return None;
+            }
+        };
+        Some(local_pointer)
+    }
+
     /// Validate and store a pointer record
-    pub(crate) fn validate_and_store_pointer_record(
+    pub(crate) async fn validate_and_store_pointer_record(
         &self,
         pointer: Pointer,
         key: RecordKey,
@@ -696,6 +761,18 @@ impl Node {
             return Err(Error::RecordKeyMismatch);
         }
 
+        // Keep the pointer with the highest counter
+        if let Some(local_pointer) = self.get_local_pointer(pointer.network_address()).await {
+            if pointer.counter() <= local_pointer.counter() {
+                info!(
+                    "Ignoring Pointer PUT at {key:?} with counter less than or equal to the current counter ({} <= {})",
+                    pointer.counter(),
+                    local_pointer.counter()
+                );
+                return Ok(());
+            }
+        }
+
         // Store the pointer
         let record = Record {
             key: key.clone(),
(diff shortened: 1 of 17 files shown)


ant-node/src/put_validation.rs
```diff
@@ -241,28 +241,44 @@ impl Node {
                 res
             }
             RecordKind::DataOnly(DataTypes::Pointer) => {
-                // Pointers should always be paid for
-                error!("Pointer should not be validated at this point");
-                Err(Error::InvalidPutWithoutPayment(
-                    PrettyPrintRecordKey::from(&record.key).into_owned(),
-                ))
```
grumbach, line 248: This would reject any updates meaning pointers can only be set once, making them useless!

ant-protocol/src/storage/pointer.rs
```diff
@@ -118,91 +130,30 @@ impl Pointer {
...
-        bytes_array.copy_from_slice(&bytes);
-
-        let owner = PublicKey::from_bytes(bytes_array).map_err(PointerError::BlsError)?;
-
-        let mut rng = thread_rng();
-        let target = PointerTarget::ChunkAddress(ChunkAddress::new(XorName::random(&mut rng)));
-
-        // Create a temporary secret key just for hex decoding test purposes
-        let sk = SecretKey::random();
-        Ok(Self::new(owner, 0, target, &sk))
```
grumbach, line 141: Discarded this as it is wrong AND unused. Removed the associated tests below too!

autonomi/src/client/data_types/chunk.rs
```diff
@@ -113,10 +109,10 @@ fn hash_to_short_string(input: &str) -> String {
 
 impl Client {
     /// Get a chunk from the network.
-    pub async fn chunk_get(&self, addr: ChunkAddr) -> Result<Chunk, GetError> {
+    pub async fn chunk_get(&self, addr: ChunkAddress) -> Result<Chunk, GetError> {
```
grumbach, line 112: Pointers point to `ChunkAddress` which is why we need the chunk API to return chunk addresses.

autonomi/src/lib.rs
```diff
@@ -81,9 +84,26 @@ pub use libp2p::Multiaddr;
 
 #[doc(inline)]
 pub use client::{
-    data_types::chunk::Chunk, data_types::graph::GraphEntry, data_types::pointer::Pointer,
-    data_types::scratchpad::Scratchpad, files::archive_private::PrivateArchive,
-    files::archive_public::PublicArchive, files::Metadata, Client, ClientConfig,
+    // Data types
+    data_types::chunk::Chunk,
+    // Addresses for the data types
+    data_types::chunk::ChunkAddress,
```
b-zee, line 90: Great addition. Makes the API super usable now!

autonomi/src/client/data_types/chunk.rs
```diff
@@ -208,7 +204,7 @@ impl Client {
         &self,
         chunk: &Chunk,
         payment: ProofOfPayment,
-    ) -> Result<(), PutError> {
+    ) -> Result<ChunkAddress, PutError> {
```
b-zee, line 207: The returned address is the same as the address passed with `&Chunk`. I was thinking we might eventually return more information about the upload (e.g. costs), but we're already allowing devs to track that with the client events (though sparsely documented)?

autonomi/src/client/data_types/graph.rs
```diff
@@ -42,25 +43,29 @@ pub enum GraphError {
     InvalidQuote,
     #[error("Entry already exists at this address: {0:?}")]
     AlreadyExists(GraphEntryAddress),
+    #[error("Graph forked! Multiple entries found at this address: {0:?}")]
```
b-zee: The `{0:?}` will debug print the `Vec<GraphEntry>`, not only the address.

autonomi/src/client/data_types/graph.rs
```diff
@@ -42,25 +43,29 @@ pub enum GraphError {
...
         address: GraphEntryAddress,
-    ) -> Result<Vec<GraphEntry>, GraphError> {
+    ) -> Result<GraphEntry, GraphError> {
         let graph_entries = self.network.get_graph_entry(address).await?;
-
-        Ok(graph_entries)
+        match &graph_entries[..] {
+            [entry] => Ok(entry.clone()),
+            multiple => Err(GraphError::Fork(multiple.to_vec())),
+        }
```
b-zee, line 60: Bit convoluted, but thought it could be done without cloning:

```suggestion
        match TryInto::<[_; 1]>::try_into(graph_entries) {
            Ok([a]) => Ok(a),
            Err(vec) => Err(GraphError::Fork(vec)),
        }
```

autonomi/src/client/data_types/pointer.rs
```diff
@@ -29,21 +40,48 @@ pub enum PointerError {
...
     #[error("Pointer already exists at this address: {0:?}")]
     PointerAlreadyExists(PointerAddress),
+    #[error("Pointer cannot be updated as it does not exist, please create it first or wait for it to be created")]
+    CannotUpdateNewPointer,
 }
 
 impl Client {
     /// Get a pointer from the network
     pub async fn pointer_get(&self, address: PointerAddress) -> Result<Pointer, PointerError> {
+        info!("Getting pointer: {address:?}");
```
b-zee: Isn't this log trace better left to do for the caller?

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -35,12 +35,21 @@ pub enum ScratchpadError {
 impl Client {
     /// Get Scratchpad from the Network
     /// It is stored at the owner's public key
-    pub async fn scratchpad_get(
+    pub async fn scratchpad_get_from_public_key(
```
b-zee, line 38: I don't like the name change here. Isn't it implied for the scratchpad that the address is derived from the public key? This function name would have made sense if there are multiple options. (E.g. `scrathpad_get_from_chunk_address`)

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -139,12 +148,14 @@ impl Client {
     }
 
     /// Create a new scratchpad to the network
+    /// Returns the cost of the scratchpad and the address of the scratchpad
     pub async fn scratchpad_create(
         &self,
         scratchpad: Scratchpad,
         payment_option: PaymentOption,
-    ) -> Result<AttoTokens, PutError> {
+    ) -> Result<(AttoTokens, ScratchpadAddress), PutError> {
```
b-zee, line 156: Same comment as on the `chunk_upload_with_payment`. Why return the address?

b-zee (commented):
Left a few questions, but they're nitpicking.

Overall this is a really awesome PR, great work @grumbach!

ant-node/src/put_validation.rs
maqi: Not exactly.
`PointData` can still be updated via payment.
i.e. here just block `free updates`

then, the qustion is actually is: is PointData designed to be `pay for update` or `pay once, then free updates`

autonomi/src/client/data_types/pointer.rs
```diff
@@ -29,21 +40,48 @@ pub enum PointerError {
...
+
         let key = NetworkAddress::from_pointer_address(address).to_record_key();
-        let record = self.network.get_local_record(&key).await?;
+        debug!("Fetching pointer from network at: {key:?}");
```
maqi, line 51: this is somehow `duplicated` with the log at the beginning of the function?
NetworkAddress debug format already show the key is about a Pointer

autonomi/src/client/data_types/pointer.rs
```diff
@@ -71,60 +109,167 @@ impl Client {
...
 
         // verify payment was successful
-        let (proof, _price) = match payment_proofs.get(&xor_name) {
-            Some((proof, price)) => (proof, price),
+        let (proof, price) = match payment_proofs.get(&xor_name) {
+            Some((proof, price)) => (Some(proof), price),
             None => {
-                error!("Pointer at address: {address:?} was already paid for");
-                return Err(PointerError::PointerAlreadyExists(address));
+                info!("Pointer at address: {address:?} was already paid for, update is free");
```
maqi, line 113: need to cross check with the design that whether an update to the PointData is free or shall be paid.
if need to `pay for update` , the `quoting` flow need to be double checked as well.

autonomi/src/client/data_types/pointer.rs
```diff
@@ -71,60 +109,167 @@ impl Client {
...
+        }
+
+        let pointer = Pointer::new(owner, 0, target);
+        self.pointer_put(pointer, wallet).await
+    }
+
+    /// Update an existing pointer to point to a new target on the network
+    /// The pointer needs to be created first with [`Client::pointer_put`]
+    /// This operation is free as the pointer was already paid for at creation
+    pub async fn pointer_update(
```
maqi, line 202: there are now `pointer_create`, `pointer_update` and `point_put` pub functions
it may only need  `pointer_create` and `pointer_update` functions to be public, and `point_put` being a private helper function to be used by both others ?

autonomi/src/client/data_types/pointer.rs
```diff
@@ -71,60 +109,167 @@ impl Client {
...
 
-        Ok(address)
+        Ok(())
     }
 
     /// Calculate the cost of storing a pointer
-    pub async fn pointer_cost(&self, key: SecretKey) -> Result<AttoTokens, PointerError> {
-        let pk = key.public_key();
-        trace!("Getting cost for pointer of {pk:?}");
+    pub async fn pointer_cost(&self, key: PublicKey) -> Result<AttoTokens, PointerError> {
```
maqi, line 267: not seeing this to be used by the pointer_put ?

autonomi/src/client/mod.rs
```diff
@@ -28,6 +28,7 @@ pub use high_level::files;
 pub use high_level::vault;
 
 pub mod address;
+pub mod key_derivation;
```
maqi, line 31: not seeing this to be used by other mod?

autonomi/src/client/data_types/pointer.rs
```diff
@@ -71,60 +109,167 @@ impl Client {
...
+            .await
+            .inspect_err(|err| {
+                error!("Failed to put record - pointer {address:?} to the network: {err}")
+            })?;
+
+        Ok((total_cost, address))
+    }
+
+    /// Create a new pointer on the network
+    /// Make sure that the owner key is not already used for another pointer as each key is associated with one pointer
```
happybeing, line 172: <strike>This is regression from Registers where one key was able to be associated with many Registers, and makes Pointer significantly less useful than it could be. It at least adds complexity in forcing an application to manage multiple keys for different data, and store that information somewhere rather than be able to rely on a single default key and use it when updating a Pointer based on its address.

I hope that this limitation can be overcome though I'm not sure how difficult that would be!

maqi (changes requested):
ideally, put Pointdata as an additional data_type in the data_with_churn test as well

autonomi/src/client/data_types/pointer.rs
happybeing: > This is regression from Registers where one key...

I think I see how this is not in fact a regression due to the properties of BLS keys. Can you confirm that:
- the app can generate the corresponding signing_key and owner from a main/default SecretKey and an `index: &[u8]`
- which means that the `index: &[u8]` can act as a proxy for the PointerAddress and the key needed to sign an update, providing you have the main/default SecretKey
- for Pointer access, only the PointerAddress is needed

This may be similar to how Registers worked but I think was hidden. Anyway, if I'm correct here just hit :+1: and I'll be :smile:

autonomi/src/client/data_types/chunk.rs
grumbach: Yeah maybe, although in this case the payment is provided as a parameter. 
Im trying to make our api more uniform and have put return:
- price
- address where to get it

autonomi/src/client/mod.rs
grumbach: Yes not used just yet but might prove to be useful for key management as Happybeing noted

autonomi/src/client/data_types/pointer.rs
grumbach: Every type has cost estimation function. This estimate is not used by the put no.

autonomi/src/client/data_types/pointer.rs
grumbach: Maybe, wondering if some folk want to go the manual way and do more creative stuff with pointers?

autonomi/src/client/data_types/graph.rs
grumbach: Oh good find!

autonomi/src/client/data_types/graph.rs
grumbach: Oh wow! Gonna keep it simple for now if you don't mind as _premature optimization is the root of all evil_!

autonomi/src/client/data_types/scratchpad.rs
grumbach: Yeah trying for all `put`s to return price + address. 
Even though we know that `Scratchpad::address` also does the trick, I think it's a user friendly API to return the address were it was put

autonomi/src/client/data_types/pointer.rs
grumbach: Ooh exactly!

autonomi/src/client/data_types/pointer.rs
maqi: sry, I mean the pointer_put carries the `cost quoting` work as well, which actually can just call this `pointer_cost` function to reduce the duplication, and avoid future code update glitch (one updated, the other not)

combined with the [point_put to be private](https://github.com/maidsafe/autonomi/pull/2658/files#r1926834712) comment, I think the pointer API shall be:
```rust
pub fn pointer_create() { creations; pointer_cost(); pointer_put(with_payment); }
pub fn pointer_update() { fetch; update; pointer_put(without_payment); }
pub fn pointer_cost() { quoting only; }
fn pointer_put() { if with_payment { make_payment; put PointerWithPayment } else { put Pointer } }
```
To make pointer_put() public to users, will create more `confusions` than the convenince, as to my experience.

maqi (approved):

happybeing:
@grumbach Hey! Great to see this merged and have updated my code to use this ready for the next network update. I've yet to test locally but optimistic for that. I have a question related to the comment from you that I saw about `client::key_derivation`.

My code now passes a `name: String` along with the main/default signing key, and I create a new one for each Pointer using:

```
let child_secret_key = main_secret_key.derive_child(name.as_bytes());
```

I took a look at the key_derivation code and see it is more complex and mentions it ensures keys cannot be linked. Is that the point of that over my approach above? I guess the idea would be to use `MainPubKey::derive_key(name.as_bytes())` instead of `derive_child()` but is it significantly different?

I'm not knowledgeable about this area and a quick read of the docs left me bamboozled. That first explanation :rofl: is a bit much for me, though I was up late last night:
> This is used to generate a new DerivedPubkey from a MainPubkey, and the corresponding DerivedSecretKey from the MainSecretKey of that MainPubkey.

Great to see Pointers are coming. This will keep me :smile: for a while and I'll work on using `GraphEntry` to give a history soon... without Registers :tada: Both my projects are ready, we just need reliable uploads, esp of directories so I'm hopef!ul the PRs I just saw merged will help with that (more quotes etc). :pray:

PR #2665, opened 2025-01-24
chore: increase quotes fetch sample to `10`

mickvandijke:


ant-networking/src/lib.rs
```diff
@@ -1075,6 +1075,7 @@ impl Network {
         &self,
         key: &NetworkAddress,
         client: bool,
+        expanded_close_group_size: Option<usize>,
```
mickvandijke: I did not want unintentional side effects on the node side for increasing the close group size here, thus an option.

maqi: as long as the value is under KValue(20), it's still the local calculation.
just result in extra small messagings to be exchanged.

I don't think it needs to be an `option` ? as we cann't set it as None anyway ?

maqi (changes requested):

maqi (approved):

PR #2666, opened 2025-01-24
fix: get market price retry condition

jacderida:
This was merged into `main`, but it could help with some errors that users are seeing, so it can go out with our next hotfix.


mickvandijke (approved):
🚀

PR #2670, opened 2025-01-25
Replace rpc in antctl with metric service

ermineJose:
### Description

ant-node-manager currently uses rpc server to collect node_information and network_information, replace the rpc server with metric server. 

### Related Issue

### Type of Change

- [ ] Documentation update

### Checklist


ant-networking/src/driver.rs
```diff
@@ -438,6 +438,7 @@ impl NetworkBuilder {
         is_client: bool,
         req_res_protocol: ProtocolSupport,
         upnp: bool,
+        root_dir: Option<PathBuf>
```
RolandSherwin: Since this is optional, maybe should this be a method inside the `NetworkBuilder`?

ant-node/src/node.rs
```diff
@@ -217,6 +218,22 @@ impl NodeBuilder {
...
+                let listeners = state.listeners.iter().map(|m| m.to_string()).collect();
+                let network_info = NetworkInfoMetrics::new(connected_peers, listeners);
+
+                write_network_metrics_to_file(
+                    runing_node_metrics.root_dir_path.clone(),
+                    network_info,
+                    runing_node_metrics.network.peer_id().to_string()
+                );
+        });
+
```
RolandSherwin: Two things here:
- It would be better if we don't import things from `service_management` inside ant_node. They should be isolated.
- The listeners can be added or removed, and it can happen throughout the lifetime of the program (for private nodes), it would be better if you could move the logic inside the `NetworkEvent::NewListenAddr` event, which is handled inside `antnode` to write the listen addr to the file. 
	- Since we also have to deal with removal of the listen addr (for private nodes), we should also create a new `NetworkEvent::ListenerClosed` event which will be emitted from `SwarmEvent::ListenerClosed`. This newly created event has to be then handled inside antnode and you should remove that particular listenaddr from the file.

ant-service-management/src/metric.rs
```diff
@@ -0,0 +1,216 @@
...
+            let all_metrics = prometheus_parse::Scrape::parse(lines.into_iter())?;
+
+            Ok(all_metrics)
+    }
+
+    pub fn get_node_info_from_metadata_extended(&self ,scrape: &prometheus_parse::Scrape, node_info: &mut NodeInfoMetrics) {
+        for sample in scrape.samples.iter() {
+            for (key, value) in sample.labels.iter() {
+                match key.as_str() {
+                    "peer_id" => node_info.peer_id = value.parse().unwrap(),
```
RolandSherwin: We don't use `unwrap` or `expect` in our code as it may cause the program to crash instantly, we should return error types here.

RolandSherwin: All the unwraps have to be removed.

ant-service-management/src/metric.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+impl MetricClient {
+    pub fn new(endpoint_port: u16) -> Self {
+        Self {
+            endpoint_port: endpoint_port.to_string(),
+        }
+    }
+
+    pub async fn get_endpoint_metrics(&self, endpoint_name: &str) -> Result<prometheus_parse::Scrape, Box<dyn std::error::Error>> {
```
RolandSherwin: You should probalby create a new `error.rs` or if one already exists, add new enums for the errors returned from these functions. `Box<dyn std::error::Error>` is a dynamic object and we cannot easily downcast it back to the actual error. 
We use something called as `thiserror` for error management, it is a good read.

ant-service-management/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub mod error;
 pub mod faucet;
 pub mod node;
 pub mod rpc;
```
RolandSherwin, line 16: I think all the references to `rpc` should be removed from our codebase.

ant-service-management/src/metric.rs
```diff
@@ -0,0 +1,216 @@
...
+                    _ => {}
+                }
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl RpcActions for MetricClient {
+    async fn node_info(&self) -> Result<NodeInfo> {
```
RolandSherwin: Regarding the `Action` here:
- We must rename `RpcActions` -> `MetricActions`
- remove the unused methods
- Also the `expect` has to be removed.

ant-service-management/src/metric.rs
```diff
@@ -0,0 +1,216 @@
...
+        let scrape = self.get_endpoint_metrics("metadata_extended").await.expect("Failed to get endpoint metadata_extended");
+        let mut node_info = NodeInfoMetrics::default();
+        self.get_node_info_from_metadata_extended(&scrape, &mut node_info);
+        let scrape = self.get_endpoint_metrics("metrics").await.expect("Failed to get endpoint metrics");
+        self.get_node_info_from_metrics(&scrape, &mut node_info);
+        println!("node_info: {:?}", node_info);
```
RolandSherwin: I assume these `println` are for debugging? It would be great if these can be tidied up.

ant-service-management/src/metric.rs
jacderida: Yeah, I was also puzzled about the use of `RpcActions`. We shouldn't have references to RPC any more.

ant-networking/src/driver.rs
ermineJose: Done

ant-node/src/node.rs
ermineJose: Done!

ant-service-management/src/metric.rs
ermineJose: Done

ant-service-management/src/metric.rs
ermineJose: Done

ant-service-management/src/metric.rs
ermineJose: Done

ant-service-management/src/lib.rs
ermineJose: ant-service-management has dependency on ant-node-rpc-client module of our source, should i refactor there as well?, i left it intact so that existing changes with ant-node-rpc-client are not broken! or else, i will have to remove them as well!

ant-node-manager/src/add_services/mod.rs
```diff
@@ -108,27 +103,13 @@ pub async fn add_node(
...
     while node_number <= target_node_count {
         trace!("Adding node with node_number {node_number}");
-        let rpc_free_port = if let Some(port) = rpc_port {
-            port
-        } else {
-            service_control.get_available_port()?
-        };
         let metrics_free_port = if let Some(port) = metrics_port {
             Some(port)
-        } else if options.enable_metrics_server {
```
RolandSherwin, line 122: Should `enable_metrics_server` be removed, as `antctl` would now need metrics server by default to function?

ant-node-manager/src/bin/daemon/main.rs
```diff
@@ -10,7 +10,7 @@
 extern crate tracing;
 
 use ant_logging::LogBuilder;
```
RolandSherwin, line 12: The `ant-node-manager/src/bin/daemon/` can now be removed, as it directly depends on the `rpc` feature to function. This should be removed from the deployer as well, so this can be a follow up PR.

ant-node-manager/src/helpers.rs
```diff
@@ -381,7 +381,9 @@ pub fn check_port_availability(port_option: &PortRange, nodes: &[NodeServiceData
         if let Some(port) = node.node_port {
             all_ports.push(port);
         }
-        all_ports.push(node.rpc_socket_addr.port());
+        if let Some(addr) = node.rpc_socket_addr {
+            all_ports.push(addr.port());
+        }
```
RolandSherwin, line 386: Should we remove `rpc_socket_addr` from `NodeServiceData` here?

ant-node-manager/src/bin/daemon/main.rs
ermineJose: Created a Linea ticket for this. https://linear.app/autonominetwork/issue/AUTO-301/remove-antctl-daemon-from-autonomi-and-deployer

PR #2671, opened 2025-01-27
feat: useable and tested scratchpads

grumbach:
- Now scratchpad have their own module clearly separated from Vault and Vaults use the generic Scratchpad API.
- Exactly like Pointers, Scratchpads have a put/get/create/update API that allows for manual or user friendly use of the scratchpad.
- E2E tests for Scratchpad to prove they work as intended. They could serve as usage examples as well.
- GraphEntry, Pointer and Scratchpad use PaymentOption now, unifying the payment process

### DIFF
--- a/ant-protocol/src/storage/scratchpad.rs
+++ b/ant-protocol/src/storage/scratchpad.rs
@@ -23,32 +23,81 @@ pub struct Scratchpad {
     /// Network address. Omitted when serialising and
     /// calculated from the `encrypted_data` when deserialising.
     address: ScratchpadAddress,
-    /// Data encoding: custom apps using scratchpad should use this so they can identify the type of data they are storing
+    /// Data encoding: custom apps using scratchpad should use this so they can identify the type of data they are storing in the bytes
     data_encoding: u64,
     /// Contained data. This should be encrypted
     #[debug(skip)]
     encrypted_data: Bytes,
     /// Monotonically increasing counter to track the number of times this has been updated.
     counter: u64,
-    /// Signature over `Vec<counter>`.extend(Xorname::from_content(encrypted_data).to_vec()) from the owning key.
-    /// Required for scratchpad to be valid.
-    signature: Option<Signature>,
+    /// Signature over the above fields
+    signature: Signature,
 }
 
 impl Scratchpad {
-    /// Creates a new instance of `Scratchpad`.
-    pub fn new(owner: PublicKey, data_encoding: u64) -> Self {
+    /// Creates a new instance of `Scratchpad`. Encrypts the data, and signs all the elements.
+    pub fn new(
+        owner: &SecretKey,
+        data_encoding: u64,
+        unencrypted_data: &Bytes,
+        counter: u64,
+    ) -> Self {
+        let pk = owner.public_key();
+        let encrypted_data = Bytes::from(pk.encrypt(unencrypted_data).to_bytes());
+        let addr = ScratchpadAddress::new(pk);
+        let signature = owner.sign(Self::bytes_for_signature(
+            addr,
+            data_encoding,
+            &encrypted_data,
+            counter,
+        ));
+        let s = Self {
+            address: addr,
+            encrypted_data,
+            data_encoding,
+            counter,
+            signature,
+        };
+        debug_assert!(s.is_valid(), "Must be valid after being signed. This is a bug, please report it by opening an issue on our github");
+        s
+    }
+
+    /// Create a new Scratchpad without provding the secret key
+    /// It is the caller's responsibility to ensure the signature is valid (signs [`Scratchpad::bytes_for_signature`]) and the data is encrypted
+    /// It is recommended to use the [`Scratchpad::new`] method instead when possible
+    pub fn new_with_signature(
+        owner: PublicKey,
+        data_encoding: u64,
+        encrypted_data: Bytes,
+        counter: u64,
+        signature: Signature,
+    ) -> Self {
         Self {
             address: ScratchpadAddress::new(owner),
-            encrypted_data: Bytes::new(),
+            encrypted_data,
             data_encoding,
-            counter: 0,
-            signature: None,
+            counter,
+            signature,
         }
     }
 
-    /// Return the current count
-    pub fn count(&self) -> u64 {
+    /// Returns the bytes to sign for the signature
+    pub fn bytes_for_signature(
+        address: ScratchpadAddress,
+        data_encoding: u64,
+        encrypted_data: &Bytes,
+        counter: u64,
+    ) -> Vec<u8> {
+        let mut bytes_to_sign = data_encoding.to_be_bytes().to_vec();
+        bytes_to_sign.extend(address.to_hex().as_bytes());
+        bytes_to_sign.extend(counter.to_be_bytes().to_vec());
+        bytes_to_sign.extend(encrypted_data.to_vec());
+        bytes_to_sign
+    }
+
+    /// Get the counter of the Scratchpad, the higher the counter, the more recent the Scratchpad is
+    /// Similarly to counter CRDTs only the latest version (highest counter) of the Scratchpad is kept on the network
+    pub fn counter(&self) -> u64 {
         self.counter
     }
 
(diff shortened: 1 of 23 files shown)


autonomi/src/client/data_types/pointer.rs
```diff
@@ -96,10 +101,10 @@ impl Client {
         debug!("Paying for pointer at address: {address:?}");
         let (payment_proofs, _skipped_payments) = self
             // TODO: define Pointer default size for pricing
-            .pay(
-                DataTypes::Pointer.get_index(),
-                std::iter::once((xor_name, 128)),
-                wallet,
+            .pay_for_content_addrs(
+                DataTypes::Pointer,
+                std::iter::once((xor_name, size_of::<Pointer>())),
```
b-zee: `size_of::<Pointer>()` is the in-memory (Rust) size. Is that the same as the size of the data we store (serialized)? Or doesn't this need to be that accurately calculated/specified?

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -25,30 +25,79 @@ pub struct Scratchpad {
     address: ScratchpadAddress,
     /// Data encoding: custom apps using scratchpad should use this so they can identify the type of data they are storing
     data_encoding: u64,
-    /// Contained data. This should be encrypted
+    /// Encrypted data stored in the scratchpad, it is encrypted automatically by the [`Scratchpad::new`] and [`Scratchpad::update_and_sign`] methods
```
maqi: any mut op (even create) will be signed, hence using consistent name of `update` shall be enough?

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -25,30 +25,79 @@ pub struct Scratchpad {
...
     #[debug(skip)]
     encrypted_data: Bytes,
-    /// Monotonically increasing counter to track the number of times this has been updated.
+    /// Monotonically increasing counter to track the number of times this has been updated. When pushed to the network, the scratchpad with the highest counter is kept.
```
maqi: better to have a new line for the new sentence (80/100 chars per line)

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -25,30 +25,79 @@ pub struct Scratchpad {
...
+            signature,
+        };
+        debug_assert!(s.is_valid(), "Must be valid after being signed. This is a bug, please report it by opening an issue on our github");
+        s
+    }
+
+    /// Create a new Scratchpad without provding the secret key
+    /// It is the caller's responsibility to ensure the signature is valid (signs [`Scratchpad::bytes_for_signature`]) and the data is encrypted
+    /// It is recommended to use the [`Scratchpad::new`] method instead when possible
+    pub fn new_with_signature(
```
maqi, line 70: to make the public API cleaner, this function shall not be exposed ?
also, this function means input_data is `not` encrypted, which is un-consistent with `new` function and claims made in other places.
also, this one doesn't call the `is_valid` self check ?

I'd suggest:
  * not use this in the public API
  * make `new` function use this (if retained as a private helper)

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -57,43 +106,32 @@ impl Scratchpad {
         self.data_encoding
     }
 
-    /// Increments the counter value.
-    pub fn increment(&mut self) -> u64 {
+    /// Updates the content, re-signs the scratchpad
```
maqi: better expand `Updates the content` to more detailed work done:
* increment the counter
* encrypt the input content and replace the existing encrypted data
* sign the new fields and replace the existing signature

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -112,170 +126,209 @@ impl Client {
...
-        let scratch_address = scratchpad.network_address();
-        let address = *scratchpad.address();
-        let scratch_key = scratch_address.to_record_key();
+    ) -> Result<(AttoTokens, ScratchpadAddress), ScratchpadError> {
+        let address = scratchpad.address();
 
         // pay for the scratchpad
-        let (receipt, _skipped_payments) = self
+        let xor_name = address.xorname();
+        let size = size_of::<Scratchpad>() + scratchpad.payload_size();
```
maqi: better make this as a `utility function` within the protocol ScratchPad definition mod.
to avoid future mismatched usage, for example, other places take `scratchpad.bytes_for_sign().len()` as the `size of a scratchpad instance`

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -112,170 +126,209 @@ impl Client {
...
+        if size > MAX_SCRATCHPAD_SIZE {
```
maqi: make this as part of the `ScratchPad::is_valid()`, so that it can always be checked during create and update ?

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -25,30 +25,79 @@ pub struct Scratchpad {
...
+            counter,
+        ));
+        let s = Self {
+            address: addr,
+            encrypted_data,
+            data_encoding,
+            counter,
+            signature,
+        };
+        debug_assert!(s.is_valid(), "Must be valid after being signed. This is a bug, please report it by opening an issue on our github");
```
maqi: not recommended to use `assert` in the production code.
better make this returns an `error` to report back detailed failures

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -112,170 +126,209 @@ impl Client {
...
-        let scratch_xor = scratch.address().xorname();
+        let scratch_xor = ScratchpadAddress::new(*owner).xorname();
 
-        // TODO: define default size of Scratchpad
         let store_quote = self
             .get_store_quotes(
-                DataTypes::Scratchpad.get_index(),
-                std::iter::once((scratch_xor, 256)),
+                DataTypes::Scratchpad,
+                std::iter::once((scratch_xor, size_of::<Scratchpad>())),
```
maqi: better make this as a const

autonomi/src/client/data_types/pointer.rs
grumbach: That's an excellent question, the Pointer size is fixed though so this field is useless for pointers, I was in favour of completely removing it

ant-protocol/src/storage/scratchpad.rs
grumbach: Yeah make sense

ant-protocol/src/storage/scratchpad.rs
grumbach: I was thinking of keeping this for folks that cannot provide us with the Secret key but can sign messages/encrypt. This is typically the case with hardware wallets, where one cannot get the secret key out, but can perfectly sign messages.

ant-protocol/src/storage/scratchpad.rs
grumbach: Agreed!

autonomi/src/client/data_types/scratchpad.rs
grumbach: Make sense

autonomi/src/client/data_types/scratchpad.rs
grumbach: 100%

ant-protocol/src/storage/scratchpad.rs
grumbach: This is a debug assertion so it is not included in production code (--release mode).
I could remove it though, it's not that useful

autonomi/src/client/data_types/pointer.rs
grumbach: For now, made it a universally accessible information: `Pointer::size()` that we can decide to change down the line.

ant-protocol/src/storage/scratchpad.rs
maqi: even we accept such input, the signature and size shall still be checked.
i.e. the `verify` function shall be called

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -25,30 +25,81 @@ pub struct Scratchpad {
...
+            counter,
+        ));
+        Self {
+            address: addr,
+            encrypted_data,
+            data_encoding,
+            counter,
+            signature,
+        }
+    }
```
maqi, line 65: shall the `self.verify()` function still to be called ?

ant-protocol/src/storage/scratchpad.rs
```diff
@@ -57,43 +108,32 @@ impl Scratchpad {
...
-        if let Some(signature) = &self.signature {
-            let mut signing_bytes = self.counter.to_be_bytes().to_vec();
-            signing_bytes.extend(self.encrypted_data_hash().to_vec()); // add the count
-
-            self.owner().verify(signature, &signing_bytes)
-        } else {
-            false
-        }
+    /// Verifies that the Scratchpad signature is valid
+    pub fn verify(&self) -> bool {
```
maqi, line 129: shall the size to be checked as well?
same to the other data_types

Also, I notice there is `scratchpad_verify` which checks both signature and size for ScratchPad
I think normally, the `verify` indicates both, `verify_signature` and `verify_size (or is_size_too_big)` verifies individual part.
I'd suggest update the names/function calls of all data_types' `verify` to make them consistant.

maqi (approved):

PR #2674, opened 2025-01-27
WIP: Initial Logging Reconfiguration

jacderida:
- a4af60390 **feat: `ant` logs to stderr**

  BREAKING CHANGE: the `--log-output-dest` argument is removed. Logging to file is a bit strange for a
  short-lived program, so we now log to stderr. The user can redirect to file if they want to.

  Logging is off by default. It is switched on by using the `verbosity` argument in the standard `-v`
  to `-vvvvv` form. This increases the levels from `WARN` to `TRACE`.

  The default logging targets were changed to crates that seem a bit more relevant for using `ant`,
  and to remove particularly noisy crates, like `ant_bootstrap`. All the crates will log at the same
  level, as defined by the `verbosity` argument; previously the default logging targets used different
  levels for different crates, and those seemed a little arbitrary.

  The behaviour of `ANT_LOG` is retained, so the user can still get output for particular crates, or
  use the special `all` option to get trace-level logging for all crates.

- 65a8ead41 **chore: remove `otlp` feature**

  This was a vestige of our attempt to use OpenSearch, so we don't need it any more.

  The `opentelemetry` crate references have also been removed.

- dc60e17df **feat: switch off logging by default in `antnode`**

  Previously, `antnode` would log to file by default.

  Now logging will be enabled if:

  * The `ANT_LOG` variable has been used.
  * The `--log-output-dest` argument has been used.
  * The `-v` argument has been used.

  The verbosity argument can take different counts, e.g., `-vvv`, each count increasing the level of
  the logging from `ERROR` through to `TRACE`.

  The logs will go to `stderr` by default. If the user wants them to go to file, they can use the
  `--log-output-dest` argument and specify either the default `data-dir` or a custom path.

  This setup should be backwards compatible for anybody who has tooling setup to use
  `--log-output-dest`, but it will be a breaking change for users who are expecting logs to file by
  default.

- 243a030f1 **chore: remove the `verbosity` argument from `antctl`**

  This was an argument that was implemented in a clumsy manual way that was intended to vary the
  amount of output on the node manager. We will need to provide another `verbose` argument on the
  `local run` command to control the output of the logs for local networks.

  We still need to retain the use of it, but instead of setting it via a command line argument, it
  will just be set to `Normal`. The `node-launchpad` sets it to `Minimal` in the places where it needs
  the text output reduced.

- 162d66bd2 **feat: provide `verbosity` arg for `local run` cmd**

  This also provides the `-v` mechanism for varying the logging output for the local network. The
  value that is set here is passed to each of the node processes.

  The logging output destination is to file, otherwise on the terminal you would get a blizzard of
  output from all the node processes.

### DIFF
--- a/.github/workflows/benchmark-prs.yml
+++ b/.github/workflows/benchmark-prs.yml
@@ -47,7 +47,7 @@ jobs:
         timeout-minutes: 30
 
       - name: Start a local network
-        uses: maidsafe/ant-local-testnet-action@main
+        uses: jacderida/ant-local-testnet-action@chore-enable_logging
         env:
           ANT_LOG: "all"
         with:
@@ -71,7 +71,9 @@ jobs:
 
       - name: Start a client instance to compare memory usage
         shell: bash
-        run: ./target/release/ant --local file upload "./the-test-data.zip"
+        run: |
+          mkdir -p $CLIENT_DATA_PATH/logs
+          ./target/release/ant --local file upload "./the-test-data.zip" 2> $CLIENT_DATA_PATH/logs/ant.log
         env:
           ANT_LOG: "all"
         timeout-minutes: 5
--- a/.github/workflows/benchmark-prs.yml
+++ b/.github/workflows/benchmark-prs.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Start a client instance to compare memory usage
         shell: bash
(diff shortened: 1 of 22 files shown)


jacderida:
This is too out of date now to be useful.

PR #2680, opened 2025-01-29
feat(pricing): add records_per_type into QuotingMetrics

maqi:
### Description

add records_per_type into QuotingMetrics

### DIFF
--- a/ant-networking/src/record_store.rs
+++ b/ant-networking/src/record_store.rs
@@ -19,7 +19,7 @@ use aes_gcm_siv::{
 use ant_evm::{QuotingMetrics, U256};
 use ant_protocol::{
     convert_distance_to_u256,
-    storage::{RecordHeader, RecordKind, ValidationType},
+    storage::{DataTypes, RecordHeader, RecordKind, ValidationType},
     NetworkAddress, PrettyPrintRecordKey,
 };
 use hkdf::Hkdf;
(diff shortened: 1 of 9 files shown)


ant-networking/src/record_store.rs
```diff
@@ -138,7 +138,7 @@ pub struct NodeRecordStore {
     /// The configuration of the store.
     config: NodeRecordStoreConfig,
     /// Main records store remains unchanged for compatibility
-    records: HashMap<Key, (NetworkAddress, ValidationType)>,
+    records: HashMap<Key, (NetworkAddress, ValidationType, DataTypes)>,
```
RolandSherwin, line 141: Would it be possible to ditch `ValidationType` and instead use the `DataTypes` to derive it instead? Just a nit pick tbh, will make things a bit cleaner and can be a follow up if sounds reasonable to you..

maqi: `ValidationType` also need to hold the `content_hash` for non-chunk data.
Meanwhile, DataTypes are `public API` which supposed to keep as simple as possible.
So, don't think `merge them` is good to the public user.

RolandSherwin: Ah sorry, overlooked that! Thanks!

RolandSherwin (approved):

PR #2681, opened 2025-01-30
Registers on top of graph entry

grumbach:
By popular demand, `Registers` are back, but this time built on top of `GraphEntry`. Which:
- removes the update limitation (previously limited to 1024 updates)
- simplifies the underlying structure
- register value is a 32 bytes entry
- underlying structure looks like a doubly linked list so history traversal is simpler
- comes with a default way to access history in the API (although users can also manually access it through the underlying graph)
- pay for each update, although pretty cheap

How they work:
```
Pointer to register Head:                              [points to here]
                                                                    |
                                                                    V
GraphEntries: [register root] <-> [value2] <-> [value3] <-> [latest value]
```

> Usage examples in `autonomi/tests/registers.rs`

CLI integration coming soon! 

cc: @happybeing hope you like them!

### DIFF
--- a/ant-protocol/src/storage/graph.rs
+++ b/ant-protocol/src/storage/graph.rs
@@ -17,34 +17,49 @@ pub use bls::{PublicKey, Signature};
 pub type GraphContent = [u8; 32];
 
 /// A generic GraphEntry on the Network
+/// Graph entries are stored at the owner's public key. Note that there can only be one graph entry per owner.
+/// Graph entries can be linked to other graph entries as parents or descendants.
+/// Applications are free to define the meaning of these links, those are not enforced by the protocol.
+/// The protocol only ensures that the graph entry is immutable once uploaded and that the signature is valid and matches the owner.
+/// For convenience it is advised to make use of BLS key derivation to create multiple graph entries from a single key.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash, Ord, PartialOrd)]
 pub struct GraphEntry {
+    /// The owner of the graph. Note that graph entries are stored at the owner's public key
     pub owner: PublicKey,
+    /// Other graph entries that this graph entry refers to as parents
     pub parents: Vec<PublicKey>,
+    /// The content of the graph entry
     pub content: GraphContent,
-    pub outputs: Vec<(PublicKey, GraphContent)>,
+    /// Other graph entries that this graph entry refers to as descendants/outputs along with some data associated to each one
+    pub descendants: Vec<(PublicKey, GraphContent)>,
     /// signs the above 4 fields with the owners key
     pub signature: Signature,
 }
 
 impl GraphEntry {
-    /// Maximum size of a graph entry
-    pub const MAX_SIZE: usize = 1024;
+    /// Maximum size of a graph entry: 100KB
+    pub const MAX_SIZE: usize = 100 * 1024;
 
     /// Create a new graph entry, signing it with the provided secret key.
     pub fn new(
-        owner: PublicKey,
+        owner: &SecretKey,
         parents: Vec<PublicKey>,
         content: GraphContent,
-        outputs: Vec<(PublicKey, GraphContent)>,
-        signing_key: &SecretKey,
+        descendants: Vec<(PublicKey, GraphContent)>,
     ) -> Self {
-        let signature = signing_key.sign(Self::bytes_to_sign(&owner, &parents, &content, &outputs));
+        let key = owner;
+        let owner = key.public_key();
+        let signature = key.sign(Self::bytes_to_sign(
+            &owner,
+            &parents,
+            &content,
+            &descendants,
+        ));
         Self {
             owner,
             parents,
             content,
-            outputs,
+            descendants,
             signature,
         }
     }
@@ -54,14 +69,14 @@ impl GraphEntry {
         owner: PublicKey,
         parents: Vec<PublicKey>,
         content: GraphContent,
-        outputs: Vec<(PublicKey, GraphContent)>,
+        descendants: Vec<(PublicKey, GraphContent)>,
         signature: Signature,
     ) -> Self {
         Self {
             owner,
             parents,
             content,
-            outputs,
+            descendants,
             signature,
         }
     }
@@ -71,7 +86,7 @@ impl GraphEntry {
         owner: &PublicKey,
         parents: &[PublicKey],
         content: &[u8],
-        outputs: &[(PublicKey, GraphContent)],
+        descendants: &[(PublicKey, GraphContent)],
     ) -> Vec<u8> {
         let mut bytes = Vec::new();
         bytes.extend_from_slice(&owner.to_bytes());
@@ -85,9 +100,9 @@ impl GraphEntry {
         );
         bytes.extend_from_slice("content".as_bytes());
         bytes.extend_from_slice(content);
-        bytes.extend_from_slice("outputs".as_bytes());
+        bytes.extend_from_slice("descendants".as_bytes());
         bytes.extend_from_slice(
-            &outputs
+            &descendants
                 .iter()
                 .flat_map(|(p, c)| [&p.to_bytes(), c.as_slice()].concat())
                 .collect::<Vec<_>>(),
@@ -101,11 +116,11 @@ impl GraphEntry {
 
     /// Get the bytes that the signature is calculated from.
     pub fn bytes_for_signature(&self) -> Vec<u8> {
-        Self::bytes_to_sign(&self.owner, &self.parents, &self.content, &self.outputs)
+        Self::bytes_to_sign(&self.owner, &self.parents, &self.content, &self.descendants)
     }
 
     /// Verify the signature of the graph entry
-    pub fn verify(&self) -> bool {
+    pub fn verify_signature(&self) -> bool {
         self.owner
             .verify(&self.signature, self.bytes_for_signature())
     }
@@ -114,7 +129,7 @@ impl GraphEntry {
     pub fn size(&self) -> usize {
         size_of::<GraphEntry>()
             + self
-                .outputs
+                .descendants
                 .iter()
                 .map(|(p, c)| p.to_bytes().len() + c.len())
                 .sum::<usize>()
--- a/ant-protocol/src/storage/pointer.rs
+++ b/ant-protocol/src/storage/pointer.rs
@@ -7,29 +7,13 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::storage::{ChunkAddress, GraphEntryAddress, PointerAddress, ScratchpadAddress};
-use bls::{Error as BlsError, PublicKey, SecretKey, Signature};
-use hex::FromHexError;
+use bls::{PublicKey, SecretKey, Signature};
 use serde::{Deserialize, Serialize};
-use thiserror::Error;
 use xor_name::XorName;
-
-#[derive(Error, Debug)]
-pub enum PointerError {
-    #[error("Failed to decode hex string: {0}")]
-    HexDecoding(#[from] FromHexError),
-    #[error("Failed to create public key: {0}")]
-    BlsError(#[from] BlsError),
-    #[error("Invalid public key bytes length")]
-    InvalidPublicKeyLength,
-    #[error("Invalid signature")]
-    InvalidSignature,
-    #[error("Serialization error: {0}")]
-    SerializationError(String),
-}
 
 /// Pointer, a mutable address pointing to other data on the Network
 /// It is stored at the owner's public key and can only be updated by the owner
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Ord, PartialOrd)]
 pub struct Pointer {
     owner: PublicKey,
     counter: u32,
@@ -37,7 +21,7 @@ pub struct Pointer {
     signature: Signature,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Ord, PartialOrd)]
 pub enum PointerTarget {
     ChunkAddress(ChunkAddress),
     GraphEntryAddress(GraphEntryAddress),
@@ -107,6 +91,11 @@ impl Pointer {
         PointerAddress::from_owner(self.owner)
     }
 
+    /// Get the target of the pointer
+    pub fn target(&self) -> &PointerTarget {
+        &self.target
+    }
+
     /// Get the bytes that were signed for this pointer
     pub fn bytes_for_signature(&self) -> Vec<u8> {
         Self::bytes_to_sign(&self.owner, self.counter, &self.target)
@@ -128,7 +117,7 @@ impl Pointer {
     }
 
     /// Verifies if the pointer has a valid signature
-    pub fn verify(&self) -> bool {
+    pub fn verify_signature(&self) -> bool {
         let bytes = self.bytes_for_signature();
         self.owner.verify(&self.signature, &bytes)
     }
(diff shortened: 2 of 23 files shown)


ant-protocol/src/storage/pointer.rs
```diff
@@ -7,37 +7,21 @@
...
-    HexDecoding(#[from] FromHexError),
-    #[error("Failed to create public key: {0}")]
-    BlsError(#[from] BlsError),
-    #[error("Invalid public key bytes length")]
-    InvalidPublicKeyLength,
-    #[error("Invalid signature")]
-    InvalidSignature,
-    #[error("Serialization error: {0}")]
-    SerializationError(String),
-}
```
grumbach, line 28: Remove Protocol level errors, all errors are API-side so code changes don't affect nodes

ant-protocol/src/storage/graph.rs
```diff
@@ -17,34 +17,49 @@ pub use bls::{PublicKey, Signature};
...
+    pub descendants: Vec<(PublicKey, GraphContent)>,
     /// signs the above 4 fields with the owners key
     pub signature: Signature,
 }
 
 impl GraphEntry {
-    /// Maximum size of a graph entry
-    pub const MAX_SIZE: usize = 1024;
+    /// Maximum size of a graph entry: 100KB
+    pub const MAX_SIZE: usize = 100 * 1024;
```
maqi, line 41: this is a huge increase might affects the pricing.
may need to notify the pricing guys as this will affect the cost charging.

meanwhile, the main affectors of a graph_entry's size is `parents` and `number of descandents` .
is it really necessary to increase the `max_size` that much ?

autonomi/src/client/high_level/mod.rs
```diff
@@ -9,3 +9,21 @@
...
 pub mod files;
 pub mod vault;
+
+/// Registers are a mutable piece of data on the Network.
+/// They can be read by anyone and updated only by the register owner.
+/// Each entry is signed by the owner and all value history is kept on the Network.
+/// They can be accessed on the Network using the RegisterAddress which is effectively the hash of the owner's [`crate::PublicKey`].
+/// This means there can only be one Register per key.
+///
+/// The underlying structure of registers is a graph, where each version is a new [`crate::GraphEntry`]
```
maqi, line 19: it may worth to mention that: as Pointer and GraphEntry are both stored at the address of it's owner's public_key.
It's suggested to use the bls derivation infrastructure, to make the Register's address being root key's public_key, meanwhile the underlying Pointer and GraphEntry(s) are all using the drived keys from the root.

autonomi/src/client/high_level/register/history.rs
```diff
@@ -0,0 +1,74 @@
...
+        Self {
+            client,
+            register_owner,
+            current: root,
+        }
+    }
+
+    /// Fetch and go to the next register value from the history
+    /// Returns `Ok(None)` when we reached the end
+    pub async fn next(&mut self) -> Result<Option<RegisterValue>, RegisterError> {
```
maqi, line 37: according to the `Register underline structure` as illustrated,
the `self.current` shall always be the `Pointer` ?
to make RegisterHistory achieve the `iterator` like operates, it shall have somthing like:
``` rust
pub struct RegisterHistory {
    client: Client,
    register_owner: PublicKey,
    pointer: PointerAddress,     // tail of the register update history
    head: GraphEntryAddress,     // head of the register update history
    current: GraphEntryAddress,  // current iterator position
}
```

autonomi/src/client/high_level/register/history.rs
```diff
@@ -0,0 +1,74 @@
...
+        };
+        let next_entry_pk: PublicKey = MainPubkey::from(self.register_owner)
+            .derive_key(&next_derivation)
+            .into();
+        self.current = GraphEntryAddress::from_owner(next_entry_pk);
+        Ok(Some(entry.content))
+    }
+
+    /// Get all the register values from the history
+    pub async fn collect(&mut self) -> Result<Vec<RegisterValue>, RegisterError> {
```
maqi, line 57: this is based on the assumption that `self.current` is currently pointing to the `head` of the update history, which may not be the case.

Hence, suggested to refactor the struct, as proposed previously.

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /// Get the owner of the register
+    pub fn owner(&self) -> PublicKey {
+        self.owner
+    }
+
+    /// To underlying graph representation
+    pub fn to_underlying_graph_root(&self) -> GraphEntryAddress {
+        GraphEntryAddress::from_owner(self.owner)
```
maqi, line 51: As GraphEntry addresses keeps `changing`, but Pointer address remains the same.
Will it better to :
* make the PointerAddress using the root key
* make the head GraphEntry using the root_key.derive(`0`)

autonomi/src/client/high_level/register/history.rs
```diff
@@ -0,0 +1,74 @@
...
+            Err(e) => return Err(e),
+        };
+        let next_entry_pk: PublicKey = MainPubkey::from(self.register_owner)
+            .derive_key(&next_derivation)
+            .into();
+        self.current = GraphEntryAddress::from_owner(next_entry_pk);
+        Ok(Some(entry.content))
+    }
+
+    /// Get all the register values from the history
```
maqi: may need to note that the returned Vec is ordered by the `update order` ?

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+    CannotUpdateNewRegister,
+    #[error(
+        "Invalid register value length: {0}, expected something within {REGISTER_VALUE_SIZE} bytes"
+    )]
+    InvalidRegisterValueLength(usize),
+}
+
+/// Hard coded derivation index for the register head pointer
+/// Derive the register's main public key by it to get the pointer owner/address
+const REGISTER_HEAD_DERIVATION_INDEX: [u8; 32] = [0; 32];
```
maqi, line 85: as `query a register's latest value` shall be the most frequent op, shall the `PointerAddress` to use the root_key.public
meanwhile make `head of GraphEntry` to use this `default all 0 derivation index` ?

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+        if bytes.len() > REGISTER_VALUE_SIZE {
+            return Err(RegisterError::InvalidRegisterValueLength(bytes.len()));
+        }
+        let mut content: RegisterValue = [0; REGISTER_VALUE_SIZE];
+        content[..bytes.len()].copy_from_slice(bytes);
+        Ok(content)
+    }
+
+    /// Create a new register with an initial value
+    /// Note that two payments are required, one for the underlying [`GraphEntry`] and one for the [`crate::Pointer`]
```
maqi, line 109: also one more thing worth to be noted is that:
the input `owner` shall be the key got from `register_key_from_name`.

if pass a client's key in, that means that client can only create one Register for self.

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+        payment_option: PaymentOption,
+    ) -> Result<(AttoTokens, RegisterAddress), RegisterError> {
+        let main_key = MainSecretKey::new(owner.clone());
+        let public_key = main_key.public_key();
+
+        // create the first entry and decide on the next key
+        let index = DerivationIndex::random(&mut rand::thread_rng());
+        let next_key = main_key.public_key().derive_key(&index);
+        let parents = vec![];
+        let descendants = vec![(next_key.into(), index.into_bytes())];
```
maqi, line 123: To allow a Register restore from fork or other issue, and can keeps growing.
It may worth to allow Pointer to acutally pointing to a `new GraphEntry` link ?
i.e. the head of the `new GraphEntry link` is not derived from 0, but from other ?

Another argument to suppot the suggestion of make `Pointer` using the main root.public .

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+        };
+
+        // get the next derivation index from the current head entry
+        debug!("Getting register head graph entry at {graph_entry_addr:?}");
+        let (parent_entry, new_derivation) = self
+            .register_get_graph_entry_and_next_derivation_index(graph_entry_addr)
+            .await?;
+
+        // create a new entry with the new value
+        let main_key = MainSecretKey::new(owner.clone());
```
maqi, line 186: shall not need to `clone` a secret_key ? 
just call derive below shall be ok ?
if not, it may indicate something wrong ?

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+        let descendants = vec![(next_pk.into(), next_derivation.into_bytes())];
+        let new_entry = GraphEntry::new(&new_key.into(), parents, new_value, descendants);
+
+        // put the new entry in the graph
+        let (cost, new_graph_entry_addr) = self.graph_entry_put(new_entry, payment_option).await?;
+
+        // update the pointer to point to the new entry
+        let target = PointerTarget::GraphEntryAddress(new_graph_entry_addr);
+        let pointer_key = self.register_head_pointer_sk(&main_key.into());
+        self.pointer_update(&pointer_key, target).await?;
```
maqi, line 215: as the graph_entry has been updated, the `failure to update Pointer` shall be handled carefully to avoid mis-pointing

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+        // get the content of the entry
+        let content = entry.content;
+        Ok(content)
+    }
+
+    /// Get the cost of a register operation.
+    /// Returns the cost of creation if it doesn't exist, else returns the cost of an update
+    pub async fn register_cost(&self, owner: &PublicKey) -> Result<AttoTokens, CostError> {
+        let pointer_pk = self.register_head_pointer_pk(&RegisterAddress { owner: *owner });
+        let scratchpad_cost = self.scratchpad_cost(owner);
```
maqi: why scratch_pad ?

autonomi/src/client/high_level/register/mod.rs
```diff
@@ -0,0 +1,352 @@
...
+                entry_by_smallest_derivation
+            }
+            Err(err) => return Err(err.into()),
+        };
+        let new_derivation =
+            entry
+                .descendants
+                .iter()
+                .map(|d| d.1)
+                .min()
```
maqi: For Register, if have multiple descendants, shall raise error instead ?

maqi (changes requested):

ant-protocol/src/storage/graph.rs
grumbach: The previous was mistaken, the size we agreed on was this one

autonomi/src/client/high_level/mod.rs
grumbach: That is the case. The root is the first graph entry, from which one can access the other graph entries as it is a chained list of graph entries at different addresses (also different keys, all derived from the root one). 
The pointer is simply the root key derived by 0.

autonomi/src/client/high_level/register/history.rs
grumbach: Indeed maybe current is a bad name. Wanted to say the current in therms of iteration. Every time we call next, current goes to next. 
But agree that we might need head information etc

grumbach: Although note that it can all be derived from the owner pubkey

autonomi/src/client/high_level/register/history.rs
grumbach: Make sense, agreed, let's make collect take all from the head instead of current

autonomi/src/client/high_level/register/mod.rs
grumbach: So technically graphentries never change (immutable). We create new ones with the new versions, and the pointer points to the latest version.

grumbach: I would think keeping the root graphentry as the default makes the underlying graph structure accessible more easily to folk wanting to see the history manually. The pointer itself would better be hidden as it is a mutable type (so dangerous)

autonomi/src/client/high_level/register/history.rs
grumbach: Agreed

autonomi/src/client/high_level/register/mod.rs
grumbach: Derivation is very cheap, what is slow is getting both the pointer and its pointed content (graphentry). 
This would not be an optimization really

autonomi/src/client/high_level/register/mod.rs
grumbach: This fn derives a new key from the owner, meaning that the owner has as many registers as names the owner can invent!

grumbach: So the owner key is never returned by this fn. A new key for each name, like the previous registers!

autonomi/src/client/high_level/register/mod.rs
grumbach: The pointer does point to a new graphentry created at every update. 
Although the root is the first element in the chain (so we can remember the head) the pointer only knows of the last entry (so we have quick access to the tail).

autonomi/src/client/high_level/register/mod.rs
grumbach: MainSecretKey requires ownership of the key unfortunately

autonomi/src/client/high_level/register/mod.rs
grumbach: Agreed, I will need to think of what happens in that case

autonomi/src/client/high_level/register/mod.rs
grumbach: That's a mistake! Nice find!!

autonomi/src/client/high_level/register/mod.rs
grumbach: Yeah maybe that would be good indeed. Agreed I'll make an error here

autonomi/src/client/high_level/register/mod.rs
grumbach: Ok added a fix for that, in update, if we realize we aren't at the HEAD, update the pointer so we 'heal' that problem!

autonomi/src/client/high_level/register/mod.rs
grumbach: We could hide the clone inside the constructor, but that would be even worse

maqi: then we shall derive from the passed in owner directly, instead of `construct a MainSecretKey from the owner` ?
or, ask a `MainSecretKey` to be passed in instead ?

autonomi/src/client/high_level/mod.rs
maqi: I just mean add as comment   : )

maqi (approved):

autonomi/src/client/high_level/register/mod.rs
happybeing: @grumbach Your Register implementation is very helpful in understanding GraphEntry and how to use it for this so I'm making notes of how to update my History<T> which currently is just one entry using Pointer directly to the data, so not really a History yet.

It looks fairly straightforward, but I have a question. Since each GraphEntry references its descendents before they are created, their addresses are computed in advance.

Would it be feasible for me to compute the address of each GraphEntry based on its position in the linked list (ie using 0, 1, 2... to create the DerivationIndex). Something like:
```
let index_bytes: [0u8; 32] = some_function(pointer.counter());
let next_derivation = DerivationIndex::from_bytes(index_bytes);
let next_pk = main_key.public_key().derive_key(&next_derivation);
let descendants = vec![(next_pk.into(), next_derivation.into_bytes())];
let new_entry = GraphEntry::new(&new_key.into(), parents, new_value, descendants);
```
If that's possible, it would make it fast to access any entry in a very large history without having to iterate the graph.

EDIT: I think this would not work for a public history because it requires access to the main_key, which is a SecretKey. Presumably this is checked on the network in some way, so only the owner of the History could compute GE addresses in this way!

PR #2682, opened 2025-01-30
refactor: fix Python bindings to use async

b-zee:
The Python bindings are mostly constructing temporary async runtimes, but that doesn't work most of the times. This PR uses a special helper crate to aide in correctly using async Rust from Python.


autonomi/python/examples/autonomi_example.py
```diff
@@ -9,23 +10,18 @@ def main():
     print(f"Wallet balance: {wallet.balance()}")
 
     # Connect to the network
-    # These should be valid multiaddresses of network nodes
-    peers = [
-        "/ip4/127.0.0.1/tcp/12000",
-        "/ip4/127.0.0.1/tcp/12001"
-    ]
-    client = Client.connect(peers)
+    client = await Client.init()
```
dirvine, line 13: Nice, this is much nicer

autonomi/src/python.rs
```diff
@@ -307,23 +346,21 @@ impl PyPointerTarget {
...
-            ))),
+            inner: PointerTarget::ChunkAddress(ChunkAddress::new(XorName::from_content(xorname))),
         })
     }
 
     #[staticmethod]
     fn from_chunk_address(addr: &PyChunkAddress) -> Self {
         Self {
-            inner: RustPointerTarget::ChunkAddress(addr.inner),
+            inner: PointerTarget::ChunkAddress(addr.inner),
```
dirvine, line 363: Nice work on removing the rust name there. Much cleaner

dirvine (approved):

PR #2683, opened 2025-01-30
chore: upgrade libp2p to 0.55.0

jacderida:
- 04c8a6f62 **chore: upgrade libp2p to 0.55.0**

  They introduced a new `connection_id` field on some of the their types and this required updating
  our references to those. For now, I opted to not use the field.

- b627ec5d0 **chore: remove u256 conversion after libp2p upgrade**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,9 +142,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-chains"
-version = "0.1.51"
+version = "0.1.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


maqi (approved):

b-zee:
@jacderida Just curious about something. This PR also updated a lot of other unrelated dependencies in the `Cargo.lock` file. I wonder if you ran `cargo update`?

We just encountered a really weird edge-case here: https://github.com/maidsafe/autonomi/pull/2778#issuecomment-2681751642

I think actually the alloy project is at fault here with semantic versioning or something, but I do think it's useful to do `cargo update` separate from updating a single dependency. Just noting, this really is a weird edge case, but I think it's worth pointing out.

PR #2685, opened 2025-01-30
chore: increase quotes fetch sample to `10`

jacderida:
This was originally #2265 and was for a hotfix which we didn't release.


maqi:
obsoleted. re-raise if required.

PR #2687, opened 2025-01-31
fix: add shutdown signal to event receiver loop

mickvandijke:


autonomi/src/client/mod.rs
```diff
@@ -245,7 +245,11 @@ impl Client {
 
         // Wait until we have added a few peers to our routing table.
         let (sender, receiver) = futures::channel::oneshot::channel();
-        ant_networking::time::spawn(handle_event_receiver(event_receiver, sender));
+        ant_networking::time::spawn(handle_event_receiver(
+            event_receiver,
+            sender,
+            shutdown_tx.subscribe(),
```
maqi, line 251: with this `shutdown_rx` expected to be used,
the `_shutdown_tx` field of the Client shall get the underscore removed and be exposed as public as well, so that it doesn't got negelected ?

if it is just for the internal usage, the `_shutdown_tx` shall have clear comment to explain it.

mickvandijke: We consider the shutdown sender being dropped as a valid shutdown signal. So if `Client` is dropped, and thus its field `_shutdown_tx` is dropped as well, a shutdown signal in the form of an error will go out automatically.

I'll add a comment to clarify that behaviour.

maqi (approved):

mickvandijke:
Closed https://github.com/maidsafe/autonomi/issues/2669

PR #2690, opened 2025-01-31
fix: not enough quotes by increasing sample size

mickvandijke:
- Increases quote fetch sample size from 7 to 10.
- Widens the amount of nodes close enough to consider as valid payees in the node check.


maqi:
obsoleted. re-raise if required.

PR #2693, opened 2025-02-02
feat(Vault): infinite capacity expandable Vault

### LINKED ISSUE #2456: Vaults are limited in size (because scratchpad is limited to record size)
This means that ultimately, user data or the amount of files a user uploads is limited to the size of the underlying scratchpad. 
We need to offer a way for users to have more either by paying for an extension or something else. Some users are bound to want to upload TB of files!

maqi:
### Description

Refactor Vault to the following infrasture:
```text
GraphEntry root [outputs1, outputs2, outputs3, outputs4, ...]
                    |         |         |         |
                    |         V         V         V
ScratchPad:         |        sp1       sp2       sp3
                    V
GraphEntry       expand [outputs1, outputs2, outputs3, outputs4, ...]
                                      |         |         |
                                      V         V         V
ScratchPad:                          sp1       sp2       sp3
```
Which makes Vault having infinite capacity, allowing to be stepped expanded on use.

Closes https://github.com/maidsafe/autonomi/issues/2456

### DIFF
--- a/ant-cli/src/commands/vault.rs
+++ b/ant-cli/src/commands/vault.rs
@@ -12,15 +12,15 @@ use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use color_eyre::Section;
 
-pub async fn cost(peers: NetworkPeers) -> Result<()> {
+pub async fn cost(peers: NetworkPeers, expected_max_size: u64) -> Result<()> {
     let client = crate::actions::connect_to_network(peers).await?;
     let vault_sk = crate::keys::get_vault_secret_key()?;
 
     println!("Getting cost to create a new vault...");
-    let total_cost = client.vault_cost(&vault_sk).await?;
+    let total_cost = client.vault_cost(&vault_sk, expected_max_size).await?;
 
     if total_cost.is_zero() {
-        println!("Vault already exists, modifying an existing vault is free");
+        println!("Vault already exists, updating an existing vault is free unless the new content exceeds the current vault's paid capacity.");
     } else {
         println!("Cost to create a new vault: {total_cost} AttoTokens");
     }
--- a/autonomi/src/client/high_level/vault/mod.rs
+++ b/autonomi/src/client/high_level/vault/mod.rs
@@ -9,16 +9,22 @@
 pub mod key;
 pub mod user_data;
 
-use ant_protocol::storage::ScratchpadAddress;
 pub use key::{derive_vault_key, VaultSecretKey};
 pub use user_data::UserData;
 
 use crate::client::data_types::scratchpad::ScratchpadError;
+use crate::client::key_derivation::{DerivationIndex, MainSecretKey};
 use crate::client::payment::PaymentOption;
 use crate::client::quote::CostError;
 use crate::client::Client;
-use ant_evm::AttoTokens;
+use crate::graph::GraphError;
+use ant_evm::{AttoTokens, U256};
+use ant_networking::{GetRecordError, NetworkError};
+use ant_protocol::storage::{
+    GraphContent, GraphEntry, GraphEntryAddress, Scratchpad, ScratchpadAddress,
+};
 use ant_protocol::Bytes;
+use bls::PublicKey;
 use std::hash::{DefaultHasher, Hash, Hasher};
 use tracing::info;
 
@@ -29,7 +35,18 @@ use tracing::info;
 /// The value 0 is reserved for tests
 pub type VaultContentType = u64;
 
-/// For custom apps using Scratchpad, this function converts an app identifier or name to a [`VaultContentType`]
+/// Defines the max size of content can be written into per ScratchPad
+const MAX_CONTENT_PER_SCRATCHPAD: usize = Scratchpad::MAX_SIZE - 1024;
+
+/// Defines the max number of Scratchpads that one GraphEntry can point to
+/// The current value is assuming GraphEntry max_size to be 100KB.
+const NUM_OF_SCRATCHPADS_PER_GRAPHENTRY: usize = 1_000;
+
+/// Hard coded derivation index for the Vault's root GraphEntry.
+/// Derive the Vault's main secret/public key by it to get the root GraphEntry owner/address
+const VAULT_HEAD_DERIVATION_INDEX: [u8; 32] = [0; 32];
+
+/// For custom apps using Vault, this function converts an app identifier or name to a [`VaultContentType`]
 pub fn app_name_to_vault_content_type<T: Hash>(s: T) -> VaultContentType {
     let mut hasher = DefaultHasher::new();
     s.hash(&mut hasher);
@@ -38,10 +55,18 @@ pub fn app_name_to_vault_content_type<T: Hash>(s: T) -> VaultContentType {
 
 #[derive(Debug, thiserror::Error)]
 pub enum VaultError {
-    #[error("Vault error: {0}")]
+    #[error("Vault Scratchpad related error: {0}")]
     Scratchpad(#[from] ScratchpadError),
+    #[error("Vault GraphEntry related error: {0}")]
+    GraphEntry(#[from] GraphError),
+    #[error("Vault Cost related error: {0}")]
+    Cost(#[from] CostError),
     #[error("Protocol: {0}")]
     Protocol(#[from] ant_protocol::Error),
+    #[error("Vault doesn't have enough graph descendants: {0}")]
+    VaultNotEnoughGraphDescendants(String),
+    #[error("Vault with empty content")]
+    VaultWithZeroContentSize,
 }
 
 impl Client {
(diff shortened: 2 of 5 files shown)


ant-cli/src/commands/vault.rs
```diff
@@ -12,15 +12,15 @@ use color_eyre::eyre::Context;
...
     let client = crate::actions::connect_to_network(peers).await?;
     let vault_sk = crate::keys::get_vault_secret_key()?;
 
     println!("Getting cost to create a new vault...");
-    let total_cost = client.vault_cost(&vault_sk).await?;
+    let total_cost = client.vault_cost(&vault_sk, expected_max_size).await?;
 
     if total_cost.is_zero() {
-        println!("Vault already exists, modifying an existing vault is free");
+        println!("Vault already exists, modifying an existing vault is partially free, unless need to expand capacity.");
```
grumbach: I'd remove the "partially". Modifying is free but expanding the size is not. Maybe update shouldn't expand automatically so we don't have bad surprises? Maybe vault upgrade should be a separate process?
Maybe we need a way to price upgrades down the line?

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +71,71 @@ impl Client {
...
+        let mut cur_graph_entry_addr = GraphEntryAddress::from_owner(public_key.into());
+        let mut decrypted_full_text = vec![];
+        let mut content_type = 0;
+        let mut has_end_reached = false;
+
+        while !has_end_reached {
+            let graph_entry = self.graph_entry_get(cur_graph_entry_addr).await?;
+
+            // The first descendant is reserved for `expand GraphEntry`.
+            cur_graph_entry_addr = GraphEntryAddress::from_owner(graph_entry.descendants[0].0);
```
grumbach: descendants[0] is a dangerous pattern. I could create a malicious graphentry to crash the client when there are no descendants. 
Id use match instead

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +71,71 @@ impl Client {
...
+
+            let scratchpad_addresses = graph_entry.descendants[1..].to_vec();
```
grumbach: Same here

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +71,71 @@ impl Client {
...
 
-        let data = pad.decrypt_data(secret_key)?;
         debug!("vault data is successfully fetched and decrypted");
-        Ok((data, pad.data_encoding()))
+        Ok((Bytes::from(decrypted_full_text.concat()), content_type))
     }
 
     /// Get the cost of creating a new vault
-    pub async fn vault_cost(&self, owner: &VaultSecretKey) -> Result<AttoTokens, CostError> {
+    /// A quick estimation of cost: m * graph_entry_cost + n * scratchpad_cost
```
grumbach: Please explain what m and n are

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +71,71 @@ impl Client {
...
+            Ok(graph_entry_cost)
+        } else {
+            let scratchpad_cost = self.scratchpad_cost(&public_key.into()).await?;
+
+            let num_of_scratchpads = max_size / MAX_CONTENT_PER_SCRATCHPAD as u64 + 1;
+            let num_of_graph_entry =
+                num_of_scratchpads / NUM_OF_SCRATCHPADS_PER_GRAPHENTRY as u64 + 1;
+
+            let total_cost = U256::from(num_of_graph_entry) * graph_entry_cost.as_atto()
+                + U256::from(num_of_scratchpads) * scratchpad_cost.as_atto();
```
grumbach, line 151: The cost of a scratchpad is not universal accross the network. It depends on the pk that owns it. Each address space might have different prices!

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +71,71 @@ impl Client {
...
+            Ok(AttoTokens::from_atto(total_cost))
+        }
     }
 
     /// Put data into the client's VaultPacket
     ///
-    /// Pays for a new VaultPacket if none yet created for the client.
-    /// Provide the bytes to be written to the vault and the content type of those bytes.
+    /// Dynamically expand the Vault's capacity with uploading un-created GraphEntry.
+    /// Un-created Scratchpad will be uploaded as well if necessary.
```
grumbach: Im not sure what you mean, think we need to reword this. Id say:
Dynamically expand the vault capacity by paying for more space when needed.

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+
+        // scratchpad_derivations ordered by the collection order
+        let (mut cur_free_graphentry_derivation, mut scratchpad_derivations) = self
+            .current_capacity(
+                &main_secret_key,
+                DerivationIndex::from_bytes(VAULT_HEAD_DERIVATION_INDEX),
+            )
+            .await?;
+
+        let content_chunks = chunk_bytes(data);
```
grumbach: Maybe chunk is confusing here as we are not dealing with Chunks

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+
+        let scratchpad_derivations = descendants.split_off(1);
+        Ok((
+            new_graphentry_derivation,
+            scratchpad_derivations,
+            graph_cost,
+        ))
+    }
+
+    // Collectes the current claimed capacity (i.e. the uploaded `GrapthEntry`s)
```
grumbach: Typo collects*

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+    // The returned value is:
+    //   * cur_free_graphentry_derivation: i.e. the root if no graph_entry uploaded,
+    //       otherwise, the first un-used one (the output[0] of the tail of the linked GraphEntry)
+    //   * scratchpad_derivations: ordered by the collection order
+    async fn current_capacity(
```
grumbach: Not sure the name says what this fn does.

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+            let public_key = main_secret_key
+                .derive_key(&cur_free_graphentry_derivation)
+                .public_key();
+            let cur_graph_entry_addr = GraphEntryAddress::from_owner(public_key.into());
+
+            match self.graph_entry_get(cur_graph_entry_addr).await {
+                Ok(entry) => {
+                    cur_free_graphentry_derivation =
+                        DerivationIndex::from_bytes(entry.descendants[0].1);
+                    scratchpad_derivations.extend(&entry.descendants[1..]);
```
grumbach, line 319: Dangerous to assume descendants is not empty

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+
+        // claim more capacity if short of.
+        // Note: as the Scratchpad is `created on use`, hence during the `claim stage`,
+        //       NUM_OF_SCRATCHPADS_PER_GRAPHENTRY to be claimed in one newly created GraphEntry.
+        while scratchpad_derivations.len() < content_chunks.len() {
+            let (new_free_graphentry_derivation, new_scratchpad_derivations, grap_cost) = self
+                .expand_capacity(
+                    &main_secret_key,
+                    &cur_free_graphentry_derivation,
+                    payment_option.clone(),
```
grumbach, line 200: Use of payment option in a loop where it could be a receipt. This code would only work if payment option was a wallet as one cannot know in advance what addresses to pay for.

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+                )
+                .await?;
+            cur_free_graphentry_derivation = new_free_graphentry_derivation;
+            scratchpad_derivations.extend(&new_scratchpad_derivations);
+            total_cost = AttoTokens::from_atto(total_cost.as_atto() + grap_cost.as_atto());
+        }
+
+        for (i, content) in content_chunks.into_iter().enumerate() {
+            let sp_secret_key = main_secret_key
+                .derive_key(&DerivationIndex::from_bytes(scratchpad_derivations[i].1));
```
grumbach, line 210: Again access with [i] is dangerous. 
Maybe we could zip it into the iteration above? 
`(i, content, dx)`

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+            match self
+                .scratchpad_update(&sp_secret_key.clone().into(), content_type, &content)
+                .await
+            {
+                Ok(()) => continue,
```
grumbach, line 215: Why update scratchpad in a loop? Can't we do it in one go?

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+                Err(ScratchpadError::CannotUpdateNewScratchpad) => {
+                    let (price, addr) = self
+                        .scratchpad_create(
+                            &sp_secret_key.into(),
+                            content_type,
+                            &content,
+                            payment_option.clone(),
```
grumbach, line 222: Again this would only work for wallet payment option

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -78,21 +144,203 @@ impl Client {
...
+            let (new_free_graphentry_derivation, new_scratchpad_derivations, grap_cost) = self
+                .expand_capacity(
+                    &main_secret_key,
+                    &cur_free_graphentry_derivation,
+                    payment_option.clone(),
+                )
+                .await?;
+            cur_free_graphentry_derivation = new_free_graphentry_derivation;
+            scratchpad_derivations.extend(&new_scratchpad_derivations);
+            total_cost = AttoTokens::from_atto(total_cost.as_atto() + grap_cost.as_atto());
```
grumbach: typo graph_cost*

grumbach (changes requested):
Great poc! Left many comments about security and payment usage. 
Good that you included some tests! Although I think Id love more tests to prove that all these loops actually work! We want to make sure Vaults still work after that!

ant-cli/src/commands/vault.rs
maqi: > Maybe we need a way to price upgrades down the line

The cost is now being dynamicaly charged during the upgrade, i.e. capacity growing along with the content being written.

autonomi/src/client/high_level/vault/mod.rs
maqi: True. Hence this is just an `estimation`, and for `expected max_size`
As the ScratchPad will only be paid and upload when have content written to, there is no point to give `exact cost` for `max_size` .

autonomi/src/client/high_level/vault/mod.rs
maqi: as the ScratchPad is now generated on need, it may means only accept Wallet to allow dynamic paying makes more sense.
The current API also doesn't have the facility for the `addresses to pay` to be collected in advance (though not a bad thing, as to keep the API simple)

Will prefer to leave the work of `make Vault support pay by receipt` as a separate follow-up PR, if necessary.

autonomi/src/client/high_level/vault/mod.rs
maqi: as mentioned in the slack thread, `parallelism for optimization` is expected to be done in a separate follow-up PR

autonomi/src/client/high_level/vault/mod.rs
maqi: same comment as at [above](https://github.com/maidsafe/autonomi/pull/2693#discussion_r1939219980)

autonomi/src/client/high_level/vault/mod.rs
maqi: there is a check_and_create above `while scratchpad_derivations.len() < contents.len() {`
hence shall be safe here to access with index.

also, `zipping` means we have to carry out extra `cloning`

grumbach: If extra cloning can avoid manual checks vs compile-time checks then I'd go for extra cloning. The rust compiler does an excellent job of ignoring `clone` when it can in release mode, I wouldn't worry about performance here.

ant-cli/src/commands/vault.rs
```diff
@@ -12,15 +12,15 @@ use color_eyre::eyre::Context;
...
     let client = crate::actions::connect_to_network(peers).await?;
     let vault_sk = crate::keys::get_vault_secret_key()?;
 
     println!("Getting cost to create a new vault...");
-    let total_cost = client.vault_cost(&vault_sk).await?;
+    let total_cost = client.vault_cost(&vault_sk, expected_max_size).await?;
 
     if total_cost.is_zero() {
-        println!("Vault already exists, modifying an existing vault is free");
+        println!("Vault already exists, updating an existing vault could be charged if new content exceeding the current vault's capacity.");
```
grumbach:
```suggestion
        println!("Vault already exists, updating an existing vault is free unless the new content exceeds the current vault's capacity.");
```

autonomi/src/client/high_level/vault/mod.rs
```diff
@@ -52,24 +77,86 @@ impl Client {
...
 
     /// Get the cost of creating a new vault
-    pub async fn vault_cost(&self, owner: &VaultSecretKey) -> Result<AttoTokens, CostError> {
+    /// A quick estimation of cost:
+    ///   num_of_graph_entry * graph_entry_cost + num_of_scratchpad * scratchpad_cost
+    pub async fn vault_cost(
+        &self,
+        owner: &VaultSecretKey,
+        max_size: u64,
+    ) -> Result<AttoTokens, VaultError> {
```
grumbach, line 130: Here please return a `CostError` if possible. If possible this will allow to remove `CostError` from the `VaultError`

autonomi/src/client/high_level/vault/mod.rs
grumbach: Agreed, follow up work. Although I wonder if we should keep Vaults auto-upgrading? Instead of returning an error for max capacity reached and offering an API specifically to upgrade? No hidden costs this way but maybe too complex? Stuff to think about...

autonomi/src/client/high_level/vault/mod.rs
grumbach: This has not been addressed, forgotten maybe?

grumbach (commented):
Good fixes, agree that most can go as follow up, just a couple comments specifically I'd love it if we removed any risky `descendants[0]`

autonomi/src/client/high_level/vault/mod.rs
maqi: Forgot to reply, there is check of `if entry.descendants.len() <= NUM_OF_SCRATCHPADS_PER_GRAPHENTRY {` just above to ensure the `descendants ` is not empty.

autonomi/src/client/high_level/vault/mod.rs
maqi: There is now `VaultError::VaultWithZeroContentSize` returned to give alert of updating with empty content.

Also, from the API perspective, having a consistent of VaultError across all public API will be better to return different (some CostError, some VaultError) ?

grumbach (approved):
Great work Qi! LGTM

PR #2694, opened 2025-02-03
Register in cli and ci

grumbach:
- CLI registers integration matching previous API known to users
- Tests to prove and verify they work in CI
- Add registers in Vault's user data so folk don't need to keep ANY extra key at all for registers

cc: @happybeing


mickvandijke (approved):
Great work! 🔥

maqi (approved):

PR #2695, opened 2025-02-03
feat: update payment vault contract interface

mickvandijke:
- Updates the payment vault contract interface in preparation of the upgrade / TGE.
- Updates the local payment vault contract for testing.
- Changes the default chain to `ArbitrumOne`.


evmlib/src/contract/payment_vault/interface.rs
```diff
@@ -44,3 +54,13 @@ impl From<QuotingMetrics> for IPaymentVault::QuotingMetrics {
...
 }
+
+fn data_type_conversion(data_type: u32) -> u8 {
+    match data_type {
+        0 => 2, // Chunk
+        1 => 0, // GraphEntry
+        2 => 3, // Pointer
+        3 => 1, // Scratchpad
+        _ => 4, // Does not exist
+    }
```
grumbach, line 65: Ouch, we already have such a conversion upstream from DataType to u32, this feels a bit awkward.

grumbach: Not blocking in any way though we can always fix that later.

grumbach (approved):
LGTM

PR #2696, opened 2025-02-03
feat(client): implement ClientOperationCfg to control any client oper…

RolandSherwin:


ant-cli/src/commands.rs
```diff
@@ -65,6 +68,8 @@ pub enum FileCmd {
         addr: String,
         /// The destination file path.
         dest_file: String,
+        /// Optionally specify the quorum for the verification of the download.
+        read_quorum: Option<ResponseQuorum>,
```
grumbach, line 72: Could we have a CI test proving this is use-able in the CLI? I have no idea how to provide this argument in the CLI.

ant-cli/src/actions/connect.rs
```diff
@@ -7,13 +7,17 @@
...
 use autonomi::{get_evm_network, Client, ClientConfig};
 use color_eyre::eyre::bail;
 use color_eyre::eyre::Result;
 use indicatif::ProgressBar;
 use std::time::Duration;
 
-pub async fn connect_to_network(peers: NetworkPeers) -> Result<Client> {
+pub async fn connect_to_network(
+    peers: NetworkPeers,
+    client_operation_config: ClientOperationConfig,
```
grumbach, line 19:
```suggestion
    client_operation_config: Option<ClientOperationConfig>,
```
Make this an option so we don't need to do the check in every single CLI function that calls this. We can then 
```
client_operation_config.unwrap_or_default()
```

ant-protocol/Cargo.toml
```diff
@@ -22,7 +22,6 @@ color-eyre = "0.6.3"
 crdts = { version = "7.3", default-features = false, features = ["merkle"] }
 custom_debug = "~0.6.1"
 dirs-next = "~2.0.0"
-exponential-backoff = "2.0.0"
```
grumbach, line 25: Very good that we removed that from the protocol layer!

autonomi/src/client/config.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub(crate) write_retry_strategy: RetryStrategy,
+    /// The number of scratchpads to wait for before considering the read after write operation successful.
+    pub(crate) verification_quorum: ResponseQuorum,
+    /// The retry strategy to use if the read after write operation fails.
+    pub(crate) verification_retry_strategy: RetryStrategy,
+    /// The number of scratchpads to wait for before considering the read operation successful.
+    pub(crate) read_quorum: ResponseQuorum,
+    /// The retry strategy to use if the read operation fails.
+    pub(crate) read_retry_strategy: RetryStrategy,
+}
```
grumbach: Those are exactly the same please make this one type. If in the future we need to make them different, we will. But for now, please let's keep this simple.

autonomi/src/client/mod.rs
```diff
@@ -256,7 +239,8 @@ impl Client {
         Ok(Self {
             network,
             client_event_sender: Arc::new(None),
-            evm_network: config.evm_network,
+            evm_network: Arc::new(config.evm_network),
+            operation_config: Arc::new(config.operation_config),
```
grumbach, line 243: Why do we need `Arc` here?

autonomi/src/lib.rs
```diff
@@ -85,6 +88,14 @@ pub use libp2p::Multiaddr;
 
 #[doc(inline)]
 pub use client::{
+    // Client Configs
+    config::ChunkOperationConfig,
+    config::ClientConfig,
+    config::ClientOperationConfig,
+    config::GraphOperationConfig,
+    config::PointerOperationConfig,
+    config::ScratchpadOperationConfig,
```
grumbach: Making this one type will make the exposed types much simpler, 6 exposed public structs related to config is huge

grumbach (commented):
Thanks for making this PR @RolandSherwin! Apart form the config types duplication LGTM!

autonomi/src/client/config.rs
RolandSherwin: Updated. But this would mean that we lose the ability to set the cfgs in one place. We reuse the configs at some places and those might be prone to errors.

autonomi/src/client/mod.rs
RolandSherwin: Every component Network is arc'ed to be freely clonable without incurring much cost, so I assume we'd want to do the same for Clients.

autonomi/src/client/config.rs
grumbach: Ok will address that in a follow up then

grumbach (approved):
✅ 🚢

PR #2700, opened 2025-02-04
feat: manual chunk API and test, improved config for client

grumbach:
- Adds a manual chunk API so all autonomi data types have a `put`/`get`/`cost` suite with the same API and the same behaviour. 
- The docs for `chunk_put` encourages the users to use the higher level `data_put` instead so users don't use the lower level manual chunk API accidentally. 
- Improve config

### DIFF
--- a/ant-cli/src/commands.rs
+++ b/ant-cli/src/commands.rs
@@ -71,6 +71,11 @@ pub enum FileCmd {
         addr: String,
         /// The destination file path.
         dest_file: String,
+        /// Experimental: Optionally specify the quorum for the download (makes sure that we have n copies for each chunks).
+        ///
+        /// Possible values are: "one", "majority", "all", n (where n is a number greater than 0)
+        #[arg(short, long)]
+        quorum: Option<ResponseQuorum>,
     },
 
     /// List previous uploads
@@ -203,9 +208,11 @@ pub async fn handle_subcommand(opt: Opt) -> Result<()> {
                 public,
                 quorum,
             } => file::upload(&file, public, peers.await?, quorum).await,
-            FileCmd::Download { addr, dest_file } => {
-                file::download(&addr, &dest_file, peers.await?).await
-            }
+            FileCmd::Download {
+                addr,
+                dest_file,
+                quorum,
+            } => file::download(&addr, &dest_file, peers.await?, quorum).await,
             FileCmd::List => file::list(),
         },
         Some(SubCmd::Register { command }) => match command {
--- a/ant-cli/src/commands.rs
+++ b/ant-cli/src/commands.rs
@@ -58,6 +58,11 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Experimental: Optionally specify the quorum for the verification of the upload.
+        ///
+        /// Possible values are: "one", "majority", "all", n (where n is a number greater than 0)
+        #[arg(short, long)]
+        quorum: Option<ResponseQuorum>,
     },
 
     /// Download a file from the given address.
@@ -66,8 +71,6 @@ pub enum FileCmd {
         addr: String,
         /// The destination file path.
         dest_file: String,
-        /// Optionally specify the quorum for the verification of the download.
-        read_quorum: Option<ResponseQuorum>,
     },
 
     /// List previous uploads
@@ -195,12 +198,14 @@ pub async fn handle_subcommand(opt: Opt) -> Result<()> {
     match cmd {
         Some(SubCmd::File { command }) => match command {
             FileCmd::Cost { file } => file::cost(&file, peers.await?).await,
-            FileCmd::Upload { file, public } => file::upload(&file, public, peers.await?).await,
-            FileCmd::Download {
-                addr,
-                dest_file,
-                read_quorum,
-            } => file::download(&addr, &dest_file, peers.await?, read_quorum).await,
+            FileCmd::Upload {
+                file,
+                public,
+                quorum,
+            } => file::upload(&file, public, peers.await?, quorum).await,
+            FileCmd::Download { addr, dest_file } => {
+                file::download(&addr, &dest_file, peers.await?).await
+            }
             FileCmd::List => file::list(),
         },
         Some(SubCmd::Register { command }) => match command {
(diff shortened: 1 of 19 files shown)


ant-protocol/src/storage/chunks.rs
```diff
@@ -25,6 +25,9 @@ pub struct Chunk {
 }
 
 impl Chunk {
+    /// The maximum size of a chunk is 1MB
+    pub const MAX_SIZE: usize = 1024 * 1024;
```
b-zee: What is the relation between this and the `self_encryption::MAX_CHUNK_SIZE`? Should they not be the same?

autonomi/src/client/data_types/chunk.rs
```diff
@@ -140,6 +145,115 @@ impl Client {
         }
     }
 
+    /// Manually upload a chunk to the network.
+    /// It is recommended to use the [`Client::data_put`] method instead to upload data.
```
b-zee, line 135: Just a nit, but if this comment line is meant as a new paragraph, insert an empty doc line (`///`) between them. Otherwise rustdoc will concatenate these sentences as a single line.

ant-protocol/src/storage/chunks.rs
grumbach: They should be equal but the alternative to this constant here is to add self_encryption as a new dependency to sn_protocol which I didn't want to

autonomi/src/client/data_types/chunk.rs
grumbach: Yeah it was just for our readability it doesn't have to be 2 different lines

ant-protocol/src/storage/chunks.rs
maqi: given our current code aiming to have max 4MB, specify a const here with different vallue is bit risky?
or, it needs to be something like: DEFAULT_MAX_SIZE ?

autonomi/tests/chunk.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

ant-protocol/src/storage/chunks.rs
b-zee: Can we make this configurable? We're currently setting chunk size to 4MiB on CI.

grumbach: 100% percent agree on both your points. Will make it `DEFAULT_MAX_SIZE` as you suggest. And in the future when we have decided on a fixed size. I hope that we have a better solution.

ant-cli/src/commands.rs
```diff
@@ -66,8 +71,6 @@ pub enum FileCmd {
         addr: String,
         /// The destination file path.
         dest_file: String,
-        /// Optionally specify the quorum for the verification of the download.
-        read_quorum: Option<ResponseQuorum>,
```
grumbach, line 70: It makes no sense to add a quorum option for download of chunks as 1 is enough and there is no verification, so removed as the example is irrelevant.

ant-cli/src/commands.rs
```diff
@@ -58,6 +58,11 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Experimental: Optionally specify the quorum for the verification of the upload.
+        ///
+        /// Possible values are: "one", "majority", "all", n (where n is a number greater than 0)
+        #[arg(short, long)]
+        quorum: Option<ResponseQuorum>,
```
grumbach, line 65: Added a better useable proof of concept of the Quorum choice here

ant-protocol/src/storage/chunks.rs
```diff
@@ -25,6 +25,9 @@ pub struct Chunk {
 }
 
 impl Chunk {
+    /// The default maximum size of a chunk is 1MB
+    pub const DEFAULT_MAX_SIZE: usize = 1024 * 1024;
```
grumbach, line 29: Made this `DEFAULT_MAX_SIZE` for now, we can/should review this once we have stabilized the size.

autonomi/src/client/config.rs
```diff
@@ -59,46 +39,154 @@ impl ClientConfig {
...
-    /// Use the `set_verification_quorum` and `set_verification_retry_strategy` to configure the verification
-    /// operation.
-    pub fn set_write_retry_strategy(&mut self, strategy: RetryStrategy) {
-        self.write_retry_strategy = Some(strategy);
+/// The default configuration for the client.
+///
+/// It is optimized for faster chunk put and get, benefiting from the chunk content addressed property.
+/// Other data types are optimized for fast verification, and resilience in case of forks, which are impossible for chunks.
+impl Default for ClientOperatingStrategy {
+    fn default() -> Self {
```
grumbach, line 69: All our strategies are here, 100% of them. @RolandSherwin suggested a centralized area for all this information, I believe it is a good idea so we know what's going on. Doing these changes I realized we set contradicting values in a lot of areas. Now this is impossible.

autonomi/src/client/mod.rs
```diff
@@ -81,13 +80,11 @@ pub use ant_protocol::CLOSE_GROUP_SIZE;
...
-    pub(crate) client_event_sender: Arc<Option<mpsc::Sender<ClientEvent>>>,
+    pub(crate) client_event_sender: Option<mpsc::Sender<ClientEvent>>,
     /// The EVM network to use for the client.
-    evm_network: Arc<EvmNetwork>,
+    evm_network: EvmNetwork,
     /// The configuration for operations on the client.
-    ///
-    /// This will be shared across all clones of the client.
-    operation_config: Arc<ClientOperationConfig>,
+    config: ClientOperatingStrategy,
```
grumbach, line 87: Removed all the `Arc`s:
- `mpsc::Sender`s are [MADE TO BE CLONED](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html), `Arc` here is probably a mistake.
- `EvmNetwork` and `ClientOperatingStrategy` are simpler/faster to access like this and since reasonably sized they will not be a burden to the `clone`s if any. As _premature optimization is the root of all evil_, I suggest we simplify this like that.

maqi (approved):

ant-cli/src/commands.rs
RolandSherwin: This is used with our performance monitoring system.

grumbach: Restored!

autonomi/src/client/config.rs
RolandSherwin: Perfect, this is much nicer!

PR #2702, opened 2025-02-04
chore(client): update Vault's scratchpads in parrallel

maqi:
### Description

update Vault's scratchpads in parrallel


grumbach (approved):
LGTM!

PR #2713, opened 2025-02-06
feat(optimization): optimize dir upload and rpc calls

mickvandijke:
Merge after a successful testnet!

- Batches payment of all files in a directory (saves gas fees & RPC calls).
- Fetches quotes for all files in one go (also stops `get_market_price` RPC calls being made in parallel, which was being a pain for the free rate limit).
- Batches `get_market_price` requests (massively reduces amount of RPC calls when needing to fetch a large amount of quotes).

### DIFF
--- a/autonomi/src/client/high_level/files/fs_private.rs
+++ b/autonomi/src/client/high_level/files/fs_private.rs
@@ -219,7 +219,7 @@ impl Client {
 
         let total_cost = self
             .process_upload_results(uploads, receipt, skipped_payments_amount)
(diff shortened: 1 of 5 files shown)


happybeing:
@mickvandijke  This is great news! Is there any way of extending this behaviour for apps, so they can queue a number of operations and batch them together?

Ideally this would be for heterogeneous operations, but homogeneous would be helpful (i.e. operations of different/same type).

jacderida (approved):

PR #2714, opened 2025-02-06
Reduce resource usage

maqi:
### Description

This PR contains couple of commits aiming to reduce the resource (CPU/traffic) usage:
* reduced some logging places
* refactored NetworkAddress output format, for a shorter logging
* Node no longer carry out fresh_replicate as client already uploads to All payees
* Reduce MIN_ACCEPTABLE_HEALTHY_SCORE according to the collected statistics from the testnet run.

### Type of Change

### DIFF
--- a/ant-node/src/node.rs
+++ b/ant-node/src/node.rs
@@ -69,7 +69,7 @@ const HIGHEST_SCORE: usize = 100;
 
 /// Any nodes bearing a score below this shall be considered as bad.
 /// Max is to be 100 * 100
-const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 5000;
+const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 3000;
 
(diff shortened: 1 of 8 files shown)


ant-node/src/node.rs
```diff
@@ -69,7 +69,7 @@ const HIGHEST_SCORE: usize = 100;
 
 /// Any nodes bearing a score below this shall be considered as bad.
 /// Max is to be 100 * 100
-const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 5000;
+const MIN_ACCEPTABLE_HEALTHY_SCORE: usize = 3000;
```
mickvandijke, line 72: Just checking if this was an intended change? Did not see it in the PR description

mickvandijke (approved):

ant-node/src/node.rs
maqi: added in the PR description.
thx for remind

PR #2717, opened 2025-02-06
fix(network): do not consider a single connection error as a critical…

RolandSherwin:


ant-networking/src/cmd.rs
```diff
@@ -1026,46 +1037,64 @@ impl SwarmDriver {
...
+                let is_connection_issue =
+                    matches!(bad_behaviour, NodeIssue::CriticalConnectionIssue);
+
+                // If it is a connection issue, then we should not block it completely, give it more chances.
+                // Since we still keep the Vec<issue> around, a new connection issue will again result in RT eviction,
+                // the node will be evicted time after time.
+                // So we'll set is_bad to false to give chance to faulty connected nodes.
+                if is_connection_issue {
+                    warn!("Peer {peer_id:?} is considered as bad due to connection issue. Will be given more chances.");
+                    *is_bad = false;
```
maqi: it's already inside the `if *is_bad {` clause ?
change `is_bad` doesn't make any sense.
Also, the following `record_metrics` and `ConsidedAsBad` block will be executed as well?
I think here we shall just `remove peer` then return ?

maqi (changes requested):

maqi (approved):

PR #2726, opened 2025-02-08
fix(network): remove steam data env var and set it to 128KB

RolandSherwin:


dirvine (approved):
LGTM Perfect @RolandSherwin nice one

RolandSherwin:
Closed as we've decided to go with the def values for now.

PR #2730, opened 2025-02-10
refactor: add `get_raw_quotes` function from `get_store_quotes`

mickvandijke:
Needed for the rewards service. Does not have to be in the RC.


grumbach (approved):
LGTM

PR #2737, opened 2025-02-11
fix(python): resolve workspace configuration for local and PyPI builds

dirvine:
- Set default maturin config to workspace=false and profile=release for PyPI builds
- Add sdist-include patterns to package workspace dependencies
- Document environment variable overrides for local development
- Fix editable install issues with workspace configuration

This change ensures the package can be installed both locally with -e flag and from PyPI while maintaining correct workspace relationships.


grumbach (approved):
LGTM!

jacderida (approved):

PR #2739, opened 2025-02-12
chore: add spawn single node example

mickvandijke:
Adds an example of how to programmatically spawn a single node and bootstrap it to the network.

### DIFF
--- a/ant-node/examples/spawn_node.rs
+++ b/ant-node/examples/spawn_node.rs
@@ -1,0 +1,35 @@
+// Copyright 2025 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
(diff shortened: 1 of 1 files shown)


RolandSherwin (approved):

ant-node/examples/spawn_node.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use ant_node::spawn::node_spawner::NodeSpawner;
+use autonomi::Multiaddr;
+use std::str::FromStr;
+
+#[tokio::main]
+async fn main() {
+    // Using the genesis node as a bootstrap node for example
+    let bootstrap_peers = vec![
+        Multiaddr::from_str("/ip4/209.97.181.193/udp/57402/quic-v1/p2p/12D3KooWHygG9a7inESky2KpvHQmbX5o2UC8D29B5njdshAcv1p6")
```
RolandSherwin: Is this the multiaddr of a node from the current PROD? Do you think it would be better to do `PeersArg::get_bootstrap_addr(None, 1)` here? It would fetch a peer from the live PROD network. So no worries of the hardcoded ip/node going down.

PR #2743, opened 2025-02-13
Reapply "feat(network): enable dcutr"

jacderida:
This re-enables DCUtR.

Do not merge this PR for now. It should target a hotfix branch, or possibly something else, when one is available.


jacderida:
We may or may not come back to using DCUtR at some later stage. Closing for now.

PR #2745, opened 2025-02-14
Enable UPnP by default in antctl & show UPnP support in Launchpad

mickvandijke:
Changes:
- Enable UPnP by default when starting nodes using `antctl`.
- Add `--no-upnp` flag to `antctl` to disable UPnP.
- Rename `antctl`'s `--home-network` flag to `--relay`.
- Fixes an argument error in `antctl` (conflicts with ..).
- Remove `Home Network` from launchpad connection modes. Relay can only be selected by using `Automatic` in case UPnP fails. If users already set it to `Home Network`, it will now switch to `Automatic` automatically. This is to avoid users from using relays in case when UPnP is also available.
- Add a UPnP supported indicator to the launchpad.
- Add a connection mode column to nodes in the launchpad that shows which mode was selected by `Automatic`.
- Nodes log their startup/runtime failure reason to a "critical_failure.log".
- Node manager shows node failure reason in status command.
- Launchpad shows node failure reason in status window.
- Nodes no longer restart on crash.

Screenshot 2025-02-21 at 17 20 54
Screenshot 2025-02-14 at 15 45 48
Screenshot 2025-02-14 at 15 54 28


node-launchpad/src/components/status.rs
```diff
@@ -798,19 +813,54 @@ impl Component for Status<'_> {
         ]);
 
         let connection_mode_string = match self.connection_mode {
-            ConnectionMode::HomeNetwork => "Home Network",
-            ConnectionMode::UPnP => "UPnP",
-            ConnectionMode::CustomPorts => &format!(
+            ConnectionMode::HomeNetwork => "Home Network".to_string(),
```
jacderida, line 816: So are we still presenting it as 'home network' to the user? Or should this also be renamed to 'relay'?

jacderida (approved):

PR #2755, opened 2025-02-17
chore(node): not fetch from network when replicate fetch failed

maqi:
### Description

not fetch from network when replicate fetch failed


maqi:
to be included in the PR https://github.com/maidsafe/autonomi/pull/2754

PR #2759, opened 2025-02-19
feat: archive and files improvements

grumbach:
- long awaited `Archive::merge`: allows to merge one archive into another
- `Metadata::empty` constructor for users who don't want to reveal any file metadata
- `ant file download` worked for `Archives` only, this PR adds support for:
    - direct file download directly from a `XorName` (compatible with `data_put_public`)
    - direct data download into a file directly from a `DataMapChunk` (compatible with `data_put`)

### DIFF
--- a/ant-cli/src/actions/download.rs
+++ b/ant-cli/src/actions/download.rs
@@ -8,10 +8,13 @@
 
 use super::get_progress_bar;
 use autonomi::{
+    chunk::DataMapChunk,
     client::{
-        address::str_to_addr, files::archive_private::PrivateArchiveAccess,
-        files::archive_public::ArchiveAddr,
+        address::str_to_addr,
+        files::{archive_private::PrivateArchiveAccess, archive_public::ArchiveAddr},
+        GetError,
     },
+    data::DataAddr,
     Client,
 };
 use color_eyre::{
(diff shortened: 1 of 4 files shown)


ant-cli/src/actions/download.rs
```diff
@@ -21,19 +24,25 @@ use color_eyre::{
...
+
+    let try_datamap = DataMapChunk::from_hex(addr).ok();
+    if let Some(datamap) = try_datamap {
+        return download_from_datamap(addr, datamap, dest_path, client).await;
+    }
+
+    Err(eyre!("Failed to parse data address {addr}"))
             .with_suggestion(|| "Public addresses look like this: 0037cfa13eae4393841cbc00c3a33cade0f98b8c1f20826e5c51f8269e7b09d7")
             .with_suggestion(|| "Private addresses look like this: 1358645341480028172")
-            .with_suggestion(|| "Try the `file list` command to get addresses you have access to"),
```
maqi, line 35: lack of suggestion of `Data_map_chunk address look like`  ?

ant-cli/src/actions/download.rs
```diff
@@ -88,10 +97,16 @@ async fn download_public(
...
     client: &Client,
 ) -> Result<()> {
-    let archive = client
-        .archive_get_public(&address)
-        .await
-        .wrap_err("Failed to fetch data from address")?;
+    let archive = match client.archive_get_public(&address).await {
+        Ok(archive) => archive,
+        Err(GetError::Deserialization(_)) => {
+            info!("Failed to deserialize Public Archive from address, trying to fetch data assuming it is a single file instead");
```
maqi: better show the address within the log

ant-cli/src/actions/download.rs
```diff
@@ -127,3 +142,51 @@ async fn download_public(
...
+    addr: &str,
+    address: DataAddr,
+    dest_path: &str,
+    client: &Client,
+) -> Result<()> {
+    let bytes = match client.data_get_public(&address).await {
+        Ok(bytes) => bytes,
+        Err(e) => {
+            let err = format!("Failed to fetch file at {addr:?}: {e}");
+            return Err(eyre!(err)).wrap_err("Failed to fetch file content from address");
```
maqi, line 157: better to include the address in the error msg

autonomi/src/client/high_level/files/archive_private.rs
```diff
@@ -129,6 +129,13 @@ impl PrivateArchive {
 
         Ok(root_serialized)
     }
+
+    /// Merge with another archive
+    ///
+    /// Note that if there are duplicate entries for the same filename, the files from the other archive will be the ones that are kept.
+    pub fn merge(&mut self, other: &PrivateArchive) {
```
maqi, line 136: this new function is not being used?
at least some E2E CI test to confirm it shall work as expected ?

ant-cli/src/actions/download.rs
grumbach: There is no limit to the size of a data map chunk so the suggestion would take too many lines

ant-cli/src/actions/download.rs
grumbach: It's included line 155 no?

autonomi/src/client/high_level/files/archive_private.rs
grumbach: Will make a unit test right

ant-cli/src/actions/download.rs
grumbach: Added a suggestion mentioning them

maqi (approved):

PR #2763, opened 2025-02-19
fix(metrics): caclulate peers in rt directly from kbuckets

RolandSherwin:


ant-networking/src/cmd.rs
```diff
@@ -626,15 +626,8 @@ impl SwarmDriver {
...
-                    _index,
-                    _total_peers,
-                    peers_in_non_full_buckets,
-                    num_of_full_buckets,
-                    _kbucket_table_stats,
-                ) = self.kbuckets_status();
-                let estimated_network_size =
-                    Self::estimate_network_size(peers_in_non_full_buckets, num_of_full_buckets);
+                let kbucket_status = self.get_kbuckets_status();
+                self.update_on_kbucket_status(&kbucket_status);
```
RolandSherwin, line 630: This is optional, we do this on peer add/remove. But we calculate the value here anyway, won't hurt if we update SwarmDriver here as well.

maqi (approved):

RolandSherwin:
superseded by https://github.com/maidsafe/autonomi/pull/2767

PR #2764, opened 2025-02-19
fix(bootstrap): simplify the bootstrap cache to be a fifo

RolandSherwin:
- Simplify the bootstrap cache to be a FIFO. 
- Since there is a change in struct, we've also introduced some backwards compatibility such that we are able to write the cache in the old format. 
  - This is required mostly by our public bootstrap cache, which can still be used by nodes running the older version of the code.
- Introduce `--write-older-cache-file` argument for `antnode` and `antctl`. This also writes the argument inside the antctl's `NodeRegistry` to preserve the arg during upgrades.
- Add more tests for the ant bootstrap crate.


mickvandijke:
Hey @RolandSherwin , is this still a WIP?

RolandSherwin:
@mickvandijke , the PR is complete and it is ready for testing. I'll rebase it to main now. c.c @jacderida

jacderida (approved):

PR #2767, opened 2025-02-20
improvement of relay related metrics; remove relay clients if we have no more connections

RolandSherwin:
- Remove relay clients from the `SwarmDriver::connected_relay_clients` tracker if the `Reservation` has been closed. This happens when all the connection with the relay client has been lost. Libp2p does not have an event for this yet, so we have to do a manual check for this. 
- Calculate the `peers_in_rt` value directly from the kbuckets instead of acting upon the libp2p events. The latter gave incorrect values, and it is very prone to errors, as everyone should be careful to call the `update` fn everytime we manually remove a peer from the RT.
- Add `connected_relay_clients` metric.
- Add `relay_peers_in_routing_table`, `peers_in_non_full_buckets`, `relay_peers_in_non_full_buckets` and `percentage_of_relay_peers` metrics.

### DIFF
--- a/ant-networking/src/event/mod.rs
+++ b/ant-networking/src/event/mod.rs
@@ -37,8 +37,29 @@ use std::{
 };
 use tokio::sync::oneshot;
 
-// (total_buckets, total_peers, peers_in_non_full_buckets, num_of_full_buckets, kbucket_table_stats)
-type KBucketStatus = (usize, usize, usize, usize, Vec<(usize, usize, u32)>);
+#[derive(Debug, Clone)]
+pub(crate) struct KBucketStatus {
+    pub(crate) total_buckets: usize,
+    pub(crate) total_peers: usize,
+    pub(crate) peers_in_non_full_buckets: usize,
+    pub(crate) num_of_full_buckets: usize,
+    pub(crate) kbucket_table_stats: Vec<(usize, usize, u32)>,
+    pub(crate) estimated_network_size: usize,
+}
+
+impl KBucketStatus {
+    pub(crate) fn log(&self) {
+        info!(
+            "kBucketTable has {:?} kbuckets {:?} peers, {:?}, estimated network size: {:?}",
+            self.total_buckets,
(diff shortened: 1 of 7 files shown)


maqi (approved):

ant-networking/src/event/mod.rs
```diff
@@ -324,51 +358,55 @@ impl SwarmDriver {
...
-        // update_on_peer_addition or update_on_peer_removal when adding or removing a peer.
-        // Only log every 10th peer to avoid spamming the logs.
-        if total_peers % 10 == 0 && total_peers != self.peers_in_rt {
-            warn!(
-                "Total peers in routing table: {}, but kbucket table has {total_peers} peers",
-                self.peers_in_rt
-            );
+            let _ = metrics_recorder
+                .percentage_of_relay_peers
+                .set((status.total_relay_peers as f64 / status.total_peers as f64) * 100.0);
```
maqi: may need to be `relay_peers_in_non_full_buckets / peers_in_non_full_buckets`

ant-networking/src/metrics/mod.rs
```diff
@@ -42,9 +41,14 @@ pub(crate) struct NetworkMetricsRecorder {
...
     // metrics from ant-networking
     pub(crate) connected_peers: Gauge,
+    pub(crate) connected_relay_clients: Gauge,
     pub(crate) estimated_network_size: Gauge,
+    pub(crate) percentage_of_relay_peers: Gauge<f64, AtomicU64>,
     pub(crate) open_connections: Gauge,
     pub(crate) peers_in_routing_table: Gauge,
+    pub(crate) relay_peers_in_routing_table: Gauge,
+    pub(crate) peers_in_non_full_buckets: Gauge,
+    pub(crate) relay_peers_in_non_full_buckets: Gauge,
```
maqi: `peers_in_non_full_buckets` and `relay_peers_in_non_full_buckets` may can be removed before merge

PR #2768, opened 2025-02-20
feat: organically bootstrap network and limit concurrency

RolandSherwin:
- Move `NetworkDiscovery` related components into a single module and expose just the required components.
- Make the initial bootstrapping process baked inside the network code. Earlier, we exposed the `Network::dial()` fn that allowed the client and node to have very different way of bootstrapping and it was mostly just dialling ALL the peers given to them concurrently. This was highly inefficient and led to overloading of the peers in the `bootstrap cache`.
  - To make things simpler, the initial bootstrapping process is now part of the networking code and common for the client and the nodes.
  - We dial 3 of the provided initial bootstrap addrs  concurrently at any time.
  - Once the RT reaches 5 peers, we let network discovery take over the rest of the bootstrapping, and we don't dial anymore bootstrap peers.
  - The bootstrap peers are just used to get inside the network, we cannot use them to discover the network (i.e, fill our RT).
- Remove the `Network::dial()` public fn. 
- Network discovery rate has been increased during the startup phase, but it should slowdown exponentially as we have more peers added to the RT.
- Implements a `InitialBootstrapTrigger` that tracks events and bool to make the decision on when it is "okay" to start the bootstrapping process. This struct is seprate from the `InitialBootstrap` struct.
  - Clients starts immediately
  - Non upnp nodes will start bootstrapping once they have a listen addr
  - Upnp nodes will start bootstrapping once they've recieved either an ok/err event from the upnp behaviour.


ant-networking/src/driver.rs
```diff
@@ -761,6 +761,11 @@ impl SwarmDriver {
             );
         }
 
+        if self.is_client {
+            self.initial_bootstrap
+                .trigger_initial_bootstrap(&mut self.swarm, self.peers_in_rt);
+        }
+
```
RolandSherwin: This is fine to do I assume, even though the swarm has yet to run inside the loop, we could still queue tasks to the swarm. 
The `trigger_initial_bootstrap` would just add the `Dial` action to the swarm'ms inner queue.

ant-networking/src/event/swarm.rs
```diff
@@ -266,6 +278,21 @@ impl SwarmDriver {
                 let _ = self.live_connected_peers.remove(&connection_id);
                 self.record_connection_metrics();
             }
+            SwarmEvent::OutgoingConnectionError {
```
maqi, line 315: when `peer_id` is `none` , what does it mean ?

it seems this can be `merged` with the one below? just have an inner check of whether peer_id is None/Some to switch handling branch?

If does need to be split out as current, the `event_string` better to be differentiated.

maqi (approved):

ant-networking/src/event/swarm.rs
RolandSherwin: > when peer_id is none , what does it mean ?

I think this happens if we try to dial an add without a PeerId? 

> it seems this can be merged with the one below? just have an inner check of whether peer_id is None/Some to switch handling branch?

Yeah, that is what I tried first, but that block is already shifted to the right alot (too many match statements). We cannot return early if its `None` too, that is why the separate block.

The other option was to `match (err, peerid)` and deal with `(Err::type, Some(peer_id)` inside the matching arms.
If you like this approach, I can go with this.

> If does need to be split out as current, the event_string better to be differentiated.

Sure will update it!

ant-networking/src/bootstrap.rs
```diff
@@ -18,12 +18,57 @@ use libp2p::{
...
 /// The max number of peers to be added to the routing table before stopping the initial bootstrap process.
 const MAX_PEERS_BEFORE_TERMINATION: usize = 5;
 
+/// This is used to track the conditions that are required to trigger the initial bootstrap process once.
+pub(crate) struct InitialBootstrapTrigger {
+    pub(crate) upnp: bool,
+    pub(crate) client: bool,
+    pub(crate) upnp_gateway_result_obtained: bool,
+    pub(crate) listen_addr_obtained: bool,
+}
```
RolandSherwin, line 38: Chose to not merge this with the other struct as I get pretty confused when I try to do that..

PR #2773, opened 2025-02-24
feat: print out uploaded chunk addresses on CLI

b-zee:


autonomi/src/client/data_types/chunk.rs
```diff
@@ -238,22 +238,38 @@ impl Client {
...
                         .chunk_upload_with_payment(chunk, proof.clone())
                         .await
                         .inspect_err(|err| error!("Error uploading chunk {address:?} :{err:?}"))
                         // Return chunk reference too, to re-use it next attempt/iteration
-                        .map_err(|err| (chunk, err))
+                        .map_err(|err| (chunk, err));
+                    #[cfg(feature = "loud")]
+                    println!(
+                        "({}/{}) Chunk stored at: {}",
+                        i + 1,
```
grumbach: Here I would reword this to 
```diff
- "({}/{}) Chunk stored at: {}",
+ "({}/{}) Uploading Chunk at: {}",
```
As we did not store it yet at time of printing!

grumbach (approved):
Small comment but generally LGTM!

autonomi/src/client/data_types/chunk.rs
b-zee: If I understand correctly, `put_record` waits for it to be verified, this is called by `chunk_upload_with_payment`. Which also includes this line:
```
debug!("Successfully stored chunk: {chunk:?} to {storing_nodes:?}");
```

Though now that I look at it, the added `println!` statement should only be executed if `res.is_ok()`.

PR #2776, opened 2025-02-25
feat!: accurate useable uniform address management

grumbach:
Address management was clunky, messy and hard to use. Made it uniformly useable across all types, and in Python too!

- now all address types have the same methods: `to_hex`, `from_hex`
- all pubkey addressed datatypes actually have the pubkey in their address
- high level `DataAddress` which shares the values above instead of the raw low level `XorName` that can't even be `from_hex`'d
- Python now uses actual accurate addresses instead of clunky hex strings sometimes, and actual addresses for other types
- Fix inaccurate/missing python bindings for addresses, now all have `to_hex`, `from_hex`


happybeing:
@grumbach I'm wondering if this still allows me to take a string from the command line, which can be either a name, hex register/history address, or hex Archive address, and still tell the difference from the String representations? 

I have implemented this to simplify the CLI for dweb so the user can just provide a name or either kind of address and not be concerned about the differences or provide different '--options'.

ant-protocol/src/storage/address/chunk.rs
```diff
@@ -20,18 +23,30 @@ impl ChunkAddress {
...
         &self.0
     }
 
+    /// Returns the hex string representation of the address.
     pub fn to_hex(&self) -> String {
         hex::encode(self.0)
     }
+
+    /// Creates a new ChunkAddress from a hex string.
+    pub fn from_hex(hex: &str) -> Result<Self, AddressParseError> {
```
mickvandijke:
```suggestion
    pub fn try_from_hex(hex: &str) -> Result<Self, AddressParseError> {
```

mickvandijke (approved):
Very happy to see these changes, great work!! Left one pedantic comment

grumbach:
@happybeing although this PR does not address it and keeps addresses short, I indeed wondered about adding type information to the addresses. It would make them slightly larger but would allow as you say for apps to immediately know what an address is. 
As I wanted to keep the breaking change to a minimum I did not include this extra change here. But definitely something I'm keeping in mind.

PR #2778, opened 2025-02-25
fix(evmlib): remove `0x` prefix

b-zee:


mickvandijke (approved):
I'm super confused how our production build is fine with this, but Dave isn't 😅

b-zee:
Looks like this behavior was changed recently.

https://github.com/alloy-rs/core/commit/0d8fe1811ea16359ca74b8f7e7abea9dd7045363

So it's a discrepancy between the indirect dependencies used in Dave vs this repo.

b-zee:
<s>I've pushed a commit to require the `alloy-primitives` crate to be `^0.8.18`. Might prevent others that update `autonomi` to the latest version from breaking their builds.</s> Wait, that's not helping anything because the first commit already fixes that basically. Sorry, brain farted there!

Also, @mickvandijke, maybe we should try and update `alloy`? We're using `0.7.3` (December 5), last version is `0.11.1` (February 12).

mickvandijke:
> Looks like this behavior was changed recently.
> 
> [alloy-rs/core@0d8fe18](https://github.com/alloy-rs/core/commit/0d8fe1811ea16359ca74b8f7e7abea9dd7045363)
> 
> So it's a discrepancy between the indirect dependencies used in Dave vs this repo.

Ahhh, thanks for figuring that out!

PR #2782, opened 2025-02-25
fix: support running multiple EVM testnets in parallel

mickvandijke:
No longer use port `4343` by default for EVM testnets. Instead let the OS decide which port to use. This allows us to run multiple testnets in parallel.


ermineJose (approved):
LGTM!

PR #2784, opened 2025-02-26
fix: records_store metrics count

### LINKED ISSUE #2775: Possible under-counting of records stored
Metrics seems to undercount the actual number of records stored.
I believe that it is counting only the second and subsequent chunks in each `record_store` directory.

I have an instance of Formicaio running the past few days which consistently showed only 2 records stored and then finally three. I looked at the actual directories and could see that many of the record_store dirs did have chunks stored. On closer inspection I could see three record_store dirs with two chunks. From the date stamps I could see one chunk was relatively recent. This might explain the jump from 2 to 3 records reported

See  https://forum.autonomi.community/t/formicaio/40445/274?u=southside

I got DeepSeek to write a script that will look at what the filesystem reports for the actual contents of each record_store dir for each of the  main methods of node wrangling, Launchpad, antctl, Formicaio and NTracking

https://github.com/safenetforum-community/chunk_checker

ermineJose:
### Description

The records_store metric counts one less than the actual stored records in the nodes. The root cause being, the  records_store metric is updated before storing the records into the Hash-map.

Fix: Update the records_store after storing the records into the hash-map.

### Related Issue

Fixes #2775 .

### Type of Change

- [ ] Documentation update

### DIFF
--- a/ant-networking/src/record_store.rs
+++ b/ant-networking/src/record_store.rs
@@ -635,6 +635,11 @@ impl NodeRecordStore {
         self.records
             .insert(key.clone(), (addr.clone(), validate_type, data_type));
 
+        #[cfg(feature = "open-metrics")]
+        if let Some(metric) = &self.record_count_metric {
+            let _ = metric.set(self.records.len() as i64);
+        }
+
         // Update bucket index
         let _ = self.records_by_distance.insert(distance, key.clone());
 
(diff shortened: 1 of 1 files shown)


jacderida:
Going to try closing and re-opening this PR to kick start the CI checks.

PR #2791, opened 2025-02-27
feat!: add gas fee limit for transactions

mickvandijke:
Adds a default gas fee limit for transactions. The limit is specified in a `TransactionConfig`.

CLI users can override this limit using `--max-fee-per-gas <WEI_AMOUNT>`.


happybeing:
@mickvandijke This is great :tada:. Will there be a similar limit for ANT? :pray:

mickvandijke:
> @mickvandijke This is great 🎉. Will there be a similar limit for ANT? 🙏

That's a great idea, I'll make a ticket for it 👍

PR #2802, opened 2025-03-03
feat(client): initial POC of smart_equation

maqi:
### Description

Initial POC of SmartEquation, which using Pointer pointing to a chunk.
This can be used to replace part of the quote verification flow: get a cost based on the QuotingMetrices
To be the sole place holding the equations of how to calculate the cose, which doesn't need to use smart_contract. 

The PR contains an E2E test to demo that:
1. publish an equation of `a + b` to the network
2. fetch the equation and use it to calculate
3. update the equation to `(a+5)*b`
4. fetch the equation from the network again and use it to calculate (result shall be new)

This answers one concern in the original native notken re-introduction RFC regarding having a smart_contract styple quoting verification.
This also shows a potential that with further design and expansion, we could have a non-block-chain-based smart_contract

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1604,6 +1604,7 @@ dependencies = [
  "rmp-serde",
  "self_encryption",
  "serde",
+ "serde_json",
  "serial_test",
  "sha2",
  "test-utils",
--- a/autonomi/Cargo.toml
+++ b/autonomi/Cargo.toml
@@ -47,6 +47,7 @@ rayon = "1.8.0"
 rmp-serde = "1.1.1"
 self_encryption = "~0.30.0"
 serde = { version = "1.0.133", features = ["derive", "rc"] }
+serde_json = "1.0"
 sha2 = "0.10.6"
 thiserror = "1.0.23"
 tokio = { version = "1.35.0", features = ["sync", "fs"] }
--- a/autonomi/src/client/high_level/mod.rs
+++ b/autonomi/src/client/high_level/mod.rs
@@ -8,6 +8,7 @@
 
 pub mod data;
 pub mod files;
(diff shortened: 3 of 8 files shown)


mickvandijke:
Hey @maqi,

Interesting PoC for moving towards native smart contracts! However, I have some concerns about its current scope. The use case, as it stands, seems quite limited. For instance, our smart contract queries the latest ANT/USDT token price from the chain to determine a quote price (and I believe it also stores that in a state). Because of this, we wouldn't be able to replace the current get-quote mechanism with this approach unless we incorporate similar functionality.

I’m also wondering who else might benefit from this in its current form. If we introduce this new type to the public API, I believe we should be committed to supporting it in the long run. If the use cases are minimal, maintaining it may not be the best use of our resources.

But I do really think we should develop this further so that we can actually replace parts of our EVM smart contract with a native smart contract, because it would be so much more scalable. Maybe we should put the work behind a feature flag so that it is clear this is experimental work?

If we do move forward with adding it to the public API, I have one suggestion:

Would it make sense to name `SmartEquation` simply `Equation` or `EquationChunk`, as a subtype of `Chunk`? I think the `Pointer` in front of it should be an optional proxy address. Else the current function seems more like a `ProxyEquation`.

Looking forward to hearing your thoughts and others!

maqi:
> Because of this, we wouldn't be able to replace the current get-quote mechanism with this approach unless we incorporate similar functionality

This PoC is actually part of the native token proposal.
It doesn't aim to replace the entire current quoting flow on smart_contract.

Regarding the concern that this becomase a public type and we supposed to support in future.
I do agree with that.

As the PR already got recorded in the native token proposal for any future reference back,
I am now got it closed.

PR #2808, opened 2025-03-05
fix(test): check for peers in routing table inside network spawner test

RolandSherwin:
- The network spawner test was failing on CI sometimes because we were trying to assert that we were `connected` to ALL nodes.
- Instead, we should assert that we have added all the nodes to the RT. We will not necessarily hold a live connection with all the peers we dialed.
- Also some tiny clippy fixes.


mickvandijke (approved):
Nice one!

PR #2809, opened 2025-03-05
test: enable data_with_churn using networkspawner

ermineJose:
### Description
Replace the current local testnet network for the data_with_churn testcase, as data_with_churn requires restart functionality which can no longer be supported by node once rpc is removed.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -653,6 +653,51 @@ jobs:
         run: cargo build --release --bin antnode
         timeout-minutes: 30
 
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: install foundary
+        shell: bash
+        run: curl -L https://foundry.paradigm.xyz | bash
+
+      - name: move foundry bins to path (linux)
+        if: matrix.os == 'ubuntu-latest'
+        shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/merge.yml
```diff
@@ -653,125 +653,68 @@ jobs:
...
+        env:
+          # only set the target dir for windows to bypass the linker issue.
+          # happens if we build the node manager via testnet action
+          CARGO_TARGET_DIR: ${{ matrix.os == 'windows-latest' && './test-target' || '.' }}
+        timeout-minutes: 30
+
+      - name: Chunks data integrity during nodes churn
+        run: cargo test --release -p ant-node --test data_with_churn -- --nocapture
+        env:
+          TEST_DURATION_MINS: 8
```
mickvandijke: Why the increase of `TEST_DURATION_MINS` from 5 to 8 mins?

mickvandijke (approved):
Looks good to me, nice work! Just left one comment

.github/workflows/merge.yml
ermineJose: it was experimental, when i was debugging a failure, forgot to change it back, Good find!, let me fix it now!!

PR #2815, opened 2025-03-06
feat(node): refresh RT with random targets

maqi:
### Description

refresh RT with random targets


maqi:
closed as proved to be invalid

PR #2818, opened 2025-03-06
fix(network): resend in case of OutboundFailure::DialFailure

maqi:
### Description

Occasionally, request will be terminated due to dial out failure as the peer turn down the connection due to CONNECTION_KEEP_ALIVE_TIMEOUT.
In that case, a re-attempt shall be carried out.


RolandSherwin (approved):

jacderida:
This will be replaced by #2820.

PR #2820, opened 2025-03-07
fix(network): dial peer before sending a req

RolandSherwin:
- If we don't have an active connection to a peer while sending a request, the `req/resp` behaviour would try to dial the peer. But this assumes that the swarn could find the multiaddrs of that peer.
- If the peer is part of kad's RT, then the swarm could find the addr. But we cannot just add all the peers to the RT, i.e, say a client querying for store costs for hundreds of chunks. 
- There is a `PeerAddresses` which is a LRU cache as part of the `identify` Behaviour, where we could add these addresses to, but that directly ties in the # of requests a client could do with the size of the cache. Hence the cache approach is dropped.
- Thus, this PR modifies the `ant_networking` fn calls to accommodate for an `Addresses(Vec<MulitAddr>)` struct wherever it is necessary. 
- Replication fetcher and other components that work based on the `local RT peers` would use `empty vec` as swarm could find these addrs from the RT.

### DIFF
--- a/ant-networking/src/cmd.rs
+++ b/ant-networking/src/cmd.rs
@@ -193,7 +193,8 @@ pub enum NetworkSwarmCmd {
     SendRequest {
         req: Request,
         peer: PeerId,
-        addrs: Addresses,
+        /// If the address is provided, we will try to perform a dial before sending the request.
+        addrs: Option<Addresses>,
 
         // If a `sender` is provided, the requesting node will await for a `Response` from the
         // Peer. The result is then returned at the call site.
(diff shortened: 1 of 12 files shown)


ant-networking/src/cmd.rs
```diff
@@ -82,16 +83,10 @@ pub enum LocalSwarmCmd {
     GetKBuckets {
         sender: oneshot::Sender<BTreeMap<u32, Vec<PeerId>>>,
     },
-    /// Returns the replicate candidates in range.
-    /// In case the range is too narrow, returns at lease CLOSE_GROUP_SIZE peers.
-    GetReplicateCandidates {
-        data_addr: NetworkAddress,
-        sender: oneshot::Sender<Result<Vec<PeerId>>>,
-    },
```
RolandSherwin, line 90: Unused

ant-networking/src/driver.rs
```diff
@@ -442,23 +442,25 @@ impl SwarmDriver {
...
-    pub(crate) fn get_closest_k_value_local_peers(&mut self) -> Vec<PeerId> {
+    pub(crate) fn get_closest_k_value_local_peers(&mut self) -> Vec<(PeerId, Addresses)> {
         let k_bucket_key = NetworkAddress::from_peer(self.self_peer_id).as_kbucket_key();
 
         // get closest peers from buckets, sorted by increasing distance to us
         let peers: Vec<_> = self
             .swarm
             .behaviour_mut()
             .kademlia
-            .get_closest_local_peers(&k_bucket_key)
```
RolandSherwin: Both the fns are the same, but the other one returns the `KadPeer` which hold the addresses as well. It also ignores the passed in `self.peer_id`, but we already chain that at the end.

ant-networking/src/cmd.rs
```diff
@@ -1212,13 +1233,13 @@ impl SwarmDriver {
 
         // get closest peers from buckets, sorted by increasing distance to the target
         let kbucket_key = target.as_kbucket_key();
-        let closest_k_peers: Vec<PeerId> = self
+        let closest_k_peers: Vec<(PeerId, Addresses)> = self
             .swarm
             .behaviour_mut()
             .kademlia
-            .get_closest_local_peers(&kbucket_key)
+            .find_closest_local_peers(&kbucket_key, &self.self_peer_id)
```
RolandSherwin, line 1238: We ignore `self_peer_id` here. This seems correct when trying to fetch candidates to send out replicate cmd to.

maqi: better put this as a comment in code ?

ant-networking/src/cmd.rs
```diff
@@ -551,6 +549,26 @@ impl SwarmDriver {
                         trace!("Replicate cmd to self received, ignoring");
                     }
                 } else {
+                    // dial the peer and send the request
+                    if addrs.0.is_empty() {
+                        info!("No addresses for peer {peer:?} to send request. This could cause dial failure if swarm could not find the peer's addrs.");
+                    } else {
+                        let opts = DialOpts::peer_id(peer)
```
maqi: shall we carry out the additional dial `AFTER` got an error ?
i.e. make dial a separate driver cmd, and to be executed on network level, instead of within the driver instance

maqi (approved):

ant-networking/src/cmd.rs
RolandSherwin: > shall we carry out the additional dial AFTER got an error ?

yeah sounds good. 

> i.e. make dial a separate driver cmd, and to be executed on network level, instead of within the driver instance

Dial is not blocking, it just adds the item to the queue for the swarm, so we should be fine? moving it to the networking layer could create a lot of indirections which seems unnecessary for this instance?

PR #2821, opened 2025-03-07
feat nodejs napi rs bindings

b-zee:
- **feat: implement basic NodeJS bindings with napi-rs**
- **fix: format; release profile removal in crate**


nodejs/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "0.1.0"
+description = "NodeJS bindings for the autonomi client"
+license = "GPL-3.0"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+autonomi = { path = "../autonomi" }
+bls = { package = "blsttc", version = "8.0.1" }
```
grumbach: To avoid compatibility issues across versions, I would suggest we remove this dependency and re-export whatever we need from bls in autonomi.

grumbach (approved):
Amazing work Benno! 👏

nodejs/Cargo.toml
b-zee: Good point, will refactor that

vphongph:
Looks great! GG!

To sum up:
90% (60/66) coverage of the functions from Rust!!

4 are referenced in the docs/comments as ToDo:
- upload_chunks_with_retries()
- get_raw_quotes()
- get_store_quotes()
- init_with_config()

2 are not referenced in the docs/comments:
- enable_client_events()
- get_quotes_for_content_addresses(), but this one is from external signer, it's low priority as Anselme told me

PR #2829, opened 2025-03-10
Improve client quoting performance

maqi:
### Description

With super large production network and nodes may contain certain outdated peer infos.
`kad::get_closest` could timed out, which slow down the quoting and uploading process a lot.

Meanwhile, the client doesn't set up the query_timeout.
The default value that got used is 60s, which is way too long.

Hence, to improve the quoting performance, this PR contains following work:
1.  using `re-dial` only during the re-attempt, to avoid un-necessary time out.
2.  set client query_timeout to be 10s

With the above, the quoting time against production network reduced to 30s (from the previous 3 min 20s).


grumbach (approved):
LGTM

PR #2831, opened 2025-03-11
feat: cli analyze addr

grumbach:
Analyze any address using:

```
ant analyze 360597a525a9900770c5f6716a0a406a0bea10a2dbe1abe5be9f1b0b0f993851
```

Prints out the result of the analysis.

```
Chunk stored at: 360597a525a9900770c5f6716a0a406a0bea10a2dbe1abe5be9f1b0b0f993851
Chunk content in hex: 806ee175e90830b14ad8f596979815f7
```

Works for all data types including:
- chunk
- pointer
- scratchpad
- graphentry

Also including high level data types:
- data
- public archives
- private archives
- registers

Also works for raw DataMap:

```
ant analyze 81a54669727374939400dc0020ccf9cca612cceccca53611ccf7ccf67f6226ccb848493a122dcc9acceb21cca7ccef0024ccabccf7104812ccfdccecdc0020086135cc91700eccbeccbbcc931dcce8113dcceb1d5e4ccca80859ccb22ccca5ccabccd0ccf310ccc5cccc67ccb8ccbd049401dc002038ccdc4bccbb777e130b4cccf7cc8d2265cc9b0bcca2ccbdcccdccdbcc9dccf174011eccb7cce2520d224fccf542dc0020cc92cca5cceeccc5ccb5ccf4241c59cc874c42703c54582eccddcca2cce0cc98ccc2ccc5ccb36642ccaeccf0cc8f6318ccc8049402dc00203605cc97cca525cca9cc900770ccc5ccf6716a0a406a0bccea10cca2ccdbcce1ccabcce5ccbecc9f1b0b0fcc993851dc0020cc98ccf55c40ccd05a13cc91ccb172427818ccce31ccd0445acc9fccd9536e2d05ccef27466a671acce2cce305
```
```
DataMap containing 3 Chunks
Content is another DataMap: false
[
    f9a612eca53611f7f67f6226b848493a122d9aeb21a7ef0024abf7104812fdec,
    38dc4bbb777e130b4cf78d22659b0ba2bdcddb9df174011eb7e2520d224ff542,
    360597a525a9900770c5f6716a0a406a0bea10a2dbe1abe5be9f1b0b0f993851,
]
Decrypted Data in hex: 48656c6c6f2c20576f726c643f
```

Can be run in verbose mode to print extra information:

```
ant analyze -v 360597a525a9900770c5f6716a0a406a0bea10a2dbe1abe5be9f1b0b0f993851
```

### DIFF
--- a/autonomi/src/client/analyze.rs
+++ b/autonomi/src/client/analyze.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 MaidSafe.net limited.
+// Copyright 2025 MaidSafe.net limited.
 //
 // This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
 // Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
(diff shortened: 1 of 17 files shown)


autonomi/src/client/analyze.rs
```diff
@@ -0,0 +1,464 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

autonomi/src/client/analyze.rs
```diff
@@ -0,0 +1,464 @@
...
+fn chunk_list_from_datamap(map: DataMap) -> Vec<ChunkAddress> {
+    let mut chunks = Vec::new();
+    for info in map.infos() {
+        chunks.push(ChunkAddress::new(info.dst_hash));
+    }
+    chunks
+}
+
+fn data_hex(data: &Bytes) -> String {
+    if data.len() <= MAX_HEX_PRINT_LENGTH {
```
maqi, line 386: this seems doesn't need to be 4k ?
maybe 32 or even 16 will be enough

autonomi/src/client/analyze.rs
```diff
@@ -0,0 +1,464 @@
...
+    let register_res = client.register_get(&register_address);
+    let scratchpad_res = client.scratchpad_get(&scratchpad_address);
+
+    let (maybe_graph_entry, maybe_pointer, maybe_register, maybe_scratchpad) =
+        tokio::join!(graph_entry_res, pointer_res, register_res, scratchpad_res);
+
+    match (
+        maybe_graph_entry,
+        maybe_pointer,
+        maybe_register,
```
maqi, line 341: vault not covered ?

autonomi/src/client/analyze.rs
```diff
@@ -0,0 +1,464 @@
...
+            ($($arg:tt)*) => {
+                if verbose {
+                    println!($($arg)*);
+                }
+            };
+        }
+        let hex_addr = address.trim_start_matches("0x");
+
+        // data addresses
+        let maybe_xorname = ChunkAddress::from_hex(address).ok();
```
maqi, line 176: XorName and PublicKey has the same length.
In theory, the output of `to_hex()` from these two, can be converted to the other via `from_hex()` decoder.

Is there any test to confirm these two can be differentiated ?

grumbach: XorName is 32 bytes and a bls pubkey is 48 bytes but will make sure in a test

autonomi/src/client/analyze.rs
grumbach: Yeah this will need to come in an update

autonomi/src/client/analyze.rs
grumbach: I asked chatGPT what size would be reasonable to print out. I think we need to be able to print large contents too up to a reasonable size.

maqi: yeah, we `able to` print out that large size, but most part of it will be useless.
I think for the data itself, a file flushed to disk will be more helpful than print it out ?

maqi (approved):

PR #2834, opened 2025-03-11
feat(client): use PeersArgs inside ClientConfig

RolandSherwin:
- Fixes an issue where the CLI would try to fetch `mainnet` contacts even though we have specified the `--testnet` value to it. This was because the CLI was fetching the `initial peers` at two places (which was unnecessary) and also had misconfigurations at the other place.
- The `PeersArgs` has also been renamed to `InitialPeersConfig` for better clarity and re-usability across the codebase.
- Clients now write bootstrap cache on a local network. This keeps the clients same for both the local and non-local network.


ant-cli/src/access/network.rs
```diff
@@ -1,46 +0,0 @@
...
-    let addrs = peers.get_addrs(None, Some(100)).await
-        .wrap_err("Please provide valid Network peers to connect to")
-        .with_suggestion(|| format!("make sure you've provided network peers using the --peers option or the {ANT_PEERS_ENV} env var"))
-        .with_suggestion(|| "a peer address looks like this: /ip4/42.42.42.42/udp/4242/quic-v1/p2p/B64nodePeerIDvdjb3FAJF4ks3moreBase64CharsHere")?;
-
-    let net = if peers.local {
-        NetworkPeers::Local(addrs)
-    } else {
-        NetworkPeers::Public(addrs)
-    };
```
RolandSherwin, line 43: The variants `NetworkPeers::Local` or `NetworkPeers::Public` is set based on the peers.local and nothing else. So this could be removed.

Also, this is where we were fetching the addrs/initial contacts, which is not required now.

autonomi/src/client/mod.rs
```diff
@@ -198,19 +209,13 @@ impl Client {
...
     /// ```
     pub async fn init_with_config(config: ClientConfig) -> Result<Self, ConnectError> {
-        let peers_args = PeersArgs {
-            disable_mainnet_contacts: config.local,
-            addrs: config.peers.unwrap_or_default(),
-            local: config.local,
-            ..Default::default()
-        };
-        let initial_peers = match peers_args.get_addrs(None, None).await {
+        let initial_peers = match config.init_peers_config.get_addrs(None, None).await {
```
RolandSherwin, line 212: The `initial peers` are now obtained while building the client.

PR #2837, opened 2025-03-13
Client fixes to improve the client upload success rate and performance

maqi:
### Description

This PR contains following tweaks aiming to improve the client upload success rate and performance:
* increase client KAD_QUERY_TIMEOUT from 10s to 60s, to allow more accurate peer sets got returned from the get_closest query
* reduce some un-necessary tryouts, to reduce the overall upload process, especially during failure case.


grumbach (approved):
LGTM

PR #2838, opened 2025-03-13
Use fixed set for existence check

maqi:
### Description

This is on top of the PR https://github.com/maidsafe/autonomi/pull/2837

Use fixed set for existence check, to reduce the chance that the paid payees mismatch with the peers got challenged.

### DIFF
--- a/ant-evm/src/data_payments.rs
+++ b/ant-evm/src/data_payments.rs
@@ -38,22 +38,13 @@ impl From<PeerId> for EncodedPeerId {
     }
 }
 
-/// The proof of payment for a data payment
+/// The proof of payment for a data payment, only to be used on client side
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Serialize, Deserialize)]
-pub struct ProofOfPayment {
+pub struct ClientProofOfPayment {
     pub peer_quotes: Vec<(EncodedPeerId, Vec<Multiaddr>, PaymentQuote)>,
 }
 
-impl ProofOfPayment {
-    /// returns a short digest of the proof of payment to use for verification
-    pub fn digest(&self) -> Vec<(QuoteHash, QuotingMetrics, RewardsAddress)> {
-        self.peer_quotes
-            .clone()
-            .into_iter()
-            .map(|(_, _, quote)| (quote.hash(), quote.quoting_metrics, quote.rewards_address))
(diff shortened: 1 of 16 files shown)


ant-node/src/put_validation.rs
```diff
@@ -665,7 +665,7 @@ impl Node {
         // push self in as the returned list doesn't contain self
         closest_k_peers.push((self_peer_id, Default::default()));
         let mut payees = payment.payees();
-        payees.retain(|peer_id| !closest_k_peers.iter().any(|(p, _)| p == peer_id));
+        payees.retain(|(peer_id, _addrs)| !closest_k_peers.iter().any(|(p, _)| p == peer_id));
         if !payees.is_empty() {
```
grumbach, line 669: This is a breaking change! Will break all old nodes when they receive this!

ant-evm/src/data_payments.rs
```diff
@@ -50,30 +50,36 @@ impl ProofOfPayment {
...
             .clone()
             .into_iter()
-            .map(|(_, quote)| (quote.hash(), quote.quoting_metrics, quote.rewards_address))
+            .map(|(_, _, quote)| (quote.hash(), quote.quoting_metrics, quote.rewards_address))
             .collect()
     }
 
     /// returns the list of payees
-    pub fn payees(&self) -> Vec<PeerId> {
+    pub fn payees(&self) -> Vec<(PeerId, Vec<Multiaddr>)> {
```
grumbach, line 58: This is a massive change and will be breaking many things, just want to understand why we are doing this, could you mind explaining why we are doing this?

grumbach (commented):
LGTM, some comments though:
- the payees change will have a lot of conflicts with the relay rewards work
- the `Addresses` (vec of multiaddresses?) type is risky as easy to confuse with the `Address` (evm address) type which has nothing to do with it! Could we rename it?

maqi:
> * the payees change will have a lot of conflicts with the relay rewards work

something have to be resolved

> * the `Addresses` (vec of multiaddresses?) type is risky as easy to confuse with the `Address` (evm address) type which has nothing to do with it! Could we rename it?

That's not a new one?
anyway, I am ok with the renameing. 
However would suggest it to be done in a separate PR

maqi:
closed as the result of testing against production network is not good.

PR #2840, opened 2025-03-15
further attempts to improve the client upload success rate

maqi:
### Description

Contains following works to improve the client upload success rate:

- increase client KAD_QUERY_TIMEOUT from 10s to 60s, to allow more accurate peer sets got returned from the get_closest query (From PR #2837)
- reduce some un-necessary tryouts, to reduce the overall upload process, especially during failure case. (From PR #2837)
- Check both payees and network queried closest nodes for chunk existence check (on top of PR #2838)
- Dial before put_record_to, to ensure connection to be re-established, following the model within the req/rsp communication.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -405,7 +405,7 @@ jobs:
           done
         env:
           ANT_LOG: "v"
-        timeout-minutes: 25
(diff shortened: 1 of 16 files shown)


maqi:
closed as the result of testing against production network is not good.

PR #2842, opened 2025-03-16
perf(tokio): switch to rt-multi-thread for better concurrency

### LINKED ISSUE #2841: Not using multi-thread tokio runtime.
I don't know if this is intentional due to some limitation, but generally the performance when spawning would improve by using the `rt-multi-thread` feature.
The most important places I identified would be `ant-node` and `ant-networking`, but there may be more.

gnytezero:
### Description

Generally the performance when spawning would improve by using the `rt-multi-thread` feature over `rt`.
These have been switched for `ant-node` and `ant-networking` that were assumed to be in most need of it.

This change may or may not have deeper and wider consequences including solving, unveiling or give rise to bugs, or otherwise affect functionality in unintended ways, as the system has been built relying on single-threaded tokio runtime.

I would say this change requires thorough validation in test environment to make sure all such unintended consequences are identified.

### Related Issue

Fixes #2841.

### Notes

I have not added tests that prove the fix is effective or does not break anything.
I have not verified in a test network that the fix is effective or does not break anything.

I did not find a branch `alpha` to make the PR to, as requested in `README`, and went for `main` instead, as it seemed most suitable.

### DIFF
--- a/ant-cli/Cargo.toml
+++ b/ant-cli/Cargo.toml
@@ -25,7 +25,7 @@ harness = false
 [dependencies]
 ant-build-info = { path = "../ant-build-info", version = "0.1.26" }
 ant-logging = { path = "../ant-logging", version = "0.2.48" }
-autonomi = { path = "../autonomi", version = "0.4.3", features = [ "loud" ] }
+autonomi = { path = "../autonomi", version = "0.4.3", features = ["loud"] }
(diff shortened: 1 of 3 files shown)


gnytezero:
I realized this needs additional work, as `record_store::tests::can_store_after_restart` fails when running all tests together, even though it passes when run separately.

mickvandijke (approved):

PR #2848, opened 2025-03-18
fix(node): using distance_range to verify payment payee as well

maqi:
### Description

There might be payee got blocked by us or churned out from our perspective.
We shall still consider the payment is valid whenever payees are close enough.
Hence using distance_range to verify payment payee as well

### Type of Change


mickvandijke (approved):
Looks good!

PR #2852, opened 2025-03-19
feat(autonomi): add `get_node_version` method to `Client`

mickvandijke:
- adds `get_node_version` method to `Client`.
- adds `impl From<ChunkAddress> for NetworkAddress`. (Makes it easier to use `Client::get_closest_for_address`, as `NetworkAddress` is a private type.)
- moves `get_closest_to_address` method from `quote` mod to `network`.

Changes needed for the emission service and relayer reward feature.

### DIFF
--- a/ant-networking/src/lib.rs
+++ b/ant-networking/src/lib.rs
@@ -30,6 +30,7 @@ mod relay_manager;
 mod replication_fetcher;
 pub mod time;
 mod transport;
+pub mod version;
 
 use cmd::LocalSwarmCmd;
 use xor_name::XorName;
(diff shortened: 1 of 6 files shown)


ant-networking/src/lib.rs
```diff
@@ -728,6 +730,28 @@ impl Network {
         Ok(None)
     }
 
+    /// Request the node version of a peer on the network.
+    pub async fn get_node_version(&self, peer_id: PeerId) -> Result<Version, String> {
+        let request = Request::Query(Query::GetVersion(NetworkAddress::from_peer(peer_id)));
+        match self
+            .send_request(request, peer_id, Default::default())
```
maqi, line 737: node only try to fetch version from near range peers, which shall sit in the RT and swarm driver can figure out peer's multi_addr internally.

However, for client, in most case, the targeted node is not in the RT, hence the multi_addr is not available.
Sending a `default` here will result in the request failed to be sent out in most cases.
which result in a wrong interception of nodes' version.

maqi (changes requested):

mickvandijke:
Will be replaced by another approach.

PR #2855, opened 2025-03-20
refactor!: network address and get closest address

mickvandijke:
BREAKING CHANGE (ant-protocol): Removed some associated functions for `NetworkAddress` and replaced them with a more conventional `impl<From>`.

- Replaces all simple `NetworkAddress::from_chunk_address`, `NetworkAddress::from_peer_id` etc.. with `impl<From<ChunkAddress>> for NetworkAddress` etc... Makes it easier to auto-convert between types.
- Moves `Client::get_closest_to_address` to a new module called `network`.
- Adds a test for `Client::get_closest_to_address`.


autonomi/src/client/network.rs
```diff
@@ -0,0 +1,17 @@
+use crate::Client;
```
maqi, line 9: copyright declaration

autonomi/tests/network.rs
```diff
@@ -0,0 +1,45 @@
+use ant_bootstrap::InitialPeersConfig;
```
maqi, line 9: copyright declaration

maqi (approved):

PR #2874, opened 2025-04-02
refactor: remove manual libp2p replication factor configuration

mickvandijke:
Replication factor is now managed by libp2p and defaults to the `K_VALUE` = `20`.


mickvandijke:
Shows a performance regression in tests. Likely because of the influence on quorum. Libp2p fixed the problem that this PR was trying to solve anyway, so can close this.

PR #2879, opened 2025-04-03
Refresh scheme further tweaking

maqi:
### Description

There are couple of places can be tweaked further to achieve a better efficiency:

- always add self in as a get_closest target
- fill up as many as possible (with cap of 10) get_closest targets among empty_buckets
- avoid un-necessary random generation
- reduce network_discovery frequency
- avoid `% 0` due to edge case of empty non_full_buckets

### DIFF
--- a/ant-networking/src/network_discovery.rs
+++ b/ant-networking/src/network_discovery.rs
@@ -113,20 +113,20 @@ impl SwarmDriver {
             .partition(|kb| kb.num_entries() >= K_VALUE.get());
 
         // Process non-full buckets, collect get_closest candidates
-        let non_full_buckets_indexes = non_full_buckets
+        let non_full_non_empty_buckets_indexes = non_full_buckets
             .iter()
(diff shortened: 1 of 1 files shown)


mickvandijke:
Hi @maqi ,

Besides the addition of a self lookup (by adding self peer id to targets), should we also avoid adding a generated target for the bucket with the closest neighbour(s) to self peer id?

It might clash with the results of the self lookup I think 🤔 .

maqi:
> should we also avoid adding a generated target for the bucket with the closest neighbour(s) to self peer id?

you mean for the `full buckets` ? 
they have been filtered, and no candidate from `full buckets` will be picked.

> It might clash with the results of the self lookup

don't get what you mean.
there might be `waste of effort`, however there is no thing of `clash` matters ?

mickvandijke:
Hey @maqi , nevermind, I think I mixed up some thoughts. PR looks good!

mickvandijke (approved):

PR #2883, opened 2025-04-04
feat!: add versioning to NodeServiceData (BREAKING CHANGE)

RolandSherwin:
[BREAKING CHANGE]

- Renames `--home-network` flag to `--realy` inside antnode binary. Also updates the antctl node registry field.
- We now use `no-upnp` flag everywhere and also fixed some undefined values when passing multiple network flags.
- Rename `peers_args` to `initial_peers_config` inside antctl node registry file.
- Implement versioning for antctl `NodeServiceData` to maintain backwards compatibility with older versions of antctl.


ant-service-management/src/node/node_service_data_v1.rs
```diff
@@ -0,0 +1,152 @@
...
+};
+
+pub const NODE_SERVICE_DATA_SCHEMA_V1: u32 = 1;
+
+fn schema_v1_value() -> u32 {
+    NODE_SERVICE_DATA_SCHEMA_V1
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct NodeServiceDataV1 {
```
jacderida, line 28: Just a tiny optional thing. Is it worth maybe just using a small comment to point out which fields are new/different in V1?

jacderida (approved):

ant-service-management/src/node/node_service_data_v1.rs
RolandSherwin: Updated it!

PR #2886, opened 2025-04-08
test: add data test for Node.js

b-zee:


vphongph (approved):
Hi, good job Benno!

I read the code and ran the tests, it looks good to me!

Small comment: 2 dependencies from ava are deprecated, it's inflight and glob, I hope they won't generate issues in the future!

PR #2891, opened 2025-04-10
client light networking

mickvandijke:


autonomi/src/networking/driver/mod.rs
```diff
@@ -36,15 +37,12 @@ use tokio::sync::mpsc;
 // Autonomi Network Constants, this should be in the ant-protocol crate
 const KAD_STREAM_PROTOCOL_ID: StreamProtocol = StreamProtocol::new("/autonomi/kad/1.0.0");
 const MAX_PACKET_SIZE: usize = 1024 * 1024 * 5;
-const REPLICATION_FACTOR: NonZeroUsize =
-    NonZeroUsize::new(7).expect("REPLICATION_FACTOR must be > 0");
-
 /// Libp2p defaults to 10s which is quite fast, we are more patient
-pub const REQ_TIMEOUT: Duration = Duration::from_secs(30);
+pub const REQ_TIMEOUT: Duration = Duration::from_secs(10);
```
grumbach: Any reason for changing this? 10s is why we have so many timeouts on prod

autonomi/src/networking/driver/mod.rs
```diff
@@ -104,7 +102,6 @@ impl NetworkDriver {
             .set_max_packet_size(MAX_PACKET_SIZE)
             .set_parallelism(KAD_ALPHA)
             .set_query_timeout(KAD_QUERY_TIMEOUT)
-            .set_replication_factor(REPLICATION_FACTOR)
```
grumbach, line 107: Was this useless? I thought libp2p based the range of queries on this?

autonomi/src/networking/mod.rs
```diff
@@ -80,6 +83,7 @@ pub enum NetworkError {
 #[derive(Debug, Clone)]
 pub struct Network {
     task_sender: tokio::sync::mpsc::Sender<NetworkTask>,
+    _driver_task: Arc<JoinHandle<()>>,
```
grumbach: Why keep this one around?

grumbach: Removed it as I didn't observe any difference. Although made the task receiver an `Arc` so we never drop the original receiver (this fixed the dropped receiver error).

mickvandijke: Was a quick fix for the dropped network client error.

This was mostly an experimental branch to figure out why we didn't get any quote responses from non-local nodes.

> Removed it as I didn't observe any difference. Although made the task receiver an Arc so we never drop the original receiver (this fixed the dropped receiver error).

Hmm that's interesting, it did fix it for me (maybe only in the quoting flow). But happy with your solution!

PR #2896, opened 2025-04-16
refactor: client light networking

mickvandijke:


grumbach:
replaced by #2925

grumbach (approved):
LGTM, since it is quite big, would be good to have another review @maqi or @RolandSherwin?

autonomi/src/networking/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// When the `to` field is empty, the record is stored at the closest nodes to the record address,
+    /// else it is specifically stored to the nodes in the `to` field
+    pub async fn put_record(
+        &self,
+        record: Record,
+        to: Vec<PeerId>,
+        quorum: Quorum,
+    ) -> Result<(), NetworkError> {
+        let (tx, rx) = oneshot::channel();
+        let task = NetworkTask::PutRecord {
```
maqi, line 198: the `ant_network::put_record` carry out ChunkProofVerification as well.
However, seems with the CLN, the `verification` step got entirely skipped.

This will create a false user experience that: a success uploaded file may failed to be downloaded, due to some of chunks failed to be uploaded, say due to quote expiration.
In that situation, `kad::put_record` still return with success, but the copy won't got stored.

This step of verification can only be skipped AFTER we swtich from `kad::put_record_to` to `request/response messaging` (which will be used in the future work of `separate payment and upload`)

maqi (changes requested):
Skipping of verification could result in false user experience.
Meanwhile, the future work of `separate payment and upload` may allow skipping the verification.

Suggest to only got this PR merged AFTER bring back the verification (even on higher level), and tested with mainnet.
Otherwise, may have to wait for the `separate payment and upload` work got merged in.

autonomi/src/networking/driver/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+                    query_id,
+                    NetworkTask::PutRecord {
+                        record,
+                        to,
+                        quorum,
+                        resp,
+                    },
+                );
+            }
+            NetworkTask::GetQuote {
```
maqi, line 285: This better to be a nore generic task for `Request/Response messaging`, for the ease of other future usage.
Instead of currently too binded with `Quote`

autonomi/src/networking/utils.rs
```diff
@@ -0,0 +1,22 @@
+use crate::Multiaddr;
```
maqi, line 1: copyright statement

maqi (changes requested):

autonomi/src/networking/driver/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+                    key: addr.clone(),
+                    data_type,
+                    data_size,
+                    nonce: None,
+                    difficulty: 0,
+                });
+
+                // Add the peer addresses to our cache before sending a request.
+                for addr in &peer.addrs {
+                    self.swarm.add_peer_address(peer.peer_id, addr.clone());
```
maqi, line 302: we previously experienced many req/resp failure, and it was resolved by re-dail attempt.
will this `add_peer_address` approach achieve the same?
may need certain devnet tests or against mainnet to confirm?

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -0,0 +1,325 @@
...
+            NetworkTask::GetClosestPeers { resp, .. } => {
+                self.closest_peers.insert(id, resp);
+            }
+            NetworkTask::GetRecord { resp, .. } => {
+                self.get_record.insert(id, resp);
+            }
+            NetworkTask::PutRecord { resp, .. } => {
+                self.put_record.insert(id, resp);
+            }
+            _ => {}
```
maqi, line 89: `wild card` match shall be avoided.

maqi (changes requested):

autonomi/src/networking/driver/mod.rs
grumbach: As we don't use other `Request/Response` messaging now, this is premature in my opinion. Also, when we do add a new task using `Request/Response`, this shouldn't be a change in `NetworkTask`, but an implementation detail of how those tasks are handled. The fact that we use `Request/Response` should not bubble up and pollute the `NetworkTask` abstraction. 
Although I agree that once we do have such requests, they can use the same underlying code.

autonomi/src/networking/utils.rs
grumbach: I'm not sure I understand? Could you be more specific?

autonomi/src/networking/driver/mod.rs
grumbach: I'm not sure about this one, @mickvandijke do you know about this one?

autonomi/src/networking/driver/task_handler.rs
grumbach: Oh I didn't know, sure can fix that! What was the reason for that ban again?

autonomi/src/networking/driver/mod.rs
mickvandijke: This worked just as good or better than dialing before sending a request when I tested.

autonomi/src/networking/utils.rs
maqi:
```
// Copyright 2025 MaidSafe.net limited.
//
// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
```

autonomi/src/networking/driver/task_handler.rs
maqi: to avoid future new entry got slipped through in some hadling process

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -0,0 +1,347 @@
...
+                let peers = holders.keys().cloned().collect();
+                let records_uniq = holders.values().cloned().fold(Vec::new(), |mut acc, x| {
+                    if !acc.contains(&x) {
+                        acc.push(x);
+                    }
+                    acc
+                });
+
+                let res = match &records_uniq[..] {
+                    [] => responder.send(Ok((None, peers))),
```
maqi: this shall be an `error` ?
as no entry returned for a `get` ?

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -0,0 +1,347 @@
...
+                    [one] => responder.send(Ok((Some(one.clone()), peers))),
+                    [_one, _two, ..] => responder.send(Err(NetworkError::SplitRecord(holders))),
```
maqi: For mutable datatypes, situation of `split` could happen, especially with large scaled network with the mutable data got certain rounds of updates.

In such case, we shall `aggregate` the multiple versions.

Hence, here, we shall at least report back all versions

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -0,0 +1,347 @@
...
+                    .send(Ok((None, peers)))
+                    .map_err(|_| TaskHandlerError::NetworkClientDropped)?;
+            }
+            Err(kad::GetRecordError::QuorumFailed {
+                key,
+                records,
+                quorum,
+            }) => {
+                trace!(
+                    "QueryId({id}): GetRecordError::QuorumFailed {:?}, records: {:?}, quorum: {:?}",
```
maqi, line 186: For mutable datatypes, QuorumFailed scenario could happen with multiple versions exists.
Hence, such fetched records shall be `aggregated`.
here, at least shall report back such multiple versions.

maqi (changes requested):
For mutable datatypes, network get_record could fail with split or quorum_failed
but they shall be handled, instead of current `simple fail`

autonomi/src/networking/driver/task_handler.rs
grumbach: At the record get layer this is not an error, just no record.

autonomi/src/networking/driver/task_handler.rs
grumbach: The `SplitRecord` error is managed by the datatypes layer, and is different for each type.

autonomi/src/networking/driver/task_handler.rs
grumbach: This is up to the data layer calls to decide. The record layer has no knowledge of data types but does report splits when they happen.

autonomi/src/client/data_types/pointer.rs
```diff
@@ -49,58 +53,51 @@ pub enum PointerError {
...
+            Ok(None) => Err(GetError::RecordNotFound)?,
+            Err(NetworkError::SplitRecord(result_map)) => {
+                warn!("Pointer at {key:?} is split, trying resolution");
+                select_highest_pointer_version(result_map, key)?
+            }
+            Err(err) => {
+                error!("Error fetching pointer: {err:?}");
+                return Err(PointerError::GetError(err.into()));
+            }
         };
```
grumbach, line 74: @maqi this is where is it handled for Pointers for example

autonomi/src/networking/driver/task_handler.rs
grumbach: `NetworkError::SplitRecord(holders)` contains all versions and holders per version. The data layer can then use this data and do as it sees fit.

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -0,0 +1,347 @@
...
+                let (responder, holders) = self.take_responder_and_holders_for_task(id)?;
+                let peers = holders.keys().cloned().collect();
+
+                responder
+                    .send(Ok((None, peers)))
+                    .map_err(|_| TaskHandlerError::NetworkClientDropped)?;
+            }
+            Err(kad::GetRecordError::QuorumFailed {
+                key,
+                records,
```
maqi, line 182: this `records` returned with the failure better to be reported back ?

maqi: seems the QuorumFailed error never got raised from libp2p anyway
refer further to https://github.com/libp2p/rust-libp2p/discussions/6044

PR #2905, opened 2025-04-18
impl: automatic reachability status

RolandSherwin:
Depends on https://github.com/maidsafe/autonomi/pull/2921

- Adds a `--reachability-check` flag to the nodes that would override the network flags such as: `--relay`, `--ip` and `--no-upnp`.
- Determines if a node is reachable via UPnP.
- If not, asks other peers to dial us back after 180s and determines our reachability status with that.
- Maps what others observe us at, along with which local adapter they used to dial us to finally determine the reachability status.

### DIFF
--- a/ant-node/src/bin/antnode/main.rs
+++ b/ant-node/src/bin/antnode/main.rs
@@ -76,8 +76,6 @@ pub fn parse_log_output(val: &str) -> Result<LogOutputDestArg> {
 #[command(disable_version_flag = true)]
 #[clap(name = "antnode cli", version = env!("CARGO_PKG_VERSION"))]
 struct Opt {
-    #[clap(long, default_value_t = false)]
-    nat_detection: bool,
     /// Set to connect to the alpha network.
     #[clap(long)]
     alpha: bool,
@@ -186,6 +184,12 @@ struct Opt {
     /// The special value `0` will cause the OS to assign a random port.
     #[clap(long, default_value_t = 0)]
     port: u16,
+
+    /// Enabling this will run an optional reachability check before starting the node.
+    ///
+    /// Enabling this will cause the node to override some of the network flags like `--home-network`, `--upnp`, `--ip`.
+    #[clap(long, default_value_t = false)]
+    reachability_check: bool,
 
     /// Specify the rewards address.
     /// The rewards address is the address that will receive the rewards for the node.
@@ -325,7 +329,7 @@ fn main() -> Result<()> {
             node_socket_addr,
             root_dir,
         );
-        node_builder.with_nat_detection(opt.nat_detection);
+        node_builder.with_reachability_check(opt.reachability_check);
         node_builder.local(opt.peers.local);
         node_builder.no_upnp(opt.no_upnp);
         node_builder.bootstrap_cache(bootstrap_cache);
@@ -377,20 +381,24 @@ async fn run_node(
     reset_critical_failure(log_output_dest);
 
     info!("Starting node ...");
-    if node_builder.nat_detection {
-        let status = node_builder.run_nat_det().await?;
+    if node_builder.reachability_check {
+        info!("Running reachability check ...");
+        let status = node_builder.run_reachability_check().await?;
         match status {
             ReachabilityStatus::Upnp => {
+                info!("Reachability check: UPnP detected. Starting node with UPnP enabled.");
                 node_builder.no_upnp(false);
             }
             ReachabilityStatus::Reachable {
                 local_adapter: mut socket_addr,
             } => {
+                info!("Reachability check: Reachable. Starting node with socket addr: {socket_addr} and UPnP disabled.");
                 socket_addr.set_port(0);
                 node_builder.no_upnp(true);
                 node_builder.with_socket_addr(socket_addr);
             }
             ReachabilityStatus::Unreachable { .. } => {
+                info!("Reachability check: Unreachable. Starting node in Relay mode.");
                 node_builder.relay_client(true);
             }
         }
--- a/ant-node/src/bin/antnode/main.rs
+++ b/ant-node/src/bin/antnode/main.rs
@@ -367,7 +367,7 @@ fn main() -> Result<()> {
 /// - `Ok(None)` if we want to shutdown the node.
 /// - `Err(_)` if we want to shutdown the node with an error.
 async fn run_node(
-    node_builder: NodeBuilder,
+    mut node_builder: NodeBuilder,
     rpc: Option<SocketAddr>,
     log_output_dest: &str,
     log_reload_handle: ReloadHandle,
@@ -378,7 +378,20 @@ async fn run_node(
 
     info!("Starting node ...");
     if node_builder.nat_detection {
-        let _status = node_builder.run_nat_det().await?;
+        let status = node_builder.run_nat_det().await?;
+        match status {
+            NatStatus::Upnp => {
+                node_builder.upnp(true);
+            }
+            NatStatus::Public(mut socket_addr) => {
+                socket_addr.set_port(0);
+                node_builder.upnp(false);
+                node_builder.with_socket_addr(socket_addr);
+            }
+            NatStatus::NonPublic => {
+                node_builder.is_behind_home_network(true);
+            }
+        }
     }
 
     let running_node = node_builder.build_and_run()?;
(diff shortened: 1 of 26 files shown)


ant-node/src/bin/antnode/main.rs
```diff
@@ -133,6 +134,12 @@ struct Opt {
     #[clap(long, verbatim_doc_comment)]
     network_id: Option<u8>,
 
+    /// Enabling this will run an optional reachability check before starting the node.
+    ///
+    /// Enabling this will cause the node to override some of the network flags like `--home-network`, `--upnp`, `--ip`.
+    #[clap(long, default_value_t = false)]
```
maqi, line 191: the default may need to be set as `true` for the convenience of the normal users.

ant-node/src/node.rs
```diff
@@ -151,6 +166,28 @@ impl NodeBuilder {
...
+    pub async fn run_reachability_check(&self) -> Result<ReachabilityStatus> {
+        let mut network_builder = NetworkBuilder::new(
+            self.identity_keypair.clone(),
+            self.local,
+            self.initial_peers.clone(),
+        );
+
+        network_builder.listen_addr(self.addr);
+        #[cfg(feature = "open-metrics")]
+        network_builder.metrics_server_port(self.metrics_server_port);
```
maqi: this seems a `temp` node instance to carry out the reachability check.
enabling metrics for this might interfere with the metrics collected?
also, it's using the same ID as to the `normal` node instance, right?

ant-node/src/bin/antnode/main.rs
```diff
@@ -371,6 +379,29 @@ async fn run_node(
     reset_critical_failure(log_output_dest);
 
     info!("Starting node ...");
+    if node_builder.reachability_check {
+        info!("Running reachability check ...");
```
maqi: As it will take 3 mins for the reachability_check to complete.
A more `explicit` notice shall be given out to users?
especialy for the script/launch_pad users. otherwise they may easily think the node is stalled ?

ant-networking/src/network_builder.rs
```diff
@@ -283,20 +284,100 @@ impl NetworkBuilder {
...
+            })
+            .boxed();
+
+        // Listen on the provided address
+        let listen_socket_addr = self
+            .listen_addr
+            .ok_or(NetworkError::ListenAddressNotProvided)?;
+
+        // Identify Behaviour
+        let agent_version = IDENTIFY_REACHABILITY_CHECK_CLIENT_VERSION_STR
```
maqi: This means the reachability only got supported by the new version nodes ?

Wondering why this is required.

This could result in backward compability issue ? i.e. not only old nodes actually blocks or got blocked, but also only after most of nodes got upgraded, can this feature to be fully functional ?

ant-networking/src/network_builder.rs
```diff
@@ -384,6 +465,12 @@ impl NetworkBuilder {
                 .expect("Failed to obtain read lock for REQ_RESPONSE_VERSION_STR")
                 .clone();
 
+            let req_res_protocol = if is_client {
```
maqi: in case of being client, there is no point to carry out the reachability check ?

ant-networking/src/event/identify.rs
```diff
@@ -63,37 +72,29 @@ impl SwarmDriver {
             return;
         }
 
-        if info.agent_version.contains("client") {
+        // return early for reachability-check-client / clients
+        if info.agent_version.contains("reachability-check-client") {
```
maqi: This means only if most of nodes got upgraded to latest version, then the function can be functional usable by others.

otherwise, if testing with old version nodes, the check will fail ?

ant-networking/src/driver.rs
```diff
@@ -65,6 +68,9 @@ pub(crate) const CLOSET_RECORD_CHECK_INTERVAL: Duration = Duration::from_secs(15
 /// Interval over which we query relay manager to check if we can make any more reservations.
 pub(crate) const RELAY_MANAGER_RESERVATION_INTERVAL: Duration = Duration::from_secs(30);
 
+/// Interval over which we check if we could dial any peer in the dial queue.
+const DIAL_QUEUE_CHECK_INTERVAL: Duration = Duration::from_secs(2);
```
maqi: just a remind the dial back interval hence is now in a range of 180s ~ 182s .
the requestor side hence need to wait for bit longer, say 190s for safe.

ant-networking/src/reachability_check.rs
```diff
@@ -0,0 +1,991 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

ant-networking/src/reachability_check.rs
```diff
@@ -0,0 +1,991 @@
...
+            HashMap::from([(listen_id, HashSet::from([listen_socket_addr.ip()]))]);
+        Self {
+            swarm,
+            state: ReachabilityCheckState::WaitingForUpnp,
+            initial_listener,
+            initial_contacts,
+        }
+    }
+
+    /// Asynchronously drives the swarm event loop. This function will run indefinitely,
```
maqi: for the reachability check, `indefinitely` sounds risky
shall some time limitation to be given avoid edge case that really cause `indefinite loop` ?

also, the loop will be terminated in case of error or completion, hence the comment shall be updated anyway ?

ant-networking/src/reachability_check.rs
```diff
@@ -0,0 +1,991 @@
...
+                                            return Ok(status);
+                                        } else {
+                                           info!("We are retrying the WaitingForExternalAddr workflow. We will not return a status yet.");
+                                        }
+                                    }
+                                    Err(err) => {
+                                        warn!("Error while getting reachability status: {err}");
+                                        return Err(err.into());
+                                    }
+                                }
```
maqi: make the handling of dialing_completion a separate function shall avoid this bit scary cascaded brackets usage ?

ant-networking/src/reachability_check.rs
```diff
@@ -0,0 +1,991 @@
...
+                                external_to_local_addr_map
+                                    .entry(reachable_addr)
+                                    .or_default()
+                                    .insert(SocketAddr::new(
+                                        *another_listener_ip,
+                                        local_adapter_addr.port(),
+                                    ));
+                            }
+
+                            // 2. else try to find 10.0.0.0 address
```
maqi: entire 10.x.x.x is reserved for private, not only 10.0.0.0 ?
may need to update any comment/logging using `10.0.0.0` to `10.x.x.x` ?

ant-networking/src/driver.rs
RolandSherwin: Yea! I've added a 20s buffer time there.

ant-networking/src/event/identify.rs
RolandSherwin: As discussed in slack, we should be fine with "letting go" of backwards compatibility here, as the 180s dialback is very crucial and the check won't work without it. Hence, adding another non-backwards compatible change to the mix is fine.

ant-networking/src/network_builder.rs
RolandSherwin: > but also only after most of nodes got upgraded, can this feature to be fully functional ?

That is right, older nodes cannot help us in testing out the "reachability" of nodes as they don't have the 180s wait before dialing back. Hence adding another dependency here is fine.

ant-networking/src/network_builder.rs
RolandSherwin: Yes, clients do not have to be externally reachable? They initiate all the connections first and that should be mostly fine with any router?

ant-networking/src/reachability_check.rs
RolandSherwin: Ah that is a copy paste from the actual node's swarm driver. The code should terminate if there are no dial backs for 180s + 20s from the other peers.

RolandSherwin: Lemme update the doc there

ant-node/src/bin/antnode/main.rs
RolandSherwin: The feature will not work as intended unless most of the nodes we're dialing are upgraded, so this will be turned off for now.

ant-node/src/node.rs
RolandSherwin: > enabling metrics for this might interfere with the metrics collected?

The things are not wired under the hood, so we should be fine. But will remove this redundant line for clarity. Thank you! 

> also, it's using the same ID as to the normal node instance, right? 

Yes that is right..

PR #2906, opened 2025-04-19
chore(network): extra distance info during query iteration

maqi:
### Description

extra distance info during libp2p network query iteration


maqi:
closed as investigation work completed

PR #2915, opened 2025-04-22
feat(network): default chunk_batch_size to 1

maqi:
### Description

Default both CHUNK_UPLOAD_BATCH_SIZE and CHUNK_DOWNLOAD_BATCH_SIZE to 1.

During the upload/download test against the Alpha network, it proved that reducing the batch_size help with lifting up the success rate, especially for users with poor connection.
We prefer the default value to be the most robust and stable one.
Users can tweak batch_size value for a better performance, if the connection condition allows.


jacderida (approved):

maqi:
The work will be merged into main via the stable merge back PR https://github.com/maidsafe/autonomi/pull/2916.

PR #2920, opened 2025-04-23
passing the data_with_churn CI runs

maqi:
### Description

* revert the use of NodeSpawner in CI runs, so that logging Artifacts of data_with_churn and verify_location tests can be properly generated
* use `libp2p::kad::BucketInserts::OnConnected` for client network to pass the data_with_churn test


grumbach (approved):
QI YOU'RE THE MAN!! 👏👏🎉🎉🤩🤩

PR #2933, opened 2025-04-25
Easy alpha for cli

grumbach:
The `network-id` flag can now be used as the SOLE config option to connect to various networks (no need to provide other flags):
```
ant --network-id=2 file download ...
```

More info:

```
      --network-id <NETWORK_ID>
          Specify the network ID to use. This will allow you to run the CLI on a different network.
          Note that this overrides all other network config options (except in the Custom Network
          case).
          
          Valid values are:
           - 0: Local Network
           - 1: Mainnet (default)
           - 2: Alpha Network
           - 3: Custom Network (obtained through environment variables and other network config flags)
          
          [default: 1]
```

### DIFF
--- a/ant-cli/src/actions/connect.rs
+++ b/ant-cli/src/actions/connect.rs
@@ -66,13 +66,14 @@ pub async fn connect_to_network_with_config(
     match res {
         Ok(client) => {
             info!("Connected to the Network");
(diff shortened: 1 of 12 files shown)


mickvandijke (approved):
LGTM!

PR #2935, opened 2025-04-25
feat: sync register key in vault

grumbach:


vphongph:
I'm testing the PR and I'm facing issues

I removed the register key and launched `ant vault load`
While vault loading I can't get the signing key from the vault.

On the opposite side when I copy back the register key, and `ant vault sync`
I have the error `The register key in the vault does not match the local register key, aborting sync to prevent loss of current register key`
But it's the same key, I can edit previous register with this key!

Let's discuss further after the weekend

vphongph (approved):
Vault is fixed! GG!

PR #2943, opened 2025-04-30
fix: don't repay for failed put

grumbach:


ant-cli/src/commands/file.rs
```diff
@@ -80,6 +82,16 @@ pub async fn upload(
...
+            Err(UploadError::PutError(PutError::Batch(upload_state))) => {
+                // NB TODO: save payment to local disk for re-use
+                let exit_code =
+                    upload_exit_code(&UploadError::PutError(PutError::Batch(Default::default())));
+                return Err((
+                    eyre!(UploadError::PutError(PutError::Batch(upload_state)))
+                        .wrap_err("Failed to upload file".to_string()),
+                    exit_code,
+                ));
+            }
```
grumbach: CLI integration TODO here

ant-cli/src/commands/file.rs
```diff
@@ -95,6 +107,16 @@ pub async fn upload(
...
```
grumbach: CLI integration TODO here as well

autonomi/src/client/data_types/chunk.rs
```diff
@@ -225,100 +228,82 @@ impl Client {
...
+                    }
+                }
+                res
+            });
+        }
+        let uploads =
+            process_tasks_with_max_concurrency(upload_tasks, *CHUNK_UPLOAD_BATCH_SIZE).await;
+
+        // return errors
+        if uploads.iter().any(|res| res.is_err()) {
```
maqi, line 288: can use `split` to avoid un-necessary iteration

autonomi/src/client/put_error_state.rs
```diff
@@ -0,0 +1,72 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

ant-evm/src/data_payments.rs
```diff
@@ -17,7 +17,7 @@ pub use std::time::SystemTime;
 use xor_name::XorName;
 
 /// The time in seconds that a quote is valid for
-pub const QUOTE_EXPIRATION_SECS: u64 = 3600;
+pub const QUOTE_EXPIRATION_SECS: u64 = 3600 * 24;
```
maqi: changing the quote expiration will have certain impact on the backward compartibility side.
especially against the mainnet.
The value itself may also need to be further discussed.

maqi (changes requested):
changing the quote expiration will have certain impact on the backward compartibility side.
especially against the mainnet.
The value itself may also need to be further discussed.

Hence put the PR in the Request changes

ant-evm/src/data_payments.rs
grumbach: Indeed, quoting will still be limited to 1 hour on older nodes while newer nodes will be more patient. 
With time, the network as a whole will be upgraded. 
In the meantime we can implement parallel fixes to better support larger file uploads.

autonomi/src/client/data_types/chunk.rs
grumbach: Although I agree that would be an optimization, here split would reduce readability. I'd prefer to avoid premature optimization here as I doubt this is the bottleneck in our code.

ant-evm/src/data_payments.rs
maqi: As David suggested to separate payment and upload, this quoting expiration doesn't need to be extened.
I'd suggest remain this as it is to avoid extra hassle.

grumbach: Ok make sense, will revert to 1 hour then.

maqi (approved):

PR #2945, opened 2025-04-30
feat!: more max fee per gas options

mickvandijke:
Adds more options for setting the max fee per gas in transactions: 

```rust
pub enum MaxFeePerGas {
    /// Use the current market price for fee per gas. WARNING: This can result in unexpected high gas fees!
    #[default]
    Auto,
    /// Use the current market price for fee per gas, but with an upper limit.
    LimitedAuto(u128),
    /// Use no max fee per gas. WARNING: This can result in unexpected high gas fees!
    Unlimited,
    /// Use a custom max fee per gas in WEI.
    Custom(u128),
}
```

BREAKING CHANGE: Updates the publicly exported struct `TransactionConfig`. Affects `evmlib`, `ant-evm` & `autonomi`. Minor version bump will be sufficient.

New options for the `--max-fee-per-gas` CLI arg are:

- `low`: Use the average max gas price bid.
-  `market`: Use the current max gas price bid, with a max of 4 * the average gas price bid. (default)
- `auto`: Use the current max gas price bid. WARNING: Can result in high gas fees!
- `limited-auto:<WEI AMOUNT>`: Use the current max gas price bid, with a specified upper limit.
- `unlimited`: Do not use a limit for the gas price bid. WARNING: Can result in high gas fees!
- `<WEI AMOUNT>`: Set a custom max gas price bid.

TODO:
- [x] Update Python bindings
- [x] Update NodeJS bindings
- [x] ~~Early fail batched transactions if account can't pay gas fees for all~~ Low priority after batching PR: https://github.com/maidsafe/autonomi/pull/2978


ant-cli/src/commands/file.rs
```diff
@@ -43,22 +44,26 @@ pub async fn upload(
...
 
     let mut wallet = load_wallet(client.evm_network()).map_err(|err| (err, IO_ERROR))?;
 
-    if let Some(max_fee_per_gas) = max_fee_per_gas {
-        wallet.set_transaction_config(TransactionConfig::new(max_fee_per_gas))
-    }
+    // todo: @roland which exit code should be used here?
+    let max_fee_per_gas = max_fee_per_gas_param
+        .into_max_fee_per_gas(client.evm_network())
+        .map_err(|err| (err, 61))?;
```
mickvandijke: Hey @RolandSherwin , could you check if using this exit code here is alright?

ant-cli/src/args/max_fee_per_gas.rs
```diff
@@ -0,0 +1,92 @@
+use autonomi::{MaxFeePerGas, Network};
```
maqi, line 9: copyright statement ?

maqi (changes requested):
there shall be some check like
```
let has_enough_fund = wallet.balance > num_of_transaction * max_transaction_fee
``` 
When wallet doesn't have enough fund to cover entire gas_fee, failure shall be thrown at the very beginning BEFORE make any transaction (instead of failed at the 11th, as at https://forum.autonomi.community/t/current-limitations-estimating-eth-costs-to-upload-files-proper-way-to-find-answers/41755/18)
```
Quoting for 1034 chunks..
Paying for 1034 chunks..

   0: Failed to upload file
   1: Failed to upload file
   2: Error occurred during payment.
   3: Wallet error: ChunkPaymentsContract(RpcError(ErrorResp(ErrorPayload { code: -32000, message: "execution reverted", data: None })))
   4: Chunk payments contract error: server returned an error response: error code -32000: execution reverted
   5: server returned an error response: error code -32000: execution reverted

Location:
   ant-cli/src/commands/file.rs:86
```

mickvandijke:
> there shall be some check like
> 
> ```
> let has_enough_fund = wallet.balance > num_of_transaction * max_transaction_fee
> ```
> 
> When wallet doesn't have enough fund to cover entire gas_fee, failure shall be thrown at the very beginning BEFORE make any transaction (instead of failed at the 11th, as at https://forum.autonomi.community/t/current-limitations-estimating-eth-costs-to-upload-files-proper-way-to-find-answers/41755/18)
> 
> ```
> Quoting for 1034 chunks..
> Paying for 1034 chunks..
> 
>    0: Failed to upload file
>    1: Failed to upload file
>    2: Error occurred during payment.
>    3: Wallet error: ChunkPaymentsContract(RpcError(ErrorResp(ErrorPayload { code: -32000, message: "execution reverted", data: None })))
>    4: Chunk payments contract error: server returned an error response: error code -32000: execution reverted
>    5: server returned an error response: error code -32000: execution reverted
> 
> Location:
>    ant-cli/src/commands/file.rs:86
> ```

Hey @maqi,

I've been digging into this and it's been quite tricky. Accurately predicting whether a wallet has enough funds to cover all gas fees upfront is hard, especially since gas prices can fluctuate significantly between transactions and considering all the different gas preset options. The max tx fee for example is almost never the real fee. The real fee is usually significantly lower.

With the batching improvements coming in, I see this as a lower priority. Since only one transaction will be made per batch before uploading. The case where multiple transactions succeed before a final one fails (and invalidates all prior progress) will no longer occur, as uploads will follow each successful payment immediately.

If a transaction fails after this PR and the batching PR, the user will be met with a more descriptive error. When uploading the file again with restocked gas funds, they will just continue where they left off as all previous chunks will already be uploaded.

maqi:
> With the batching improvements coming in, I see this as a lower priority. Since only one transaction will be made per batch before uploading. The case where multiple transactions succeed before a final one fails (and invalidates all prior progress) will no longer occur, as uploads will follow each successful payment immediately.

Yeah, agreed with the PR https://github.com/maidsafe/autonomi/pull/2978 to batching, the `multiple transaction` scenario will then gone.
However, this then means this PR better raised into CLN instead of main branch.

Regarding the `tricking` to figure out, I think the current simple of `num_of_transactions * max_gas_fee` shall be enough to prevent most of case.
Even with certain `margin`, but that's to protect user's waste.
Instead of current that user will for sure waste its fund when go into multi-transaction scenario.

mickvandijke:
Requires CLN: https://github.com/maidsafe/autonomi/pull/2896 to be merged into `main` first.

vphongph:
It's good for me for the Python part (python.rs)

I will add your implementation to my PR, your version is more user friendly and Pythonic

In my version I was just wrapping the original functions, your version is smarter!

mickvandijke:
Working on NodeJS bindings

maqi (approved):

PR #2949, opened 2025-05-04
feat: provide `--alpha` argument

jacderida:
- 3c7896cd2 **chore!: remove the `--testnet` argument**

  BREAKING CHANGE: the `ant`, `antctl` and `antnode` binaries will no longer support the `--testnet`
  argument.

  This argument will no longer be required because we can designate we are using a testnet by setting
  `--network-id` to any value above 2, where 1 will be reserved for `mainnet` and 2 will be reserved
  for `alphanet`.

  Due to maintaining backwards compatibility with the node service data that is used by `antctl`, we
  had to introduce backwards compatibility with the `InitialPeersConfig` struct, thus it now has two
  versions. A new version of `NodeServiceData` was introduced for the removal of the argument, and we
  will also add a new `--alpha` flag argument to this same version.

- 98483e2a7 **feat: provide `--alpha` argument**

  The `ant`, `antnode` and `antctl` binaries will all now support an `--alpha` argument as part of
  their peer configurations.

  When the argument is used, a network ID of 2 will be applied, and the initial peers will be fetched
  from a set of five static IP addresses that have been allocated for the alpha network. So if users
  want to connect to this network, they need only use this switch rather than manually supplying
  `--network-id`, `--network-contacts-url` and/or `--peer` arguments.

- 8440449b0 **feat: select evm network based on network id**

  The doc comments on the `get_evm_network` function provide more detail, but the summary is:
  * The `EVM_NETWORK` and associated variables are checked first. They allow the selection of a preset
    or a custom configuration.
  * Otherwise if these variables are not being used:
      + If no network ID is being used and it's not a local network, we use network ID `1` and so we
        select Arbitrum One.
      + If the `--alpha` argument has been used we use network ID `2` and so we select Arbitrum
        Sepolia Test.
      + If a network ID has been explicitly supplied we are in a testnet scenario and therefore we
        will assert that the `EVM_NETWORK` variables must be used.

  If these conditions are not met the binary will error. I think this is better than selecting a
  default network because when I've been testing with users I've seen this result in confusion about
  exactly which EVM network was being selected.

### DIFF
--- a/ant-cli/src/main.rs
+++ b/ant-cli/src/main.rs
@@ -26,6 +26,7 @@ use color_eyre::Result;
 
 use ant_logging::metrics::init_metrics;
 use ant_logging::{LogBuilder, LogFormat, ReloadHandle, WorkerGuard};
+use ant_protocol::version::ALPHANET_ID;
 use autonomi::version;
 use opt::Opt;
 use tracing::Level;
@@ -34,7 +35,9 @@ use tracing::Level;
 async fn main() -> Result<()> {
     color_eyre::install().expect("Failed to initialise error handler");
     let opt = Opt::parse();
-    if let Some(network_id) = opt.network_id {
+    if opt.peers.alpha {
+        version::set_network_id(ALPHANET_ID);
+    } else if let Some(network_id) = opt.network_id {
         version::set_network_id(network_id);
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -855,6 +855,7 @@ version = "0.3.13"
 dependencies = [
  "ant-build-info",
  "ant-logging",
+ "ant-protocol",
  "autonomi",
  "clap",
  "color-eyre",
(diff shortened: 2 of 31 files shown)


ant-cli/src/main.rs
```diff
@@ -34,7 +35,9 @@ use tracing::Level;
 async fn main() -> Result<()> {
     color_eyre::install().expect("Failed to initialise error handler");
     let opt = Opt::parse();
-    if let Some(network_id) = opt.network_id {
+    if opt.peers.alpha {
+        version::set_network_id(ALPHANET_ID);
+    } else if let Some(network_id) = opt.network_id {
         version::set_network_id(network_id);
     }
```
RolandSherwin, line 42: I'm not sure how these lines crept in, as it was removed in this PR and it is now set inside the `autonomi` lib, rather than at the cli. https://github.com/maidsafe/autonomi/pull/2914/files#diff-6601c049ff7babe23590437ea97dc668f98f20b261768b146788a7019663888aR220

This change has to be moved inside `Autonomi::init_with_config()`

jacderida:
This is being closed in favour of #2952.

PR #2950, opened 2025-05-05
chore(network): extended libp2p::kad::OutboundSubstream timeout

maqi:
### Description

extend libp2p::kad::OutboundSubstream timeout

NOTE: This PR just for test runs to showing the effect of changing such parameter. 
           A new released libp2p version is required to allow the change become effective.


maqi:
Not required as in favour of https://github.com/maidsafe/autonomi/pull/2969, to switch from `kad::put_record_to` to `request/response messaging` .

PR #2954, opened 2025-05-07
feat(API python): added missing classes and methods

vphongph:
Big update on Python API to match Rust API

### DIFF
--- a/autonomi/src/python.rs
+++ b/autonomi/src/python.rs
@@ -1,40 +1,180 @@
-use std::{path::PathBuf, str::FromStr, sync::Arc};
+// Standard library imports
+use std::{path::PathBuf, str::FromStr, sync::Arc, time::Duration};
 
-use crate::client::data::DataAddress;
-use crate::client::files::archive_private::PrivateArchiveDataMap;
-use crate::client::files::archive_public::ArchiveAddress;
-use crate::client::pointer::PointerTarget;
+// External dependencies
+use ant_bootstrap::BootstrapCacheConfig;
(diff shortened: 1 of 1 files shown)


autonomi/src/python.rs
```diff
@@ -1067,6 +1298,195 @@ impl PyClient {
...
+                            .into_iter()
+                            .map(|(peer_id, quote)| {
+                                (
+                                    peer_id.to_string(),
+                                    quote.content.to_string(),
+                                    format!("{:?}", quote.timestamp),
+                                    format!("{:?}", quote.quoting_metrics),
+                                    format!("0x{}", hex::encode(quote.rewards_address.as_slice())),
+                                    hex::encode(&quote.pub_key),
+                                    hex::encode(&quote.signature),
```
grumbach: Would be better to return the actual quotes (wrapped) not stringified to we can use it in other functions here

autonomi/src/python.rs
```diff
@@ -1,40 +1,186 @@
...
+#[pyclass(name = "DataTypes", eq, eq_int)]
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum PyDataTypes {
+    Chunk = 0,
+    GraphEntry = 1,
+    Pointer = 2,
+    Scratchpad = 3,
+}
+
+impl From<PyDataTypes> for DataTypes {
```
b-zee, line 173: Nice impl!

b-zee:
Massive effort Victor, nice one!

- edit - woops closed by accident

PR #2955, opened 2025-05-08
Client light networking rebased

mickvandijke:


ant-cli/src/actions/connect.rs
```diff
@@ -7,73 +7,81 @@
...
-            Client::init_local().await
-        }
-        NetworkId::Main => {
-            progress_bar.set_message("Connecting to The Autonomi Network...");
-            Client::init().await
-        }
-        NetworkId::Alpha => {
-            progress_bar.set_message("Connecting to the Alpha Autonomi Network...");
-            Client::init_alpha().await
-        }
```
grumbach: Are we sure we want to remove that and go back to confusing many-lines client initialization?

ant-cli/src/commands.rs
```diff
@@ -69,6 +70,11 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Experimental: Optionally specify the quorum for the verification of the upload.
+        ///
+        /// Possible values are: "one", "majority", "all", n (where n is a number greater than 0)
+        #[arg(short, long, value_parser = parse_quorum)]
+        quorum: Option<Quorum>,
```
grumbach: Isn't this unused?

ant-cli/src/commands/wallet.rs
```diff
@@ -81,8 +80,8 @@ pub fn export() -> Result<()> {
     Ok(())
 }
 
-pub async fn balance(local: bool, network_id: NetworkId) -> Result<()> {
-    let network = network_id.evm_network(local)?;
+pub async fn balance(local: bool) -> Result<()> {
+    let network = get_evm_network(local, None)?;
```
grumbach: What about alpha? Here we only support live and local!
Cli arguments are ignored!

autonomi/Cargo.toml
```diff
@@ -26,8 +26,9 @@ extension-module = ["pyo3/extension-module", "pyo3-async-runtimes"]
 loud = []
 
 [dependencies]
-ant-bootstrap = { path = "../ant-bootstrap", version = "0.2.1" }
+ant-bootstrap = { path = "../ant-bootstrap", version = "0.2.2" }
 ant-evm = { path = "../ant-evm", version = "0.1.13" }
+ant-networking = { path = "../ant-networking", version = "0.3.11" }
```
grumbach: We shouldn't be needing _ant-networking_ here!

autonomi/src/client/quote.rs
```diff
@@ -104,7 +114,9 @@ impl Client {
             })
             .collect();
 
-        process_tasks_with_max_concurrency(futures, *FILE_UPLOAD_BATCH_SIZE).await
+        let parallism = std::cmp::min(*CHUNK_UPLOAD_BATCH_SIZE * 8, 128);
+
+        process_tasks_with_max_concurrency(futures, parallism).await
```
grumbach, line 119: Why are we ignoring the constant here and doing our own maths?

grumbach (commented):
Left a bunch of comments! Else LGTM! Maybe we should fix the CI before merge? 
Good work!

ant-cli/src/actions/connect.rs
mickvandijke: No we don't, good catch! Restored here: https://github.com/maidsafe/autonomi/pull/2955/commits/e05fdddf833cc0eeb333ed72b68449bc3a49852c

ant-cli/src/commands.rs
mickvandijke: Sharp!

ant-cli/src/commands/wallet.rs
mickvandijke: Another good observation! Restored here: https://github.com/maidsafe/autonomi/pull/2955/commits/5a7867e5431bbc0fc472614b9f52d9803dd9783c

autonomi/Cargo.toml
mickvandijke: Yeeted: https://github.com/maidsafe/autonomi/pull/2955/commits/da01124d9750b009619ae00889cf0ef8b5c34dd2

autonomi/src/client/quote.rs
mickvandijke: Change was introduced here: https://github.com/maidsafe/autonomi/pull/2929

Not sure why we don't allow users to set it lower than 8?

Maybe @maqi can shed some light here

maqi: a quoting task is supposed to be much lighter than the chunk_upload task.
hence got such correlation.

grumbach (approved):
LGTM!

PR #2964, opened 2025-05-13
feat: pointer counter using u64

grumbach:


ant-protocol/src/storage/pointer.rs
```diff
@@ -96,7 +96,10 @@ impl Pointer {
     }
 
     /// Get the bytes that the signature is calculated from
-    fn bytes_to_sign(owner: &PublicKey, counter: u32, target: &PointerTarget) -> Vec<u8> {
+    fn bytes_to_sign(owner: &PublicKey, counter: u64, target: &PointerTarget) -> Vec<u8> {
+        // to support retrocompatibility with old pointers, we need to cast the counter to u32
+        let counter = counter as u32;
```
b-zee: There might be an attack vector here?

Imagine a pointer with counter `0` pointing to target `A`. Then an attacker can reset the pointer to target `A` by using a higher counter that casts to `0u32`.

Or does the counter update have to be `+1` compared to the previous?

grumbach: Yeah absolutely, good catch Benno!

ant-protocol/src/storage/pointer.rs
```diff
@@ -16,7 +16,7 @@ use xor_name::XorName;
 #[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Ord, PartialOrd)]
 pub struct Pointer {
     owner: PublicKey,
-    counter: u32,
+    counter: u64,
```
maqi, line 19: deserialize an old Pointer Record with u32 counter into new u64 one will be problematic as well ?

maqi (approved):

b-zee (approved):

PR #2967, opened 2025-05-14
feat: plain text json archives with retro comp

grumbach:
Makes archives plain text JSON instead of an unreadable blob of bytes.

### DIFF
--- a/autonomi/src/client/high_level/files/archive_private.rs
+++ b/autonomi/src/client/high_level/files/archive_private.rs
@@ -114,7 +114,11 @@ impl PrivateArchive {
 
     /// Deserialize from bytes.
     pub fn from_bytes(data: Bytes) -> Result<PrivateArchive, rmp_serde::decode::Error> {
-        let root: PrivateArchiveVersioned = rmp_serde::from_slice(&data[..])?;
+        let root: PrivateArchiveVersioned = match serde_json::from_slice(&data[..]) {
+            Ok(root) => root,
+            Err(_) => rmp_serde::from_slice(&data[..])?,
+        };
+
         // Currently we have only `V0`. If we add `V1`, then we need an upgrade/migration path here.
         let PrivateArchiveVersioned::V0(root) = root;
 
@@ -124,7 +128,8 @@ impl PrivateArchive {
     /// Serialize to bytes.
     pub fn to_bytes(&self) -> Result<Bytes, rmp_serde::encode::Error> {
         let versioned = PrivateArchiveVersioned::V0(self.clone());
-        let root_serialized = rmp_serde::to_vec_named(&versioned)?;
+        let root_serialized = serde_json::to_vec(&versioned)
+            .map_err(|e| rmp_serde::encode::Error::Syntax(e.to_string()))?;
         let root_serialized = Bytes::from(root_serialized);
 
         Ok(root_serialized)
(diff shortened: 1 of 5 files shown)


autonomi/src/client/high_level/files/archive_private.rs
```diff
@@ -114,7 +114,11 @@ impl PrivateArchive {
 
     /// Deserialize from bytes.
     pub fn from_bytes(data: Bytes) -> Result<PrivateArchive, rmp_serde::decode::Error> {
-        let root: PrivateArchiveVersioned = rmp_serde::from_slice(&data[..])?;
+        let root: PrivateArchiveVersioned = match serde_json::from_slice(&data[..]) {
```
maqi, line 117: Does this mean old node can't support new format ?
this many cause a little bit chaos during upgrade ?

may mark this PR as BREAKING CHANGE ?

maqi (changes requested):

happybeing:
Won't this make it hard to standardise the format? I've not thought about the implications, but that seems an issue with JSON for areas where this needs to be validated because people will no-doubt bypass this if they can go straight to JSON.

I think a blob of unreadable bytes has advantages :rofl:

autonomi/src/client/high_level/files/archive_private.rs
grumbach: Nodes do not read the content of the (encrypted) chunks this will be in. So it shouldn't be an issue for nodes?

grumbach:
Hey @happybeing, glad you're joining the conversation! 
The Archives are stored in encrypted chunks in all cases so the change here is once you've downloaded the data using the datamap: you get the Archive in json format INSTEAD of just bytes that need specific code/software to parse. 
This means if I download the Archive unknowingly of its type in a file, I get something a human can read with the naked eye instead of a serialized archive in binary format. 
This will make archives more accessible. In all cases the security is identical as Archives are stored in encrypted chunks!
Hope that makes it clearer?

> Won't this make it hard to standardise the format?

Good point — also, JSON serialization will result in a larger size.

b-zee:
> Won't this make it hard to standardise the format? I've not thought about the implications, but that seems an issue with JSON for areas where this needs to be validated because people will no-doubt bypass this if they can go straight to JSON.

I think both JSON and MessagePack are equally well standardized. So if we're parsing them, there might even be extra fields in the data that are ignored IIRC. This applies to both msgpack and JSON deserialization libraries.

Do you mean that because it's human readable, people will edit these manually? If so, then the same will happen as if someone serializes a different MessagePack data structure. The data will either be parsed successfully or not. If the archive misses an XOR or the XOR is in an invalid hex string etc, it will just fail to parse. With MessagePack that might not happen practically, because if I recall correctly, it encodes string lengths etc.

autonomi/src/client/high_level/files/archive_private.rs
maqi: Ah, right, on node side, the concern is invalid, as node doesn't need to know about the detailed content of an archive-chunk.

However, the Client has to be updated to support new format.
Meanwhile, the old archives cann't be upgraded, and will be persisted in the network (somehow the same as to the standarize issue that Happybeing pointed out),
as any archive format upgrade will result in new chunk and ripple effect to folders (archive) holding it.

Hence, I think the PR shall at least have a Breaking Change label , and some more notes in the PR description, to highlight this?

happybeing:
> Do you mean that because it's human readable, people will edit these manually? If so, then the same will happen as if someone serializes a different MessagePack data structure

My concern is that if people see JSON they will read *and write* their own JSON and store it "as if it is an Archive" when that may or may not be the case, and that it encourages "tweaks" that can make it harder for apps to interoperate at this fundamental level. Imagine if apps could easily write directory structures to your hard disk instead of using the POSIX APIs. 

Keeping things tied to Rust/serde serialisation makes it hard to bypass the Rust API and "get it wrong".

But maybe your aim is to decouple the types from that API? I also don't know where MessagePack comes into this. Is this about providing APIs that don't sit on top of the Rust libs?

Maybe I'm not clear what the goal is here.

As for keeping to a standard, another 'human readable' format could be used. RDF perhaps, and using the Solid protocol as a model that has containers of objects which can be treated like files. So that's an alternative that may have the benefit of both an existing, more validatable standard, and being human readable rather than a blob.

grumbach:
Closing this for now. Leaving this to future work to unify/improve on Archives.

PR #2972, opened 2025-05-15
Status spinner error antctl

ermineJose:
### Description

adds a progress bar to status command of antctl
improves the verbose of error condition for removing a running service

### DIFF
--- a/ant-node-manager/src/lib.rs
+++ b/ant-node-manager/src/lib.rs
@@ -47,6 +47,8 @@ use ant_service_management::{
 };
 use colored::Colorize;
 use semver::Version;
+use tokio::sync::mpsc;
+use tokio::task;
 use tracing::debug;
 
 pub const DAEMON_DEFAULT_PORT: u16 = 12500;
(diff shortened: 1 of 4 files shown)


ant-node-manager/src/lib.rs
```diff
@@ -558,12 +560,32 @@ pub async fn refresh_node_registry(
...
+
+    // Progress printer task
+    let handle = task::spawn(async move {
+        let mut last_percent = 0;
+        while let Some(processed) = rx.recv().await {
+            let percent = (processed * 100) / total_nodes.max(1);
+            if percent != last_percent {
+                use std::io::{stdout, Write};
+                stdout().flush().unwrap();
+                print!("\rRefresh progress - {:>3}%", percent);
```
RolandSherwin: Can we use `indicatif::ProgressBar` here? It is already being used inside antnode manager.

ant-node-manager/src/lib.rs
```diff
@@ -558,11 +561,23 @@ pub async fn refresh_node_registry(
...
     }
     info!("Refreshing the node registry");
 
+    let total_nodes = node_registry.nodes.len() as u64;
+    // Create a progress bar
+    let pb = ProgressBar::new(total_nodes);
+    pb.set_style(
+        ProgressStyle::default_bar()
+            .template("{spinner:.green} [{elapsed_precise}] [{bar:40.cyan/blue}] {pos}/{len} ({percent}%) ETA: {eta_precise}")
+            .unwrap()
```
RolandSherwin: We cannot use unwraps in the code.. instead, we need to use proper errors.

ant-node-manager/src/lib.rs
```diff
@@ -631,7 +646,11 @@ pub async fn refresh_node_registry(
                 }
             }
         }
+        pb.inc(1);
     }
+
+    pb.finish_with_message("Node registry refresh complete!");
+
```
RolandSherwin: Do not print if `print_refresh_message` is not set.

jacderida (approved):

PR #2975, opened 2025-05-21
feat: remove quote expiration

grumbach:


maqi (changes requested):
better cherry-pick the work from main (which got PR 2976 merged in, that having the same work)

PR #2978, opened 2025-05-21
feat(cli): batch entire quote-pay-upload flow to overcome quote expiracy

maqi:
### Description

The current quote all, pay all, then upload all flow put user in a higher risk in wasting extra fund when have to carry out re-upload due to in-middle failure,
especially for user with poor connection to upload big files or directory containing many files.

This PR change to batch entire quote-pay-upload flow, with a rolling window on chunks level.
It also brings in extra benefits of:
* quote expire issue is no longer a hurdle
* small files got combined together to be processed, so that saving transaction fees
* more user friendly progressing print out
* no more concern of gas_fee payment failure due to multiple transactions, as discussed at [here](https://github.com/maidsafe/autonomi/pull/2945#issuecomment-2900413588)

Changes to the environment settings:

*  FILE_ENCRYPT_BATCH_SIZE : newly added, default to num_of_thread_parallism * 8
*  FILE_UPLOAD_BATCH_SIZE : no longer effective for most cases, only used within Vault upload.
*  UPLOAD_FLOW_BATCH_SIZE: newly added, default to MAX_TRANSFERS_PER_TRANSACTION / 3 . Suggested to be multiples of the value (records-payouts-per-transaction).

The work is to replace https://github.com/maidsafe/autonomi/pull/2944, and will also resolve the `quote expire` issue mostly.

### DIFF
--- a/autonomi/src/client/high_level/files/fs_shared.rs
+++ b/autonomi/src/client/high_level/files/fs_shared.rs
@@ -1,57 +1,201 @@
+// Copyright 2025 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use super::CombinedChunks;
+use crate::client::high_level::data::DataAddress;
+use crate::client::payment::PaymentOption;
 use crate::client::payment::Receipt;
-use crate::client::{ClientEvent, UploadSummary};
+use crate::client::{ClientEvent, PutError, UploadSummary};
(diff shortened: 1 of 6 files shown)


autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -1,57 +1,200 @@
...
+
+/// Number of batch size of an entire quote-pay-upload flow to process.
+/// Suggested to be multiples of 80 (records-payouts-per-transaction).
+///
+/// Can be overridden by the `UPLOAD_FLOW_BATCH_SIZE` environment variable.
+static UPLOAD_FLOW_BATCH_SIZE: LazyLock<usize> = LazyLock::new(|| {
+    let batch_size = std::env::var("UPLOAD_FLOW_BATCH_SIZE")
+        .ok()
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(80);
```
mickvandijke: `UPLOAD_FLOW_BATCH_SIZE` should default to `85`, or even better `MAX_TRANSFERS_PER_TRANSACTION / 3` (= 85 atm).

mickvandijke (approved):
Great work, other from one suggested change, LGTM!

PR #2979, opened 2025-05-22
feat: fix failing uploads with routing table refresh on retries

grumbach:


autonomi/src/client/data_types/chunk.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) static CHUNK_UPLOAD_BATCH_SIZE: LazyLock<usize> = LazyLock::new(|| {
     let batch_size = std::env::var("CHUNK_UPLOAD_BATCH_SIZE")
         .ok()
         .and_then(|s| s.parse().ok())
-        .unwrap_or(1);
+        .unwrap_or(64);
```
maqi, line 42: It is the most robust one preferred.
increasing this value has proved causeing problem to user with normal connection

autonomi/src/client/data_types/chunk.rs
```diff
@@ -51,7 +51,7 @@ pub static CHUNK_DOWNLOAD_BATCH_SIZE: LazyLock<usize> = LazyLock::new(|| {
     let batch_size = std::env::var("CHUNK_DOWNLOAD_BATCH_SIZE")
         .ok()
         .and_then(|s| s.parse().ok())
-        .unwrap_or(1);
+        .unwrap_or(64);
```
maqi, line 54: same, shall not be increased

maqi (changes requested):
Also the PR description is lacking of words to explain what's the reason of the change

maqi:
in favour of https://github.com/maidsafe/autonomi/pull/2980, as the extra get_closest call introduce more work and can not guarantee the issue got resolved (when multiple uploading tasks to be undertaken in parallel)

maqi:
closed as in favour of https://github.com/maidsafe/autonomi/pull/2980,

PR #2980, opened 2025-05-22
fix(client): add target peers address to libp2p before `put_record_to`

mickvandijke:
Adds the target nodes' known addresses into libp2p, before putting records.

Close https://github.com/maidsafe/autonomi/pull/2979


maqi (approved):

grumbach (approved):
LGTM!

mickvandijke:
Reduced the peer cache size from `10_000` to `2_000` because the `verify data location` test kept failing.

PR #2990, opened 2025-05-28
chore(network): remove un-used client only ant_networking methods

maqi:
### Description

With ant_networking crate got the client side work moved out, the crate can be simplified.

The PR contains following simplifications:
* removed no longer used API methods that to be used by client only
* simplified record_store as no need to accomadate the calls from client.
* merge get_closest_from_local methods
* remove dead network sampling code
* remove unused FifoRegister
* remove unused network get_record
* remove unused config API


grumbach:
Nice one Qi! Added some more stripping out client related code in ant-networking here:
https://github.com/maqi/autonomi/pull/3

PR #2992, opened 2025-05-30
feat: get check after put

mickvandijke:
Ignore the `PutRecordQuorumFailed` error on `PutRecordTo` as it is not reliable. Instead, introduce a manual get record check to verify the put. This will likely slow down the upload throughput, but it will be more reliable.

The issue with the `PutRecordQuorumFailed` errors being thrown comes from increasing the `CHUNK_UPLOAD_BATCH_SIZE`. The error is thrown even though the put can still have been successful.

### DIFF
--- a/autonomi/src/networking/mod.rs
+++ b/autonomi/src/networking/mod.rs
@@ -189,6 +189,7 @@ impl Network {
         quorum: Quorum,
     ) -> Result<(), NetworkError> {
         let (tx, rx) = oneshot::channel();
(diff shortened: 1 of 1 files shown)


grumbach (changes requested):
I suggest we do the GET only when we get a PutQuorum error. This way we keep good performance when there is no error and eliminate false positives when this error occurs. Will open a PR with the changes.

mickvandijke:
Replaced by #2994

PR #2993, opened 2025-05-30
feat: support network id from 0-255

jacderida:
- 75c6aa7d8 **feat: support network id from 0-255**

  There was a bit of a misunderstanding in the way the `NetworkId` was implemented. It was reserving
  the values 0, 1, and 2, for local, main, and alpha networks, respectively, then assigning 3 for any
  other 'custom' network.

  We change it here such that 0, 1, and 2 are reserved as above, but custom networks are permitted to
  be between 3 and 255.

- a2b85174e **chore: standardise autonomi usage in connection messages**

  Our network is called 'Autonomi', not 'the Autonomi Network', and therefore only 'Autonomi' should
  be capitalised to denote it as a proper noun.

  The words 'alpha', 'local', and 'custom' are variants of Autonomi networks, so they should not be
  capitalised.

  Other usages of the word 'network' being capitalised were changed because 'network' is not a proper
  noun in that context.

- ad9ecd27f **chore: do not log result for `GetRecord`**

  Writing the result was causing excessive logging for `file download` commands.

### DIFF
--- a/ant-cli/src/actions/connect.rs
+++ b/ant-cli/src/actions/connect.rs
@@ -39,25 +39,25 @@ pub async fn connect_to_network_with_config(
 ) -> Result<Client, ExitCodeError> {
     let progress_bar = ProgressBar::new_spinner();
     progress_bar.enable_steady_tick(Duration::from_millis(120));
-    progress_bar.set_message("Connecting to The Autonomi Network...");
+    progress_bar.set_message("Connecting to the Autonomi network...");
     let new_style = progress_bar.style().tick_chars("⠁⠂⠄⡀⢀⠠⠐⠈🔗");
     progress_bar.set_style(new_style);
 
     let res = match network_context.network_id.as_u8() {
         LOCAL_NETWORK_ID => {
-            progress_bar.set_message("Connecting to a local Autonomi Network...");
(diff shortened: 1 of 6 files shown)


ant-cli/src/actions/connect.rs
```diff
@@ -39,28 +39,28 @@ pub async fn connect_to_network_with_config(
 ) -> Result<Client, ExitCodeError> {
     let progress_bar = ProgressBar::new_spinner();
     progress_bar.enable_steady_tick(Duration::from_millis(120));
-    progress_bar.set_message("Connecting to The Autonomi Network...");
+    progress_bar.set_message("Connecting to Autonomi...");
```
maqi: `Autonomi` refers to client app as well
maybe change to `Connection to an Autonomi network ...` ?

maqi (approved):

ant-cli/src/actions/connect.rs
b-zee: [autonomi.com](https://autonomi.com) does mention 'The Autonomi Network', which can be referred to as 'Autonomi' in the above case IMO.

> Autonomi refers to client app

Then I think that would be referred to as an Autonomi client.

PR #2994, opened 2025-06-02
On quorum error try get records

grumbach:
Don't trust PutRecordQuorumFailed error on PutRecordTo as it is not reliable. Introduce a manual get record check to verify the put. This will NOT slow down the upload throughput, and it will be more reliable. 

The issue with the PutRecordQuorumFailed errors being thrown comes from increasing the CHUNK_UPLOAD_BATCH_SIZE. The error is thrown even though the put can still have been successful.

I suggest we replace #2992 by this PR. What do you think @mickvandijke?


mickvandijke (approved):
LGTM!

PR #3003, opened 2025-06-04
feat: early return `GetRecord` request in CLN if quorum is met

mickvandijke:
CLN did not really do anything with the `GetRecord` quorum. This change will make it early return a record and its holders if the specified quorum is met.

CLN download speed is now faster on average compared to `stable`.

TODO:
- [x] Gracefully end the `GetRecord` internal Kad query when early returning.

### DIFF
--- a/autonomi/src/networking/driver/task_handler.rs
+++ b/autonomi/src/networking/driver/task_handler.rs
@@ -8,13 +8,14 @@
 
 use ant_evm::PaymentQuote;
 use ant_protocol::{NetworkAddress, PrettyPrintRecordKey};
-use libp2p::kad::{self, PeerInfo, QueryId, Record};
+use libp2p::kad::{self, PeerInfo, QueryId, Quorum, Record};
 use libp2p::request_response::OutboundRequestId;
 use libp2p::PeerId;
 use std::collections::HashMap;
 use thiserror::Error;
 
 use crate::networking::interface::NetworkTask;
+use crate::networking::utils::get_quorum_amount;
 use crate::networking::NetworkError;
 use crate::networking::OneShotTaskResult;
 
@@ -46,7 +47,7 @@ pub(crate) struct TaskHandler {
             PeerInfo,
         ),
     >,
-    get_record: HashMap<QueryId, OneShotTaskResult<RecordAndHolders>>,
+    get_record: HashMap<QueryId, (OneShotTaskResult<RecordAndHolders>, Quorum)>,
     get_record_accumulator: HashMap<QueryId, HashMap<PeerId, Record>>,
 }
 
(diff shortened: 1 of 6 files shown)


autonomi/src/networking/driver/task_handler.rs
```diff
@@ -132,35 +133,32 @@ impl TaskHandler {
...
+
                 if let Some(peer_id) = record.peer {
                     holders.insert(peer_id, record.record);
                 }
+
+                // If we have enough holders, finish the task.
+                if let Some((_resp, quorum)) = self.get_record.get(&id) {
+                    let expected_holders = get_quorum_amount(quorum);
+
+                    if holders.len() >= expected_holders {
```
maqi, line 157: for the situation of mutable data having too many versions (or nasty nodes respond chunk with corrupted data),
such simple check of `number of holders` could introduce potential risk ?

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -201,6 +199,44 @@ impl TaskHandler {
         Ok(())
     }
 
+    pub fn finish_get_record(&mut self, id: QueryId) -> Result<(), TaskHandlerError> {
+        // TODO: gracefully terminate the kad query in case we finish early.
```
maqi: could try the following kad facility to stop network query earilier:
```rust
                // Stop the query; possibly stops more nodes from being queried.
                if let Some(mut query) = self.swarm.behaviour_mut().kademlia.query_mut(&query_id) {
                    query.finish();
                }
```

maqi (changes requested):

autonomi/src/networking/driver/task_handler.rs
mickvandijke: Would increasing the `Quorum` that can be passed to the GetRecord request mitigate that risk?

mickvandijke: The current logic is aligned with the `stable`/`main` branch:
https://github.com/maidsafe/autonomi/blob/b5dbfa4568e56b84035f1a1c0889722a36ffa38e/ant-networking/src/event/kad.rs#L392

However, I agree that we should address the possibility of malicious nodes. In that case, I think we should add an exception for split records: if only a single (malicious) node has a different record, we should return the record held by the majority of peers. We should then also increase the GetRecord request quorum/minimum holders to `3` or higher.

In my opinion, though, this should be handled in a separate PR, as it introduces changes to the logic compared to `stable`.

maqi: >  we should return the record held by the majority of peers.

no, the `split` records shall always be returned

I just mean, in case of split, instead of `return on Quorum holders`, we shall `return on single version has Quorum holders` with all versions collected

> this should be handled in a separate PR

right, ok with leave this to separate PR, as I's also suggest to use req/rsp DM to replace kad::get_record

maqi (approved):

PR #3004, opened 2025-06-04
feat: tests with nodes spawner

vphongph:
1st PR about nodes spawner in unit tests:

- No more [serial] in Autonomi crate, all the tests are independent and can be ran in parallel! And testing speed is faster!
- Also added nodes spawner to -> ant-node/tests/data_with_churn.rs

/!\ Terminal/Shell can limit the number of opened file descriptors so increase the limit before launching multi threading tests!

To be done/check:
- ant-networking/src/record_store.rs -> Seems to have data race, can_store_after_restart() fails in multi threaded test, add [serial] to this set as temporary fix
- ant-node/tests/verify_data_location.rs -> Can't get RPC addresses, can't get them from nodes and no registry created with spawning network. The test fails with background network as well
- ant-node-manager/tests/e2e.rs -> The test fails with spawned or background network, is it a CI only test?
- evmlib/tests/wallet.rs -> Not sure how to handle the client/spawned network. The test fails with background network as well

EDIT of commit 1ad9a26873c572dac793a2c4d915aad3bf824ace:

ant-node/network_spawner.rs
- Readability refactor
- Attempt of creating independent spans but hard stuck (read my issue on top of this message) :x:

ant-logging/layers.rs
- Readability refactor
- Added test modules auto detection to have the necessary targets, otherwise the Log Router would filter out logs from the test modules/targets
- Added client span detection for the writer in the original Log Router from PR 3046
- Added whole new Log Router to have spans with unique name (span name + id + test name)
It fixed data race during multi-threaded tests with node spawner ✅

ant-logging/lib.rs
- Readability refactor
- Data structure refactor to have all the vars and guards in one struct
- Added functions to use the Log router with unique name spans (cf layers.rs)
- Added functions to copy the temporary node directories (where the logs are written) to the actual node directories created by ant-node during node spawning
It mirrors the original behavior of the logs with a background network
I had to do it this way because the logging system has to be initialized before the node spawning, otherwise some logs would be lost
- Added functions to delete the temporary directories
- Added appenders to create unique directories with test name
It fixed data race during multi-threaded tests (when the temporary directories are deleted) ✅

ant-logging/appender.rs
- Added new file rotater with test name to create unique log file
It fixed tests within the same module mixing their logs ✅

ant-logging/multi_node_logging.rs
- Refactored this test to adapt it to the new logging paths
autonomi/address.rs
- Implemented multi node logging in this test
Once everything is confirmed as correct, this implementation will be spread on the other tests


mickvandijke:
> ant-node/tests/verify_data_location.rs -> Can't get RPC addresses, can't get them from nodes and no registry created with spawning network. The test fails with background network as well

I believe @ermineJose was working on removing RPC alltogether. Maybe he has already resolved this issue?

mickvandijke:
>evmlib/tests/wallet.rs -> Not sure how to handle the client/spawned network. The test fails with background network as well

Ah yeah I need to fix this one, it's not related to the spawning networks. You could add an #[ignore] to it for this PR.

mickvandijke:
> ant-node-manager/tests/e2e.rs -> The test fails with spawned or background network, is it a CI only test?

This test is also broken lol. You can add an ignore here and a `TODO: fix this`.

mickvandijke:
Nice work Victor!! I think everyone will be really happy with this PR. No more having to start nodes manually before testing!! 🚂

mickvandijke (approved):

jacderida (approved):

RolandSherwin (changes requested):
Hey @vphongph! Really great work on making the tests run using cargo test! 
I would just like to point out that we'd still have the same issue of the logs going into a single file as mentioned in this prior PR https://github.com/maidsafe/autonomi/pull/2922/, which made us revert some CI changes that tried to use `NodeSpawner`

I tried some custom logging setup to solve this earlier, but failed / got distracted with other tasks. But I think we should address that issue first before merging this PR..

b-zee:
I do want to point out that the `cargo test --package autonomi --tests -- --nocapture` step in CI increases from roughly 10 minutes to 25 minutes. Now this has the the benefit that no local network has to be built separately and started (which varies from 2-10 minutes).

Maybe there are some small optimizations to reduce the footprint of starting a network? Maybe start one network per test suite or file? Or optimize the interval etc?

vphongph:
@b-zee 

Good catch! It's true now the Windows VM is super slow.

I didn't notice it right away as on my mac I gained speed on local tests (probably I can multi thread a lot).

I can try a version with one spawn per test module. It should reduce the ressource use but less parallelism will be allowed since they will share same wallet and evm network.

maqi:
Closed, as explained by Victor that though this PR brings different features, but it was blocked by the logs issue that remained unresolved.

Full report is in the slack discussions.

PR #3011, opened 2025-06-08
making no-archive the default and adding the --archive / -a flag 

rid-dim:
### Description

Since I'm doing single file uploads repeatedly atm (uploading friends themes, background images, JS Webcomponents I want to include in other WebApps, ...) where I always just use the datamap and the chunks of the file but never the archive I created this mod for me and from what I hear there's multiple people out there who would like this feature too.

Archives have their place and are good for folder upload/download/syncing ... but if I just want an object or content online on autonomi there is absolutely no reason to create and upload an archive too.

---

side note: I tested downloading a file from raw datamap and the client already supports it - so there doesn't seem to be a breaking change on the other side that would prevent changing default behavior imho

### Related Issue

only general complaints / usability issues around not being able to upload content without creating the archive with the ant client

### Type of Change

- [x] Documentation update

### Checklist

- [x] I've tested locally with my changes.
- [x] I have tried to add commit messages with [commitlint](https://commitlint.js.org/#/).

### DIFF
--- a/ant-cli/src/commands.rs
+++ b/ant-cli/src/commands.rs
@@ -70,6 +70,13 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Create and upload archive after file upload. This enables traditional archive functionality.
+        /// The archive preserves directory structure, filenames, and metadata of uploaded folders.
+        /// This ensures the original folder structure and file naming are maintained after download.
+        /// Without this flag, all files are uploaded as individual objects, losing filenames, metadata, and directory structure.
+        /// Using this option incurs additional costs.
+        #[arg(short, long)]
+        archive: bool,
         /// Optional: Specify the maximum fee per gas in u128.
         #[arg(long)]
         max_fee_per_gas: Option<u128>,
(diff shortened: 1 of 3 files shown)


jacderida:
Awesome, thanks for the contribution! Would you mind squashing your commits into one please?

ant-cli/src/commands.rs
```diff
@@ -70,6 +70,9 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Create and upload archive after file upload. This enables traditional archive functionality.
```
jacderida, line 73: Could we perhaps state the purpose of the archive here? I think it's for storing metadata.

We could also mention here that it will cost more for the archive.

rid-dim: I've added a more lengthy description - not sure it's too long now - but might not be an issue since I think it clearly states now what happens (which is good imho)

good thought - nobody new to this will have an idea what the archive flag is about

grumbach:
Partially covered in #3017
- kept archives as default but add `--no-archive` flag to avoid them
- single files are not archives by default!

PR #3016, opened 2025-06-10
chore: add debug trait to client and wallet

### LINKED ISSUE #2996: Debug trait for Client and Wallet
Rust guidelines suggest, that every type should implement Debug.
https://rust-lang.github.io/api-guidelines/debuggability.html#c-debug

Is there a reason for these types (and possibly more) do not implement it?

grumbach:
fixes https://github.com/maidsafe/autonomi/issues/2996


mickvandijke (approved):
Much prefer this one over #3012 !

PR #3017, opened 2025-06-10
feat: no archive flag for cli and default to that for single file upload

grumbach:
Based on the excellent suggestions by @rid-dim here:
- https://github.com/maidsafe/autonomi/pull/3010
- https://github.com/maidsafe/autonomi/pull/3011

This PR:
- adds the `--no-archive` flag as excellently suggested by @rid-dim
- but keep archives as the default for directories so upload and download commands stay symmetric
- single file uploads never generate `archive` (user can only put single file in a directory to have an archive, if really wanted)

### DIFF
--- a/ant-cli/README.md
+++ b/ant-cli/README.md
@@ -17,7 +17,7 @@ ant [OPTIONS] <COMMAND>
 
 ### File
 - `file cost <file>`
-- `file upload <file> [--public]`
+- `file upload <file> [--public] [--no-archive]`
 - `file download <addr> <dest_file>`
 - `file list`
 
@@ -155,15 +155,16 @@ Expected value:
 
 #### Upload a file
 ```
-file upload <file> [--public]
+file upload <file> [--public] [--no-archive]
 ```
 Uploads a file to the network.
 
 Expected value: 
 - `<file>`: File path (accessible by current user)
 
-The following flag can be added:
-`--public` (Optional) Specifying this will make this file publicly available to anyone on the network
+The following flags can be added:
+- `--public` (Optional) Specifying this will make this file publicly available to anyone on the network
+- `--no-archive` (Optional) Skip creating local archive after upload. Only upload files without saving archive information
 
 #### Download a file
 ```
--- a/ant-cli/README.md
+++ b/ant-cli/README.md
@@ -164,7 +164,7 @@ Expected value:
 
 The following flags can be added:
 - `--public` (Optional) Specifying this will make this file publicly available to anyone on the network
-- `--no-archive` (Optional) Skip creating local archive after upload. Only upload files without saving archive information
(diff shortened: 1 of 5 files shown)


ant-cli/README.md
```diff
@@ -155,15 +155,16 @@ Expected value:
 
 #### Upload a file
 ```
-file upload <file> [--public]
+file upload <file> [--public] [--no-archive]
```
maqi, line 158: the PR description says:
```
although make --no-archive the default for single file uploads (folk can still put singles files as an archive by putting them in a directory)
```

which indicates a totally different `default` behaviour of archiving.

This ReadMe might need to be updated to reflect the difference by:
having separate sections of : file and directory

ant-cli/src/commands.rs
```diff
@@ -70,6 +70,12 @@ pub enum FileCmd {
         /// Upload the file as public. Everyone can see public data on the Network.
         #[arg(short, long)]
         public: bool,
+        /// Skip creating an archive after uploading a directory.
+        /// When uploading a directory, files are normally grouped into an archive for easier management.
+        /// This flag uploads the files individually without creating the archive metadata.
+        /// Note: This option only affects directory uploads - single file uploads never create archives.
```
maqi, line 76: > single file uploads never create archives

the PR description or help doc of the opt both indicates `no-archive` is a flag of an optional archive.
i.e. even when upload single file, there could be `archive` created.

at least one of the sides is incorrect and need to be corrected/synced.

ant-cli/src/commands/file.rs
```diff
@@ -159,20 +131,80 @@ pub async fn upload(
     }
     info!("Summary for upload of file {file} at {local_addr:?}: {summary:?}");
 
-    // save to local user data
-    let writer = if public {
-        crate::user_data::write_local_public_file_archive(archive, &name)
+    // save archive to local user data
+    if !no_archive && not_single_file {
```
maqi, line 135: this means for a single file, there will be no archive, not matter public or private.

ant-cli/src/commands/file.rs
```diff
@@ -159,20 +131,80 @@ pub async fn upload(
...
+            .wrap_err("Failed to save file to local user data")
+            .with_suggestion(|| "Local user data saves the file address above to disk, without it you need to keep track of the address yourself")
+            .map_err(|err| (err, IO_ERROR))?;
+        info!("Saved file to local user data");
+    }
+
+    Ok(())
+}
+
+/// Uploads a directory to the network and prints the content and addresses. Returns the archive address and the address to access the data.
```
maqi: > Uploads a directory to the network

the input `dir_path` could be a single file

maqi (changes requested):
some comments/expample across PR description/ReadMe/help_doc/code_comment better to be updated and synced.

ant-cli/src/commands/file.rs
grumbach: Yes that's intended, will fix the docs!

maqi (approved):

PR #3021, opened 2025-06-10
chore(network): move ConnectionClosed logging to debug level

RolandSherwin:
- Do not log `ConnectionClosed` as an error msg.
- Reduce the logging level for connection errors (and other misc) from `warn/error` to `debug`. We still log the errors from `dial_error_to_str` & `listen_error_to_str` at the `Level::Error`. Thus, we're just avoiding double logging here. And here is the reasoning:
	- Old connection_error logs: Contained `connection id` and was useful to track errors from various peers. BUT to correlate the `connection id`, the other events like `ConnectionEstablished` / `IncomingConnection` etc were logged at `DEBUG` level. Thus this log line at ERROR level is kinda redundant.
	- New connection_error logs: These don't contain `connection_id` but are used instead to track errors src/dst in our internal dashboard. These are already in `ERROR` level. The loss of usable information would be just from the error string, like `TransportError::Other`, at which point we'd need connection ids to correlate, so we can always change the log level if we want to dig deeper.

### DIFF
--- a/ant-networking/src/event/swarm.rs
+++ b/ant-networking/src/event/swarm.rs
@@ -296,13 +296,21 @@ impl SwarmDriver {
                 error,
             } => {
                 event_string = "OutgoingConnErrWithoutPeerId";
+
                 warn!("OutgoingConnectionError on {connection_id:?} - {error:?}");
+
                 let remote_peer = "";
-                for error_str in dial_error_to_str(&error) {
-                    error!(
-                        "Node {:?} Remote {remote_peer:?} - Outgoing Connection Error - {error_str:?}",
-                        self.self_peer_id,
-                    );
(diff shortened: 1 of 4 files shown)


ant-networking/src/event/swarm.rs
```diff
@@ -648,6 +664,7 @@ impl SwarmDriver {
     // Do not log IncomingConnectionError if the the send_back_addr is the same on the adjacent established connections.
     //
     // We either check by IP address or by `/p2p/<peer_id>` for relayed nodes.
+    #[allow(dead_code)]
```
RolandSherwin, line 667: Will be used for #3007

maqi (approved):

PR #3024, opened 2025-06-11
Revert "Merge pull request #3022 from RolandSherwin/revert_addrs_from…

RolandSherwin:
- We had to revert https://github.com/maidsafe/autonomi/pull/2947 back before the release. So this PR aims to "revert the revert" to test this feature again.
- Also has changes from https://github.com/maidsafe/autonomi/pull/2981
- And adds in commit from https://github.com/maidsafe/autonomi/pull/3007


ant-networking/src/driver.rs
```diff
@@ -61,6 +61,9 @@ pub(crate) const CLOSET_RECORD_CHECK_INTERVAL: Duration = Duration::from_secs(15
 /// Interval over which we query relay manager to check if we can make any more reservations.
 pub(crate) const RELAY_MANAGER_RESERVATION_INTERVAL: Duration = Duration::from_secs(30);
 
+/// Interval over which we check if we could dial any peer in the dial queue.
+const DIAL_QUEUE_CHECK_INTERVAL: Duration = Duration::from_secs(2);
```
grumbach, line 65: Does that mean we dial peers every 2 seconds?

RolandSherwin: No. We check if any peers are ready to dial from the list (every 2s).

RolandSherwin: We could do something else to make the check better here? Maybe I'll ask claude later..

maqi (approved):

PR #3025, opened 2025-06-11
chore(docs): update readme and contributing files

RolandSherwin:


CLAUDE.md
```diff
@@ -0,0 +1,135 @@
...
+All data operations require EVM payments. When developing:
+- Use the local EVM testnet for development (`cargo run --bin evm-testnet`)
+- Set SECRET_KEY environment variable for client operations
+- Payments are handled through the evmlib crate
+
+### Network Architecture
+- Nodes form a Kademlia DHT for routing and data location
+- Data is stored as encrypted chunks across multiple nodes
+- QUIC is the primary transport protocol
+- Network discovery uses bootstrap peers from ant-bootstrap
```
maqi: that's `initial` network discovery
there is also `peroidical network discovery` (on node side only), which is achieved via querying certain network distance space with a round-robin style.

CLAUDE.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Payment System Integration
+All data operations require EVM payments. When developing:
+- Use the local EVM testnet for development (`cargo run --bin evm-testnet`)
+- Set SECRET_KEY environment variable for client operations
+- Payments are handled through the evmlib crate
+
+### Network Architecture
+- Nodes form a Kademlia DHT for routing and data location
+- Data is stored as encrypted chunks across multiple nodes
```
maqi: what nodes stores are `records`, `necrypted chunks` is just one type.
there is also other `types`.

records will be encrypted by holder before flush to disk, to avoid `malicious raw record upload`.

maqi (approved):

PR #3036, opened 2025-06-13
retry strategy for ant file upload

### LINKED ISSUE #3026: [CLI] Retry strategy for upload failures
Image

the current (of the upcoming ant client) behavior of just erroring out on every error of a batched upload does make it pretty useless to really go for large uploads.

i personally would prefer a "just retry with same settings 1 minute later" flag
then we may try to go for uploads for some hours every minute while gas prices are high ... but who cares ...
(ideally retrying the chunk with failed chunk-proof wouldn't happen isolated but simply with the next (shortened by the failed chunks) batch of chunks) ... from what i observe it usually works on next upload attempt ... but i'm sure that's more difficult so just "keep pushing" would be cooler than the current retrying in my terminal and seeing it always start from the beginnning :D)

rid-dim:
### Description

as mentioned in discord I am running a modified ant client that does retry (once a minute) on upload-fails and just repeats the failed batch resulting in reliable and resilient uploads without wasting eth on high fees or time on restarting the upload from scratch on fail. It works really well and makes uploading to main suddenly a really good experience. 

this is a significantly cleaner version than the one I used before ... so maybe this could be used for basing a mod yourself on it. To me it still somehow looks like changes on too many places that look too similar but it really might be because of the architecture of the file upload ... not sure. I'm sure you're more qualified to judge and create a retrying client based on these ideas or decide to push it back due to priorities.

I tested with different combination of flags and it works for me pretty well (but I'm sure your tests are more complete than mine ;-) )

---

fee error:

> ⚠️  ❌ Gas fee too low!
💰 Your max fee per gas: 1000 wei
📈 Network base fee: 10000000 wei
💡 Increase your --max-fee-per-gas if you want the upload to be executed faster. Retrying after 1 minute pause...

chunk proof error:

> ⚠️  ❌ Upload batch failed: 1 chunks failed. Retrying after 1 minute pause...
🔄 Retrying upload...
....
Paying for 1 chunks..
Chunk payments of 1 chunks completed. 31 chunks were free / already paid for
35 chunks were free in this batch 32

### Related Issue

Fixes #3026 .

### Type of Change

- [x] Documentation update

### Checklist

### DIFF
--- a/ant-cli/src/commands/file.rs
+++ b/ant-cli/src/commands/file.rs
@@ -48,6 +48,7 @@ pub async fn upload(
     no_archive: bool,
     network_context: NetworkContext,
     max_fee_per_gas_param: Option<MaxFeePerGasParam>,
+    retry_failed: bool,
 ) -> Result<(), ExitCodeError> {
     let config = ClientOperatingStrategy::new();
 
@@ -85,28 +86,34 @@ pub async fn upload(
 
     // upload dir
     let not_single_file = !dir_path.is_file();
-    let (archive_addr, local_addr) =
-        match upload_dir(&client, dir_path, public, no_archive, payment).await {
-            Ok((a, l)) => (a, l),
-            Err(UploadError::PutError(PutError::Batch(upload_state))) => {
-                let res = cached_payments::save_payment(file, &upload_state);
-                println!("Cached payment to local disk for {file}: {res:?}");
-                let exit_code =
-                    upload_exit_code(&UploadError::PutError(PutError::Batch(Default::default())));
-                return Err((
-                    eyre!(UploadError::PutError(PutError::Batch(upload_state)))
-                        .wrap_err("Failed to upload file".to_string()),
-                    exit_code,
-                ));
-            }
-            Err(err) => {
-                let exit_code = upload_exit_code(&err);
-                return Err((
-                    eyre!(err).wrap_err("Failed to upload file".to_string()),
-                    exit_code,
-                ));
-            }
-        };
+    let result = if retry_failed {
+        println!("🔄 Retry mode enabled - will persistently retry failed chunks until successful");
+        upload_dir(&client, dir_path, public, no_archive, payment, true).await
+    } else {
+        upload_dir(&client, dir_path, public, no_archive, payment, false).await
+    };
+
+    let (archive_addr, local_addr) = match result {
+        Ok((a, l)) => (a, l),
+        Err(UploadError::PutError(PutError::Batch(upload_state))) => {
+            let res = cached_payments::save_payment(file, &upload_state);
+            println!("Cached payment to local disk for {file}: {res:?}");
+            let exit_code =
+                upload_exit_code(&UploadError::PutError(PutError::Batch(Default::default())));
+            return Err((
+                eyre!(UploadError::PutError(PutError::Batch(upload_state)))
+                    .wrap_err("Failed to upload file".to_string()),
+                exit_code,
+            ));
+        }
+        Err(err) => {
+            let exit_code = upload_exit_code(&err);
+            return Err((
+                eyre!(err).wrap_err("Failed to upload file".to_string()),
+                exit_code,
+            ));
+        }
+    };
 
     // wait for upload to complete
     if let Err(e) = upload_completed_tx.send(()) {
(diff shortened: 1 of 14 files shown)


autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -130,15 +207,28 @@ impl Client {
...
         // Take next batch of chunks (up to UPLOAD_FLOW_BATCH_SIZE)
-        let batch: Vec<_> = remaining_chunks
-            .drain(..std::cmp::min(remaining_chunks.len(), *UPLOAD_FLOW_BATCH_SIZE))
-            .collect();
+        let batch_size = std::cmp::min(aggregated_chunks.len(), *UPLOAD_FLOW_BATCH_SIZE);
+        
+        // Important: Don't drain chunks yet - we might need to retry them
+        let batch: Vec<_> = if retry_on_failure {
+            // For retry mode, clone the chunks so we can retry if needed
+            aggregated_chunks[..batch_size]
```
maqi, line 177: this actually carry out retries on ALL chunks in the batch, which is bit a waste of time/traffic, as need to carry out quoting at least for those succeeded ones.
Also, carry out retry immediately (even after 1 min) in most case will just be a waste, as the network wide general status (gas_fee or nodes' distribution/connection status) won't change that much.

Hence, only those `failed` chunks shall be appended to the tail, to be re-tried later on.

autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -101,13 +122,69 @@ impl Client {
...
+                retry_on_failure,
             )
-            .await?;
+            .await;
+
+            match batch_result {
+                Ok(()) => continue,
+                Err(err) if retry_on_failure => {
+                    // Helper function to extract gas values from error message
+                    fn extract_gas_values(err_str: &str) -> Option<(String, String)> {
```
maqi: prefer move this out as a separate utility function

autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -101,13 +122,69 @@ impl Client {
...
+                                    return Some((max_fee.to_string(), base_fee.to_string()));
+                                }
+                            }
+                        }
+                        None
+                    }
+
+                    // Format error message for user
+                    let err_str = format!("{err:?}");
+                    let error_msg = if err_str.contains("max fee per gas less than block base fee") {
```
maqi: prefer make this `error_msg` construction as a separate utiility function

autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -101,13 +122,69 @@ impl Client {
...
+                    };
+                    
+                    println!("⚠️  {}. Retrying after 1 minute pause...", error_msg);
+                    info!("Upload error: {}. Retrying in 1 minute...", err);
+                    
+                    // Wait 1 minute before retry
+                    sleep(Duration::from_secs(60)).await;
+                    println!("🔄 Retrying upload...");
+                    
+                    // Continue the loop to retry with the same chunks
```
maqi, line 138: the retry shall be carried out on entire chunks, instead of per batch.
also, better with a controllable terminate condiation, such as no more than 3 times ?

ant-cli/src/commands/file.rs
```diff
@@ -160,13 +167,20 @@ async fn upload_dir(
     public: bool,
     no_archive: bool,
     payment_option: PaymentOption,
+    retry_failed: bool,
```
maqi: make `retry_failed` a member field of Client, shall avoid have to passing this flag around and avoid many of refactors across the file.rs, fs_private.rs and fs_public.rs

maqi (changes requested):

autonomi/src/client/high_level/files/fs_shared.rs
rid-dim: > Also, carry out retry immediately (even after 1 min) in most case will just be a waste, as the network wide general status (gas_fee or nodes' distribution/connection status) won't change that much.

actually I have to disagree with this. for failed chunks on retry it often results in either them suddenly being skipped because they "exist" or the upload works on the next attempt. I've not seen uploads fail more than once.

fees tend to fluctuate often too and I've often seen success on 2nd attempt even if it's just 1 minute later

on the wasteful part of retrying the whole batch I agree 100% but for my mod I didn't care and I'm glad it retries only the batch and doesn't start with self-encrypting gigabytes from start again ...

autonomi/src/client/high_level/files/fs_shared.rs
rid-dim: > better with a controllable terminate condiation, such as no more than 3 times ?

if so then please configurable because infinite is okay - I can always hit ctrl+c later when it hangs forever - but getting back to the point where it started failing due to high fees can be very painful and a long wait ...

ant-cli/src/commands/file.rs
rid-dim: that really makes it significantly simpler!

autonomi/src/client/high_level/files/fs_shared.rs
rid-dim: I did move them to the client src/client/utils.rs (and left the tests in that Claude added) I like that change

autonomi/src/client/high_level/files/fs_shared.rs
rid-dim: I did move them to the client src/client/utils.rs (and left the tests in that Claude added) I like that change

maqi:
Closed as in favour of PR https://github.com/maidsafe/autonomi/pull/3038, which was created on top of this PR.

PR #3037, opened 2025-06-16
feat: make chunk_batch_upload public

b-zee:
This method is useful in case users want to manually pay and separately upload chunks. This is the case for Project Dave where private files are manually encrypted, paid for and uploaded.


maqi (changes requested):
Related document, even the maidsafe\docs repo needs to be updated.

autonomi/examples/chunk_batch_upload.rs
```diff
@@ -0,0 +1,47 @@
+/// This example demonstrates how to upload multiple chunks of data in a single transaction using the `chunk_batch_upload` method.
```
maqi: need a copyright claim

maqi (approved):

PR #3038, opened 2025-06-16
Retry on upload failure

### LINKED ISSUE #3026: [CLI] Retry strategy for upload failures

maqi:
### Description

This work is on top of PR https://github.com/maidsafe/autonomi/pull/3036 to carry out auto retry on upload failure.
Which avoids forcing user to carry out re-uploads manually and have to go through the encryption and quoting uploaded chunks again and again.

Newly added command line opt:
* `retry-failed` :  default to be 0 indicating not carry out retry on upload failure. 
                          set to be the times of total content chunks.
                          Will carry out round-robin style (push the failed chunks to the tail of the upload task queue as scheduled for retry) upto the max threshold (X times total content chunks).

Example usage: `--retry-failed 3`

Thanks for the @rid-dim , here is a screenshot (with detailed chunk/file info blurred) of a upload run against the production net.
Showing the retry works and as an example of the output
image

### Related Issue

Fixes https://github.com/maidsafe/autonomi/issues/3026 .

### Type of Change

### DIFF
--- a/autonomi/src/client/high_level/files/fs_shared.rs
+++ b/autonomi/src/client/high_level/files/fs_shared.rs
@@ -10,6 +10,7 @@ use super::CombinedChunks;
 use crate::client::high_level::data::DataAddress;
 use crate::client::payment::PaymentOption;
 use crate::client::payment::Receipt;
+use crate::client::utils::format_upload_error;
 use crate::client::{ClientEvent, PutError, UploadSummary};
 use crate::files::UploadError;
 use crate::Client;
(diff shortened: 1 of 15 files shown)


autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -66,7 +69,6 @@ impl Client {
     }
 
     /// Processes file uploads with payment in batches
-    /// Returns total cost of uploads or error if any upload fails
     pub(crate) async fn pay_and_upload(
```
grumbach: Is that not the case anymore?

autonomi/src/client/high_level/files/fs_shared.rs
```diff
@@ -99,15 +101,48 @@ impl Client {
...
+            let candidate_chunks = aggregated_chunks.len();
+
+            let batch_result = self
+                .process_chunk_batch(
+                    &mut aggregated_chunks,
+                    &mut receipts,
+                    &mut free_chunks_counts,
+                    payment_option.clone(),
+                    retry_on_failure,
+                )
```
grumbach: Not introduced in this PR but significantly more affected with it: is there any way we can remove all these "mut"? They make it hard to know what's going on.

grumbach (commented):
The upload code might need some refactoring as it is getting more and more complex.

autonomi/src/client/high_level/files/fs_shared.rs
maqi: oh, I think that's removed by the community member, which was using infinite loop to retry

will add some more comments in make it clearer

grumbach (approved):
LGTM

PR #3039, opened 2025-06-16
feat(node): disable payee range check

maqi:
### Description

Disable payee range check on node side, due to high record rejection rate during the upscale private nodes testnet run.


maqi:
Closed as not preferred

PR #3041, opened 2025-06-17
feat(anctl): allow concurrent access of node registry

RolandSherwin:
- Remove unused auditor and faucet code.
- The `NodeRegistry` is the state of the `antnode` services currently running on the machine. But this struct does not allow concurrent access to it's field, instead the field are always passed in as mutable references `&mut NodeServiceData`.
- This would not allow us to monitor >1 node at a time and limits what we could do.
- This PR introduces a new `NodeRegistryManager` which has `Arc<RwLock<_>>>` inside it to allow concurrent access to all its fields. 
- This is required for #2968


node-launchpad/src/components/status.rs
```diff
@@ -350,14 +353,19 @@ impl Status<'_> {
...
+        self.is_nat_status_determined = self.node_registry.nat_status.blocking_read().is_some();
+        let mut nodes = Vec::new();
+        for node in self.node_registry.nodes.blocking_read().iter() {
+            let node = node.blocking_read();
+            if node.status != ServiceStatus::Running {
+                nodes.push(node.clone());
+            }
+        }
+        self.node_services = nodes;
+
```
RolandSherwin: This does not work. Fixing it.

RolandSherwin: Fixed with latest commit

ermineJose (approved):
LTGM!!

PR #3046, opened 2025-06-18
feat(logging): use separate file per spawned node

RolandSherwin:
- Writes logs to a separate file per spawned node


maqi:
closed as no longer necessary

PR #3048, opened 2025-06-18
Verify existence and retry failed uploads on default

maqi:
### Description

This is the combination work of PR https://github.com/maidsafe/autonomi/pull/3038 and https://github.com/maidsafe/autonomi/pull/3045,
aiming to reduce the upload/download errors during the upscale_private_nodes devnet run.

Close https://github.com/maidsafe/autonomi/pull/3045

### DIFF
--- a/autonomi/src/networking/retries.rs
+++ b/autonomi/src/networking/retries.rs
@@ -57,7 +57,9 @@ impl Network {
                 }
             }
         }
-        Err(NetworkError::InvalidRetryStrategy)
+        Err(NetworkError::PutRecordError(
+            "Put record {addr} failed after retries".to_string(),
+        ))
     }
 
     /// Get a record from the network with retries
@@ -99,7 +101,9 @@ impl Network {
                 }
             }
         }
-        Err(NetworkError::InvalidRetryStrategy)
+        Err(NetworkError::GetRecordError(
+            "Get record {addr} failed after retries".to_string(),
+        ))
     }
 
     /// Get quotes from the network with retries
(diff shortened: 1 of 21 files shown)


autonomi/src/networking/retries.rs
```diff
@@ -43,17 +57,19 @@ impl Network {
                 }
             }
         }
-        Err(NetworkError::InvalidRetryStrategy)
+        Err(NetworkError::PutRecordError(
+            "Put record {addr} failed after retries".to_string(),
+        ))
```
grumbach: We will end up here if we have an invalid retry strategy too!

autonomi/src/networking/retries.rs
```diff
@@ -43,17 +57,19 @@ impl Network {
...
 
     /// Get a record from the network with retries
     pub async fn get_record_with_retries(
         &self,
         addr: NetworkAddress,
         strategy: &Strategy,
+        quorum: Quorum,
     ) -> Result<Option<Record>, NetworkError> {
         let mut errors = vec![];
-        let quorum = strategy.get_quorum;
```
grumbach: The strategy contains a quorum. Why add this here? Should we drop the strategy altogether?

autonomi/src/networking/retries.rs
```diff
@@ -85,7 +101,9 @@ impl Network {
                 }
             }
         }
-        Err(NetworkError::InvalidRetryStrategy)
+        Err(NetworkError::GetRecordError(
+            "Get record {addr} failed after retries".to_string(),
+        ))
```
grumbach: Why change the error name while the flow is unchanged?

PR #3049, opened 2025-06-18
feat: avoid printing progress when verbose is minimal

ermineJose:
Progress bar reflects on node-launchpad, Add support to remove progress bar from displaying on LP.

### DIFF
--- a/ant-node-manager/src/lib.rs
+++ b/ant-node-manager/src/lib.rs
@@ -388,6 +388,7 @@ pub async fn status_report(
         service_control,
         !output_json,
         is_local_network,
+        VerbosityLevel::Normal,
     )
     .await?;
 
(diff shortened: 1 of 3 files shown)


ant-node-manager/src/lib.rs
```diff
@@ -646,9 +651,17 @@ pub async fn refresh_node_registry(
                 }
             }
         }
-        pb.inc(1);
+        if verbosity != VerbosityLevel::Minimal {
```
RolandSherwin: Tiny nit-pick, but this line is redundant and can be removed.

ant-node-manager/src/lib.rs
```diff
@@ -646,9 +651,17 @@ pub async fn refresh_node_registry(
...
+            if let Some(ref pb) = pb {
+                pb.inc(1);
+            }
+        }
+    }
+    if verbosity != VerbosityLevel::Minimal {
```
RolandSherwin: Same here.

node-launchpad/src/components/status.rs
```diff
@@ -167,6 +167,7 @@ impl Status<'_> {
             &ServiceController {},
             false,
             true,
+            ant_node_manager::VerbosityLevel::Normal,
```
RolandSherwin: Should we not make it `VerbosityLevel::Minimal` here?

PR #3052, opened 2025-06-24
fix: remove extra retry layer

grumbach:


maqi (changes requested):
The overall level retry is introduced to overcome two reasons that make holders reject uploaded record :
* un-synced records' closest nodes' knowledge between client and nodes, during churn, result in rejection of out-scope-payee
* invalid payment due to EVM rate limit
* gas-fee spike

All of them can only be resolved on mins level, and may require replacing some payees.

None of them can be achieved on this `inner chunk put_record retry` level

PR #3053, opened 2025-06-24
Switch to request_rsponse DM

maqi:
### Description

Switch to request_rsponse DM

### Type of Change

### DIFF
--- a/ant-node/src/node.rs
+++ b/ant-node/src/node.rs
@@ -26,7 +26,12 @@ use ant_protocol::{
 };
 use bytes::Bytes;
 use itertools::Itertools;
-use libp2p::{identity::Keypair, kad::U256, request_response::OutboundFailure, Multiaddr, PeerId};
+use libp2p::{
+    identity::Keypair,
+    kad::{Record, U256},
+    request_response::OutboundFailure,
+    Multiaddr, PeerId,
+};
 use num_traits::cast::ToPrimitive;
 use rand::{
     rngs::{OsRng, StdRng},
@@ -467,11 +472,12 @@ impl Node {
             }
             NetworkEvent::QueryRequestReceived { query, channel } => {
                 event_header = "QueryRequestReceived";
-                let network = self.network().clone();
+                let node = self.clone();
                 let payment_address = *self.reward_address();
 
                 let _handle = spawn(async move {
-                    let res = Self::handle_query(&network, query, payment_address).await;
+                    let network = node.network().clone();
+                    let res = Self::handle_query(node, query, payment_address).await;
 
                     // Reducing non-mandatory logging
                     if let Response::Query(QueryResponse::GetVersion { .. }) = res {
@@ -580,11 +586,8 @@ impl Node {
         Ok(())
     }
 
-    async fn handle_query(
-        network: &Network,
-        query: Query,
-        payment_address: RewardsAddress,
-    ) -> Response {
+    async fn handle_query(node: Self, query: Query, payment_address: RewardsAddress) -> Response {
+        let network = node.network();
         let resp: QueryResponse = match query {
             Query::GetStoreQuote {
                 key,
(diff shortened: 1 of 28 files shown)


autonomi/src/networking/driver/mod.rs
```diff
@@ -315,6 +315,29 @@ impl NetworkDriver {
...
+                self.pending_tasks.insert_query(
+                    req_id,
+                    NetworkTask::Request {
+                        peer_id,
+                        addresses,
+                        req,
+                        resp,
+                    },
+                );
+            }
```
grumbach, line 340: Please let's call this `UploadRecord` and keep `NetworkTask` a high level abstraction. This will help keep the networking layer API clean without low level details like the `Request`s and other types.

autonomi/src/networking/driver/swarm_events.rs
```diff
@@ -120,14 +120,18 @@ impl NetworkDriver {
...
+                peer_address,
+                storage_proofs: _,
+            }) => self
+                .pending_tasks
+                .update_get_quote(request_id, quote, peer_address)?,
+            Response::Query(QueryResponse::UploadRecord { result, .. }) => {
+                self.pending_tasks.update_request(request_id, result)?
+            }
+            _ => warn!("Unsupported response of request({request_id:?}): {response:?}"),
         }
```
grumbach, line 135: This will avoid such code blocks

autonomi/src/networking/driver/swarm_events.rs
```diff
@@ -147,8 +151,18 @@ impl NetworkDriver {
...
+            self.pending_tasks.update_request(
+                request_id,
+                Err(ant_protocol::error::Error::OtherFailure(format!(
+                    "{error:?}"
+                ))),
+            )?;
+        } else {
+            self.pending_tasks
+                .terminate_get_quote(request_id, peer, error)?;
+        }
```
grumbach, line 165: And this too.

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -59,6 +61,7 @@ impl TaskHandler {
             closest_peers: Default::default(),
             put_record: Default::default(),
             get_cost: Default::default(),
+            requests: Default::default(),
```
grumbach, line 64: Let's call this `put_record_req` as an alternative way to put a record.

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -86,20 +93,36 @@ impl TaskHandler {
...
+                addresses,
+                ..
+            } => {
+                self.requests.insert(id, (resp, peer_id, addresses));
+            }
+            NetworkTask::GetClosestPeers { .. }
+            | NetworkTask::GetRecord { .. }
+            | NetworkTask::PutRecord { .. } => {
+                warn!("A network query task shall not be handled as a request");
+            }
```
grumbach, line 125: We can avoid such code blocks.

ant-node/src/node.rs
```diff
@@ -580,11 +586,8 @@ impl Node {
         Ok(())
     }
 
-    async fn handle_query(
-        network: &Network,
-        query: Query,
-        payment_address: RewardsAddress,
-    ) -> Response {
+    async fn handle_query(node: Self, query: Query, payment_address: RewardsAddress) -> Response {
+        let network = node.network();
```
grumbach, line 590: Shouldn't this become a proper method of node? Maybe `&self` as we don't need ownership here do we?

maqi:
PutOnHold as requires the libp2p new release

autonomi/src/client/data_types/chunk.rs
```diff
@@ -242,126 +246,158 @@ impl Client {
...
+                serialized_record: serialized_record.clone(),
+            });
+            let self_clone = self.clone();
+            let serialized_record_clone = serialized_record.clone();
+            let payment_clone = payment.clone();
+            tasks.push(async move {
+                let result = self_clone
+                    .network
+                    .send_request(peer_id, Addresses(addrs.clone()), request)
+                    .await;
```
grumbach, line 348: I think we should hide `Request` type from client code. Client code is too high level to see any of low level implementation detail this IMO. Although I understand this makes the `NetworkTask` more generic, I feel that small benefit has huge implications in readability of high level data type code.

autonomi/src/client/mod.rs
```diff
@@ -143,6 +140,10 @@ pub enum PutError {
     },
     #[error("Batch upload: {0}")]
     Batch(ChunkBatchUploadState),
+    #[error("Error occurred during network operation: {0}")]
+    NetworkError(#[from] NetworkError),
+    #[error("Error occurred during serialize/deserialize: {0}")]
+    ProtocolError(#[from] ant_protocol::Error),
```
grumbach, line 146: There is already a `NetworkError` wrapper above, as well as a `Serialization` error.
Those should be removed unless really necessary as they make error management upstream unpractical.

grumbach (changes requested):
Multiple notes, will see what I can do and open a PR with suggested improvements.

grumbach:
replaced by #3071

PR #3060, opened 2025-06-26
chore: bump nodejs versions

jacderida:
From now on, this is something we'll have to do as part of the release.


vphongph (approved):
Looks good to me!

But /!\ Clippy triggered an error on this PR

PR #3061, opened 2025-06-26
chore: bump nodejs versions

jacderida:
Also merging the same version bump into `main`.


vphongph (approved):
Looks good to me!

But /!\ Clipper triggers an error!

PR #3071, opened 2025-07-03
Retro compatible generic record puts using req resp

grumbach:
- use latest version of libp2p
- use request response for clients to send records to nodes so we have better feedback from them
- retro compatibility with old nodes
- generic for all records

Replaces #3053


mickvandijke (approved):
Great work!!! Much faster uploads indeed!

PR #3074, opened 2025-07-03
Revert "refactor: network and swarm initialization code"

RolandSherwin:


grumbach (changes requested):
I would suggest we only revert changes when they are proven to be negative, not just when we have conflicts on another branch, this doesn't solve the problem, it moves it to the other branch.

PR #3079, opened 2025-07-04
Reapply "Merge pull request #3071 from grumbach/req_resp_record_put"

jacderida:
This is reapplying #3071 and keeping it open for further testing.


ant-protocol/src/error.rs
```diff
@@ -33,16 +27,19 @@ pub enum Error {
...
     /// The provided SecretyKey failed to decrypt the data
     #[error("Failed to derive CipherText from encrypted_data")]
     ScratchpadCipherTextFailed,
     /// The provided cypher text is invalid
     #[error("Provided cypher text is invalid")]
     ScratchpadCipherTextInvalid,
 
+    // ---------- Record Put errors
+    #[error("Error handling record put: {0}")]
+    PutRecordFailed(String),
```
maqi: changing the sequence of the enum entries here might cause communication issue between new client and old nodes.
for example, if old node reply with `ChunkDoesNotExist`, then the new client won't be recognize it, as the enum pos index got changed.
ideally, there shall only be new entries got appended to the enum entries tail.

ant-protocol/src/messages/query.rs
```diff
@@ -16,7 +16,7 @@ use serde::{Deserialize, Serialize};
 /// Network, and their semantics.
 ///
 /// [`protocol`]: crate
-#[derive(Eq, PartialEq, PartialOrd, Clone, Serialize, Deserialize, Debug)]
+#[derive(Eq, PartialEq, PartialOrd, Clone, Serialize, Deserialize, custom_debug::Debug)]
 pub enum Query {
```
maqi, line 20: we do better rename this `Query` to say `Request` to reflect the newly added `PutRecord` action, which is not a `query` but a `write` now.

autonomi/src/networking/driver/mod.rs
```diff
@@ -275,7 +277,38 @@ impl NetworkDriver {
...
+                let record_address = NetworkAddress::from(&record.key);
+                let peer_address = NetworkAddress::from(to.peer_id);
+                let req = Request::Query(Query::PutRecord {
+                    holder: peer_address,
+                    serialized_record: record.value.clone(),
+                    address: record_address,
+                });
+
+                // Add the peer addresses to our cache before sending a request.
+                for addr in &to.addrs {
```
maqi: there is now new facility of `send_request_with_address` to use, so that no longer require this manual add_address

autonomi/src/networking/driver/mod.rs
```diff
@@ -275,7 +277,38 @@ impl NetworkDriver {
...
+                        to,
+                        quorum,
+                        resp,
+                    },
+                );
+            }
+            NetworkTask::PutRecordReq {
+                record,
+                to,
+                quorum,
```
maqi: unlike `PutRecordKad`, this `PutRecordReq` is only targeting single target.
no idea why require this `quorum`

autonomi/src/networking/driver/swarm_events.rs
```diff
@@ -125,14 +125,24 @@ impl NetworkDriver {
...
 
-        if let Response::Query(QueryResponse::GetStoreQuote {
-            quote,
-            peer_address,
-            storage_proofs: _,
-        }) = response
-        {
-            self.pending_tasks
-                .update_get_quote(request_id, quote, peer_address)?;
+        match response {
```
maqi, line 128: unlike QueryResult, which on swarm_events/task_handler level need to `accumulate` multiple results, than call back,
request/response is a one-to-one DM, which on swarm_events/task_handler level no need to accumulate anything, and can call back directly.
Hence, if there is a generic `pending_request_tasks`,
here only need to do 
```rust
self.pending_tasks.update_request(request_id, resp)?;

fn update_request(self, request_id, resp: QueryResponse) {
   let call_back = self.remove(request_id);
   call_back.send(resp)?
}
```
Then on upper_level caller side, it only needs to `match` its correspondent response, say `QueryResponse::GetStoreQuote` or `QueryResponse::PutRecord`

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -36,7 +36,8 @@ type RecordAndHolders = (Option<Record>, Vec<PeerId>);
 #[allow(clippy::type_complexity)]
 pub(crate) struct TaskHandler {
     closest_peers: HashMap<QueryId, OneShotTaskResult<Vec<PeerInfo>>>,
-    put_record: HashMap<QueryId, OneShotTaskResult<()>>,
+    put_record_kad: HashMap<QueryId, OneShotTaskResult<()>>,
+    put_record_req: HashMap<OutboundRequestId, OneShotTaskResult<()>>,
```
maqi, line 40: `put_record_req` can be merged with `get_cost` to be: `pending_requests: HashMap<OutboundRequestId, OneShotTaskResult<QueryResponse>>`,

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -326,22 +371,39 @@ impl TaskHandler {
...
-                .remove(&id)
-                .ok_or(TaskHandlerError::UnknownQuery(format!(
-                    "OutboundRequestId {id:?}"
-                )))?;
-
-        trace!("OutboundRequestId({id}): initially sent to peer {original_peer:?} got fatal error from peer {peer:?}: {error:?}");
-        resp.send(Err(NetworkError::GetQuoteError(error.to_string())))
-            .map_err(|_| TaskHandlerError::NetworkClientDropped)?;
+        // Get quote case
+        if let Some((resp, _data_type, original_peer)) = self.get_cost.remove(&id) {
```
maqi, line 381: if using the generic pending_requests, then here no need to have such cacaded if clauses.
imaging in future we will need req/rsp based get_closest_peers to support RBS

autonomi/src/networking/mod.rs
```diff
@@ -189,17 +205,123 @@ impl Network {
...
+        let mut ok_res = vec![];
+        let mut err_res = vec![];
+        let mut old_nodes_tasks = vec![];
+        while let Some((res, peer)) = tasks.next().await {
+            match res {
+                // redirect to old protocol on old nodes
+                Err(NetworkError::IncompatibleNetworkProtocol) => {
+                    let record_clone = record.clone();
+                    let self_clone = self.clone();
+                    let handle = tokio::spawn(async move {
```
maqi, line 243: fall back to old shall only got carried out when there is not enough succeeded new nodes

autonomi/src/networking/mod.rs
```diff
@@ -189,17 +205,123 @@ impl Network {
...
+                            .await;
+                        (res, peer)
+                    });
+                    old_nodes_tasks.push(handle);
+                }
+                // accumulate oks until Quorum is met
+                Ok(()) => {
+                    ok_res.push(peer);
+                    if ok_res.len() >= expected_holders.get() {
+                        return Ok(());
```
maqi, line 255: this return could result those tasks pushed into old_nodes_tasks become phantom ones ?

autonomi/src/networking/retries.rs
```diff
@@ -55,7 +55,7 @@ impl Network {
                     }
                 }
                 // return fatal errors
-                Err(err) if err.is_fatal() => {
+                Err(err) if err.cannot_retry() => {
```
maqi, line 58: with the switch to the req/rsp, the above code block of `get_record_with_retries` on success, to verify the existence,
may can be removed on this level.

There might be some chance that when fall back to put_record_kad due to old nodes, there is risk of missing storage, but that shall be an acceptable ?

on higher level, if user concerns with node not store as claimed, the extra fetch-to-verify can be carried out by themselves.

maqi (dismissed):

ant-protocol/src/error.rs
grumbach: Put it to the end along with `OutdatedRecordCounter` which was added recently too. 
But good point. I think more generally we need to deal with how we send errors. And how retro compatibility is handled.

ant-protocol/src/messages/query.rs
grumbach: I suggest we do this in another PR as a lot of renamings in other areas will need to occur

autonomi/src/networking/driver/mod.rs
grumbach: Nice one!

autonomi/src/networking/driver/mod.rs
grumbach: Excellent point!

autonomi/src/networking/driver/task_handler.rs
grumbach: `QueryResponse` is too generic, most of it we don't use, this will pollute the namespace and make error management a nightmare.

autonomi/src/networking/driver/task_handler.rs
grumbach: There are 2 "if"s. 
Which saves us a lot of confusing pattern matching in higher level code if we do the generic version you're suggesting.

autonomi/src/networking/mod.rs
grumbach: In a real network with lots of old nodes still, waiting for all new nodes to respond before doing any requests to old nodes would be ridiculously slow. To maintain good performance this is necessary. Also we don't need to wait for all these answers if new nodes replied first. 
Once we deem old nodes don't exist anymore this code can be safely removed.

autonomi/src/networking/mod.rs
grumbach: Yes we ignore the result and they are dropped. It's a race, but that is also true for requests to new nodes. We just pick the fastest ones, old or new.

autonomi/src/networking/retries.rs
grumbach: > There might be some chance that when fall back to put_record_kad due to old nodes, there is risk of missing storage, but that shall be an acceptable ?

I agree that the get is not needed for new nodes, and that this check will indeed be a performance downer for a small decrease in reliability for older nodes sometimes. 
On the other hand our current priority is RELIABILITY over performance so I say let's keep it for now, and once we deem enough nodes have upgraded, this can safely go and we will enjoy a performance JUMP without any impact on reliability.

ant-protocol/src/error.rs
grumbach: Handled in https://github.com/maidsafe/autonomi/pull/3088
Good call Qi!

jacderida:
For some reason I am not able to approve this PR. Going to just merge anyway. It's been through testing.

PR #3080, opened 2025-07-04
feat(ci): impl custom actions to just upload logs

RolandSherwin:
- Introduce a GH action to upload test log files
- Fix the issue of logs from different tests being written to the same file. Also remove some config arguments from the init method to make it less error prone.


jacderida (approved):

vphongph:
It's a good thing to auto detect thread/test name for log files!

i did the same with the multi node logging from network spawner

PR #3088, opened 2025-07-08
feat: retro compatible protocol errors

grumbach:
Makes errors in Protocol retro compatible and resistant to changes.


ant-protocol/src/error.rs
```diff
@@ -82,3 +82,107 @@ impl From<store::Error> for Error {
...
+            RecordExists(PrettyPrintRecordKey<'static>),
+            // New variant that doesn't exist in the original Error enum
+            #[error("New error variant for testing")]
+            NewErrorVariant,
+            #[error("Unknown error variant")]
+            #[serde(other)]
+            Unknown,
+        }
+
+        // Test with a struct that has a missing variant (simulating older version)
```
maqi, line 150: the example doesn't need to have so many entries,
it's even bit difficult to figure out which one got `reduced` .
several entries shall be enough, and a comment to explicitly state which entry got `reduced` will be better.

ant-protocol/src/error.rs
```diff
@@ -69,6 +56,19 @@ pub enum Error {
...
     #[error("The record already exists, so do not charge for it: {0:?}")]
     RecordExists(PrettyPrintRecordKey<'static>),
+
+    // ---------- Record Put errors
+    #[error("Error handling record put: {0}")]
+    PutRecordFailed(String),
+    #[error("Outdated record: with counter {counter}, expected any above {expected}")]
+    OutdatedRecordCounter { counter: u64, expected: u64 },
+
+    // Dev Note: add new variants above this one for backward compatibility with older protocol versions
```
maqi, line 76: > above this one for backward compatibility

but still need to be AFTER the last valid one, i.e. OutdatedRecordCounter in this case ?

ant-protocol/src/error.rs
```diff
@@ -18,24 +18,11 @@ pub type Result<T> = std::result::Result<T, Error>;
 #[derive(Error, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[non_exhaustive]
 pub enum Error {
-    // ---------- Misc errors
-    #[error("Could not obtain user's data directory")]
-    UserDataDirectoryNotObtainable,
```
maqi, line 23: removal of this will cause disorder when try to parse old `Error::ChunkDoesNotExist` (entry index 4) into new (where entry index 4 is `QuoteGenerationFailed` )   ?
which won't fall back to `unkown` ?

maqi (changes requested):
the refactor scenario (with certain entries removed) in this PR need at least some unit test to show the old error can be parsed into new and vice-versa.
Entries exists in both old and new shall be parsed correctly across.

ant-protocol/src/error.rs
grumbach: There is a comment for that look below

ant-protocol/src/error.rs
grumbach: I'm not sure that is the case, we've been changing this struct for a while now, so I thought it was time to clean that up. Lots of dead code too.

ant-protocol/src/error.rs
```diff
@@ -82,3 +92,221 @@ impl From<store::Error> for Error {
...
+            rmp_serde::from_slice(&serialized_unknown).unwrap();
+        assert_eq!(deserialized_unknown, ReducedError::Unknown);
+    }
+
+    #[test]
+    fn test_error_retro_compatibility_many_missing() {
+        // test with many missing variants
+        #[derive(Error, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+        #[non_exhaustive]
+        enum ManyMissingVariants {
```
maqi, line 242: to prove the compatibility best, we need :
* have this ManyMissingVariants enum contains another complex entry
* confirm such complex entry can be converted between Error and ManyMissingVariants in dual directions.

ant-protocol/src/error.rs
```diff
@@ -82,3 +92,221 @@ impl From<store::Error> for Error {
...
+        }
+
+        // Test with ManyMissingVariants
+        let current_error = Error::ScratchpadCipherTextInvalid;
+        let serialized_current = rmp_serde::to_vec(&current_error).unwrap();
+        let deserialized_many_missing: ManyMissingVariants =
+            rmp_serde::from_slice(&serialized_current).unwrap();
+        assert_eq!(deserialized_many_missing, ManyMissingVariants::Unknown);
+
+        // Test that GetStoreQuoteFailed works with ManyMissingVariants
```
maqi, line 260: need to confirm the other direction as well,
i.e. a serialized ManyMissingVariants::GetStoreQuoteFailed can be parsed as Error::GetStoreQuoteFailed properly

ant-protocol/src/error.rs
```diff
@@ -82,3 +92,221 @@ impl From<store::Error> for Error {
...
+            #[error("Unknown error variant")]
+            #[serde(other)]
+            Unknown,
+        }
+
+        // Test serialization and deserialization of current Error
+        let current_error = Error::ScratchpadCipherTextInvalid;
+        let serialized_current = rmp_serde::to_vec(&current_error).unwrap();
+
+        // Test that we can deserialize into ReducedError (older version)
```
maqi, line 188: also need to use a complex entry to confirm works in such case

ant-protocol/src/error.rs
```diff
@@ -82,3 +92,221 @@ impl From<store::Error> for Error {
...
+            RecordParsingFailed,
+            #[error("The record already exists, so do not charge for it: {0:?}")]
+            RecordExists(PrettyPrintRecordKey<'static>),
+            // Missing some variants that exist in the current Error enum
+            #[error("Unknown error variant")]
+            #[serde(other)]
+            Unknown,
+        }
+
+        // Test serialization and deserialization of current Error
```
maqi, line 184: also need to cover the other direction, i.e. serialized ReducedError to be parsed as Error, with both simple and complex entry.

maqi (changes requested):

ant-protocol/src/error.rs
```diff
@@ -82,3 +92,221 @@ impl From<store::Error> for Error {
...
+        let holder = NetworkAddress::from(PeerId::random());
+        let key = NetworkAddress::from(PeerId::random());
+        let complex_error = Error::ReplicatedRecordNotFound {
+            holder: Box::new(holder),
+            key: Box::new(key),
+        };
+        let serialized_complex = rmp_serde::to_vec(&complex_error).unwrap();
+        let deserialized_complex: ComplexTypesRemoved =
+            rmp_serde::from_slice(&serialized_complex).unwrap();
+        assert_eq!(deserialized_complex, ComplexTypesRemoved::Unknown);
```
maqi: which one is currently failing for the complex removal scenario ?

is this assert, or the later on one ?

grumbach: The deserilazation `unwrap`

maqi (approved):

PR #3092, opened 2025-07-09
chore(node): utilise send_request_with_addresses on node side

maqi:
### Description

utilise send_request_with_addresses on node side

### DIFF
--- a/ant-node/src/networking/driver/cmd.rs
+++ b/ant-node/src/networking/driver/cmd.rs
@@ -24,7 +24,6 @@ use libp2p::{
         store::{Error as StoreError, RecordStore},
         KBucketDistance as Distance,
     },
-    swarm::dial_opts::{DialOpts, PeerCondition},
     Multiaddr, PeerId,
 };
 use std::time::Instant;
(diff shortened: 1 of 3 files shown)


ant-node/src/networking/network/mod.rs
```diff
@@ -246,89 +245,21 @@ impl Network {
...
-        if let Err(error) = &r {
-            error!("Error in response: {:?}", error);
-
-            match error {
-                NetworkError::OutboundError(OutboundFailure::Io(_))
-                | NetworkError::OutboundError(OutboundFailure::ConnectionClosed)
-                | NetworkError::OutboundError(OutboundFailure::DialFailure) => {
-                    warn!(
-                        "Outbound failed for {req_str} .. {error:?}, dialing it then re-attempt."
-                    );
```
grumbach, line 277: Do we not need that error management anymore?

ant-node/src/networking/driver/cmd.rs
```diff
@@ -83,29 +82,42 @@ impl SwarmDriver {
...
+                                        Some(peer_entry.node.value.clone().into_vec())
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .collect::<Vec<Vec<Multiaddr>>>();
+                            if !peer_addrs.is_empty() {
+                                result = peer_addrs[0].clone();
+                                break;
                             }
```
grumbach: Can we make this into its own separate helper function? It would help with readability A LOT.

ant-node/src/networking/network/mod.rs
maqi: yes
because no longer using redial to re-establish peer's addresses info

jacderida (approved):

PR #3094, opened 2025-07-10
Req resp get

grumbach:


grumbach:
Discontinue req/resp gets for now as less promising than hoped for.

PR #3098, opened 2025-07-14
fix(client): make file_upload use same workflow as dir_upload

maqi:
### Description

make file_upload use same workflow as dir_upload

### Type of Change

### DIFF
--- a/ant-cli/src/exit_code.rs
+++ b/ant-cli/src/exit_code.rs
@@ -79,6 +79,7 @@ pub(crate) fn put_error_exit_code(err: &PutError) -> i32 {
         PutError::Wallet(_) => 42,
         PutError::Batch(_) => 44,
         PutError::PayeesMissing => 45,
+        PutError::Other(_) => 46,
     }
 }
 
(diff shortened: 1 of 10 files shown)


autonomi/src/client/high_level/data/private.rs
```diff
@@ -73,44 +73,18 @@ impl Client {
...
-                tokens_spent,
-            };
-            if let Err(err) = channel.send(ClientEvent::UploadComplete(summary)).await {
-                error!("Failed to send client event: {err:?}");
-            }
+        // Note within the `pay_and_upload`, UploadSummary will be sent to cli via event_channel.
+        match self.pay_and_upload(payment_option, combined_chunks).await {
+            Ok(total_cost) => Ok((total_cost, DataMapChunk(data_map_chunk))),
+            Err(UploadError::PutError(err)) => Err(err),
+            Err(err) => Err(PutError::Other(format!("{err}"))),
```
grumbach: Lets avoid adding error variants like Other, especially if it's just because of the refactoring

ant-cli/src/exit_code.rs
```diff
@@ -79,6 +79,7 @@ pub(crate) fn put_error_exit_code(err: &PutError) -> i32 {
         PutError::Wallet(_) => 42,
         PutError::Batch(_) => 44,
         PutError::PayeesMissing => 45,
+        PutError::Other(_) => 46,
```
grumbach: Why do we need an Other here??

grumbach (changes requested):
Good work! Let's remove that Other error variant as it reduces error quality.

grumbach (approved):
Nice work Qi!
Added a couple of improvements in code structure and error management.
LGTM
You can merge if happy with the changes!

PR #3102, opened 2025-07-16
feat: factor out encryption in fs api

grumbach:


maqi (changes requested):
Ideally, a deduplication across encryptions of dir/file and pricate/public can be undertaken.
This can be done in future anyway.

maqi (approved):
Ideally, a deduplication across encryptions of dir/file and pricate/public can be undertaken.
This can be done in future anyway.

PR #3105, opened 2025-07-18
feat(autonomi): add `get_node_version` pub fn

mickvandijke:
Adds a `get_node_version` public function to the `autonomi` lib. This is needed for the emission service and can be used by app developers.


mickvandijke:
Replaced by https://github.com/maidsafe/autonomi/pull/3106

PR #3109, opened 2025-07-22
upload flow refactor for the future streaming accomodation 

grumbach:


autonomi/src/client/encryption.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    pub fn new_in_memory_with(
+        file_path: String,
+        relative_path: PathBuf,
+        metadata: Metadata,
+        is_public: bool,
+        bytes: Bytes,
+    ) -> Result<(Self, DataMapChunk), crate::self_encryption::Error> {
+        let start = Instant::now();
+        let (data_map_chunk, mut chunks) = encrypt(bytes)?;
```
maqi, line 94: shall use the latest self_encryption's streaming facilities.
instead of still loading all chunks into memory in one go.

maqi (changes requested):

maqi (approved):

PR #3112, opened 2025-07-25
feat(antctl): spawn evm-testnet as part of local testnet cmd

RolandSherwin:
- Removes the `evm_testnet_data.csv` file and instead the values are hardcoded now.
- When `--build` is used, the `evm-testnet` binary is compiled and auto run for you.


grumbach (approved):
Nice one!

PR #3113, opened 2025-07-28
feat: add chunk caching for downloads with retry support

mickvandijke:
Caches chunks by default to the `{data_dir}/autonomi/client/chunk_cache` folder.

Get chunk calls will first try the cache.

On successful download of all chunks within a datamap, the cached chunks for that datamap are deleted.


ant-cli/src/commands.rs
```diff
@@ -116,6 +117,15 @@ pub enum FileCmd {
         /// Experimental: Optionally specify the number of retries for the download.
         #[arg(short, long)]
         retries: Option<usize>,
+        /// Enable chunk caching to resume failed downloads.
+        /// Successfully downloaded chunks are cached to disk and reused on retry.
+        /// The cache is automatically keyed by the data address being downloaded.
+        #[arg(long)]
+        cache_chunks: bool,
```
jacderida: Since caching is the default, I think this option should be presented to the user as `--disable-cache`, which can then be inverted to `cache_chunks` for the code to deal with, since it's easier to read the positive case.

The documentation for the argument could be:
```
By default, chunks will be cached to enable resuming downloads.

Set this flag to disable the cache.
```

jacderida (approved):

PR #3122, opened 2025-07-29
feat: verify bootstrap peers before adding to RT

mickvandijke:
Verifies that bootstrap peers actually support the kad protocol before adding them to the RT.

It dials the initial peers after running the network in the Client (during Client init) and it waits for their identity packets to be received. When we receive an identity packet that supports the kad protocol, we directly insert it into the RT. When we have added 20 peers to our RT, we fire a signal that the bootstrapping is complete.

The fix is quite big and I’m not happy about introducing so much code to the new CLN. So we should only see it is a temp patch until dealing with unsupported/missing protocol nodes is resolved on the libp2p end.

NOTE: Will be replaced by a new PR if tests prove to be successful: https://github.com/maidsafe/autonomi/pull/3129

### DIFF
--- a/autonomi/src/networking/driver/swarm_events.rs
+++ b/autonomi/src/networking/driver/swarm_events.rs
@@ -6,12 +6,14 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use ant_protocol::constants::KAD_STREAM_PROTOCOL_ID;
 use ant_protocol::messages::{QueryResponse, Response};
 use libp2p::autonat::OutboundFailure;
+use libp2p::identify::Event;
 use libp2p::kad::{Event as KadEvent, ProgressStep, QueryId, QueryResult, QueryStats};
 use libp2p::request_response::{Event as ReqEvent, Message, OutboundRequestId};
 use libp2p::swarm::SwarmEvent;
-use libp2p::PeerId;
+use libp2p::{PeerId, StreamProtocol};
 use thiserror::Error;
 
 #[derive(Error, Debug, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


RolandSherwin (changes requested):
We should deny connection to unsupported protocol, maybe even add them to `blocklist`. The peers are part of the RT as soon as they are connected to us. As we don't perform manual RT management on clients..

autonomi/src/networking/driver/swarm_events.rs
```diff
@@ -167,4 +174,26 @@ impl NetworkDriver {
...
+    ) -> Result<(), NetworkDriverError> {
+        if info
+            .protocols
+            .contains(&StreamProtocol::new(KAD_STREAM_PROTOCOL_ID))
+        {
+            for listen_addr in info.listen_addrs {
+                self.swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .add_address(&peer_id, listen_addr);
```
b-zee, line 200: `listen_addrs` is not a trusted source. There can be unreachable or random addresses in here. Also, if we were to do this we might want to add more restrictions to when we add these. E.g. only if they're also supporting our req/resp protocol.

b-zee: By the way, I'm unaware of the current mechanism of how we fill our RT, but we need to make sure this doesn't conflict with that existing mechanism.

mickvandijke: > listen_addrs is not a trusted source.

Would using the observed address be better?

RolandSherwin: `Observed` address is the self advertised address that nodes got by calling `swarm.add_external_addr` and that might be incorrect?

Nodes get the right address by correlating the `ConnectionId`..

jacderida:
This was replaced with #3129 which was subsequently released in a hotfix.

PR #3123, opened 2025-07-29
feat: added remaining methods and classes + fixed PyTransactionConfig

vphongph:
- Added remaining methods and classes (that was on hold with CLN)
- Rebased from main
- Updated Python API about InitialPeerConfig
- Fixed PyTransactionConfig to make its use more relevant and consistent with the original API

### DIFF
--- a/autonomi/src/python.rs
+++ b/autonomi/src/python.rs
@@ -2011,18 +2011,16 @@ impl PyInitialPeersConfig {
     }
 
     /// Get bootstrap addresses
-    #[pyo3(signature = (config=None, count=None))]
+    #[pyo3(signature = (count=None))]
     fn get_bootstrap_addr<'a>(
         &self,
         py: Python<'a>,
-        config: Option<PyBootstrapCacheConfig>,
         count: Option<usize>,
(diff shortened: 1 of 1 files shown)


autonomi/src/python.rs
```diff
@@ -1610,172 +1796,549 @@ impl PyGraphEntryAddress {
...
+    #[getter]
+    fn cache_save_scaling_factor(&self) -> u64 {
+        self.inner.cache_save_scaling_factor
     }
 
-    fn __str__(&self) -> PyResult<String> {
-        Ok(self.hex())
+    /// Return a string representation
+    fn __str__(&self) -> String {
+        format!("{:?}", self.inner)
```
grumbach, line 1915: This was correct, why change it?

grumbach (approved):
LGTM! One small comment.

autonomi/src/python.rs
vphongph: The diff is misleading

Classes have a different position in this PR but the diff thinks some methods have been changed but it's not the case!

PR #3126, opened 2025-07-30
Add evm testnet to releases

grumbach:


jacderida:
I will add this to the release as part of the next release branch.

PR #3127, opened 2025-07-30
feat(node): block peers with unsupported protocols

RolandSherwin:


ant-node/src/networking/driver/event/identify.rs
```diff
@@ -154,6 +170,43 @@ impl SwarmDriver {
...
+                !info
+                    .protocols
+                    .iter()
+                    .any(|protocol| protocol.as_ref().contains(*required))
+            })
+            .collect();
+
+        if !missing_protocols.is_empty() {
+            warn!(
+                "Peer {peer_id:?} does not support some mandatory protocols from: {REQUIRED_PROTOCOLS:?}, not adding it to our RT."
```
mickvandijke: Perhaps also log which protocols the node was missing?

mickvandijke (changes requested):

mickvandijke (approved):
LGTM!

PR #3140, opened 2025-08-05
feat: check for connectivity on client init

b-zee:


autonomi/src/networking/mod.rs
```diff
@@ -200,6 +203,36 @@ impl Network {
...
+
+        match timeout(timeout_duration, async {
+            loop {
+                match self.get_connections_made().await {
+                    Ok(count) => {
+                        if count >= CLOSE_GROUP_SIZE {
+                            return Ok(());
+                        }
+                    }
+                    Err(_err) => {}
```
maqi: ideally put a log inside the loop, so that in case got trapped in, at least easier to figure out where the trap is.

autonomi/src/networking/mod.rs
```diff
@@ -506,6 +539,17 @@ impl Network {
...
+
+    /// Get information about the routing table
+    pub async fn get_connections_made(&self) -> Result<usize, NetworkError> {
+        let (tx, rx) = oneshot::channel();
+        let task = NetworkTask::ConnectionsMade { resp: tx };
+        self.task_sender
+            .send(task)
+            .await
+            .map_err(|_| NetworkError::NetworkDriverOffline)?;
+        rx.await?
```
maqi, line 556: this could be an infinite wait ?
better at least put a log before the wait, so that at least in case of a deadlock, we can be eaiser to know where the place is.

maqi (approved):

PR #3141, opened 2025-08-05
test: self_encryption backward compatible test

maqi:
### Description

* integrate newly released self_encryption
* backward compatible facility to support old data_map scheme
* self_encryption backward compatible unit test and e2e test


grumbach (approved):
Excellent work!

PR #3142, opened 2025-08-05
fix(autonomi): retry failed chunks from the final / only batch

mickvandijke:
The exit condition of the while loop was not correct. It would exit if there was no new batch of chunks from the original batch schedule, but it didn't take into account any failed chunks. So if we had any failed chunks in the last batch or the first and only batch (usual for smaller files), they were just tossed and reported the upload as succeeded.

The fix also includes adding an early return for the insufficient tokens error as there is no point retrying in that case.

### DIFF
--- a/autonomi/src/client/high_level/data/helpers.rs
+++ b/autonomi/src/client/high_level/data/helpers.rs
@@ -6,15 +6,17 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::Client;
 use crate::client::encryption::EncryptionStream;
+use crate::client::payment::PayError::EvmWalletError;
 use crate::client::payment::PaymentOption;
 use crate::client::payment::Receipt;
 use crate::client::utils::format_upload_error;
 use crate::client::{ClientEvent, PutError, UploadSummary};
-use crate::Client;
 use ant_evm::{Amount, AttoTokens};
 use ant_protocol::storage::{Chunk, DataTypes};
 use evmlib::contract::payment_vault::MAX_TRANSFERS_PER_TRANSACTION;
+use evmlib::wallet::Error::InsufficientTokensForQuotes;
 use std::sync::LazyLock;
 use std::time::Duration;
 use tokio::time::sleep;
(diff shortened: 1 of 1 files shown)


autonomi/src/client/high_level/data/helpers.rs
```diff
@@ -244,13 +260,15 @@ impl Client {
         {
             Ok((receipt, free_chunks)) => (receipt, free_chunks),
             Err(err) => {
```
RolandSherwin, line 266: Tiny one, but if possible, it would be great to add some logging here for the failure case.

RolandSherwin (approved):

autonomi/src/client/high_level/data/helpers.rs
```diff
@@ -152,26 +154,39 @@ impl Client {
...
+                    .map(|(i, chunk)| ((path.clone(), processed_chunks + i, est_total_todo), chunk))
+                    .collect();
+                current_batch.extend(aggr_batch);
+
+                // process batch
+                processed_chunks += next_batch_len;
+            }
+
+            // If we have no chunks to process, break the loop
+            if current_batch.is_empty() {
```
maqi, line 177: the change of this code block seems not related to the PR title ?
is it really necessary ?

autonomi/src/client/high_level/data/helpers.rs
```diff
@@ -244,13 +260,15 @@ impl Client {
         {
             Ok((receipt, free_chunks)) => (receipt, free_chunks),
             Err(err) => {
-                if retry_on_failure {
+                return if let EvmWalletError(InsufficientTokensForQuotes(_, _)) = err {
```
maqi: bit confused with this `return if` syntax, and it's also not consistant with the syntax in other if clause branches.
better change it to the match branch ? :
```rust
match self
            .pay_for_content_addrs(DataTypes::Chunk, payment_info.into_iter(), payment_option)
            .await {
    Ok() => {},
    err(err) if matches!(err, EvmWalletError(InsufficientTokensForQuotes(_, _))) => return (vec![], vec![], 0, Some(PutError::from(err))),
    err(err) => { ... }
}
```

maqi (changes requested):
the commit title, PR title and PR description better to be : 
terminate upload in case of insufficient fund, to avoid un-necessary retry

autonomi/src/client/high_level/data/helpers.rs
mickvandijke: It is necessary. The original code was not working as intended. 

The exit condition of the while loop was not correct. It would exit if there was no new batch of chunks from the original batch schedule, but it didn't take into account any failed chunks. So if we had any failed chunks in the last batch, they were just tossed and reported the upload as succeeded

mickvandijke:
> the commit title, PR title and PR description better to be : terminate upload in case of insufficient fund, to avoid un-necessary retry

> terminate upload in case of insufficient fund, to avoid un-necessary retry

This part is only an extra optimization in the PR. 

The retry logic of the original code was not working. The retry chunks being returned on an error, were not actually retried. They were silently thrown away and the code continued as if the upload was successful.

maqi (approved):

PR #3143, opened 2025-08-05
fix: store single file upload access locally

mickvandijke:
Single file upload data addresses and private datamaps were not saved in the `user_data` folder like for multi-file uploads (archives).

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -463,12 +463,27 @@ jobs:
 
           ./target/release/ant --log-output-dest data-dir --local file list 2>&1 > file_list.txt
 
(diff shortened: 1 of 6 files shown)


maqi (changes requested):
Why there is so many user_data access facilities need to be added ?
Isn't the directory ones shall use the similar facilities ?

mickvandijke:
> Why there is so manay user_data access facilities need to be added ? Isn't the directory ones shall use the similar facilities ?

The PR is just following what has already been established. Archive datamaps and data addresses were already stored in `/user_data/private_file_archives` and `/user_data/file_archives`. This PR just fixes that single files follow the same behaviour, by storing them in `/user_data/public_files` and `/user_data/private_files`.

grumbach (approved):
LGTM

maqi (approved):

PR #3158, opened 2025-08-11
feat: add single node payment mode

mickvandijke:
Add an additional payment mode where only one node gets paid instead of three for each record. The one node will have to be paid three times the quoted amount as a penalty for using this mode. So uploads using the single node payment mode will be more expensive ANT token wise, but it will save gas fees by about 67%.

This mode is enabled by default. To disable this mode, simply add the `--disable-single-node-payment` flag to the `ant upload command`.

This is a client and smart contract only change.

REQUIREMENT:

- [x] Update the PROD `PaymentVault` smart contract to support single node payment mode.


maqi (changes requested):
Put as Request Change, as the new cmd opt of `single_node_payment` need a decision of `whether shall be default to be true`, i.e. whether the new scheme shall be enabled by default.

autonomi/src/client/quote.rs
```diff
@@ -193,41 +203,172 @@ impl Client {
...
+            (*p1, a1.clone(), q1.clone(), Amount::ZERO),
+            (*p2, a2.clone(), q2.clone(), Amount::ZERO),
+            quotes[2].clone(),
+            quotes[3].clone(),
+            quotes[4].clone(),
+        ])
+    }
+
+    /// Create payment structure for single node mode (pay only highest priced node with 10x)
+    fn create_single_node_quote_payment(
```
maqi, line 327: the input quotes are already `sorted by price` 
the `create_standard_quote_payment` and `create_single_node_quote_payment` can simply to be:
```
        QuoteForAddress(vec![
            (*p1, a1.clone(), q1.clone(), Amount::ZERO),
            (*p2, a2.clone(), q2.clone(), Amount::ZERO),
            quotes[2].clone(),
            quotes[3].clone(),
            quotes[4].clone(),
        ])
```
or 
```
        QuoteForAddress(vec![
            (*p1, a1.clone(), q1.clone(), Amount::ZERO),
            (*p2, a2.clone(), q2.clone(), Amount::ZERO),
            (*p3, a3.clone(), q3.clone(), Amount::ZERO),
            (*p4, a4.clone(), q4.clone(), Amount::ZERO),
            quotes[4].clone(),
        ])
```

autonomi/src/client/quote.rs
```diff
@@ -193,41 +203,172 @@ impl Client {
...
+        content_addr: XorName,
+    ) -> QuoteForAddress {
+        // Find the highest price and select node to pay
+        let highest_price = quotes
+            .iter()
+            .map(|(_, _, _, price)| *price)
+            .max()
+            .expect("quotes should not be empty as we checked len >= MINIMUM_QUOTES_TO_PAY");
+
+        let node_to_pay = self.select_highest_priced_node(quotes, highest_price);
```
maqi: already sorted, no need to re-sort

autonomi/src/client/quote.rs
```diff
@@ -193,41 +203,172 @@ impl Client {
...
+        let enhanced_price = highest_price * Amount::from(10u64);
+
+        // Sort by distance to maintain the closest 5 nodes
+        self.sort_quotes_by_distance(quotes, content_addr);
```
maqi: does this step required ?

mickvandijke: No :)

autonomi/src/client/quote.rs
mickvandijke: Sharp!

maqi (approved):

PR #3163, opened 2025-08-18
chore(client): split fundamental parts out from autonomi as autonomi_core

maqi:
### Description

🎯 Architectual Benefits

1. Simplified Learning Curve

* Fewer APIs to learn: ~15 core APIs vs ~85 in full autonomi
* Focused functionality: Only essential network operations exposed
* Consistent patterns: Generic methods reduce cognitive overhead

2. Better Maintainability

* Reduced surface area: Fewer APIs to maintain and test
* Clear separation: Core vs application-level functionality
* Easier evolution: Changes to high-level APIs don't affect core

3. Enhanced Flexibility

* Generic operations: DataContent enum allows handling all data types uniformly
* Extensible design: New data types can be added without API changes
* Composable workflows: Applications can build complex operations from simple primitives

4. Performance Optimization

* Smaller binary size: Fewer dependencies and code paths
* Faster compilation: Reduced complexity in core crate
* Targeted optimizations: Focus on core network performance

🚀 Recommendations
When to Use autonomi_core
* Building custom clients: When you need fine-grained control over network operations
* Embedded systems: When binary size and dependencies matter
* Library development: When building higher-level abstractions
* Performance-critical applications: When you need minimal overhead


autonomi-core/src/client/mod.rs
```diff
@@ -0,0 +1,488 @@
...
+pub enum ClientInitSetup {
+    /// Use default configuration
+    Default,
+    /// Connect to alpha network
+    Alpha,
+    /// Bootstrap from specific peers
+    Peers(Vec<Multiaddr>),
+    /// Use custom configuration
+    Config(ClientConfig),
+}
```
RolandSherwin, line 56: Do we need this? This just looks like a wrapper around `ClientConfig`? People who are using the core crate should be able to read through `ClientConfig` and be able to figure out the values themselves, we don't have to have another wrapper here tbh?

autonomi/src/client/data_types/chunk.rs
```diff
@@ -6,200 +6,47 @@
...
-                DataTypes::Chunk,
-            )))
-        }
-    }
-
     /// Get a chunk from the network.
     pub async fn chunk_get(&self, addr: &ChunkAddress) -> Result<Chunk, GetError> {
         info!("Getting chunk: {addr:?}");
-
-        if let Some(cached_chunk) = self.try_load_chunk_from_cache(addr)? {
```
RolandSherwin, line 172: Earlier, the `try_load_chunk_from_cache` was always called during the lower level chunk_get, but now it is being called only on `fetch_from_data_map` inside core? So we might not hit the cache always on the new code?

autonomi/src/client/put_error_state.rs
```diff
@@ -6,68 +6,5 @@
...
-                }
-            }
-            err => {
-                self.failed.push((address, err.to_string()));
-            }
-        }
-    }
-}
+// Re-export types from autonomi_core
+pub use autonomi_core::client::ChunkBatchUploadState;
```
RolandSherwin: Can this be re-exported somewhere else instead of holding it in a file?

RolandSherwin (changes requested):
Looks good to me in general! Left some comments though

autonomi/src/client/data_types/chunk.rs
maqi: `chunk_cache` is mainly for large file upload/download.
after the split out, single chunk_get/chunk_put are now mainly for much critiacal data, such as datamap/archive, which shall be fewer in number and doesn't matter much if hit cache or not?

autonomi/src/client/put_error_state.rs
maqi: just to be backward compatible, that the `namespace` path remain unchanged

autonomi-core/src/client/mod.rs
maqi: bit tricky here.
`ClientConfig` is an already published stuff, which means we can't change it.
But for the case of `Alpha`, it means `populate with some special values` (as shown in the attached code block).
which isn't obvious on API level and could be easily missed out.
hence introduced this `wrapper`.
With this `wrapper`, it is conveient if later on need to support another network, say `Belta`, just append to the enum `ClientInitSetup` and have an individual inner handling facility will be enough. 
```rust
        let bootstrap_cache_config = config::BootstrapCacheConfig::new(false)
            .inspect_err(|err| {
                warn!("Failed to create bootstrap cache config: {err}");
            })
            .ok();

        let client_config = ClientConfig {
            init_peers_config: InitialPeersConfig {
                first: false,
                addrs: vec![],
                network_contacts_url: ant_bootstrap::contacts::ALPHANET_CONTACTS
                    .iter()
                    .map(|s| s.to_string())
                    .collect(),
                local: false,
                ignore_cache: false,
                bootstrap_cache_dir: None,
            },
            evm_network: EvmNetwork::ArbitrumSepoliaTest,
            strategy: Default::default(),
            network_id: Some(2),
            bootstrap_cache_config,
        };
```

maqi:
Closing this PR as during the HO discussion, it is decided to move autonomi(client) as a separate repository,
to achieve a full speed client-only release cycle.

PR #3167, opened 2025-08-22
fix(bootstrap): dont overwrite self with peers from fs

RolandSherwin:
- earlier we were overwriting any new peers in `self` with the peers found from 'other' (i.e, fs cache). this is incorrect and has been fixed
- also enforced that the new peers from `self` are preserved and we only pop the old entries from the back
- added a few tests to assert all thses changes

Here, `self` is the new set of peers that are kept in memory. `other` is the set of peers obtained from fs. 
The fs peers are ordered from new to old.
So in the loop, we keep the self (or new set of peers) and append the other peers from new to old.

### DIFF
--- a/ant-bootstrap/src/cache_store/cache_data_v1.rs
+++ b/ant-bootstrap/src/cache_store/cache_data_v1.rs
@@ -32,7 +32,7 @@ impl CacheData {
     /// This has to be bumped whenever the cache data format changes to ensure compatibility.
     pub const CACHE_DATA_VERSION: u32 = 1;
 
-    /// Sync the self cache with another cache. This would just add the 'other' state to self.
+    /// Sync the self cache with another cache. Self peers (newer) are preserved.
     pub fn sync(&mut self, other: &CacheData, max_addrs_per_peer: usize, max_peers: usize) {
         let old_len = self.peers.len();
         let other_len = other.peers.len();
@@ -41,6 +41,8 @@ impl CacheData {
             if other_addrs.is_empty() {
                 continue;
             }
+
+            let mut found_existing = false;
             for (peer, addrs) in self.peers.iter_mut() {
                 if peer == other_peer {
                     for addr in other_addrs.iter() {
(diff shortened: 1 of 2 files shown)


ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -51,17 +53,26 @@ impl CacheData {
...
-            while self.peers.len() > max_peers {
-                self.peers.pop_front();
+            // break if limit reached
+            if self.peers.len() >= max_peers {
+                break;
             }
         }
 
+        // Apply max_peers limit by removing from back (oldest from other)
+        while self.peers.len() > max_peers {
```
maqi, line 72: this is somehow no-use as within the previous loop, the breaking out criteria is `if self.peers.len() >= max_peers {` already.

Also, using `self.peers.pop_back()` (new discovered nodes from other is old) here is conflicting with the previous `self.peers.push_back((*other_peer, other_addrs.clone()));` (new discovered nodes from other is new)

ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -51,17 +53,26 @@ impl CacheData {
...
 
-            self.peers.push_back((*other_peer, other_addrs.clone()));
+            if !found_existing {
+                self.peers.push_back((*other_peer, other_addrs.clone()));
+            }
 
-            while self.peers.len() > max_peers {
-                self.peers.pop_front();
+            // break if limit reached
+            if self.peers.len() >= max_peers {
```
maqi, line 66: the previous logic using `self.peers.pop_front();` which indicates:
1. peers are odered from `old to new`
2. once cap reached, `old` ones preferred to be replaced by the later on new ones.

This actually indicates couple of issues in this merging logic:
1.  `self` seems assumed to be always being `old` than the ones from `other` , as using `self.peers.push_back((*other_peer, other_addrs.clone()));`
2.  once cap reached, those potential `new` nodes in other won't be able to be `refreshed` in

maqi (changes requested):
as stated in the review comment, it seems there is some logic issue/conflict within the merge workflow.

maqi (approved):
Updated the PR description to clarify the confusion on the logic.
The previous cocern is obsoleted.

PR #3176, opened 2025-09-02
feat: streaming apis for data

grumbach:
Data streaming as simple as it should have been. 

```rust
let stream = client.data_stream_public(&addr).await?;
for chunk_result in stream {
    let chunk = chunk_result?;
    // Process chunk...
}
```

### DIFF
--- a/autonomi/src/client/high_level/data/private.rs
+++ b/autonomi/src/client/high_level/data/private.rs
@@ -57,7 +57,7 @@ impl Client {
     }
 
     /// Stream a blob of (private) data from the network. Returns an Iterator that yields chunks progressively.
-    /// Use this for large files to avoid loading everything into memory.
+    /// Use this for large blobs of data like videos to avoid loading everything into memory.
     ///
     /// # Example
     ///
--- a/autonomi/src/client/high_level/data/private.rs
+++ b/autonomi/src/client/high_level/data/private.rs
@@ -16,6 +16,7 @@ use crate::client::{GetError, PutError};
 
 pub use crate::Bytes;
 pub use crate::client::data_types::chunk::DataMapChunk;
+pub use crate::client::high_level::data::stream::DataStream;
 
 impl Client {
     /// Fetch a blob of (private) data from the network. In-memory only - fails for large files.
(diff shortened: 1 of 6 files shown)


autonomi/src/client/high_level/data/private.rs
```diff
@@ -55,6 +56,45 @@ impl Client {
         Ok(data)
     }
 
+    /// Stream a blob of (private) data from the network. Returns an Iterator that yields chunks progressively.
+    /// Use this for large files to avoid loading everything into memory.
```
maqi: this is slightly misleading, as the yield chunks if not got drained, will be held in memory.
For `large files`, it's still the `stream with flush to disk` being the best option and guarantee `NOT` loading all stuff into memory.

Suggest to `remove` this sentence, or maybe change to:
Use this for streaming service such as videa clip to avoid long-time loading.

autonomi/src/client/high_level/data/public.rs
```diff
@@ -34,6 +34,36 @@ impl Client {
         self.data_get(&datamap_chunk).await
     }
 
+    /// Stream a blob of public data from the network. Returns an Iterator that yields chunks progressively.
+    /// Use this for large files to avoid loading everything into memory.
```
maqi: same as the other comment, better rewording this

autonomi/src/client/high_level/data/stream.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

autonomi/src/client/high_level/data/stream.rs
```diff
@@ -0,0 +1,173 @@
...
+use crate::Bytes;
+use crate::Client;
+use crate::client::GetError;
+
+type ChunkFetcher =
+    Box<dyn Fn(&[(usize, XorName)]) -> self_encryption::Result<Vec<(usize, Bytes)>> + Send + Sync>;
+
+pub struct DataStream {
+    streaming_decrypt: self_encryption::DecryptionStream<ChunkFetcher>,
+    finished: bool,
```
maqi: oh, it's bit weird that the `wrapper` requires an additional `finished` flag.
maybe shall make `DecryptionStream` provide such status indicating getter ?
and seems if we had such getter in `DecryptionStream` , we then no need this `DataStream ` wrapper ?

autonomi/tests/streaming_download.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright 2024 MaidSafe.net limited.
```
maqi: 2025

maqi (changes requested):
Seems the `DataStream` wrapper is not required, if `self_encryption::DecryptionStream` provides a `is_finished` status indication getter ?

autonomi/src/client/high_level/data/private.rs
grumbach: Agreed

maqi (approved):

PR #3185, opened 2025-09-04
feat(scratchpad_put)!: return receipt for secret-key–free updates

### LINKED ISSUE #3184: Make `scratchpad_put` the single, secret-key–free path for both creation and updates
See #3185

### Problem
`scratchpad_put` already allows **create** without the address’s secret key, but it cannot be used to **update** because its `PaymentOption` parameter requires either a new payment or a **receipt**, and no API returns a receipt today. This forces updates through `scratchpad_update` (no payment), which **does** require the secret key and auto-encrypts/signs, coupling read and write capabilities.

### Motivation
Enable a complete **secret-key–free** workflow (create + update) using `scratchpad_put`, so clients can use custom encryption and share decrypt-only keys without exposing the address’s secret key (and thus write access).

### Proposed change
- Return a `Receipt` from `scratchpad_put` responses.

#### Semantics
- **Create via `scratchpad_put`:** `Receipt` has exactly **one** entry (the proof and total cost for the created `XorName`).
- **Update via `scratchpad_put`:** `Receipt` is **empty** (no additional payment).

**Receipt type for reference (unchanged)**
```rust
pub type Receipt = HashMap<XorName, (ClientProofOfPayment, AttoTokens)>;
```

#### Breaking changes
- Return type (within `Result`) for `scratchpad_put` changes from `(AttoTokens, ScratchpadAddress)` to `(AttoTokens, ScratchpadAddress, Receipt)`, (adds update receipt).
- Any destructuring assuming a 2-tuple will fail.

### Migration
- Update call sites:
  - From: `(cost, addr) = scratchpad_put(...)`
  - To: `(cost, addr, receipt) = scratchpad_put(...)`
- If 2-tuple semantics must be kept temporarily, use a shim:
  - `scratchpad_put_legacy() → returns (cost, addr)`

### Compatibility / Release notes
- Public API: **major** version bump.

### Considerations
- The total cost (`AttoTokens`) is also present in `Receipt`; consider removing this redundancy later so consumers derive total cost from the receipt. For now, adding `Receipt` as the third tuple minimizes churn.

oetyng:
TL;DR: `scratchpad_put` now returns a `Receipt`; enables create+update without the address’s secret key.
Closes #3184.

Changes:
- Return `(AttoTokens, ScratchpadAddress, Receipt)` from `scratchpad_put`.
- Create: one-entry `Receipt`; Update: empty `Receipt`.

Breaking change: 2-tuple → 3-tuple.
Migration: `let (cost, addr, receipt) = scratchpad_put(...);`

### DIFF
--- a/autonomi-nodejs/src/lib.rs
+++ b/autonomi-nodejs/src/lib.rs
@@ -359,7 +359,7 @@ impl Client {
         scratchpad: &Scratchpad,
         payment_option: &PaymentOption,
     ) -> Result</* (AttoTokens, ScratchpadAddress) */ tuple_result::ScratchpadPut> {
-        let (cost, addr) = self
+        let (cost, addr, _receipt) = self
             .0
             .scratchpad_put(scratchpad.0.clone(), payment_option.0.clone())
             .await
--- a/autonomi/src/client/data_types/scratchpad.rs
+++ b/autonomi/src/client/data_types/scratchpad.rs
@@ -12,7 +12,7 @@ use crate::{
     Amount, AttoTokens, Client,
     client::{
         GetError, PutError,
-        payment::{PayError, PaymentOption},
+        payment::{PayError, PaymentOption, Receipt},
         quote::CostError,
     },
     networking::{NetworkError, PeerInfo},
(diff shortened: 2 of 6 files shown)


oetyng:
Closing as based on an incorrect premise.
I assumed scratchpad_put could not be used for updates without the Receipt variant of PaymentOption. Re-reading the code and tests shows updates already work: providing a wallet as PaymentOption succeeds and returns 0 cost, with no receipt involved.

There is still the possibility to extend an explicit secret-key-free, caller-controlled update flow with a fn e.g. scratchpad_put_update, that takes a caller-constructed Scratchpad and stores it verbatim after scratchpad_verify. Closing this issue in favor of that extension.

PR #3189, opened 2025-09-05
test: fork test, BLS test, CLI and Python API update

vphongph:
- Added example in Python for printing scratchpad fork error and doing the resolution
- Added scratchpad fork error printing in CLI in the command "get"
- Added scratchpad fork error warning while updating in CLI in the command "edit"
- Added CI test to trigger a fork and print it
- Replaced vec "latest" and "dedup_latest" from resolve_split_record() with a Hashset
- Added a test to make sure BLS signature is deterministic and BLS encryption is not. The test is testing the library blsttc.


ant-cli/src/commands.rs
```diff
@@ -293,6 +293,9 @@ pub enum ScratchpadCmd {
         /// Display the data as a hex string instead of raw bytes
         #[arg(long)]
         hex: bool,
+        /// Print detailed fork error analysis if a fork is detected
+        #[arg(long)]
+        print_fork_error: bool,
```
grumbach: Any reason to make this an option instead of it being the default behavior?

ant-cli/src/commands.rs
```diff
@@ -305,6 +308,9 @@ pub enum ScratchpadCmd {
         secret_key: bool,
         /// The new data to store in the scratchpad (Up to 4MB)
         data: String,
+        /// Print detailed fork error analysis if a fork is detected
+        #[arg(long)]
+        print_fork_error: bool,
```
grumbach: Same here

ant-cli/src/commands.rs
```diff
@@ -313,6 +319,16 @@ pub enum ScratchpadCmd {
...
+
+    /// Print full fork error information for a scratchpad (for debugging fork conflicts)
+    PrintForkError {
+        /// The name of the scratchpad.
+        name: String,
+        /// Indicate that this is an external scratchpad secret key.
+        /// (Use this when interacting with a scratchpad shared with you by someone else)
+        #[arg(short, long)]
+        secret_key: bool,
+    },
```
grumbach: I wonder if an interractive SolveFork command would be more helpful here?

grumbach (commented):
Nice work, left some UX related comments

maqi (changes requested):
According to the changes in the PR, the `fork` is not due to the `encryption randomness`, but `corruent update`, which is a usage scenario not in autonomi's control.

The `example` just to demonstrate one way to resolve such fork (sorting by signature), but in most case may not be the one that user wanted.

The printout of fork record may also not be required by user, as the content could be parsed using other format instead of direct string.

Hence, suggest to : 
1. commit and PR title replace `fix` keyword with something else, say `chore` 
2. autonomi client shall not be changed, i.e. the print_for_error flag is not necessary
3. CI shall got the example executed, if that example is useful
4. include unit tests to confirm: BLS signatures ARE deterministic, BLS encryption is NOT deterministic
5. the example shall be updated to use `same content` to update concurrently, instead of `different contents`

AND, for this particular case:
> the example shall be updated to use same content to update concurrently

the FORK shall be resolved on client side, as it shall be possible as client has the key to decrypt.
AND, it shall be resolved automatically, instead of a print then interact way.

autonomi/src/client/data_types/mod.rs
```diff
@@ -91,7 +91,7 @@ where
         }
         _multi => {
             warn!("Multiple conflicting records found at latest version for {key:?}");
-            Err(fork_error(latest))
+            Err(fork_error(dedup_latest))
```
maqi: `latest` shall be returned, to allow user resolve the conflicts based on num_of_appearance

autonomi/tests/scratchpad.rs
```diff
@@ -269,3 +269,183 @@ async fn scratchpad_errors() -> Result<()> {
...
+
+        if attempt >= MAX_ATTEMPTS {
+            panic!("Maximum attempts reached without fork detection - test failed");
+        }
+        println!("Retrying with new scratchpad...");
+    }
+    Ok(())
+}
+
+fn print_fork_details(
```
maqi: this seems duplicated with the `print_fork_analysis` and the python binding one.

may consider make this as a helper function in client, so that the test here, example in ant-cli and python binding can utilize the unified one.

autonomi/python/examples/scratchpad_fork_display_and_resolution.py
```diff
@@ -0,0 +1,161 @@
...
+
+def generate_test_data():
+    words = [
+        "apple", "banana", "cherry", "dragon", "forest", "galaxy", "harbor", 
+        "island", "kitchen", "mountain", "ocean", "penguin", "rainbow", 
+        "sunset", "treasure", "village", "butterfly", "castle", "garden"
+    ]
+    return " ".join(random.sample(words, WORD_SAMPLE_SIZE))
+
+def print_fork_analysis(fork_error, owner_key):
```
maqi, line 28: duplicated with `print_fork_details`

ant-cli/src/commands/scratchpad.rs
```diff
@@ -13,11 +13,67 @@ use crate::wallet::load_wallet;
...
+use autonomi::client::data_types::scratchpad::ScratchpadError;
 use autonomi::client::scratchpad::SecretKey as ScratchpadSecretKey;
 use color_eyre::Section;
 use color_eyre::eyre::Context;
 use color_eyre::eyre::Result;
 use color_eyre::eyre::eyre;
+use color_eyre::owo_colors::OwoColorize;
+
+/// Print detailed fork analysis for conflicting scratchpads
+fn print_fork_analysis(
```
maqi: duplicated with `print_fork_details`

ant-cli/src/commands/scratchpad.rs
```diff
@@ -199,10 +268,23 @@ pub async fn edit(
     // get network current scratchpad
     println!("Retrieving scratchpad from network...");
     let address = ScratchpadAddress::new(scratchpad_key.public_key());
-    let net_scratchpad = client
-        .scratchpad_get(&address)
-        .await
-        .wrap_err("Failed to retrieve scratchpad from network")?;
+    let scratchpad_result = client.scratchpad_get(&address).await;
```
maqi, line 219: the code block here is duplicated with `scratchpad_update` after PR 3198 merged in.
here only needs to call `client.scratchpad_update(scratchpad_key , data_type, content`)

maqi (changes requested):
ant-cli and python have couple of workflows (Fork error print or resolve) duplicated implemented.
which could result in some future maintains hassle.

Such duplicated workflow shall be defined on autonomi_client level so that they can use the unified handling workflow.

autonomi/python/examples/scratchpad_fork_display_and_resolution.py
vphongph: I let the `print_fork_analysis` in the python file. It's different from the Rust one and it's example, I want to let it explicit so users can see an example to handle the fork.

autonomi/src/client/data_types/mod.rs
vphongph: Replaced by Hashset as discussed

ant-cli/src/commands/scratchpad.rs
vphongph: I let the function `scratchpad_update_from` because we also look what's inside the local cache. It's a way to deal with spamming and update latency, latest scratchpad is also stored in cache.

So we need both `scratchpad_update_from` and `scratchpad_get`

maqi (approved):

autonomi/tests/scratchpad.rs
```diff
@@ -302,21 +302,26 @@ async fn scratchpad_fork_display() -> Result<()> {
...
-            let client_clone = client.clone();
+        for i in 0..=CONCURRENT_UPDATES_COUNT {
             let base_scratchpad_clone = base_scratchpad.clone();
             let owner_key_clone = owner_key.clone();
+            let client_instance = clients[i].clone();
 
             let task = tokio::spawn(async move {
                 let data = Bytes::from("Let's fork!");
-                let result = client_clone
+                let result = client_instance
```
vphongph, line 324: I don't think creating more client instances increase chance of fork, it's really about the concurrent updates.

Looking at the previous CI run, many update errors on this test were about failing to start a client. I think creating 10 concurrent clients is too much pressure for the network.

maqi: seems the success rate got improved with multiple client instances (though cut down the total number to fit into the CI VM spec).

anyway, will keep an eye on it.
if this test turned out being too risky, will suggest mark it as `ignored` (the testself is a very good example to case showing the concurrent scenario results in Fork) to avoid hindering the mergeing process.

PR #3191, opened 2025-09-05
feat(scratchpad): add wallet-free fn for caller-controlled updates

oetyng:
Closes #3185

### Summary
- Add `scratchpad_put_update` to perform wallet-free, caller-controlled updates by storing a fully formed, pre-signed `Scratchpad` verbatim after `scratchpad_verify`.

### Details
- Preconditions: address exists; owner signature valid.
- No counter bump, encryption, or re-signing; input is stored as-is.
- `scratchpad_put` still supports create/update; this method makes the walletless update path explicit and skips the payment flow.

### DIFF
--- a/autonomi/src/client/data_types/scratchpad.rs
+++ b/autonomi/src/client/data_types/scratchpad.rs
@@ -381,6 +381,75 @@ impl Client {
         Ok(scratchpad)
     }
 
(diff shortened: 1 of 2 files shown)


grumbach:
Nice one! Included and replaced by #3198

PR #3192, opened 2025-09-06
test(CI): scratchpad CI test

maqi:
### Description

Scratchpad CI test :

* sequential create and update
* ensure counter increased as expectation
* ensure content got updated properly as expected

### DIFF
--- a/.github/workflows/scratchpad.yml
+++ b/.github/workflows/scratchpad.yml
@@ -1,0 +1,252 @@
+name: Scratchpad Checks
+
+on:
+  merge_group:
+    branches: [main]
+  pull_request:
+    branches: ["*"]
+
+jobs:
+  scratchpad:
+    name: Scratchpad tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            ant_path: /home/runner/.local/share/autonomi
+          - os: windows-latest
+            ant_path: C:\\Users\\runneradmin\\AppData\\Roaming\\autonomi
+          - os: macos-latest
+            ant_path: /Users/runner/Library/Application\ Support/autonomi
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yml
```diff
@@ -593,6 +593,250 @@ jobs:
           log_file_prefix: ant_test_logs_e2e
           build: true
 
+  scratchpad:
```
jacderida: We could consider having each data type in its own workflow file, like the way there is a separate file for the file-related tests.

It's just because this `merge.yml` file is already quite large.

However, this is purely optional.

.github/workflows/merge.yml
```diff
@@ -593,6 +593,250 @@ jobs:
...
+        if: matrix.os != 'windows-latest'
+        run: echo "SECRET_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" >> $GITHUB_ENV
+        shell: bash
+      - name: Set secret key for Windows
+        if: matrix.os == 'windows-latest'
+        run: echo "SECRET_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        shell: pwsh
+
+      - name: Generate scratchpad signing key
+        run: ./target/release/ant --log-output-dest=data-dir --local scratchpad generate-key
```
jacderida: Just a minor point: I think we can drop the `--log-output-dest` argument because `data-dir` is the default value anyway.

It just makes the commands a little less verbose.

I think all the tests do this, not just the ones you've added. So we probably just keep copying it all the time.

jacderida (approved):

PR #3195, opened 2025-09-08
test(CI): pointer CI tests

maqi:
### Description

Pointer CI test :
* sequential create and update
* ensure counter increased as expectation
* ensure content got updated properly as expected

### DIFF
--- a/.github/workflows/pointer.yml
+++ b/.github/workflows/pointer.yml
@@ -1,0 +1,299 @@
+name: Pointer Checks
+
+on:
+  merge_group:
+    branches: [main]
+  pull_request:
+    branches: ["*"]
+
+jobs:
+  pointer:
+    name: Pointer tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            ant_path: /home/runner/.local/share/autonomi
+          - os: windows-latest
+            ant_path: C:\\Users\\runneradmin\\AppData\\Roaming\\autonomi
+          - os: macos-latest
+            ant_path: /Users/runner/Library/Application\ Support/autonomi
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Build binaries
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yml
```diff
@@ -593,6 +593,297 @@ jobs:
           log_file_prefix: ant_test_logs_e2e
           build: true
 
+  pointer:
```
jacderida: As per the scratchpad tests, we can consider having these in their own file.

However, purely optional.

.github/workflows/merge.yml
```diff
@@ -593,6 +593,297 @@ jobs:
...
+
+      # Create test data files for pointer targets
+      - name: Create test data files
+        run: |
+          echo "test_data_A" > test_data_A.txt
+          echo "test_data_B" > test_data_B.txt
+
+      # Upload test data to get addresses for pointer targets
+      - name: Upload test data A
+        run: ./target/release/ant --log-output-dest=data-dir --local file upload --public test_data_A.txt > ./upload_output_A 2>&1
```
jacderida: As per scratchpads, I don't think we need the `--log-output-dest` on the commands.

.github/workflows/merge.yml
```diff
@@ -593,6 +593,297 @@ jobs:
...
+        run: |
+          $ADDRESS_B = rg "At address: ([0-9a-f]*)" -o -r '$1' ./upload_output_B
+          echo "ADDRESS_B=$ADDRESS_B" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        shell: pwsh
+
+      - name: Generate pointer signing key
+        run: ./target/release/ant --log-output-dest=data-dir --local pointer generate-key
+
+      - name: Create pointer pointing to address_A
+        run: ./target/release/ant --log-output-dest=data-dir --local pointer create test_pointer ${{ env.ADDRESS_A }} --target-data-type chunk > ./pointer_create_output 2>&1
```
jacderida: Should we test with more types than chunk? Or would that be for another PR or maybe the staging tests?

maqi: the different type will be covered by other autonomi client level e2e tests.

this is for the ant-cli e2e test, and using Chunk only just for convenience.

jacderida (approved):

PR #3198, opened 2025-09-10
Various scratchpad improvements

grumbach:
PR including various improvements to scratchpad API:
- a new `scratchpad_put_update` method that allows the upload of custom scratchpads updates without payment as suggested by @oetyng (Thanks Ed! ❤️ )
- a fix in `scratchpad_update` which handles fork errors to allow for smooth fork resolution through a regular update

### DIFF
--- a/autonomi/src/client/data_types/scratchpad.rs
+++ b/autonomi/src/client/data_types/scratchpad.rs
@@ -381,6 +381,75 @@ impl Client {
         Ok(scratchpad)
     }
 
+    /// Stores the provided, fully-formed `Scratchpad` as an update.
+    /// Requires: (a) address exists, (b) owner signature is valid.
+    /// No counter bump, encryption, or signing is performed; input is stored verbatim.
+    /// Size/signature are checked via `scratchpad_verify`; network enforces versioning.
+    /// Note: For caller-controlled scratchpads, scratchpad_put supports create and update; use scratchpad_put_update when you want an explicit wallet-free update (skips the payment path).
+    /// The owner-key–managed path with enforced encryption/signing remains scratchpad_create / scratchpad_update.
+    pub async fn scratchpad_put_update(
+        &self,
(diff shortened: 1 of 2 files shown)


vphongph (approved):
Good PR!

Looks good and it's working as intended, I just tested it!

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -292,12 +292,16 @@ impl Client {
...
             Ok(scratchpad) => Some(scratchpad),
             Err(ScratchpadError::GetError(GetError::RecordNotFound)) => None,
+            // forks should not stop updates as updates are here to resolve forks, hence the max_by_key
             Err(ScratchpadError::GetError(GetError::Network(NetworkError::SplitRecord(
                 result_map,
             )))) => result_map
                 .values()
                 .filter_map(|record| try_deserialize_record::<Scratchpad>(record).ok())
                 .max_by_key(|scratchpad: &Scratchpad| scratchpad.counter()),
+            Err(ScratchpadError::Fork(scratchpads)) => scratchpads
```
maqi, line 302: the `scratch_get` (same as to `pointer_get`) already carried out `resolve_split_records` , which means the Fork returned here being the only situation of `same counter with different content`, hence `max_by_key` won't resolve (or resolve by mistake, just picking the first entry of the highest counter)

maqi (changes requested):

maqi (approved):
The `NetworkError::SplitRecord` is not got resolved within the `scratchpad_get` and always got translated into `ScratchpadError::Fork`, if the case of `same highest counter with different content` does happen.

Hence the handling branch here can now be removed.
Leave it for the future task of `purify GetError out of ScratchpadErrors`.

During the update, only `counter` matters, hence returned `Fork` can be resolved by just pick the first entry of the highest counter

PR #3203, opened 2025-09-16
Make missing file upload api stream

grumbach:
Comes after:
https://github.com/maidsafe/autonomi/pull/3202/
Please merge the above first

- `file_content_upload` and `file_content_upload_public` now support streaming without any API changes.


maqi (changes requested):
will approve once got the CI check part pushed in

maqi (approved):

PR #3204, opened 2025-09-16
fix(client): only early return during get when got unique content from enough peers

maqi:
### Description

During get, only early return when got unique content from enough peers (>= expected_holders)

### Type of Change


grumbach (commented):
What happened to keeping all copies including the "last honest node"?

maqi:
> What happened to keeping all copies including the "last honest node"?

it slows down the performance a lot.
hence Mick raised the original PR to improve the CLN performace.

grumbach (approved):
✅🚢

PR #3205, opened 2025-09-16
chore(evmlib): update alloy from `0.15.6` to `1.0.32`

oetyng:


RolandSherwin (approved):
Thank you @oetyng 😄

oetyng:
This old version was causing issues with the newer versions of `serde`, so was needing it to keep deps up to date on my end.

PR #3207, opened 2025-09-22
feat: use request/response direct messaging to replace kad::get_record

maqi:
### Description

For the future support of RBS , during fetching record from network, kad::get_record shall be replaced by using request/response direct messaging.

A local test (`CHUNK_DOWNLOAD_BATCH_SIZE` set to `8`) of downloading a 440MB file from mainnet showing a significant performance improvement (from previous 5min35secs to 3mins10secs) meanwhle with much less memory and traffic usage (around 1/3 of the previous.)

### Type of Change


maqi:
The work is put-on-hold as having a split download performance.
During the local and VM tests against the mainnet : 
* With super fast network connection, like 2.5 Gbps on DO, the stable (2025.9.1.1) performs much better than the PR 3207. Due to the nature of kad::get_record and its savy of connection fully satisfied.
* Meanwhile, with slow home connection, for example 50Mbps, PR 3207 performs better than the stable.

The `low quality of nodes` in mainnet also affects the performance of PR 3207 more significantly.
Hence at least delay this work till the `removal of relay` work got fully deployed to mainnet.

maqi:
For Chunk, the best is using req/rsp DM as a fall back approach, due to the performance consideration.
This has been implemented as in the PR https://github.com/maidsafe/autonomi/pull/3309 .

For other mutable data types, req/rsp DM shall be used as default, however it best to be implemented as an individual PR.

PR #3212, opened 2025-09-25
feat(python): extend graphentry methods

oetyng:
This PR gives access to `GraphEntry` internals to Python users. A few tests assumed details that differ from the current bindings so they were updated. The Python bindings tests weren't previously covered in CI, so a workflow for that was added as well.

Changes
- Exposes `GraphEntry.content()`, `GraphEntry.parents()`, and `GraphEntry.descendants()` in PyO3 bindings.
- Adds tests for the extended `GraphEntry` api.
- Aligns existing tests with the current API surface to make them compile and pass.
- Adds cross-platform CI for Python bindings.

Test alignment details:
 - Assertions updated for `GraphEntryAddress.hex` and `PointerAddress.hex` having 96-char (48-byte) hex strings.
 - Replaced `repr`/`eval` round-trips with `from_hex(...)` round-trips where applicable.

Workflow:
- GitHub Actions workflow builds PyO3 with maturin, installs via uv, and runs pytest on Ubuntu, macOS, and Windows.

### DIFF
--- a/.github/workflows/python-bindings.yml
+++ b/.github/workflows/python-bindings.yml
@@ -1,0 +1,65 @@
+name: python-bindings
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python: ['3.12']  # bump/add versions later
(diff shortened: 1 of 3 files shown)


vphongph (approved):
Looks good to me!

PR #3231, opened 2025-10-07
feat: paying three median nodes

maqi:
### Description

paying three median 3 nodes


maqi:
After discussion, it is ok to have the payment scheme to be :
* default: pay median node 3 times of its quote
* option: pay three highest nodes their quotes

Hence got this PR closed to save the smart_contract update work, and other test/release related works.

PR #3232, opened 2025-10-07
feat: cli analyze closest nodes

grumbach:
```bash
ant analyze --closest-nodes a7d2fdbb975efaea25b7ebe3d38be4a0b82c1d71e9b89ac4f37bc9f8677826e0
```

### DIFF
--- a/ant-cli/src/commands/analyze.rs
+++ b/ant-cli/src/commands/analyze.rs
@@ -153,8 +153,12 @@ async fn print_closest_nodes(client: &autonomi::Client, addr: &str, verbose: boo
         ));
     };
 
-    println!("Found {} closest peers to {}:", peers.len(), addr);
-    for (i, peer) in peers.iter().enumerate() {
+    // Sort peers by peer_id for consistent output
+    let mut sorted_peers = peers;
+    sorted_peers.sort_by(|a, b| a.peer_id.cmp(&b.peer_id));
+
+    println!("Found {} closest peers to {}:", sorted_peers.len(), addr);
+    for (i, peer) in sorted_peers.iter().enumerate() {
         println!("{}. Peer ID: {}", i + 1, peer.peer_id);
         if verbose {
             println!("   Addresses:");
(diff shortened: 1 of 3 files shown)


ant-cli/src/commands/analyze.rs
```diff
@@ -104,3 +113,61 @@ fn try_other_types(addr: &str, verbose: bool) {
...
+
+    // Sort peers by peer_id for consistent output
+    let mut sorted_peers = peers;
+    sorted_peers.sort_by(|a, b| a.peer_id.cmp(&b.peer_id));
+
+    println!("Found {} closest peers to {}:", sorted_peers.len(), addr);
+    for (i, peer) in sorted_peers.iter().enumerate() {
+        println!("{}. Peer ID: {}", i + 1, peer.peer_id);
+        if verbose {
+            println!("   Addresses:");
```
maqi, line 164: prefer the Addresses to be in the same line

ant-cli/src/commands/analyze.rs
```diff
@@ -104,3 +113,61 @@ fn try_other_types(addr: &str, verbose: bool) {
...
+        ));
+    };
+
+    // Sort peers by peer_id for consistent output
+    let mut sorted_peers = peers;
+    sorted_peers.sort_by(|a, b| a.peer_id.cmp(&b.peer_id));
+
+    println!("Found {} closest peers to {}:", sorted_peers.len(), addr);
+    for (i, peer) in sorted_peers.iter().enumerate() {
+        println!("{}. Peer ID: {}", i + 1, peer.peer_id);
```
maqi, line 162: prefer the block of code to be : 
```
if verbose {
    println!("{}. Peer ID: {} - {:?}", i + 1, peer.peer_id, peer.addrs);
} else {
    println!("{}. Peer ID: {}", i + 1, peer.peer_id);
}
```

maqi (commented):
Better to include an output in the PR description as an example as well.

ant-cli/src/commands/analyze.rs
grumbach: As addresses are very long this will make the printing very ugly and hard to read.

ant-cli/src/commands/analyze.rs
grumbach: I've actually tried both, and this one really made the difference (the output is much cleaner)

grumbach:
Replaced by https://github.com/maidsafe/autonomi/pull/3233

PR #3235, opened 2025-10-08
fix(bootstrap_cache): push newer addresses to the front

RolandSherwin:
- The newer addresses should be pushed to the front. This allows the clients to read the latest addresses rather than stale ones.

### DIFF
--- a/ant-bootstrap/src/cache_store/cache_data_v1.rs
+++ b/ant-bootstrap/src/cache_store/cache_data_v1.rs
@@ -12,7 +12,7 @@ use atomic_write_file::AtomicWriteFile;
 use libp2p::{Multiaddr, PeerId};
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::VecDeque,
+    collections::{HashMap, VecDeque},
     fs::{self, OpenOptions},
     io::{Read, Write},
     path::{Path, PathBuf},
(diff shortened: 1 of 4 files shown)


ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -47,18 +48,19 @@ impl CacheData {
                 if peer == other_peer {
                     for addr in other_addrs.iter() {
                         if !addrs.contains(addr) {
-                            addrs.push_back(addr.clone());
+                            addrs.push_front(addr.clone());
```
maqi: as comment says, `self` is newer.
here `addr` is from `other`, which supposed to be `older`, then why push to the front ?

at least put a comment here

ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -47,18 +48,19 @@ impl CacheData {
...
                         }
                     }
                     while addrs.len() > max_addrs_per_peer {
-                        addrs.pop_front();
+                        addrs.pop_back();
```
maqi: put a comment of `remove old from tail` here

ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -82,7 +84,7 @@ impl CacheData {
         self.last_updated = SystemTime::now();
     }
 
-    /// Add a peer to the cachse data
+    /// Add a peer to front of the cache.
```
maqi: -> Add a peer to front of the cache as the newest, pruning old from tail

ant-bootstrap/src/cache_store/cache_data_v1.rs
```diff
@@ -32,7 +32,8 @@ impl CacheData {
     /// This has to be bumped whenever the cache data format changes to ensure compatibility.
     pub const CACHE_DATA_VERSION: u32 = 1;
 
-    /// Sync the self cache with another cache. Self peers (newer) are preserved.
+    /// Sync the self cache with another cache. Self peers (newer) are preserved by pushing
+    /// other peers to the back.
```
maqi, line 36: side note: construct a temp hashmap clone of `other`, so that the existence check can be carried out using `remove` instead of more time consuming `for loops`.

maqi (changes requested):
Put in Request Change as the block of:
```diff
            for (peer, addrs) in self.peers.iter_mut() {
                if peer == other_peer {
                    for addr in other_addrs.iter() {
                        if !addrs.contains(addr) {
-                           addrs.push_back(addr.clone());
+                           addrs.push_front(addr.clone());
```
seems being conflict with the comment in context and the PR title

RolandSherwin:
Resolved all the comments there @maqi, Thanks for all the suggestions 🙇

maqi (approved):

RolandSherwin:
closing in favour of https://github.com/maidsafe/autonomi/pull/3242

PR #3239, opened 2025-10-08
chore: roll back self_encryption to 0.34.0

maqi:
### Description

roll back self_encryption to 0.34.0


maqi:
not being the reason of performance regression

PR #3240, opened 2025-10-09
fix(node): evict peer immediately on WrongPeerId error

RolandSherwin:
- The `WrongPeerId` is triggered when someone restarts or resets their node and a new node gets started on the same `ip:port` combination, but with a different `peer_id`.
- But this old `multiaddr` with the `old_peer_id` will still be part of the other node's RT and they're not cleared for a VERY LONG time (as the nodes has to accumulate this error 3 times, which will take a long while if the nodes do not initiate contact with this peer in the RT)
- This leads to downstream effects and causes regression for the whole network.

### DIFF
--- a/ant-node/src/networking/driver/event/swarm.rs
+++ b/ant-node/src/networking/driver/event/swarm.rs
@@ -372,7 +372,7 @@ impl SwarmDriver {
                 );
 
                 // we need to decide if this was a critical error and if we should report it to the Issue tracker
-                let is_critical_error = match &error {
+                let issue_to_record = match &error {
                     DialError::Transport(errors) => {
                         // as it's an outgoing error, if it's transport based we can assume it is _our_ fault
                         //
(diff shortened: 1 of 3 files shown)


ant-node/src/networking/driver/event/swarm.rs
```diff
@@ -443,50 +443,55 @@ impl SwarmDriver {
...
+                        Some(NodeIssue::ConnectionIssue)
                     }
                 };
 
-                if is_critical_error {
+                if let Some(issue) = issue_to_record {
+                    let issue_for_logging = issue.clone();
                     warn!(
-                        "Outgoing Connection error to {failed_peer_id:?} is considered as critical. Marking it as an issue. Error: {error:?}"
+                        "Outgoing Connection error to {failed_peer_id:?} is considered as critical due to {issue_for_logging:?}. Marking it as an issue. Error: {error:?}"
```
maqi: it's not `due to {issue_for_logging:?}` , better change to:
`... is considered as critial. Marking it as an {issue_for_logging:?} . Error: {error:?}`

maqi (approved):
The PR Description better to be corrected : 

> when someone resets their node 

when someone resets or restarts their nodes

maqi (approved):

jacderida:
Closing in favour of #3244

PR #3253, opened 2025-10-15
feat(workflow): add macOS binary signing

mickvandijke:
Automatically sign all macOS binaries in the CI.

TODO:
- [ ] Do a test release and test that all the binaries for Mac are signed.


maqi:
Closed as the multiple OS native intaller is now supported, with the PR https://github.com/maidsafe/autonomi/pull/3432

PR #3256, opened 2025-10-16
fix(bootstrap): drive the dialing process via external loop

RolandSherwin:
- The dialing process prematurely stops if there are background async tasks that are still in flight. This means that the `Bootstrap` is not a "self-driving" struct, rather has to be driven by an external loop. This PR hence periodically drives the dialing process.

### DIFF
--- a/ant-bootstrap/src/bootstrap.rs
+++ b/ant-bootstrap/src/bootstrap.rs
@@ -213,23 +213,23 @@ impl Bootstrap {
     /// Error if we have exhausted all sources and have no more addresses to return.
     ///
     /// This does not start any dial attempts, it returns the next address to dial.
-    /// Use `trigger_bootstrapping_process` to start auto manage the whole bootstrapping process.
+    /// Use `trigger_bootstrapping_process` to poll the dialing process.
     pub fn next_addr(&mut self) -> Result<Option<Multiaddr>> {
         loop {
             self.process_events();
 
             if let Some(addr) = self.addrs.pop_front() {
-                info!(?addr, "next_addr returning queued address");
+                info!("Returning next bootstrap address: {addr}");
                 return Ok(Some(addr));
             }
 
-            if self.fetch_in_progress.is_some() {
-                debug!("next_addr waiting for in-flight fetch result");
+            if let Some(fetch_kind) = self.fetch_in_progress {
+                debug!("Fetch in progress: {fetch_kind:?}; waiting for addresses");
                 return Ok(None);
             }
 
             if self.cache_pending && !matches!(self.fetch_in_progress, Some(FetchKind::Cache)) {
-                info!("next_addr triggering cache fetch");
+                info!("Triggering cache fetch");
                 self.start_cache_fetch()?;
                 continue;
             }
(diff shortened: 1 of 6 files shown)


ant-bootstrap/src/bootstrap.rs
```diff
@@ -330,52 +331,41 @@ impl Bootstrap {
                 .config()
                 .max_contacted_peers_before_termination
         {
-            self.bootstrap_completed = true;
```
maqi, line 333: not fully get it.

In this case, why the `self.bootstrap_completed` no longer need to be populated with true ?

ant-bootstrap/src/bootstrap.rs
```diff
@@ -330,52 +331,41 @@ impl Bootstrap {
...
+            return true;
         }
 
-        true
+        false
     }
 
+    /// Manages the bootstrapping process by attempting to dial peers from the available addresses.
+    ///
+    /// Returns `true` if the bootstrapping process has completed successfully, otherwise `false`.
```
maqi: according to this comment, a `true` will only be returned when bootstrap `succeeded`.
However, looking into the `has_bootstrap_completed`, seems a `true` will be returned when all addresses drained up, doesn't matter if really got connected to network or not.
hence, the situation of all contacts failed to connect may also be considered as a bootstrap success ?

ant-bootstrap/src/bootstrap.rs
```diff
@@ -330,52 +331,41 @@ impl Bootstrap {
...
     pub fn trigger_bootstrapping_process<B: NetworkBehaviour>(
         &mut self,
         swarm: &mut Swarm<B>,
         contacted_peers: usize,
-    ) {
-        if !self.should_continue_bootstrapping(contacted_peers) {
-            return;
+    ) -> bool {
+        if self.has_bootstrap_completed(contacted_peers) {
+            self.bootstrap_completed = true;
```
maqi, line 366: for the function capsulation, may these three population to be remained within the `has_bootstrap_completed` function ?

ant-node/src/networking/driver/mod.rs
```diff
@@ -249,19 +247,19 @@ impl SwarmDriver {
                         let _ = self.dial_queue.remove(peer_id);
                     }
                 },
-
-                // check if we can trigger the initial bootstrap process
```
maqi, line 253: comment about this code block still being necessary to explain why doing so ?

autonomi/src/networking/driver/mod.rs
```diff
@@ -226,6 +229,16 @@ impl NetworkDriver {
                         error!("Error processing swarm event: {e}");
                     }
                 }
+                _ = async {
```
maqi, line 234: comment better to be presented

ant-bootstrap/src/bootstrap.rs
RolandSherwin: Because the fn now just checks if bootstrap has completed `has_bootstrap_completed`, which now returns a `true` if completed. The caller has to mutate the fields if required including the following:
```
self.addrs.clear();
self.ongoing_dials.clear();
```

ant-bootstrap/src/bootstrap.rs
RolandSherwin: Right yeah, will update the comments!

ant-bootstrap/src/bootstrap.rs
RolandSherwin: `has_bootstrap_completed`  has 3 places that would repeat the same mutation operations, hence better to keep a single place to mutate self.

maqi (approved):

PR #3261, opened 2025-10-22
feat: merkle tree protocol

grumbach:
Introduces Merkle Tree mechanisms in the protocol

### DIFF
--- a/ant-protocol/Cargo.toml
+++ b/ant-protocol/Cargo.toml
@@ -29,7 +29,7 @@ prometheus-client = { version = "0.23.1" }
 prost = { version = "0.9", optional = true }
 rand = "0.8"
 rmp-serde = "1.1.1"
-rs_merkle = { path = "../../rs-merkle" }
+rs_merkle = { git = "https://github.com/maidsafe/rs-merkle.git", branch = "master" }
 serde = { version = "1.0.133", features = ["derive", "rc"] }
 serde_json = "1.0"
 sha2 = "0.10.7"
--- a/ant-protocol/Cargo.toml
+++ b/ant-protocol/Cargo.toml
@@ -16,6 +16,7 @@ rpc = ["tonic", "prost"]
 [dependencies]
 ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
+blake2 = "0.10"
 bls = { package = "blsttc", version = "8.0.1" }
 bytes = { version = "1.0.1", features = ["serde"] }
 color-eyre = "0.6.3"
@@ -28,6 +29,7 @@ prometheus-client = { version = "0.23.1" }
 prost = { version = "0.9", optional = true }
 rand = "0.8"
 rmp-serde = "1.1.1"
+rs_merkle = { path = "../../rs-merkle" }
 serde = { version = "1.0.133", features = ["derive", "rc"] }
 serde_json = "1.0"
 sha2 = "0.10.7"
--- a/ant-protocol/src/messages/merkle_tree.rs
+++ b/ant-protocol/src/messages/merkle_tree.rs
@@ -1,0 +1,1821 @@
+// Copyright 2025 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use rs_merkle::Hasher;
+use serde::{Deserialize, Serialize};
+use std::time::{SystemTime, UNIX_EPOCH};
+use thiserror::Error;
+use xor_name::XorName;
+
+/// Maximum tree depth (supports 65,536 chunks = 256GB)
+pub const MAX_MERKLE_DEPTH: u8 = 16;
+
+/// Minimum number of leaves (chunks) for a Merkle tree
+pub const MIN_LEAVES: usize = 2;
+
+/// Maximum number of leaves (2^16)
+pub const MAX_LEAVES: usize = 1 << MAX_MERKLE_DEPTH;
+
+/// Maximum age of a Merkle payment (one week in seconds)
+/// Payments older than this are considered expired and nodes will reject chunks
+pub const MERKLE_PAYMENT_EXPIRATION: u64 = 7 * 24 * 60 * 60; // 7 days
+
+/// Errors that can occur when working with Merkle trees
+#[derive(Debug, Error)]
+pub enum MerkleTreeError {
+    #[error("Too few leaves: got {got}, minimum is {MIN_LEAVES}")]
+    TooFewLeaves { got: usize },
+    #[error("Too many leaves: got {got}, maximum is {MAX_LEAVES}")]
+    TooManyLeaves { got: usize },
+    #[error("Invalid leaf index: {index} (tree has {leaf_count} leaves)")]
+    InvalidLeafIndex { index: usize, leaf_count: usize },
+    #[error("Invalid midpoint index: {index} (tree has {midpoint_count} midpoints)")]
+    InvalidMidpointIndex { index: usize, midpoint_count: usize },
+    #[error("Invalid proof")]
+    InvalidProof,
+    #[error("Internal error: {0}")]
+    Internal(String),
+}
+
+pub type Result<T> = std::result::Result<T, MerkleTreeError>;
+
+/// A Merkle tree built from XorNames (chunk addresses)
+///
+/// Used for batch payment system where:
+/// - Build tree from chunk XorNames (min: 2, max: 65,536)
+/// - Tree is automatically padded to next power of 2
+/// - Root is committed on-chain for payment
+/// - Intersections at level depth/2 determine candidate pools
+/// - Individual chunk proofs verify chunks belong to paid batch
+pub struct MerkleTree {
+    /// The underlying rs_merkle tree
+    inner: rs_merkle::MerkleTree<Blake2bHasher>,
+
+    /// Original leaf count (before padding)
+    leaf_count: usize,
+
+    /// Tree depth
+    depth: u8,
+
+    /// The root hash of the tree
+    root: XorName,
+
+    /// Salts for each original leaf
+    /// Used to compute salted hashes: hash(chunk || salt)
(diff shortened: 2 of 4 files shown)


ant-protocol/Cargo.toml
```diff
@@ -28,6 +29,7 @@ prometheus-client = { version = "0.23.1" }
 prost = { version = "0.9", optional = true }
 rand = "0.8"
 rmp-serde = "1.1.1"
+rs_merkle = { git = "https://github.com/maidsafe/rs-merkle.git", branch = "master" }
```
grumbach, line 32: We're still waiting for https://github.com/antouhou/rs-merkle/pull/48
In the meantime we can use our fork.

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use rs_merkle::Hasher;
+use serde::{Deserialize, Serialize};
+use std::time::{SystemTime, UNIX_EPOCH};
+use thiserror::Error;
+use xor_name::XorName;
+
+/// Maximum tree depth (supports 65,536 chunks = 256GB)
+pub const MAX_MERKLE_DEPTH: u8 = 16;
```
maqi, line 16: still on discussion

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+
+    /// Tree depth
+    depth: u8,
+
+    /// The root hash of the tree
+    root: XorName,
+
+    /// Salts for each original leaf
+    /// Used to compute salted hashes: hash(chunk || salt)
+    salts: Vec<[u8; 32]>,
```
maqi, line 70: it has been discussed that random `salts` shall be avoided? 

the `quoting_metrics` can be salts anyway ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+        if leaf_count < MIN_LEAVES {
+            return Err(MerkleTreeError::TooFewLeaves { got: leaf_count });
+        }
+        if leaf_count > MAX_LEAVES {
+            return Err(MerkleTreeError::TooManyLeaves { got: leaf_count });
+        }
+
+        // Generate random salt for each real leaf (privacy protection)
+        let mut rng = rand::thread_rng();
+        let salts: Vec<[u8; 32]> = (0..leaf_count)
```
maqi, line 115: same as above, avoid using random salts, instead can use quoting_metrics (or its hash) ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+        // Apply salt to each real leaf: hash(chunk || salt)
+        let mut salted_leaves: Vec<[u8; 32]> = leaves
+            .iter()
+            .zip(&salts)
+            .map(|(chunk, salt)| {
+                // Compute hash(chunk || salt)
+                let mut data = Vec::with_capacity(64);
+                data.extend_from_slice(chunk.as_ref());
+                data.extend_from_slice(salt);
+                Blake2bHasher::hash(&data)
```
maqi, line 136: there is std::hash facilities.
any particular reason to use this new crate ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+            })
+            .collect();
+
+        // Add random dummy padding leaves (no salt needed - already random)
+        if leaf_count < padded_size {
+            for _ in leaf_count..padded_size {
+                let mut dummy = [0u8; 32];
```
maqi, line 143: better as fixed, instead of random ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+        let midpoints = self.midpoints()?;
+
+        midpoints
+            .into_iter()
+            .map(|midpoint| {
+                // Generate proof for this midpoint
+                let branch = self.generate_midpoint_proof(midpoint.index, midpoint.hash)?;
+
+                Ok(RewardCandidatePool {
+                    branch,
```
maqi, line 250: the `index` and `hash` of the midpoint better to be included in this struct entry as well ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+        let indices = vec![chunk_index];
+        let proof = self.inner.proof(&indices);
+
+        // Padded size is 2^depth
+        let padded_size = 1 << self.depth;
+
+        let root = self.root();
+
+        // Get the salt for this chunk
+        let salt = self.salts[chunk_index];
```
maqi, line 324: this means the `position` of chunks is vital, and leaving it being outside of the struct is bit risky.
ideally, there shall be a hashmap, which maps the chunk_name to its correspondent salt (quoting_metrics or its hash) and merkle_tree index.
So that a caller only needs to know the chunk_name

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+    index: usize,
+}
+
+/// A reward candidate derived from a midpoint
+///
+/// The candidate address is computed as hash(midpoint_hash || root || tx_timestamp).
+/// Network nodes closest to this address are eligible for batch payment rewards.
+/// Contains everything needed to verify the candidate is valid.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct RewardCandidatePool {
```
maqi, line 410: first this is just a `merkle proof` for a leave ?
second, this is just for `one` leave, not for entire list, hence naming with `Pool` is bit consufing ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+
+impl RewardCandidatePool {
+    /// Get the Merkle root from the candidate pool's branch
+    ///
+    /// Returns the root hash that the branch proves membership in.
+    pub fn root(&self) -> &XorName {
+        self.branch.root()
+    }
+
+    /// Get the candidate address for this pool
```
maqi, line 426: target address for this zone ?

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+        data.extend_from_slice(self.branch.root().as_ref());
+        data.extend_from_slice(&self.tx_timestamp.to_le_bytes());
+        XorName::from_content(&data)
+    }
+
+    /// Hash the entire pool struct for smart contract storage
+    ///
+    /// This uses deterministic serialization (MessagePack) followed by hashing.
+    /// The smart contract stores this hash to commit to the winner pool.
+    pub fn hash(&self) -> std::result::Result<XorName, rmp_serde::encode::Error> {
```
maqi, line 442: that means `num of midpoints` entries to be stored on chain ?
for 2^16 max, that means 256 entries, which could cost certain

ant-protocol/src/messages/merkle_tree.rs
```diff
@@ -0,0 +1,1821 @@
...
+/// For midpoint proofs:
+/// - leaf_index is the midpoint index at its level
+/// - total_leaves_count is the number of midpoints
+/// - salt is None (midpoints are intermediate hashes, not raw chunks)
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct MerkleBranch {
+    /// The proof hashes (sibling hashes only) from leaf/midpoint to root
+    proof_hashes: Vec<[u8; 32]>,
+
+    /// Index of the leaf/node in the tree
```
maqi, line 467: Index of the leaf/node in the tree on that level (leaves level or midpoint level)

maqi (changes requested):
put in Request Changes as:
* the work may best to be raised as a separate helper crate, like XorName; or as a separate mod, say ant-merkle-tree . instead of to be included in ant_protocol, which supposed to only define the messaging format between client and node.
* There is on-discussion parts, such as max_depth, random salts and how to fill empty leaves.

ant-protocol/src/messages/merkle_tree.rs
grumbach: For best privacy it's best to keep it random as quoting metrics can be guessed

ant-protocol/src/messages/merkle_tree.rs
grumbach: 100%

ant-protocol/src/messages/merkle_tree.rs
grumbach: Fixed ones would give away information on the position in a tree for some chunks. For example if I get a chunk and its sibling (for the branch) is identified as a padding one, I know it's the last chunk of something! Which is dangerous. Randomness here makes them unidentifiable by nodes.

ant-protocol/src/messages/merkle_tree.rs
grumbach: Why? No one except the client ever sees the tree. So once generated the rest can be kept secret to avoid leaking information on the tree. Remember that self encryption is broken if we know the position of chunks and their hashes so privacy in this tree is vital.

ant-protocol/src/messages/merkle_tree.rs
grumbach: This is the proof for a midpoint which represents a pool of nodes (20) which might get a reward. 
So not a leaf (chunk).

ant-protocol/src/messages/merkle_tree.rs
grumbach: No. Only one midpoint (hashed) is to be on chain. So total on chain is that hash.

ant-protocol/src/messages/merkle_tree.rs
grumbach: Yeah just made it a top file constant so we can edit as we go

ant-protocol/src/messages/merkle_tree.rs
grumbach: This is an internal type which cannot be accessed by the caller. We could make it a hashmap but it would SIGNIFICANTLY make it bigger as we would need to store each chunk hash too. For HUGE trees that's a lot of waste + slower retrieval.

ant-protocol/src/messages/merkle_tree.rs
grumbach: Yeah absolutely.

grumbach:
ant-protocol might not be the best crate to include this work at, moving to another PR, but good to have had this reviewed!

PR #3262, opened 2025-10-22
feat: remove relay functionality from autonomi network 

RolandSherwin:
- Deleted the entire relay_manager.rs module (~600 lines) which handled NAT traversal for private nodes   
  - Removed relay client/server behaviours from the libp2p swarm                                            
  - Simplified the transport layer (was dual relay/direct, now single direct path)                          
  - Removed relay event handlers, address crafting, and candidate selection                                 
  - Kept `--relay` flag in antnode as a no-op for backwards compatibility with existing service definitions   
  - Metrics preserved but marked deprecated


.gitignore
```diff
@@ -51,3 +51,6 @@ site/
 
 # Some test run artifacts
 autonomi/tests/file/test_dir_fetched*
+
+# Claude
+.claude/
```
maqi: This is not part of the PR.
may better raise it as separate ?

ant-node-manager/src/cmd/node.rs
```diff
@@ -128,7 +127,7 @@ pub async fn add(
         enable_metrics_server,
         evm_network: evm_network.unwrap_or(EvmNetwork::ArbitrumOne),
         env_variables,
-        relay,
+        relay: false,
```
maqi: shall the struct of `AddNodeServiceOptions` also got the `relay` field removed ?

maqi (changes requested):
Better got this PR having its own PR descriptions to cover what changed and what aimed for.

Also, seems there is some outdated stuff in the PR description of PR 3116.
For example, it says:
* ✅ Remove relay CLI options (--relay flag)
BUT, in the PR 3262 code change, the flag remains, though it does nothing.
Better to update to sync.

Put in Request Changes, as it needs sepecific testnet runs to confirm the work, before got merged in.

.gitignore
```diff
@@ -55,3 +55,6 @@ site/
 # Some test run artifacts
 autonomi/tests/file/test_dir_fetched*
 .serena/
+
+# Claude
```
maqi: best move to separate PR

ant-node/src/networking/metrics/mod.rs
```diff
@@ -41,23 +41,30 @@ pub(crate) struct VersionLabels {
 }
 
 /// The shared recorders that are used to record metrics.
+#[allow(dead_code)] // TODO: Remove this once we have removed all the deprecated metrics related to relay.
```
maqi: this suppresses warnings for ALL fields, not just deprecated ones.
Better to mark only specific deprecated fields or remove them entirely.

maqi (commented):
* Metrics cleanup incomplete: Deprecated metrics like `relay_reservation_health`, `relay_peers_percentage`, `connected_relay_clients` are kept but never updated. Consider removing in a follow-up.
* Add documentation about the breaking change for NAT users.

maqi (approved):

PR #3263, opened 2025-10-23
improvement(autonomi): improve process_tasks_with_max_concurrency performance

RolandSherwin:
Although the previous `FuturesUnordered.next().await` does return with the quickest, it is `manual` polling patten got used, and being in-efficient.
Meanwhile, `buffer_unordered` is highly optimized in the futures crate with better internal scheduling and less overhead.

Switching to use `buffer_unordered` proved can improve the download speed significatly.
In one case, previously a 4GB file downloaded in 25 mins, with a bandwidth usage pattern in `occasionally peaked`;
can now be downloaded in 3 mins, with a flat full bandwidth usage pattern.

### DIFF
--- a/autonomi/src/utils.rs
+++ b/autonomi/src/utils.rs
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::client::PutError;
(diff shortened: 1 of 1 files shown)


grumbach (approved):
LGTM

PR #3265, opened 2025-10-23
feat(launchpad): auto elevate permissions for windows

RolandSherwin:
- Remove the custom panic handler as that was not required and also use `ant-logging` to initialize the logging subscriber.
- Auto elevate permissions on windows by asking for admin rights IF we started without admin rights.


maqi (changes requested):
there is Cargo.lock change on `runas` shall be commited in as well.

maqi (approved):

PR #3268, opened 2025-10-24
feat(bootstrap): split Bootstrap::new method into two

RolandSherwin:
- Split the `async fn Bootstrap::new` method into two methods, where the `new` is made a "sync" method and `async fn new_with_preloaded_addrs` retains the same functionality as the old new method.
- The `new_with_preloaded_addrs` method is only required inside clients which tries to pre-populate its RT before starting the libp2p driver.
- Introduces `is_addr_queue_empty`, as it will be used by the reachability check implementation
- Deprecate `try_next_dial_addr` as it is just a copy of `next_addr`

### DIFF
--- a/ant-bootstrap/src/bootstrap.rs
+++ b/ant-bootstrap/src/bootstrap.rs
@@ -81,6 +81,8 @@ pub struct Bootstrap {
 impl Bootstrap {
     /// Create a new Bootstrap manager with the given configuration.
     /// Use `new_with_preloaded_addrs` to ensure that the struct contains at least MIN_INITIAL_ADDRS addresses immediately.
+    ///
+    /// Must be called from a tokio runtime context.
     pub fn new(config: BootstrapConfig) -> Result<Self> {
         let contacts_progress = Self::build_contacts_progress(&config)?;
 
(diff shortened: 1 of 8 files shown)


ant-bootstrap/src/bootstrap.rs
```diff
@@ -79,7 +79,11 @@ pub struct Bootstrap {
 }
 
 impl Bootstrap {
-    pub async fn new(mut config: BootstrapConfig) -> Result<Self> {
+    /// Create a new Bootstrap manager with the given configuration.
+    /// Use `new_with_preloaded_addrs` to ensure that the struct contains at least MIN_INITIAL_ADDRS addresses immediately.
+    ///
+    /// Must be called from a tokio runtime context.
+    pub fn new(config: BootstrapConfig) -> Result<Self> {
```
maqi, line 93: > to ensure that the struct contains at least MIN_INITIAL_ADDRS addresses immediately

Does this need to be `always enforced`, i.e. replace this `new` with `new_with_preloaded_addrs`, or raise an error if not having MIN_INITIAL_ADDRS addresses ?

ant-bootstrap/src/lib.rs
```diff
@@ -62,49 +62,30 @@ pub fn craft_valid_multiaddr(addr: &Multiaddr, ignore_peer_id: bool) -> Option<M
...
     let udp = addr
         .iter()
-        .find(|protocol| matches!(protocol, Protocol::Udp(_)));
-    let tcp = addr
+        .find(|protocol| matches!(protocol, Protocol::Udp(_)))?;
+
+    output_address.push(udp);
+    let quic = addr
         .iter()
-        .find(|protocol| matches!(protocol, Protocol::Tcp(_)));
```
maqi, line 68: just to confirm that with this removal, TCP/WebSocket transport is no longer supported.

ant-bootstrap/src/bootstrap.rs
```diff
@@ -135,11 +139,22 @@ impl Bootstrap {
         bootstrap.cache_task = Some(cache_task);
 
         if config.first {
-            info!("First node in network; clearing any existing cache");
-            bootstrap.cache_store.write().await?;
-            return Ok(bootstrap);
+            tokio::spawn(async move {
```
maqi: * The spawned task has a clone of cache_store, but the original bootstrap uses a different clone - is this intentional?
* A first node that fails to clear its cache will continue running with stale data

Bit edge cased, but may happen ?

ant-node/src/bin/antnode/main.rs
```diff
@@ -317,7 +317,7 @@ fn main() -> Result<()> {
 
     let bootstrap_config = BootstrapConfig::try_from(&opt.peers)?
         .with_backwards_compatible_writes(opt.write_older_cache_files);
-    let bootstrap = rt.block_on(Bootstrap::new(bootstrap_config))?;
+    let bootstrap = rt.block_on(async { Bootstrap::new(bootstrap_config) })?;
```
maqi: Since `Bootstrap::new()` is now sync, wrapping it in `async {}` and block_on is unnecessary. This should just be:
```
let bootstrap = Bootstrap::new(bootstrap_config)?;
```

ant-bootstrap/src/bootstrap.rs
```diff
@@ -250,6 +265,29 @@ impl Bootstrap {
         }
     }
 
+    /// Check if the address queue is empty without consuming an address.
+    /// Returns `Some(true)` if empty, `Some(false)` if not empty, and `None` if still fetching or has cache or contacts
+    /// sources pending.
+    pub fn is_addr_queue_empty(&self) -> Option<bool> {
```
maqi, line 291: This three-state return value could be confusing. Consider using an enum for clarity:
```
enum QueueStatus { HasAddresses, Empty, FetchPending }
```

ant-bootstrap/src/bootstrap.rs
```diff
@@ -135,11 +139,22 @@ impl Bootstrap {
...
+                    info!("Bootstrap cache cleared for first node");
+                }
+            });
         }
 
+        Ok(bootstrap)
+    }
+
+    /// Create a new Bootstrap manager and ensure it has at least MIN_INITIAL_ADDRS addresses preloaded.
+    pub async fn new_with_preloaded_addrs(config: BootstrapConfig) -> Result<Self> {
```
maqi, line 175: It seems it's now preferred to always use `new_with_preloaded_addrs` ?
i.e. changing `Bootstrap::new(config)` to `Bootstrap::new_with_preloaded_addrs(config).await` ?
if so, this will be a breaking change requires all downstream callers to be updated ?

maqi (approved):

ant-node/src/bin/antnode/main.rs
RolandSherwin: Nice catch, updated it!

ant-bootstrap/src/lib.rs
RolandSherwin: Yes, TCP/ws are not supported in current main. Just udp, this is just cleaning up the bootstrap method accordingly.

ant-bootstrap/src/bootstrap.rs
RolandSherwin: The struct uses `Arc` internally, so we should be fine. Yeah, the other thing is an edge case, but we usually run the `first node` on our machines, so ig we were fine earlier and we should be fine now

ant-bootstrap/src/bootstrap.rs
RolandSherwin: `is_addr_queue_empty` is used only inside the tests for now, so we should be good?

ant-bootstrap/src/bootstrap.rs
RolandSherwin: Yes, marking the commit as a breaking one

ant-bootstrap/src/bootstrap.rs
RolandSherwin: Updated the comments to reflect the changes!

PR #3269, opened 2025-10-24
feat(logging): log to file or dir dynamically based on path

RolandSherwin:
- Logs are directed to a single log file if a file with `.log` extension is provided, otherwise the logs are sent to a directory with log rotation enabled
- Also automatically create proper test file names for unit tests. This fixes an issue where running multiple unit tests could send all the logs to a single file.


ermineJose (approved):
Looks Good!

PR #3270, opened 2025-10-24
feat(node): implement the reachability check module

RolandSherwin:
Ontop of https://github.com/maidsafe/autonomi/pull/3268. This PR is just the implementation of the reachability module. The actual integration with antctl, antnode and launchpad will be done as separate PRs.

This workflow checks if a node can receive direct inbound connections from other peers on the network. If a node can't receive connections, it won't be able to participate in the network properly, so we prevent it from starting.

The workflow tries one network listener at a time, with up to 3 attempts per listener. A device could have multiple listeners/local adapters, and the workflow tries to check if they're externally reachable one at a time. This is the flow for a single adapter/listener:

1. Bind to a listener.
2. Dial out to 7 bootstrap peers concurrently
3. Each peer that connects dials us back after 180 seconds
4. If we get 4+ matching dial-backs, we're reachable
5. If it fails, retry up to 2 more times on the same listener
6. If all 3 attempts fail, move to the next listener and start over

The process exists with status = unreachable when ALL the listeners/adapters are exhausted.


ant-node/src/networking/driver/behaviour/upnp/mod.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright 2023 Protocol Labs.
```
maqi: 2026 now 

best check across other newly added file as well

also the copyright statement is different to the existing ones ?

ant-node/src/networking/reachability_check/mod.rs
```diff
@@ -0,0 +1,795 @@
+// Copyright 2025 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+#![allow(dead_code)]
```
maqi, line 10: best put a hightlight here, say a `TODO` ?, to remind when this workflow got fully integrated

ant-node/src/networking/reachability_check/mod.rs
```diff
@@ -0,0 +1,795 @@
...
+        }
+        if local_adapter_addrs.len() != 1 {
+            error!("Multiple local adapter addresses observed. Terminating the node.");
+            return Err(ReachabilityIssue::MultipleLocalAdapterAddresses);
+        }
+
+        let external_addr = *external_addrs
+            .iter()
+            .next()
+            .expect("This should not be empty, we checked it above");
```
maqi: shall not use `expect` in production code.
same to the other places.

ant-node/src/networking/network/init.rs
```diff
@@ -393,6 +400,32 @@ fn init_swarm_driver(
...
+                error!("Failed to listen on QUIC address {addr:?}: {err}");
+
+                if start_time.elapsed() > LISTEN_TIMEOUT_S {
+                    error!(
+                        "Failed to listen on QUIC address {addr:?} after {} seconds",
+                        LISTEN_TIMEOUT_S.as_secs()
+                    );
+                    return Err(NetworkError::ListenFailed(addr));
+                }
+                std::thread::sleep(Duration::from_secs(1));
```
maqi, line 417: Using std::thread::sleep in async code blocks the runtime. 
better use tokio::time::sleep instead ?

ant-node/src/networking/reachability_check/mod.rs
```diff
@@ -0,0 +1,795 @@
...
+    pub(crate) async fn new(
+        swarm: Swarm<ReachabilityCheckBehaviour>,
+        keypair: &Keypair,
+        listen_addr: SocketAddr,
+        bootstrap_config: BootstrapConfig,
+        no_upnp: bool,
+        #[cfg(feature = "open-metrics")] metrics_recorder: Option<NetworkMetricsRecorder>,
+    ) -> Result<Self, NetworkError> {
+        let swarm = swarm;
+        println!("Obtaining valid listeners for the reachability check workflow");
```
maqi, line 152: This should be a proper log statements.
in case stdout is really need, at least the same shall be logged as well.

other places need to be checked as well.

ant-bootstrap/src/lib.rs
```diff
@@ -62,49 +62,30 @@ pub fn craft_valid_multiaddr(addr: &Multiaddr, ignore_peer_id: bool) -> Option<M
...
     let udp = addr
         .iter()
-        .find(|protocol| matches!(protocol, Protocol::Udp(_)));
-    let tcp = addr
+        .find(|protocol| matches!(protocol, Protocol::Udp(_)))?;
+
+    output_address.push(udp);
+    let quic = addr
         .iter()
-        .find(|protocol| matches!(protocol, Protocol::Tcp(_)));
```
maqi: oh, I thought this part of code shall got removed already ?
why the removal happens in this PR ? which shall only about introduce the new workflow ?

ant-node/src/networking/driver/behaviour/upnp/mod.rs
RolandSherwin: The upnp code is a modified version of the libp2p code (they've not released the breaking change yet, but has been merged), so we must retain the original MIT license header (Protocol Labs) here?

ant-node/src/networking/network/init.rs
RolandSherwin: I think that is fine here and kinda intentional as the process of binding listener is sync and no other libp2p/network task would be running while we are binding..

ant-node/src/networking/reachability_check/mod.rs
RolandSherwin: Added extra log statements

ant-bootstrap/src/lib.rs
RolandSherwin: That might be from the previous commits from PR 3268, which are now rebased.

maqi (approved):

PR #3273, opened 2025-10-27
feat(ant): checking close group record holding status

maqi:
### Description

This PR expands the existing analyze tool, allow user to check the holding status of particular record.

Example to use:

```
ant analyze -v --closest-nodes f4003b5d8111a5f4fa6725e36b9a7886d2600a294a1bd3d9d572ddc23cd7b22d
``` 

The expected closest nodes' distance to the taget will be printed out.
Together with a 2-D table of distances among those closest nodes.

### Type of Change

### DIFF
--- a/ant-cli/src/commands/analyze.rs
+++ b/ant-cli/src/commands/analyze.rs
@@ -7,7 +7,8 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::actions::NetworkContext;
-use autonomi::{Multiaddr, RewardsAddress, SecretKey, Wallet, client::analyze::AnalysisError};
+use autonomi::{Multiaddr, RewardsAddress, SecretKey, Wallet, client::analyze::AnalysisError,
+    networking::NetworkAddress};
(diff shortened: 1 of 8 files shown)


RolandSherwin (approved):
LGTM!

grumbach (changes requested):
I think https://github.com/maidsafe/autonomi/pull/3277 is a more natural way (as it uses the same API as our code) to determine the holders for a record. 
Also it comes without affecting the network code by using existing infra. I suggest we replace 3273 by 3277. Would you agree? If so we can close #3273

maqi:
> I think #3277 is a more natural way (as it uses the same API as our code) to determine the holders for a record. Also it comes without affecting the network code by using existing infra. I suggest we replace 3273 by 3277. Would you agree? If so we can close #3273

As explained in the review comment of PR 3277, https://github.com/maidsafe/autonomi/pull/3277#pullrequestreview-3387652478
This PR 3277 is aiming the different thing, and being vital for the future work

PR #3280, opened 2025-10-29
refactor replication candidates range check

maqi:
### Description

refactor replication candidates range check, to make record copies stay within the closest group


maqi:
preferred for an intermedia branch to ease the testnet runs.
replaced by PR https://github.com/maidsafe/autonomi/pull/3282

PR #3286, opened 2025-11-04
feat(antnode): simulate replication and check if records are being st…

RolandSherwin:
- A simulation to test if our replication is working as intended and check if the records are being stored at the right location after uploading.


ant-node/tests/simulate_replication.rs
```diff
@@ -0,0 +1,1183 @@
...
+    println!("Phase 2: Uploading {num_chunks} chunks with {payment_mode:?} payment mode...");
+
+    let mut chunk_addresses = Vec::new();
+    let mut upload_stats = UploadStats::new();
+
+    let mut rng = rand::thread_rng();
+    for i in 0..num_chunks {
+        // Generate random chunk
+        let chunk_addr = NetworkAddress::ChunkAddress(ChunkAddress::new(XorName::random(&mut rng)));
+        let chunk_size = 1024 * 1024; // 1MB
```
maqi: to make the simulation quicker, the `content` can be very trival.
what matters is the `address` only

RolandSherwin: The chunk size is only used inside quote generation. Maybe I'll rearrange things.

ant-node/tests/simulate_replication.rs
```diff
@@ -0,0 +1,1220 @@
...
+        record_addr: &NetworkAddress,
+        payment: &SimulatedPayment,
+    ) -> Result<(), StoreResult> {
+        // Check 1: Verify we (this node) are in the payee list
+        if !payment.payees.contains(&self.peer_id) {
+            return Err(StoreResult::PaymentNotForUs);
+        }
+
+        // Check 2: Get K_VALUE (20) closest peers from our routing table to the data address
+        let closest_k = self.find_closest_local(record_addr, LIBP2P_K_VALUE);
```
maqi: this `find_closest_local` returns K peers closest to the target but exculding self
which mens the later on `out_of_k_payees` will still containing self.

this will always trigger the later on `responsible_distance` check, which being more strict and different to the production behaviour

maqi (changes requested):
Put in `Request changes` as showing different result to the testnet runs

ant-node/tests/simulate_replication.rs
```diff
@@ -0,0 +1,1373 @@
...
+            closest_5
+                .iter()
+                .map(|(peer, _)| (*peer, nodes_guard[peer].generate_quote(chunk_size)))
+                .collect()
+        };
+        quotes.sort_by_key(|(_, price)| *price);
+
+        // Select payment targets based on mode
+        let paid_nodes: Vec<PeerId> = match payment_mode {
+            PaymentMode::SingleNode => {
```
maqi: all candidates that got picked in the payment, even they got 0 paid, shall still accept the upload.
hence here, all 5 shall be included in the payment. otherwise will result in rejection of `NotInPaymentList`

maqi (changes requested):
best make the test to be executed as a separate CI run, so that the results can then be seen more explicitly.

maqi:
Simulation somehow outdated and no longer necessary.

PR #3287, opened 2025-11-04
feat: add windows binary signing support with DigiCert KeyLocker

ermineJose:
Enable windows-signing again.


jacderida:
This was implemented on an RC branch using a slightly different mechanism.

PR #3288, opened 2025-11-05
fix(node): out-of-sync record_store indexing cache shall be pruned

maqi:
### Description

out-of-sync record_store indexing cache shall be pruned


jacderida (approved):

RolandSherwin (approved):
Clean one! LGTM

PR #3295, opened 2025-11-07
Node integration

grumbach:
- integrates merkle payment in node
- mocks smart contract with a `DiskMerklePaymentContract`

### DIFF
--- a/ant-evm/src/merkle_payments/disk_contract.rs
+++ b/ant-evm/src/merkle_payments/disk_contract.rs
@@ -59,8 +59,8 @@ pub struct OnChainPaymentInfo {
     /// This is the timestamp that all nodes in the pool used for their quotes
     pub merkle_payment_timestamp: u64,
 
-    /// Addresses of the 'depth' nodes that were paid
-    pub paid_node_addresses: Vec<RewardsAddress>,
+    /// Addresses of the 'depth' nodes that were paid along with their indices in the winner pool
+    pub paid_node_addresses: Vec<(RewardsAddress, usize)>,
 }
 
 impl DiskMerklePaymentContract {
@@ -150,22 +150,25 @@ impl DiskMerklePaymentContract {
             winner_node_indices.len()
         );
 
-        // Extract paid node addresses
-        let mut paid_addresses = Vec::new();
+        // Extract paid node addresses, along with their indices
+        let mut paid_node_addresses = Vec::new();
         for (i, &node_idx) in winner_node_indices.iter().enumerate() {
             let addr = winner_pool.candidate_addresses[node_idx];
-            paid_addresses.push(addr);
+            paid_node_addresses.push((addr, node_idx));
             println!("  Node {}: {}", i + 1, addr);
         }
 
-        println!("\nSimulating payment to {} nodes...", paid_addresses.len());
+        println!(
+            "\nSimulating payment to {} nodes...",
+            paid_node_addresses.len()
+        );
         println!("=========================\n");
 
         // Store payment info on 'blockchain' (indexed by winner_pool_hash)
         let info = OnChainPaymentInfo {
             depth,
             merkle_payment_timestamp,
-            paid_node_addresses: paid_addresses,
+            paid_node_addresses,
         };
 
         let file_path = self
--- a/ant-evm/src/merkle_payments/disk_contract.rs
+++ b/ant-evm/src/merkle_payments/disk_contract.rs
@@ -64,9 +64,21 @@ pub struct OnChainPaymentInfo {
 }
 
 impl DiskMerklePaymentContract {
-    pub fn new(storage_path: PathBuf) -> Result<Self, SmartContractError> {
+    pub fn new_with_path(storage_path: PathBuf) -> Result<Self, SmartContractError> {
         std::fs::create_dir_all(&storage_path)?;
         Ok(Self { storage_path })
+    }
+
+    /// Create a new contract with the default storage path
+    /// Uses: DATA_DIR/autonomi/merkle_payments/
+    pub fn new() -> Result<Self, SmartContractError> {
+        let storage_path = if let Some(data_dir) = dirs_next::data_dir() {
+            data_dir.join("autonomi").join("merkle_payments")
+        } else {
+            // Fallback to current directory if data_dir is not available
+            PathBuf::from(".autonomi").join("merkle_payments")
+        };
+        Self::new_with_path(storage_path)
     }
 
     /// Submit batch payment (simulates smart contract logic)
--- a/ant-node/src/networking/record_store.rs
+++ b/ant-node/src/networking/record_store.rs
@@ -887,6 +887,11 @@ impl RecordStore for NodeRecordStore {
                     RecordKind::DataWithPayment(_) => {
                         debug!("Record {record_key:?} with payment shall always be processed.");
                     }
+                    RecordKind::DataWithMerklePayment(_) => {
+                        debug!(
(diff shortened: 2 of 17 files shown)


ant-evm/src/merkle_payments/disk_contract.rs
```diff
@@ -59,16 +59,28 @@ pub struct OnChainPaymentInfo {
...
     }
 
+    /// Create a new contract with the default storage path
+    /// Uses: DATA_DIR/autonomi/merkle_payments/
+    pub fn new() -> Result<Self, SmartContractError> {
+        let storage_path = if let Some(data_dir) = dirs_next::data_dir() {
+            data_dir.join("autonomi").join("merkle_payments")
+        } else {
+            // Fallback to current directory if data_dir is not available
+            PathBuf::from(".autonomi").join("merkle_payments")
```
maqi, line 79: just to confirm, is it shall be `./autonomi` ?
or this could be OS dependent ?

ant-node/src/networking/record_store.rs
```diff
@@ -887,6 +887,11 @@ impl RecordStore for NodeRecordStore {
                     RecordKind::DataWithPayment(_) => {
                         debug!("Record {record_key:?} with payment shall always be processed.");
                     }
+                    RecordKind::DataWithMerklePayment(_) => {
```
maqi, line 890: just to confirm, so this means we are going to support three types of payments:
* MerklePayment
* normal payment of 3
* normal payment of 1

right ?

even so, on Record level, there can be just one enum of DataWithPayment, then provide same payload to smart contract, and let smart contract using which detailed verification logic to handle ?

ant-node/src/node.rs
```diff
@@ -745,6 +746,22 @@ impl Node {
                     record_addr: address,
                 }
             }
+            Query::GetMerkleCandidateQuote {
```
maqi, line 749: on client <-> node communication level, this quote shall be the same as previous.
is that because of the new `timestamp` ? if so, it shall be a new member field in the privous Quote query ?

ant-node/src/node.rs
```diff
@@ -814,6 +831,97 @@ impl Node {
...
+        // Validate timestamp before signing to prevent committing to invalid times.
+        // Nodes will reject proofs with expired/future timestamps during payment verification,
+        // so signing such timestamps would create useless quotes that can't be used.
+        let now = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_secs();
+
+        // Reject future timestamps
+        if merkle_payment_timestamp > now {
```
maqi, line 858: is it safe to carry out this check across OS and machines ?

ant-node/src/put_validation.rs
```diff
@@ -1015,4 +1016,248 @@ impl Node {
...
+                // Store chunk
+                let result = self.store_chunk(&chunk, true);
+                if result.is_ok() {
+                    Marker::ValidPaidChunkPutFromClient(&pretty_key).log();
+                    self.network()
+                        .notify_fetch_completed(record_key, ValidationType::Chunk);
+                }
+                result
+            }
+            DataTypes::Scratchpad => {
```
maqi, line 1064: oh, I thought only the Chunk will use the MerklePayment scheme, and the other mutable datatypes still using the old one.
is it really necessary to make ALL datatypes using the MerklePayment ?

I think unless there is only MerklePayment leftover, otherwise it will be better to make mutable datatypes still useing the old payment scheme ?

ant-node/src/put_validation.rs
```diff
@@ -1015,4 +1016,248 @@ impl Node {
...
+        use ant_evm::merkle_payments::DiskMerklePaymentContract;
+        use std::collections::HashSet;
+
+        let record_key = target_address.to_record_key();
+        let pretty_key = PrettyPrintRecordKey::from(&record_key);
+
+        // Step 1: Query smart contract FIRST to get payment info
+        let winner_pool_hash = proof.winner_pool_hash();
+
+        let contract = DiskMerklePaymentContract::new().map_err(|e| {
```
maqi, line 1153: this code block shall be marked as TODO, as in production, an online smart_contract shall be used instead, i.e. via the facility from evm crate ?

to make the migration less painfull, maybe the disk_contract shall be inside the evm and make the disk/online contract switch as a configure setting ?

ant-node/src/put_validation.rs
```diff
@@ -1015,4 +1016,248 @@ impl Node {
...
+
+        debug!(
+            "Got payment info from smart contract: depth={}, timestamp={}, paid_nodes={}",
+            payment_info.depth,
+            payment_info.merkle_payment_timestamp,
+            payment_info.paid_node_addresses.len()
+        );
+
+        // Step 2: Verify Merkle proof structure using smart contract data
+        proof
```
maqi, line 1182: this shall be inside the contract facility funtion.
i.e. the query to the contract shall be: provide the proof payload that client provided, then the contract returns whether the proof is valid or not

maqi (commented):
The PR description shall give some detailed description of what has been integrated in this PR

ant-evm/src/merkle_payments/disk_contract.rs
grumbach: This is now just used for tests now (as we have a real smart contract), maybe /tmp would be better in fact.

ant-node/src/networking/record_store.rs
grumbach: Yes as you say. Although we could avoid a breaking change here by re-using DataWithPayment, since the smart contract logic and verification methods will be different, it's cleaner for it to be a different flow. 
Also nodes will have multiple other functionalities needed to support merkle so there will be other breaking changes too.

ant-node/src/node.rs
grumbach: It's actually completely different from the current quoting system. There's virtually nothing in common except for the QuotingMetrics, and we're even considering making them smaller for merkle to reduce smart contract calldata.

ant-node/src/node.rs
grumbach: This is an excellent comment, and as you suggested IT IS NOT (timezones, clock drift...)! I've fixed this in a later commit here: https://github.com/maidsafe/autonomi/pull/3310/commits/f71fcb603839f6bdbe4a11d9a082c8e6b2874014

ant-node/src/put_validation.rs
grumbach: So the way it is: 
- merkle payment supports all data types to allow for future support of batching for other types
- CURRENTLY is only used for files API as a start (so only chunks batches in short as this is where we need to reduce gas costs massively)
- as you suggest, the old payment method is still available and used for all types

ant-node/src/put_validation.rs
grumbach: Absolutely, the `DiskMerklePaymentContract` is just a placeholder/mock for the actual smart contract which is being introduced here: https://github.com/maidsafe/autonomi/pull/3310

ant-node/src/put_validation.rs
grumbach: To keep the smart contract's data (gas) and knowledge (privacy) small, it only keeps around hashes. So it cannot verify the proof. On the other hand, the nodes can compare and verify the proof using the data the smart contract stored.

ant-node/src/node.rs
maqi: I have doubt on that.
but we will see when the feature got fully done and to be merged.

PR #3297, opened 2025-11-10
feat(client): analyze storage health

maqi:
### Description

analyze storage health


maqi:
covered by PR https://github.com/maidsafe/autonomi/pull/3299

PR #3300, opened 2025-11-11
feat(antnode): perform network wide replication of all our keys

RolandSherwin:
For each `key` that a node holds, it tries to query the network wide closest peers for that key and check if they hold the record, else we send a `Cmd::Replicate` to them.

We perform this check for all our keys over a span of 7 days. This number is chosen such that, assuming >7 peers hold a particular record (due to the interval replication), then each record would be replicated across the network using this new method once every day on avg.

### DIFF
--- a/ant-node/src/networking/driver/mod.rs
+++ b/ant-node/src/networking/driver/mod.rs
@@ -10,11 +10,13 @@ pub(crate) mod behaviour;
 pub(crate) mod cmd;
 pub(crate) mod event;
 pub(crate) mod network_discovery;
+pub(crate) mod network_wide_replication;
 
 use ant_bootstrap::BootstrapCacheStore;
 use event::NodeEvent;
 use network_discovery::{NETWORK_DISCOVER_INTERVAL, NetworkDiscovery};
 
+use crate::networking::driver::network_wide_replication::NetworkWideReplication;
 #[cfg(feature = "open-metrics")]
 use crate::networking::metrics::NetworkMetricsRecorder;
 use crate::networking::{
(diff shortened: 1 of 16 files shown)


maqi (changes requested):
Issue 1: network_under_load Logic is Broken
Claim: Flag should be set when "records we hold are not in actual location"
Reality: Flag is ALWAYS set to true whenever ANY replication happens
Impact: Network constantly thinks it's under load, defeating the purpose

Issue 2: completed_records Never Populated
Keys are sent but never tracked as completed
Records can be replicated multiple times within the same 7-day cycle
Wasted bandwidth and ineffective filtering

Issue 3: 72-Hour Timer Never Resets (HIGH)
Timer resets every time keys are sent (every ~15 min)
Flag never becomes false during active replication
Need separate timestamps for "load detected" vs "last sent"

ant-node/src/networking/driver/mod.rs
```diff
@@ -175,6 +180,7 @@ impl SwarmDriver {
         let mut relay_manager_reservation_interval = interval(RELAY_MANAGER_RESERVATION_INTERVAL);
         let mut bootstrap_interval = Some(interval(BOOTSTRAP_CHECK_INTERVAL));
         let mut dial_queue_check_interval = interval(DIAL_QUEUE_CHECK_INTERVAL);
+        let mut network_wide_replication_interval = interval(NETWORK_WIDE_REPLICATION_INTERVAL);
```
maqi: better scrammable the interval to avoid nodes' synced actions causing spike on resource usage

ant-node/src/replication.rs
```diff
@@ -319,6 +319,13 @@ impl Node {
             }
         }
 
+        if !to_replicate.is_empty() {
```
maqi: will this be somehow to restrict ? 
it's normal that having 1 or 2 nodes not having a record, among its closest_7
maybe lift up the bar to having less than majority (i.e. <=3) holders among the closest_7

maqi (commented):
btw. I also have two general comments:
1. as the source and destination zone may have different view on `unhealthy`, the chance of `rescure` could be low ?
2. how to tracking the related metrics, then be able to test this PR's effect in testnet ?

maqi (approved):

PR #3302, opened 2025-11-12
Cli integration

grumbach:
✅ review/merge AFTER #3295

- integrates merkle payments in the autonomi lib
- integrates merkle payments in the CLI

### DIFF
--- a/ant-cli/src/commands/file.rs
+++ b/ant-cli/src/commands/file.rs
@@ -339,7 +339,7 @@ async fn upload_dir_merkle(
     };
 
     // Upload files with Merkle payment
-    let results = client
+    let (amount_paid, results) = client
         .files_put_with_merkle_payment(dir_path.clone(), public, payment_option)
         .await
         .map_err(|e| {
@@ -353,6 +353,10 @@ async fn upload_dir_merkle(
                 "Merkle payment error: {e}"
             )))
         })?;
+
+    // Display the Merkle payment cost
+    println!("Total cost: {amount_paid} AttoTokens");
+    info!("Merkle payment cost: {amount_paid}");
 
     // Create payment option for archive uploads
     let payment_option = PaymentOption::Wallet(wallet.clone());
--- a/ant-cli/src/commands/file.rs
+++ b/ant-cli/src/commands/file.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::access::cached_payments;
+use crate::access::{cached_merkle_payments, cached_payments};
 use crate::actions::NetworkContext;
 use crate::args::max_fee_per_gas::{MaxFeePerGasParam, get_max_fee_per_gas_from_opt_param};
 use crate::exit_code::{ExitCodeError, FEES_ERROR, IO_ERROR, upload_exit_code};
@@ -14,6 +14,7 @@ use crate::utils::collect_upload_summary;
 use crate::wallet::load_wallet;
 use autonomi::client::PutError;
 use autonomi::client::analyze::Analysis;
+use autonomi::client::merkle_payments::{MerkleFilePutError, MerklePaymentOption};
 use autonomi::client::payment::PaymentOption;
 use autonomi::files::UploadError;
 use autonomi::networking::{Quorum, RetryStrategy};
@@ -27,18 +28,23 @@ const MAX_ADDRESSES_TO_PRINT: usize = 3;
 pub async fn cost(
     file: &str,
     network_context: NetworkContext,
+    merkle: bool,
     use_standard_payment: bool,
 ) -> Result<()> {
     let mut client = crate::actions::connect_to_network(network_context)
         .await
         .map_err(|(err, _)| err)?;
 
-    // Configure payment mode - default is SingleNode, only override if Standard is requested
-    if use_standard_payment {
-        client = client.with_payment_mode(PaymentMode::Standard);
-        println!("💳 Using standard payment mode (pays 3 nodes individually)");
+    if merkle {
+        println!("🌳 Estimating cost with Merkle batch payment mode");
     } else {
-        println!("🎯 Using single node payment mode (default - saves gas fees)");
+        // Configure payment mode - default is SingleNode, only override if Standard is requested
+        if use_standard_payment {
+            client = client.with_payment_mode(PaymentMode::Standard);
+            println!("💳 Using standard payment mode (pays 3 nodes individually)");
+        } else {
+            println!("🎯 Using single node payment mode (default - saves gas fees)");
+        }
     }
 
     println!("Getting upload cost...");
@@ -54,6 +60,7 @@ pub async fn cost(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
 pub async fn upload(
     file: &str,
     public: bool,
@@ -61,6 +68,7 @@ pub async fn upload(
     network_context: NetworkContext,
     max_fee_per_gas_param: Option<MaxFeePerGasParam>,
     retry_failed: u64,
+    merkle: bool,
     use_standard_payment: bool,
 ) -> Result<(), ExitCodeError> {
     let config = ClientOperatingStrategy::new();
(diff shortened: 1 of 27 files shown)


ant-cli/src/commands/file.rs
```diff
@@ -27,18 +28,23 @@ const MAX_ADDRESSES_TO_PRINT: usize = 3;
...
-        client = client.with_payment_mode(PaymentMode::Standard);
-        println!("💳 Using standard payment mode (pays 3 nodes individually)");
+    if merkle {
+        println!("🌳 Estimating cost with Merkle batch payment mode");
     } else {
-        println!("🎯 Using single node payment mode (default - saves gas fees)");
+        // Configure payment mode - default is SingleNode, only override if Standard is requested
+        if use_standard_payment {
+            client = client.with_payment_mode(PaymentMode::Standard);
+            println!("💳 Using standard payment mode (pays 3 nodes individually)");
```
maqi, line 44: message shall be consistent.
`Estimating cost with ...`

ant-cli/src/commands/file.rs
```diff
@@ -278,6 +307,116 @@ async fn upload_dir(
...
+        })?;
+
+    // Display the Merkle payment cost
+    println!("Total cost: {amount_paid} AttoTokens");
+    info!("Merkle payment cost: {amount_paid}");
+
+    // Create payment option for archive uploads
+    let payment_option = PaymentOption::Wallet(wallet.clone());
+
+    if public {
```
maqi, line 364: this looks like a certain duplicated code block, that got used in other places as well ?

ant-cli/src/commands/file.rs
```diff
@@ -278,6 +307,116 @@ async fn upload_dir(
     }
 }
 
+/// Uploads a file or directory to the network using Merkle batch payment.
+/// Single files are uploaded without an archive, directories are uploaded with an archive.
+/// The no_archive argument can be used to skip the archive upload.
+/// Returns the archive address if any and the address to access the data.
```
maqi, line 313: seems in case of no-archive or multiple, a place holder string will be returned ?
```
Ok(("no-archive".to_string(), "multiple addresses".to_string()))
```
Which, better make the comment more accurate ?

ant-evm/src/merkle_payments/merkle_payment.rs
```diff
@@ -210,7 +208,7 @@ impl MerklePaymentCandidatePool {
             bytes.extend_from_slice(&node.to_bytes());
         }
 
-        XorName::from_content(&bytes)
+        XorName::from_content(&bytes).0
```
maqi, line 211: shall use proper PoolHash, instead of the XorName as a shortcut ?

ant-node/src/put_validation.rs
```diff
@@ -1147,7 +1148,38 @@ impl Node {
...
+                );
+                error!("{error_msg}");
+                return Err(PutValidationError::MerklePaymentVerificationFailed {
+                    record_key: pretty_key.clone().into_owned(),
+                    error: error_msg,
+                });
+            }
+        };
+
+        if proof.address != target_xorname {
```
maqi, line 1169: compare `proof.address.0 != target_address.as_bytes()` will be more convenient ?

autonomi/src/client/config.rs
```diff
@@ -21,7 +21,7 @@ pub(crate) static CHUNK_UPLOAD_BATCH_SIZE: LazyLock<usize> = LazyLock::new(|| {
     let batch_size = std::env::var("CHUNK_UPLOAD_BATCH_SIZE")
         .ok()
         .and_then(|s| s.parse().ok())
-        .unwrap_or(1);
+        .unwrap_or(16);
```
maqi, line 24: changing this will affect performance measurement/comparsion against the current.
better leave it as it is, and change it as a separate PR

autonomi/src/client/merkle_payments/file.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[error("Upload error: {err}")]
+    Upload {
+        err: MerklePutError,
+        receipt: MerklePaymentReceipt,
+    },
+}
+
+impl Client {
+    /// Pay for a directory of files with Merkle batch payments
+    pub async fn file_pay_with_merkle_payment(
```
maqi, line 47: I am not sure, but having entire individual file/directory handling just because of switching payment scheme indicates some infrastructure design issue.
and will cause maintainance hassle, which one problem has to be duplicated fixed in various working flows.

autonomi/src/client/merkle_payments/file.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// # Arguments
+    /// * `path` - The path to the directory to upload
+    /// * `is_public` - Whether the files are uploaded as public
+    /// * `payment` - The payment option (wallet or cached receipt)
+    ///
+    /// # Returns
+    /// * Tuple of (amount_paid, results) where:
+    ///   - `amount_paid` - Total amount paid for the Merkle batch (in AttoTokens)
+    ///   - `results` - Vector of (relative_path, datamap, metadata) tuples for each uploaded file
+    pub async fn files_put_with_merkle_payment(
```
maqi, line 125: `files_put_with_merkle_payment` and `file_put_with_merkle_payment` shall be able to be merged ?
and,
what's the diff between `directory_upload` with `files_put` ?

autonomi/src/client/merkle_payments/payments.rs
```diff
@@ -0,0 +1,283 @@
...
+                        peer,
+                        data_type_index,
+                        data_size,
+                        merkle_payment_timestamp,
+                    )
+                    .await
+            });
+        }
+
+        // Collect all candidates - error out immediately if ANY request fails
```
maqi, line 124: failure on signle error out could be risky when working against the mainnet

autonomi/src/client/merkle_payments/payments.rs
```diff
@@ -0,0 +1,283 @@
...
+        );
+
+        // Phase 4: Submit payment to smart contract
+        let pool_commitments: Vec<PoolCommitment> = candidate_pools
+            .iter()
+            .map(|pool| pool.to_commitment())
+            .collect();
+
+        // Make sure nobody else can use the wallet while we are paying
+        debug!("Waiting for wallet lock");
```
maqi, line 234: this shall be done by the evm crate, instead here ?

autonomi/src/networking/driver/task_handler.rs
```diff
@@ -489,6 +525,14 @@ impl TaskHandler {
             responder
                 .send(Ok(None))
                 .map_err(|_| TaskHandlerError::NetworkClientDropped(format!("{id:?}")))?;
+        // Get Merkle candidate quote case
+        } else if let Some(responder) = self.get_merkle_candidate_quote.remove(&id) {
```
maqi, line 529: this now becomes a longer and longer if-else clause now.
ideally, a general req/resp dispatcher shall be used instead

ant-cli/src/commands/file.rs
grumbach: Make sense!

ant-cli/src/commands/file.rs
grumbach: It is indeed very similar to the normal upload flow, only difference being the result format deep within the loop. I wanted to keep it simple for now but there is indeed room for improvement!

ant-evm/src/merkle_payments/merkle_payment.rs
grumbach: I used XorName as a hashing algo for `PoolHash` although we could indeed use something else. XorName is also used for the tree generation and is fail proof which is convenient, but I'm open to using another algo as long as it is deterministic. Any preferences?

autonomi/src/client/merkle_payments/file.rs
grumbach: The flow is very complex as it is and will need a SERIOUS refactor as you suggest, which is the reason why I created a new function from scratch to avoid adding to the existing complexity so we can clean this up more easily.

autonomi/src/client/merkle_payments/file.rs
grumbach: 100% agreed! This was fixed in a later commit!

autonomi/src/client/merkle_payments/payments.rs
grumbach: 100% agree, this was fixed here: https://github.com/maidsafe/autonomi/pull/3310/commits/c123569de41eb6c2c161a91cf3bc4e72b8ae0f1d

autonomi/src/client/merkle_payments/payments.rs
grumbach: The wallet lock is always done in the autonomi layer which is why I implemented it here, are you suggesting that we move all the wallet locks to the evm crate?

autonomi/src/networking/driver/task_handler.rs
grumbach: Agreed, I will do some refactoring in here. Although if we use a general req/resp dispatcher, we need to keep it an implementation detail that doesn't surface to the user as it will hurt the API's usability.

ant-evm/src/merkle_payments/merkle_payment.rs
maqi: we shall avoid further abuse the usage of XorName, which has been abused too much already.
better started from this kind of direct hashing usage

autonomi/src/client/merkle_payments/payments.rs
maqi: > are you suggesting that we move all the wallet locks to the evm crate?

Yes, unless there is other concerns prevents doing so ?

PR #3309, opened 2025-11-17
feat(client): for Chunk, fall back to get from closest_20 if kad query fetch failed

maqi:
### Description

When client downloading a Chunk, fall back to get from closest_20 if kad query fetch failed

### DIFF
--- a/autonomi/src/client/config.rs
+++ b/autonomi/src/client/config.rs
@@ -8,7 +8,7 @@
 
 use crate::networking::{Quorum, RetryStrategy, Strategy};
 pub use ant_bootstrap::{
-    Bootstrap, BootstrapConfig, InitialPeersConfig, error::Error as BootstrapError,
+    error::Error as BootstrapError, Bootstrap, BootstrapConfig, InitialPeersConfig,
 };
 use ant_evm::EvmNetwork;
 use evmlib::contract::payment_vault::MAX_TRANSFERS_PER_TRANSACTION;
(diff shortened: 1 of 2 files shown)


autonomi/src/client/data_types/chunk.rs
```diff
@@ -138,16 +138,77 @@ impl Client {
...
+        // Find the first successful record
+        for result in results.into_iter() {
+            match result {
+                Ok(Some(record)) => {
+                    debug!("✅ Retrieved record {key:?} from one of the closest {num_peers} peers");
+                    return Some(record);
+                }
+                _ => continue,
+            }
+        }
```
RolandSherwin, line 185: Do we want to optimize this further by stopping the rest of the tasks once a record has been obtained? `process_tasks_with_max_concurrency` will consume alot more bandwidth if multiple records are found?

We could use `stream::iter` directly instead of calling `process_tasks_with_max_concurrency`

```
  let first_record = stream::iter(query_tasks)
      .buffer_unordered(num_peers)
      .filter_map(|result| async move {
          match result {
              Ok(Some(record)) => Some(record),
              _ => None,
          }
      })
      .next()  // Takes first item then stops the stream
      .await;
```

autonomi/src/client/config.rs
```diff
@@ -160,7 +160,7 @@ impl Default for ClientOperatingStrategy {
                 put_retry: RetryStrategy::Balanced,
                 verification_quorum: Quorum::N(two),
                 get_quorum: Quorum::One, // chunks are content addressed so one is enough as there is no fork possible
-                get_retry: RetryStrategy::Quick,
+                get_retry: RetryStrategy::None, // leave to the fall-back approach of fetching from closest_20
```
RolandSherwin, line 163: Could we use `RetryStrategy::Once` to try alteast once again before falling back to the new method?

maqi: it's un-necessary and will just slow down the performance

autonomi/src/client/data_types/chunk.rs
maqi: the clippy has huge outcry on the suggested change, and seems the `early exist` can not be achieved by that way.
meanwhile, our other way to achieve early exist proved causing significant performance slowdown, hence we switched to the `steam unordered` pattern.

Also, this is a fall-back approach, which doesn't expect to be used often, hence a slight unideal performance shall be acceptable.

jacderida (approved):

PR #3310, opened 2025-11-18
Smart contract integration

grumbach:
- ✅ review/merge AFTER #3302
- implements an actual smart contract instead of the disk mock
- integration in evmlib and ant-evm
- integration in node and client of the new smart contract use
- integrates new EVM vars for this new smart contract
- makes those EVM vars available in testnets through antctl
- significantly improves CLI side printing of upload progress for better understandability
- fixes MANY BUGS


ant-evm/src/merkle_payments/merkle_payment.rs
```diff
@@ -187,9 +185,9 @@ pub struct MerklePaymentCandidatePool {
     /// The midpoint proof from ant-evm's merkle_tree module
     pub midpoint_proof: MidpointProof,
 
-    /// Candidate nodes for this pool (should be 20 nodes)
+    /// Candidate nodes for this pool (exactly 20 nodes)
```
maqi: thought it has been changed to 16 ?
anyway, better use the const of CANDIDATES_PER_POOL here ?

ant-node/src/put_validation.rs
```diff
@@ -1142,35 +1142,19 @@ impl Node {
...
-            }
-        };
-
-        if proof.address != target_xorname {
+        if proof.address.0.to_vec() != target_address.as_bytes() {
             let error_msg = format!(
                 "Merkle proof address mismatch: proof is for {:?} but data is at {:?}. \
                  This prevents reusing a proof paid for one address to store data at another address.",
-                proof.address, target_xorname
+                hex::encode(proof.address.0),
```
maqi, line 1156: both addresses has a readable debug format, which doesn't need the extra `hex::encode` here ?

ant-node/src/put_validation.rs
```diff
@@ -1230,26 +1235,27 @@ impl Node {
...
-            .network()
-            .get_n_closest_peers(&reward_pool_address, CANDIDATES_PER_POOL)
-            .await
-        {
+        // Verify network topology (>50% of paid nodes in closest)
+        // Get closest peers to the midpoint address (same address the client used to collect candidates)
+        const TOPOLOGY_CHECK_NODE_COUNT: usize = CANDIDATES_PER_POOL * 2;
+        let midpoint_address = proof.winner_pool.midpoint_proof.address();
+        let reward_pool_address = NetworkAddress::ChunkAddress(ChunkAddress::new(midpoint_address));
+        let all_closest_peers = match self.network().get_closest_peers(&reward_pool_address).await {
```
maqi, line 1242: This one only gives K_VALUE (20) peers.
Actually, the kad::get_closest_peers can only give max 20 entries in the result.

I don't think it is necessary to expand to TOPOLOGY_CHECK_NODE_COUNT (32).
Just fetch 20, and ensure:
* at least 50% still alive
* for those disappeared, they shall still be within certain distance range (say 10*responsible_range, or no farther than 1st full-bucket)

autonomi/src/client/merkle_payments/payments.rs
```diff
@@ -74,61 +75,136 @@ impl Client {
...
         data_size: usize,
         merkle_payment_timestamp: u64,
-    ) -> Result<Vec<MerklePaymentCandidateNode>, MerklePaymentError> {
-        // Get 20 closest peers to target
+    ) -> Result<[MerklePaymentCandidateNode; CANDIDATES_PER_POOL], MerklePaymentError> {
+        // Request from 25% more peers than needed to provide fault tolerance
+        // This allows up to 25% of peers to fail without blocking the payment
+        // We'll select the 20 closest successful responses to ensure we get the actual
+        // closest nodes to the target, not just any 20 that respond
+        const PEERS_TO_QUERY: usize = CANDIDATES_PER_POOL + (CANDIDATES_PER_POOL / 4);
```
maqi, line 89: kad::get_closest_peers won't give a result of having more than K_VALUE peers.

will have to stay with the 20, or have to use the req/rsp DM to build up the list manually, as to what PR https://github.com/maidsafe/autonomi/pull/3319 does.

autonomi/src/client/merkle_payments/payments.rs
```diff
@@ -74,61 +75,136 @@ impl Client {
...
         }
 
-        Ok(candidates)
+        // Sort successful candidates by distance to target and take the 20 closest
+        successful_candidates.sort_by_key(|(peer_id, _candidate)| {
+            let peer_addr = NetworkAddress::from(*peer_id);
+            network_addr.distance(&peer_addr)
+        });
+
+        // Take the 20 closest successful responses
```
maqi: CANDIDATES_PER_POOL

autonomi/src/networking/mod.rs
```diff
@@ -164,7 +164,7 @@ impl Network {
     /// This will start the network driver in a background thread, which is a long-running task that runs until the [`Network`] is dropped
     /// The [`Network`] is cheaply cloneable, prefer cloning over creating new instances to avoid creating multiple network drivers
     pub fn new(bootstrap: Bootstrap) -> Result<Self, NoKnownPeers> {
-        let (task_sender, task_receiver) = mpsc::channel(100);
+        let (task_sender, task_receiver) = mpsc::channel(10);
```
maqi: why need to change?
sounds bit risky for such a low number, when need to handle large batch of chunks ?

autonomi/src/self_encryption/stream_encryption.rs
```diff
@@ -48,6 +48,30 @@ pub struct StreamProgressState {
     total_estimated_chunks: usize,
 }
 
+/// Creates a data iterator from a file path that reads the file in chunks
+fn create_file_data_iterator(file_path: &str) -> std::io::Result<impl Iterator<Item = Bytes>> {
```
maqi, line 52: this sounds like a duplicated function to have.
requiring this may indicate some duplicated logic design

autonomi/src/self_encryption/stream_encryption.rs
```diff
@@ -196,42 +220,65 @@ impl EncryptionStream {
...
-                })?;
-                Ok(())
-            });
-
-            // once we're done, send the datamap to the datamap channel
-            match result {
-                Ok(datamap) => {
-                    // Send the DataMap result
+        // Spawn a thread to handle streaming encryption
+        std::thread::spawn(move || {
```
maqi, line 224: why not use the same previous `streaming_encrypt_from_file` function ?
so that there is no need to construct data_iter manually

evmlib/src/contract/merkle_payment_vault/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+            networkDensity: metrics
+                .network_density
+                .map(|d| U256::from_be_bytes(d))
+                .unwrap_or_default(),
+            networkSize: metrics.network_size.map(U256::from).unwrap_or_default(),
+        }
+    }
+}
+
+fn data_type_conversion(data_type: u32) -> u8 {
```
maqi: this is a duplicated function that better to use the existing one,
to avoid future maintainence hassle

evmlib/src/merkle_batch_payment.rs
```diff
@@ -85,14 +99,16 @@ pub struct OnChainPaymentInfo {
     pub paid_node_addresses: Vec<(RewardsAddress, usize)>,
 }
 
-/// Disk-based Merkle payment contract (temporary implementation)
+#[cfg(any(test, feature = "test-utils"))]
+/// Disk-based Merkle payment contract (mock for testing)
 ///
 /// This simulates smart contract behavior by storing payment data to disk.
-/// Replace this entire implementation with a real smart contract when ready.
+/// Only available for testing.
```
maqi, line 106: put a TODO to highlight ?
or put under any feature guard ?

maqi (changes requested):
Be noted that kad::get_closest_peers only returns max 20 peers

Also, seems there is some duplicated functions in evm, and un-necessary data_iter that self_encryption crate already do the job.

ant-node/src/put_validation.rs
grumbach: The debug format truncates the address, making it impossible to copy/paste for debugging later, which is why I used hex

ant-node/src/put_validation.rs
grumbach: Make sense!

autonomi/src/client/merkle_payments/payments.rs
grumbach: Here since CANDIDATES_PER_POOL=16 PEERS_TO_QUERY equals 20 which is the limit! The comment is indeed misleading as we use to have CANDIDATES_PER_POOL=20. 
Fixed!

autonomi/src/networking/mod.rs
grumbach: This was part of experiments to reduce the memory impact. Turns out it didn't change anything, so reverting to 100!

autonomi/src/self_encryption/stream_encryption.rs
grumbach: Sounds very generic indeed but couldn't find one, do you have the duplicated one in mind?

autonomi/src/self_encryption/stream_encryption.rs
grumbach: `streaming_encrypt_from_file` is the old clunky streaming API which we ought to deprecate down the road, so replaced it with the new simpler `stream_encrypt`. Although I agree that we might want to offer a file to data iter API along with it...

evmlib/src/merkle_batch_payment.rs
grumbach: It's feature gated (check above):
```
#[cfg(any(test, feature = "test-utils"))]
```

ant-node/src/put_validation.rs
maqi: if they are NetworkAddress, then it shall be the full hex to be output, not the truncated one.

autonomi/src/self_encryption/stream_encryption.rs
maqi: the self_encryption crate already do this job, if stay with the existing API calls

autonomi/src/self_encryption/stream_encryption.rs
maqi: that's somehow not related to the work of this PR.
better leave it for the later on separated PR

autonomi/src/client/merkle_payments/payments.rs
maqi: The comment better to emphasize the max returned candidates will be capped at 20.

the current 25% wording indicates the candidates fetched can be at any size

ant-node/src/put_validation.rs
grumbach: Those are XorNames are you suggesting to make them NetworkAddresses? In that case what type?

maqi (approved):

PR #3315, opened 2025-11-21
feat: `antnode` automatic upgrades for unix platforms

jacderida:
- 5c227c0ab **chore: utility target for providing binary hashes**

- 54c6993dc **feat: `antnode` automatic upgrades for unix platforms**

  quicker and safer.

  below).

    sha256 hash.
    information.

  mentioned above.

  deleted.


ant-node/src/bin/antnode/main.rs
```diff
@@ -224,23 +236,65 @@ struct Opt {
...
+}
 
-    /// Enable automatic self-restart after specified duration (in seconds).
-    ///
-    /// When set, the node will automatically restart itself after the specified
-    /// number of seconds. This is useful for testing restart behavior.
-    /// For example, `--auto-restart-delay 120` will restart the node after 2 minutes.
-    #[clap(long, verbatim_doc_comment)]
-    auto_restart_delay: Option<u64>,
+async fn calculate_restart_delay(running_node: &RunningNode) -> Duration {
```
maqi, line 259: the function needs a comment to explain the algorithm and range of the period span

ant-node/src/bin/antnode/main.rs
```diff
@@ -224,23 +236,65 @@ struct Opt {
...
+    let est_network_size = running_node
+        .get_estimated_network_size()
+        .await
+        .unwrap_or(100_000);
+
+    let mut hasher = DefaultHasher::new();
+    peer_id.hash(&mut hasher);
+    let hash_value = hasher.finish();
+
+    let time_range_hours = (est_network_size / 100_000) + 1;
```
maqi: the range_hours shall have a `cap`, i.e. better to be:
```
let time_range_hours = std::cmp::min(72, (est_network_size / 100_000) + 1);
```

ant-node/src/bin/antnode/main.rs
```diff
@@ -224,23 +236,65 @@ struct Opt {
...
-    /// number of seconds. This is useful for testing restart behavior.
-    /// For example, `--auto-restart-delay 120` will restart the node after 2 minutes.
-    #[clap(long, verbatim_doc_comment)]
-    auto_restart_delay: Option<u64>,
+async fn calculate_restart_delay(running_node: &RunningNode) -> Duration {
+    let peer_id = running_node.peer_id();
+    let est_network_size = running_node
+        .get_estimated_network_size()
+        .await
+        .unwrap_or(100_000);
```
maqi: better to make this 100_000 a constant

ant-node/src/bin/antnode/main.rs
```diff
@@ -448,24 +502,38 @@ You can check your reward balance by running:
...
-            println!("Auto-restart delay elapsed, triggering node restart");
-            if let Err(err) = ctrl_tx_restart
-                .send(NodeCtrl::Restart {
-                    delay: Duration::from_secs(1),
-                    retain_peer_id: true,
-                })
-                .await
-            {
-                error!("Failed to send auto-restart command: {err}");
+            sleep(Duration::from_secs(86400)).await;
```
maqi: at least a comment explain why this number is picked ?
also, this make it impossible to be tested in the CI runs ?

ant-node/src/bin/antnode/main.rs
```diff
@@ -448,24 +502,38 @@ You can check your reward balance by running:
...
+                            error!("Failed to send restart command: {err}");
+                        }
+                        break;
+                    }
+                    Err(e) => {
+                        debug!("Upgrade check: {e}");
+                    }
+                }
+
+                sleep(Duration::from_secs(86400)).await;
```
maqi: move this to the beginning of the loop, then the first sleep can be saved ?

also, to avoid nodes running on the same machine carry out the `upgrade` check at the same time, better scramble the duration a little bit as well

ant-node/src/lib.rs
```diff
@@ -155,6 +155,26 @@ impl RunningNode {
         Ok(kbuckets)
     }
 
+    /// Returns the estimated network size based on the current kbucket state.
+    pub async fn get_estimated_network_size(&self) -> Result<usize> {
```
maqi, line 160: there is a helper function of `estimate_network_size` , better use it directly

.github/workflows/self-restart.yml
```diff
@@ -1,595 +0,0 @@
-name: Self Restart Test
```
maqi, line 1: The new `antnode-upgrades.yml` CI run only checks:
* whether the existing executable is the latest version
* whether the existing executable can be replaced by a newer version
* whether such `executable replacement` is multithread safe

Meanwhile, the `self-restart.yml` is focused on:
* whether a node process can startup itself as a new process, instead of as a child process

Hence, I don't think the `self-restart` shall be removed.

Instead, they shall be `merged` into one test:
* carry out version check and executable replacement
* restart self using the new executable, and confirm in the CI that is the new version and process_id to be used

ant-node/src/bin/antnode/main.rs
```diff
@@ -82,6 +85,15 @@ pub fn parse_log_output(val: &str) -> Result<LogOutputDestArg> {
 #[command(disable_version_flag = true)]
 #[clap(name = "antnode cli", version = env!("CARGO_PKG_VERSION"))]
 struct Opt {
+    #[command(subcommand)]
+    command: Option<SubCmd>,
```
maqi: the default value shall be `Upgrade`, unless for windows, it shall be NOT, as we can't support it yet

If this is only for the CI purpose, as explained in the other comment, this is not required

ant-node/src/bin/antnode/main.rs
```diff
@@ -224,23 +236,65 @@ struct Opt {
...
+            eprintln!("Error: upgrade command is only supported on Unix platforms (Linux/macOS)");
+            std::process::exit(1);
+        }
+
+        #[cfg(unix)]
+        {
+            println!("Checking for updates...");
+            let rt = Runtime::new()?;
+            return rt.block_on(async {
+                match upgrade::perform_upgrade().await {
```
maqi: Oh, this sounds like the run here and the subcmd is for the CI run only?
if so, I think it can/shall be avoided, just integrate the executable upgrade flow into the existing self-restart?

maqi (changes requested):
The new `antnode-upgrades.yml` CI run only checks:
* whether the existing executable is the latest version
* whether the existing executable can be replaced by a newer version
* whether such `executable replacement` is multithread safe

Meanwhile, the `self-restart.yml` is focused on:
* whether a node process can startup itself as a new process, instead of as a child process

Hence, I don't think the `self-restart` shall be removed.

Instead, they shall be `merged` into one test:
* carry out version check and executable replacement
* restart self using the new executable, and confirm in the CI that is the new version and process_id to be used

With the above, it seems the new `upgrade` opt is not necessary.

.github/workflows/self-restart.yml
jacderida: > The new `antnode-upgrades.yml` CI run only checks:
> 
>     * whether the existing executable is the latest version
> 
>     * whether the existing executable can be replaced by a newer version
> 
>     * whether such `executable replacement` is multithread safe
> 
> 
> Meanwhile, the `self-restart.yml` is focused on:
> 
>     * whether a node process can startup itself as a new process, instead of as a child process
> 
> 
> Hence, I don't think the `self-restart` shall be removed.
> 
> Instead, they shall be `merged` into one test:
> 
>     * carry out version check and executable replacement
> 
>     * restart self using the new executable, and confirm in the CI that is the new version and process_id to be used

As discussed, we will no longer provide any `upgrade` command and opt not to test this feature in CI.

ant-node/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.2.52", features = ["process-metrics"] }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
+ant-releases = { git = "https://github.com/jacderida/ant-releases", branch = "feat-latest_release_info" }
```
RolandSherwin, line 31: Just a tiny note to make sure to use the release version of `ant-release` here before merging.

ant-node/src/bin/antnode/main.rs
```diff
@@ -448,24 +480,39 @@ You can check your reward balance by running:
...
+
+                        let delay = calculate_restart_delay(&running_node_clone).await;
+                        info!("Calculated restart delay: {delay:?}");
+
+                        if let Err(err) = ctrl_tx_restart
+                            .send(NodeCtrl::Restart {
+                                delay,
+                                retain_peer_id: true,
+                            })
+                            .await
```
RolandSherwin, line 506: The `delay` value is sent via `NodeCtrl::Restart` and the sleep is carried out within the code at line 520+ inside the same file. That loop also captures `NodeCtrl::Stop` which is used to stop nodes using Ctrl+c or during any critical error. It would be nice to maybe remove the `delay` field from the `NodeCtrl::Restart` variant and instead sleep directly inside this `upgrade checking loop` before sending out the `NodeCtrl::Restart`

RolandSherwin (changes requested):

PR #3319, opened 2025-11-24
feat(client): check target's close group from targeted nodes' perspective

maqi:
### Description

* After client got close_group candidates via `kad::get_closest_peers`, carry out a check among them
* Only those candidates that responds to the check, and also being known by others shall be considered as valid candidate
* The `Multiaddr` infos that collected from the check will be used, as these infos from neighbours will be more accurate. 
* Expanded the quoting candidates range (from 7 to 10) to tackle the severe production network status


RolandSherwin (approved):
LGTM!

PR #3322, opened 2025-11-26
feat(node): carry out fetch check among closeup peers and block bad ones

maqi:
### Description

* Node carry out periodical check against closeup nodes.
* When a peer failed to respond properly, it will be removed from node's local RT and be blocked.
* In case of disk failure detected, only prune the local indexing cache

### DIFF
--- a/ant-node/src/networking/driver/cmd.rs
+++ b/ant-node/src/networking/driver/cmd.rs
@@ -439,6 +439,7 @@ impl SwarmDriver {
             LocalSwarmCmd::AddPeerToBlockList { peer_id } => {
                 cmd_string = "AddPeerToBlockList";
                 let _ = self.swarm.behaviour_mut().blocklist.block_peer(peer_id);
+                let _ = self.blocked_peers.insert(peer_id);
             }
             LocalSwarmCmd::RecordNodeIssue { peer_id, issue } => {
                 cmd_string = "RecordNodeIssues";
--- a/ant-node/src/networking/driver/cmd.rs
+++ b/ant-node/src/networking/driver/cmd.rs
@@ -438,6 +438,7 @@ impl SwarmDriver {
             }
             LocalSwarmCmd::AddPeerToBlockList { peer_id } => {
                 cmd_string = "AddPeerToBlockList";
+                self.record_metrics(Marker::PeerConsideredAsBad { bad_peer: &peer_id });
                 let _ = self.swarm.behaviour_mut().blocklist.block_peer(peer_id);
                 let _ = self.blocked_peers.insert(peer_id);
             }
--- a/ant-node/src/networking/driver/cmd.rs
+++ b/ant-node/src/networking/driver/cmd.rs
@@ -359,6 +359,11 @@ impl SwarmDriver {
                     });
                 }
             }
+            LocalSwarmCmd::RemoveOutOfSyncEntry { key } => {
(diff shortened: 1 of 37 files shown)


ant-node/src/networking/driver/cmd.rs
```diff
@@ -430,7 +435,9 @@ impl SwarmDriver {
             }
             LocalSwarmCmd::AddPeerToBlockList { peer_id } => {
                 cmd_string = "AddPeerToBlockList";
+                self.record_metrics(Marker::PeerConsideredAsBad { bad_peer: &peer_id });
                 let _ = self.swarm.behaviour_mut().blocklist.block_peer(peer_id);
+                let _ = self.blocked_peers.insert(peer_id);
```
RolandSherwin: Why do we have to have 2 block list here?
Looking at the new `blocked_peers` field, it is used to disconnect the peer after identify, but I assume the `blocklist behaviour` would do the same? Am I missing something?

ant-node/src/node.rs
```diff
@@ -1006,12 +1020,304 @@ impl Node {
...
+                return;
+            }
+        };
+
+        let mut nearby_records: Vec<NetworkAddress> = local_records
+            .into_iter()
+            .filter_map(|(address, record_type)| {
+                if record_type != ValidationType::Chunk {
+                    return None;
+                }
```
RolandSherwin, line 1223: Should we just verify Chunks here? The method fetch_record_from_peers_with_addresses just verifies if x candidates hold the record, it does not really verify the record. So we can avoid the filter here?

ant-node/src/node.rs
```diff
@@ -1006,12 +1020,304 @@ impl Node {
...
+            return;
+        }
+
+        if successful_peers.len() < MIN_HEALTHY_REPLICA_COUNT {
+            warn!(
+                "Replica verification fetched only {} copies of {pretty_key:?}. Record is unhealthy; skipping peer classification.",
+                successful_peers.len()
+            );
+            return;
+        }
```
RolandSherwin, line 1299: Shouldn't we also call `Self::schedule_record_fetch_retry` when < 3 records are found? Why do we skip if the record is unhealthy? is it because we feel that we cannot replicate this record correctly to everyone?

ant-node/src/node.rs
maqi: We can only verify Chunk types, as the other mutable datatypes may have non-consistent copy to us.

> it does not really verify the record

The Chunk type is self-verifiable. When deserialize a Chunk from Record, if content doesn't leads to the chunk_name, an error shall be raised.

ant-node/src/node.rs
maqi: >  Why do we skip if the record is unhealthy? is it because we feel that we cannot replicate this record correctly to everyone?

Yes.
That's a record ALREADY drop below the healthy status, hence a peer doesn't holds it shall not be blamed.

ant-node/src/networking/driver/cmd.rs
maqi: Good suggestion, will switch to use the blocklist behaviour directly.

RolandSherwin (approved):
LGTM!

dirvine (approved):
Looks great now @maqi Thanks for the review fixes

dirvine (approved):

PR #3323, opened 2025-11-27
Further Support for Automatic Upgrades

jacderida:
- d63bf1271 **chore: utility target for providing binary hashes**

- 7b2d6b0fc **feat: `antnode` automatic upgrades for unix platforms**

  quicker and safer.

  below).

    sha256 hash.
    information.

  mentioned above.

  deleted.

- 6674bf164 **chore!: use 'always' restart policy for `antctl add`**

  BREAKING CHANGE: new node deployments with `antctl` will apply a different restart policy.

  With `antctl` we are going to configure services for node deployments to always restart. This is so
  we can use the node deployments with the automatic upgrades feature and let the service manager
  handle the restart of the `antnode` process. If `antnode` itself were to restart the process, the
  service manager would consider it to be dead, which would break the status tracking for `antctl`.

  We need to provide a `--stop-on-upgrade` argument for `antnode` such that it will stop the process
  on the upgrade rather than restart it, and allow the service manager to perform the restart. This
  will be coming in the next commit.

  The `node-launchpad` has been configured to use 'never' for the restart policy because we ran into
  problems with the UPnP feature if the launchpad continually restarted nodes. So automatic upgrades
  just won't be supported when using the launchpad. We will need to provide another `antnode` argument
  to disable automatic upgrades, which will come in another commit.

- 73e1de96c **feat: provide `--stop-on-upgrade` argument for `antctl`**

  If this argument is used the node will be stopped after it is upgraded, rather than restarted.

  This is to allow the service manager to restart the process if the node is being run in this
  context, which it will be if it is deployed using `antctl`, which is updated in this commit to
  always supply the argument as part of the service definition for a node.

- c5fb3e828 **feat: improve process detection for antnode services**

  The upgrade process introduced a problem with the detection of processes using the path-based
  approach we had. When the binary is overwritten, it seems the text " (deleted)" gets appended to the
  path, and thus the `antctl status` command was marking services as stopped when they were still
  running and scheduled to be upgraded.

  This new approach first tries to use the stored PID, then falls back on the path-detection
  mechanism, attempting to take the "(deleted)" suffix into account.

- a07daab7c **chore: update reference to `ant-releases`**

  The new version with the release information was published to `crates.io`, so the development
  reference can now be replaced.

- e1c1ee255 **chore: sleep in upgrade loop rather than handler**

  It was pointed out in review feedback that it would be better to perform the upgrade delay in the
  upgrade loop rather than in the node control handler because it could interfere with the node
  stopping on errors or when using ctrl+c.

- 643f14642 **chore: improve logging for upgrades**

  I added the information I found useful during the development process.

### DIFF
--- a/ant-node-manager/Cargo.toml
+++ b/ant-node-manager/Cargo.toml
@@ -51,7 +51,7 @@ semver = "1.0.20"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
 sysinfo = "0.30.12"
 thiserror = "1.0.23"
 tokio = { version = "1.43", features = ["full"] }
--- a/ant-node-manager/Cargo.toml
+++ b/ant-node-manager/Cargo.toml
@@ -34,7 +34,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.2.52" }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
-ant-releases = { version = "0.4.1" }
+ant-releases = "0.4.2"
 ant-service-management = { path = "../ant-service-management", version = "0.4.16" }
 evmlib = { path = "../evmlib", version = "0.4.5" }
 chrono = "~0.4.19"
--- a/ant-node/src/bin/antnode/main.rs
+++ b/ant-node/src/bin/antnode/main.rs
@@ -468,11 +468,12 @@ You can check your reward balance by running:
                     Ok(()) => {
                         let delay = calculate_restart_delay(&running_node_clone).await;
                         info!("Calculated delay: {delay:?}");
+                        sleep(node_restart_delay).await;
 
                         let node_ctrl = if stop_on_upgrade {
                             info!("Upgrade successful. Triggering stop...");
                             NodeCtrl::Stop {
-                                delay,
+                                delay: Duration::from_secs(0),
                                 result: StopResult::Success(
                                     "Upgrade completed successfully".to_string(),
                                 ),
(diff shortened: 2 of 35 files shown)


ant-node-manager/Cargo.toml
```diff
@@ -51,7 +51,7 @@ semver = "1.0.20"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
```
maqi, line 54: a note that this shall be updated before merge

ant-node/src/bin/antnode/main.rs
```diff
@@ -433,6 +453,58 @@ You can check your reward balance by running:
...
+        let ctrl_tx_restart = ctrl_tx.clone();
+        let running_node_clone = running_node.clone();
+        tokio::spawn(async move {
+            loop {
+                // 72 hours (259200 seconds) with ±5% randomization to prevent simultaneous upgrades
+                let base_delay = 259200;
+                let variance = rand::thread_rng().gen_range(-12960..=12960);
+                let upgrade_check_delay_secs = (base_delay + variance) as u64;
+                let upgrade_check_wake_time =
+                    chrono::Utc::now() + chrono::Duration::seconds(upgrade_check_delay_secs as i64);
```
maqi, line 467: this block of calculation can be outside of the loop

ant-node/src/bin/antnode/main.rs
```diff
@@ -433,6 +453,58 @@ You can check your reward balance by running:
...
+                            }
+                        };
+
+                        if let Err(err) = ctrl_tx_restart.send(node_ctrl).await {
+                            error!("Failed to send control command: {err}");
+                        }
+                        break;
+                    }
+                    Err(e) => {
+                        error!("Error during upgrade process: {e}");
```
maqi, line 501: most of the case that there is no upgrade shall not be counted as an `error`
ideally, the return shall be `OK(boolean` to indicate whether a new version detected or not

ant-service-management/Cargo.toml
```diff
@@ -22,7 +22,7 @@ prost = { version = "0.9" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 semver = "1.0.20"
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
```
maqi, line 25: just a note that this shall be updated before merge

jacderida:
This has been closed in favour of #3333.

PR #3333, opened 2025-11-28
Support for `antnode` Automatic Upgrades

jacderida:
- 60a3fc9c2 **chore: utility target for providing binary hashes**

- c9ed698d3 **feat: `antnode` automatic upgrades for unix platforms**

  quicker and safer.

  below).

    sha256 hash.
    information.

  mentioned above.

  deleted.

- a506e71c6 **chore!: use 'always' restart policy for `antctl add`**

  BREAKING CHANGE: new node deployments with `antctl` will apply a different restart policy.

  With `antctl` we are going to configure services for node deployments to always restart. This is so
  we can use the node deployments with the automatic upgrades feature and let the service manager
  handle the restart of the `antnode` process. If `antnode` itself were to restart the process, the
  service manager would consider it to be dead, which would break the status tracking for `antctl`.

  We need to provide a `--stop-on-upgrade` argument for `antnode` such that it will stop the process
  on the upgrade rather than restart it, and allow the service manager to perform the restart. This
  will be coming in the next commit.

  The `node-launchpad` has been configured to use 'never' for the restart policy because we ran into
  problems with the UPnP feature if the launchpad continually restarted nodes. So automatic upgrades
  just won't be supported when using the launchpad. We will need to provide another `antnode` argument
  to disable automatic upgrades, which will come in another commit.

- 5e5528283 **feat: provide `--stop-on-upgrade` argument for `antctl`**

  If this argument is used the node will be stopped after it is upgraded, rather than restarted.

  This is to allow the service manager to restart the process if the node is being run in this
  context, which it will be if it is deployed using `antctl`, which is updated in this commit to
  always supply the argument as part of the service definition for a node.

- 48ec66a98 **chore: improve logging for upgrades**

  I added the information I found useful during the development process.

- 609a2a2a8 **fix: do not truncate upgrade lock file**

  By examining the logs on a testnet run I determined all the `antnode` processes on the same
  host downloaded the upgrade binary. This is something we expressly wanted to avoid.

  The problem was the `File::create` function was truncating the lock file. It has now been changed to
  avoid this behaviour, so other processes should wait to acquire the lock.

- d84051e89 **chore: output node port to log**

  This will be used to verify all nodes participating in the automatic-upgrade process will be
  restarted with the same port.
  
- 03b5e9acd **feat: provide cache for release info**

  In a similar fashion to the cache for the upgrade binary, we provide a cache for the release
  information.

  While performing a test with 5000 nodes I got rate limited by Github when thousands of processes
  queried for the release information at approximately the same time. Now it should only be one query
  per host.

- ef9e8e6af **chore: use sha256 hash in restart delay calculation**

  In the test run with 1000 nodes we observed a distribution of restart times that seemed a bit too
  tightly clustered.

  Using this alternative hashing mechanism resulted in a distribution that was more even.

  The restart delay calculation is also moved outside the loop because it only needs to be calculated
  once.

- 7b89cac04 **chore: more elaborate error handling**

  An `UpgradeError` type is introduced to make it easier to distinguish between errors and the case
  where there is no new release available. The latter of those will now log at `INFO` level.

### DIFF
--- a/ant-node-manager/Cargo.toml
+++ b/ant-node-manager/Cargo.toml
@@ -34,7 +34,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.3.0" }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
-ant-releases = { version = "0.4.2" }
+ant-releases = "0.4.3"
 ant-service-management = { path = "../ant-service-management", version = "0.4.17" }
 evmlib = { path = "../evmlib", version = "0.4.5" }
 chrono = "~0.4.19"
--- a/ant-node-manager/Cargo.toml
+++ b/ant-node-manager/Cargo.toml
@@ -34,7 +34,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.3.0" }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
-ant-releases = { version = "0.4.1" }
+ant-releases = { version = "0.4.2" }
 ant-service-management = { path = "../ant-service-management", version = "0.4.17" }
 evmlib = { path = "../evmlib", version = "0.4.5" }
 chrono = "~0.4.19"
--- a/ant-node-manager/Cargo.toml
+++ b/ant-node-manager/Cargo.toml
@@ -51,7 +51,7 @@ semver = "1.0.20"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
 sysinfo = "0.30.12"
 thiserror = "1.0.23"
 tokio = { version = "1.43", features = ["full"] }
--- a/ant-node/Cargo.toml
+++ b/ant-node/Cargo.toml
@@ -28,7 +28,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.3.0", features = ["process-metrics"] }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
-ant-releases = { version = "0.4.2" }
+ant-releases = "0.4.3"
 ant-service-management = { path = "../ant-service-management", version = "0.4.17" }
 async-trait = "0.1"
 bls = { package = "blsttc", version = "8.0.1" }
@@ -82,6 +82,7 @@ rayon = "1.8.0"
 semver = "1.0"
 sha2 = "0.10"
 serde = { version = "1.0.133", features = ["derive", "rc"] }
+serde_json = "1.0"
 strum = { version = "0.26.2", features = ["derive"] }
 sysinfo = { version = "0.30.8", default-features = false, optional = true }
 thiserror = "1.0.23"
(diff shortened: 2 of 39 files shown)


ant-node/src/bin/antnode/main.rs
```diff
@@ -422,17 +436,71 @@ You can check your reward balance by running:
...
+    #[cfg(unix)]
+    {
+        let ctrl_tx_restart = ctrl_tx.clone();
+        let running_node_clone = running_node.clone();
+        tokio::spawn(async move {
+            loop {
+                // 72 hours (259200 seconds) with ±5% randomization to prevent simultaneous upgrades
+                let base_delay = 259200;
+                let variance = rand::thread_rng().gen_range(-12960..=12960);
+                let upgrade_check_delay_secs = (base_delay + variance) as u64;
```
maqi: this block of `upgrade_check_delay_secs` calculatin can be outside of the loop

ant-node/src/bin/antnode/main.rs
```diff
@@ -422,17 +436,71 @@ You can check your reward balance by running:
...
+                            }
+                        };
+
+                        if let Err(err) = ctrl_tx_restart.send(node_ctrl).await {
+                            error!("Failed to send control command: {err}");
+                        }
+                        break;
+                    }
+                    Err(e) => {
+                        error!("Error during upgrade process: {e}");
```
maqi: `Not to perform an upgrade` shall not be counted as an `error`, as it happens for most of the time ?

ant-node/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.3.0", features = ["process-metrics"] }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
+ant-releases = { git = "https://github.com/jacderida/ant-releases", branch = "chore-serialization_for_release_info" }
```
maqi: just a note that his shall be updated before merge

ant-node-manager/Cargo.toml
```diff
@@ -51,7 +51,7 @@ semver = "1.0.20"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
```
maqi, line 54: just a note that his shall be updated before merge

ant-node-manager/Cargo.toml
```diff
@@ -34,7 +34,7 @@ ant-build-info = { path = "../ant-build-info", version = "0.1.29" }
 ant-evm = { path = "../ant-evm", version = "0.1.17" }
 ant-logging = { path = "../ant-logging", version = "0.3.0" }
 ant-protocol = { path = "../ant-protocol", version = "1.0.9" }
-ant-releases = { version = "0.4.1" }
+ant-releases = { version = "0.4.2" }
```
maqi: just to confirm, so here doesn't need to use the same bespoken branch as ant-node ?

ant-service-management/Cargo.toml
```diff
@@ -22,7 +22,7 @@ prost = { version = "0.9" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 semver = "1.0.20"
-service-manager = "0.8.0"
+service-manager = { git = "https://github.com/jacderida/service-manager-rs", branch = "generic_restart_policy" }
```
maqi, line 25: just a note that his shall be updated before merge

ant-node/src/bin/antnode/main.rs
jacderida: An `UpgradeError` has been introduced to distinguish this scenario.

ant-node-manager/Cargo.toml
jacderida: Yeah, the change for this crate is going to need to be reviewed, so I will merge this PR into the `automatic_upgrades` branch with the Git-based reference, but it will be resolved before going back to `main`.

ant-node-manager/Cargo.toml
jacderida: It's easier to just keep all the `ant-release` references on the same version.

PR #3337, opened 2025-12-02
feat: support upload of large files in 4GB batches

grumbach:
- Max merkle tree depth set at decided `10`
- support upload of large files in 4GB batches (multiple trees/payments for anything larger)
- refactor to support pay/upload/pay next loop upload routine


mickvandijke (approved):
LGTM! 🚢

PR #3347, opened 2025-12-08
Topology verification failure logging

mickvandijke:
Adds some more logging on the topology differences between client and node if topology verification fails. It will tell us who has the most accurate view of the closest peers.


maqi (approved):
Ideally send back more detailed info to client.
As in most of case, we won't have access to node's log.

mickvandijke:
> Ideally send back more detailed info to client.
> As in most of case, we won't have access to node's log.

Yeah that's true, I'll update the PR

mickvandijke:
Hey @maqi , I've updated the PR to send more data to the client side on a topology failure and make the client log the failure in detail instead of the node.

Could you please re-review? Thanks 🙇

maqi (approved):

PR #3354, opened 2025-12-11
Merkle payments Infrastructure (without CLI)

grumbach:
# Merkle Payments Infrastructure

Adds merkle batch payment infrastructure to enable cheaper gas fees by batching on-chain payment for data uploads. This PR does not include the CLI integration which is coming soon...

## Key Changes

**Protocol & API**
- New `Query::GetMerkleCandidateQuote` for requesting merkle payment quotes from nodes
- `RecordKind::DataWithMerklePayment` variant for merkle-paid records
- Client methods for merkle-based uploads: `upload_with_merkle_payments`, `put_with_merkle_payment`

**Node**
- Merkle proof verification in `PutValidation` for all data types
- Node-side quote generation with cryptographic signatures binding PeerId to RewardsAddress
- Payment expiration validation

**EVM Integration**
- Complete `ant-evm::merkle_payments` module with merkle tree construction and verification
- Support for up to 1024 leaves per merkle tree
- Smart contract integration for on-chain payment verification

This infrastructure enables batch payment verification where multiple chunks can be paid for with a single on-chain transaction using merkle proofs, significantly reducing gas costs compared to per-chunk payments.


ant-evm/Cargo.toml
```diff
@@ -15,12 +15,14 @@ test-utils = []
...
 [dependencies]
 custom_debug = "~0.6.1"
+dirs-next = "~2.0" # NB TODO: remove this dependency once we have the real smart contract!
 evmlib = { path = "../evmlib", version = "0.4.5" }
 hex = "~0.4.3"
 libp2p = { version = "0.56.0", features = ["identify", "kad"] }
 rand = { version = "~0.8.5", features = ["small_rng"] }
 ring = "0.17.12"
 rmp-serde = "1.1.1"
+rs_merkle = { git = "https://github.com/maidsafe/rs-merkle.git", branch = "master" }
```
grumbach: @jacderida do we need to release this properly for a release?

ant-protocol/src/storage/header.rs
```diff
@@ -93,23 +102,28 @@ impl<'de> Deserialize<'de> for RecordKind {
...
-        }
+        let data_type = DataTypes::from_index(data_type_index).ok_or_else(|| {
+            serde::de::Error::custom(format!("Unexpected index {num} for RecordKind variant"))
+        })?;
+
+        Ok(match kind_type {
+            "DataOnly" => Self::DataOnly(data_type),
+            "DataWithPayment" => Self::DataWithPayment(data_type),
+            "DataWithMerklePayment" => Self::DataWithMerklePayment(data_type),
+            _ => unreachable!(),
```
maqi, line 125: define a helper enum to avoid `unreachable`, which is not perferred in the production code

ant-protocol/src/peer_id_serde.rs
```diff
@@ -0,0 +1,31 @@
...
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+//! Helper module for serializing/deserializing `Vec<PeerId>`
+
+use libp2p::PeerId;
```
maqi, line 11: we use NetworkAddress for PeerID messaging between clients/nodes.

ant-protocol/src/messages/query.rs
```diff
@@ -84,6 +84,24 @@ pub enum Query {
         /// Address of the record.
         address: NetworkAddress,
     },
+    /// Get a Merkle payment candidate quote from a node
+    ///
+    /// Node signs its current state (metrics + reward address) with the
+    /// provided timestamp, creating a commitment that cryptographically
+    /// proves the binding between PeerId and RewardsAddress.
+    GetMerkleCandidateQuote {
```
maqi, line 92: Using a this new message not only make the merkle quotes to be outstanded from normal quotes, which could have potential backward compatible issue or being vlunerable to be identified out and gaming
But also lost the `storage challenge` feature which could be useful in near future.

Hence, I'd suggest use the same Quote req/rsp.

maqi (changes requested):

maqi (approved):
Though the new `GetMerkleCandidateQuote` messging no longer supports the storage check, the same check can still be carried out by having an additional round of normal `Quote` before/after the Merkle quoting.

It may affect the performance, but given there is now only `commitment pool` nodes to be queried, which is must less than the previous `nodes close to ALL chunks`, the extra round of quoting could be acceptable.

PR #3359, opened 2025-12-12
feat: add verbosity CLI flag and refactor logging targets

RolandSherwin:
Adds a 3-level verbosity system to reduce default logging output. Users can now control log verbosity via `--verbosity` CLI flag (minimal/standard/verbose) or `ANT_LOG` keywords, with custom overrides like `ANT_LOG=std,libp2p=debug` working on top of the base level.

  - Add `--verbosity` flag to `ant-cli`, `antnode`, and `antctl` binaries
  - Support verbosity keywords in `ANT_LOG` env var (`std`, `verbose`, `minimal`)
  - CLI flag takes precedence over `ANT_LOG` keywords, but ANT_LOG custom targets always apply as overrides
  - Set all binaries to use INFO level by default

### DIFF
--- a/ant-logging/src/layers.rs
+++ b/ant-logging/src/layers.rs
@@ -29,10 +29,17 @@ use tracing_subscriber::{
 const MAX_LOG_SIZE: usize = 20 * 1024 * 1024;
 const MAX_UNCOMPRESSED_LOG_FILES: usize = 10;
 const MAX_LOG_FILES: usize = 1000;
-// Everything is logged by default
-const ALL_ANT_LOGS: &str = "all";
-// Trace at nodes, clients, debug at networking layer
-const VERBOSE_ANT_LOGS: &str = "v";
+
+// Verbosity level keywords for ANT_LOG environment variable
+// Verbose mode: TRACE/DEBUG level logging for debugging and dev testnets
+const VERBOSITY_VERBOSE: &str = "verbose";
+const VERBOSITY_VERBOSE_SHORT: &str = "v";
+// Standard mode: sets ALL crates to INFO level
+const VERBOSITY_STANDARD: &str = "standard";
+const VERBOSITY_STANDARD_SHORT: &str = "std";
+// Minimal mode: uses application's default targets (no override)
+const VERBOSITY_MINIMAL: &str = "minimal";
+const VERBOSITY_MINIMAL_SHORT: &str = "min";
 
 /// Handle that implements functions to change the log level on the fly.
 pub struct ReloadHandle(pub(crate) Handle<Box<dyn Filter<Registry> + Send + Sync>, Registry>);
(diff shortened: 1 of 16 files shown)


ant-logging/src/layers.rs
```diff
@@ -29,22 +26,35 @@ use tracing_subscriber::{
...
+// Verbosity level keywords for ANT_LOG environment variable
+// Verbose mode: TRACE/DEBUG level logging for debugging and dev testnets
+const VERBOSITY_VERBOSE: &str = "verbose";
+const VERBOSITY_VERBOSE_SHORT: &str = "v";
+// Standard mode: sets ALL crates to INFO level
+const VERBOSITY_STANDARD: &str = "standard";
+const VERBOSITY_STANDARD_SHORT: &str = "std";
+// Minimal mode: uses application's default targets (no override)
+const VERBOSITY_MINIMAL: &str = "minimal";
+const VERBOSITY_MINIMAL_SHORT: &str = "min";
```
vphongph, line 39: Inconsistency with ant-logging/src/lib.rs where "all" is parsed from a string.
It should include the "all" I think.
Here ANT_LOG=all will be seen as custom!

ant-logging/src/lib.rs
```diff
@@ -111,29 +111,80 @@ impl LogFormat {
...
+    ///
+    /// Valid values: "minimal", "min", "standard", "std", "verbose", "v", "all"
+    pub fn parse_from_str(val: &str) -> Result<Self> {
+        match val.to_lowercase().as_str() {
+            "minimal" | "min" => Ok(VerbosityLevel::Minimal),
+            "standard" | "std" => Ok(VerbosityLevel::Standard),
+            "verbose" | "v" | "all" => Ok(VerbosityLevel::Verbose),
+            _ => Err(Error::LoggingConfiguration(
+                "Valid verbosity values are: 'minimal', 'standard', or 'verbose'".to_string(),
+            )),
```
vphongph, line 139: Inconsistency with ant-logging/src/layers.rs,  where "all" is not parsed.
See the comment on ant-logging/src/layers.rs for the change request.

vphongph (changes requested):
Added comments for some minor change requests

Other comments:
- New documentation in ant-cli/README.md about flag --verbosity should be added
- This feature/refactor should add a CHANGELOG entry

Open questions:
1 - nat detection use an other system `clap_verbosity_flag::Verbosity `, should we unify it with the new system?

2 -  2 types have the same name, it can be confusing:
`ant_logging::VerbosityLevel`
`ant_node_manager::VerbosityLevel`
Should we rename the one from ant_node_manager?

3 - Unit tests for parse_ant_log() and get_logging_targets() could be added?
examples of conflicting and edge cases tests:
`ANT_LOG=verbose,libp2p=debug`
`ANT_LOG=std,minimal`
`ANT_LOG=invalid`
`ANT_LOG=""`
CLI `--verbosity standard ANT_LOG=verbose`

maqi:
After starts using `loud_X` macro, this is no longer much needed.

PR #3367, opened 2025-12-16
refactor(ant-node): increase `KAD_QUERY_TIMEOUT_S` from 10s to 120s

mickvandijke:
Increases the `KAD_QUERY_TIMEOUT_S` on the node side from 10s to 120s to address the `GetClosestTimedOut` error observed during merkle payment topology verifications.


maqi (approved):

dirvine (approved):
Thanks @mickvandijke This aligns with the client now

PR #3368, opened 2025-12-17
Merkle remove abusive xorname use

grumbach:
merge AFTER: #3364

- removes multiple instances of XorName abusive uses as a hashing algo in the merkle flow

### DIFF
--- a/ant-evm/src/merkle_payments/merkle_tree.rs
+++ b/ant-evm/src/merkle_payments/merkle_tree.rs
@@ -10,7 +10,17 @@ use ant_merkle::Hasher;
 use serde::{Deserialize, Serialize};
 use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
+use tiny_keccak::{Hasher as TinyKeccakHasher, Sha3};
 use xor_name::XorName;
+
+/// Compute SHA3-256 hash of input bytes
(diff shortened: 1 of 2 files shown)


ant-evm/src/merkle_payments/merkle_tree.rs
```diff
@@ -10,8 +10,18 @@ use ant_merkle::Hasher;
 use serde::{Deserialize, Serialize};
 use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
+use tiny_keccak::{Hasher as TinyKeccakHasher, Sha3};
 use xor_name::XorName;
 
+/// Compute SHA3-256 hash of input bytes
+fn sha3_256(input: &[u8]) -> [u8; 32] {
```
maqi: make this as a pub(crate) function to avoid duplication

maqi (changes requested):
There is functions like `split_existing_and_new_chunks` which takes in `xornames` and return `xorname`s.
which result in un-necessary conversion between NetworkAddress and Xorname.

best to reduce such case of abuse as well.

Also, the change within ant-evm means could having backward compatible issue with the released nodes/clients.
It shall be alerted.

maqi (approved):

PR #3377, opened 2025-12-18
fix(node): avoid pruning irrelevant_records too aggressively

maqi:
### Description

Using a new conservative approach : 
* Target 10% of farthest records: Instead of removing ALL out-of-range records at once, only the farthest 10% of records are considered candidates for removal.
* Respect `10 * responsible_distance_range`: Among those 10% farthest records, only remove records that are BEYOND the ` 10 * responsible_distance_range`. Records within the responsible range are retained even if they fall in the farthest 10%.

Which prevents:
* Large spikes in CPU/disk usage from removing too many entries at once
* Unnecessary refetching of records that are still within an acceptable range (since other parts of the codebase use an expanded range check, e.g., 10x responsible_distance_range)

### Type of Change


dirvine (approved):
This makes much more sense to be more conservative and less brutal , nice one

PR #3378, opened 2025-12-18
fix: failing merkle payment txs

mickvandijke:
- Adds checks for the transaction receipt status after being mined to avoid false tx confirmations.
- Increases the tx gas limit buffer from 20% to 50%.


grumbach (approved):
LGTM!

PR #3382, opened 2025-12-20
feat(ant-node): add relay_client to NodeSpawner and improve examples

vphongph:
- Add relay_client option to NodeSpawner for NAT traversal
- Add warning for zero rewards address in NodeSpawner/NetworkSpawner
- Add spawn_nodes_on_mainnet.rs example with relay and env var support
- Update spawn_local_network.rs with client connection and cost quote
- Add README.md documenting spawner APIs and options
- Remove obsolete spawn_node.rs example
- Other files: clippy formatting


maqi (changes requested):
PutOnHold as `relay_client` is going to be removed after the reachability work got merged in.

PR #3383, opened 2025-12-22
fix(test): make verify_max_parallel_fetches test deterministic

maqi:
### Description

To ensure the verify_max_parallel_fetches test asserting against deterministic result.
So that the test no longer to fail occasionallly.


grumbach (approved):
LGTM

PR #3384, opened 2025-12-22
feat(client): using req/rsp DM as a fallback during mutable_data fetch

maqi:
### Description

Implemented a common fallback mechanism for fetching records from the closest peers directly, applicable to all data types (Chunk, GraphEntry, Pointer, Scratchpad). 

For CRDT types with counter (Pointer, Scratchpad) return the multiple valid copies to carry out further CRDT merge to be carried out.

This also resolves some occasionally test failures involving mutable datatypes mutation. 

### Type of Change

### DIFF
--- a/autonomi/src/client/data_types/chunk.rs
+++ b/autonomi/src/client/data_types/chunk.rs
@@ -6,6 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use super::FALLBACK_PEERS_COUNT;
+
 use crate::client::chunk_cache::{
     default_cache_dir, delete_chunks, is_chunk_cached, load_chunk, store_chunk,
 };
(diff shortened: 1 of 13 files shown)


autonomi/src/client/data_types/chunk.rs
```diff
@@ -210,11 +166,15 @@ impl Client {
         let record = match record_result {
             Ok(Some(record)) => record,
             Ok(None) | Err(_) => {
-                // Fallback: Try fetching from closest 20 nodes directly
+                // Fallback: Try fetching from closest peers directly
                 debug!("Normal fetch failed, trying fallback for {key:?}");
```
grumbach: Shouldnt we use the fallback within the get_record_with_retries function so all data types can have it?

autonomi/src/client/data_types/graph.rs
```diff
@@ -53,35 +55,122 @@ pub enum GraphError {
...
-            PrettyPrintRecordKey::from(&record.key)
-        );
+        let graph_entries = match record_result {
+            Ok(Some(record)) => {
+                debug!(
+                    "Got record from the network, {:?}",
+                    PrettyPrintRecordKey::from(&record.key)
+                );
+                get_graph_entry_from_record(&record)?
+            }
```
grumbach: > Shouldn't we use the fallback within the get_record_with_retries function so all data types can have it?

This would avoid such repetitive fallback code

autonomi/src/client/data_types/mod.rs
```diff
@@ -89,3 +177,119 @@ where
...
+        1 => {
+            let item = best_items
+                .into_iter()
+                .next()
+                .expect("HashSet with len() == 1 must contain exactly one item");
+            debug!(
+                "Successfully resolved {key:?} with {max_copies} consistent copies"
+            );
+            Ok(item)
+        }
```
grumbach: Use:
```rs
match best_items[..] {
        [one] => 
```
To avoid using `.expect()`

autonomi/src/client/data_types/pointer.rs
```diff
@@ -57,10 +57,26 @@ pub enum PointerError {
     CannotUpdateNewPointer,
     #[error("Got multiple conflicting pointers with the latest version")]
     Fork(Vec<Pointer>),
+    #[error("Insufficient consistent copies: got {0}, need at least {1}")]
+    InsufficientCopies(usize, usize),
```
grumbach: This has nothing to do in `PointerErrors`, as it is already contained in the `GetError` sub-enum

autonomi/src/client/data_types/pointer.rs
```diff
@@ -71,7 +87,11 @@ impl Client {
             .await
         {
             Ok(Some(r)) => pointer_from_record(r)?,
-            Ok(None) => Err(GetError::RecordNotFound)?,
+            Ok(None) => {
+                // Fallback: Try fetching from closest peers directly
+                debug!("Normal fetch returned None, trying fallback for pointer at {key:?}");
+                return self.pointer_get_fallback(key.clone()).await;
+            }
```
grumbach: Again generic fallback would avoid type specific code repetition

autonomi/src/client/data_types/scratchpad.rs
```diff
@@ -61,6 +61,8 @@ pub enum ScratchpadError {
         "Got multiple conflicting scratchpads with the latest version, the fork can be resolved by putting a new scratchpad with a higher counter"
     )]
     Fork(Vec<Scratchpad>),
+    #[error("Insufficient consistent copies: got {0}, need at least {1}")]
+    InsufficientCopies(usize, usize),
```
grumbach: Same here, this shouldn't be in ScratchPad errors

grumbach (changes requested):
Good work, couple of notes:
- the fallback should be generic to records to avoid repetition in all types
- get-specific errors shouldn't be inside each datatype's high level errors as:
    - it is a duplicate error which already exists in sub-enums
    - it is not strictly related to each type in particular

autonomi/src/client/data_types/graph.rs
```diff
@@ -57,28 +57,85 @@ pub enum GraphError {
...
+            1 => {
+                let best_entry = entries_with_max
+                    .into_iter()
+                    .next()
+                    .expect("Vec with len() == 1 must contain exactly one item");
+                debug!(
+                    "Successfully resolved graph entry at {key:?} with {max_copies} consistent copies"
+                );
+                Ok(best_entry)
+            }
```
grumbach, line 130: let's use `match entries_with_max[..]` again to avoid using `expect` here too

grumbach (approved):
LGTM

PR #3388, opened 2025-12-23
refactor(evmlib): gas estimation

mickvandijke:
Improves gas estimation with buffers and avoids edge cases where no gas limit buffer would be set when gas estimation failed.


grumbach (approved):
LGTM

mickvandijke:
Rebased on `main`.

PR #3392, opened 2026-01-06
fix: avoid empty peers on timeout

grumbach:
node side fix to avoid returning an empty peers list on timeout

### DIFF
--- a/ant-node/src/networking/driver/event/kad.rs
+++ b/ant-node/src/networking/driver/event/kad.rs
@@ -106,9 +106,11 @@ impl SwarmDriver {
                         .network_discovery
                         .handle_get_closest_query(current_closest),
                     PendingGetClosestType::FunctionCall(sender) => {
+                        // Send the peers collected before timeout, not empty vec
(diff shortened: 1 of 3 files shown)


ant-node/src/networking/driver/event/kad.rs
```diff
@@ -106,9 +106,11 @@ impl SwarmDriver {
                         .network_discovery
                         .handle_get_closest_query(current_closest),
                     PendingGetClosestType::FunctionCall(sender) => {
+                        // Send the peers collected before timeout, not empty vec
+                        let peers: Vec<_> = current_closest.into_iter().collect();
                         #[allow(clippy::let_underscore_future)]
                         let _ = tokio::spawn(async move {
-                            let _ = sender.send(vec![]);
```
maqi, line 111: a partial result could be misleading/harmful for the followup ops ?

maqi (changes requested):
As to be discussed further, prefer not returning partial result, but using other `fallback` scheme.

grumbach:
Closing for now until further discussion as it affects replication too

PR #3393, opened 2026-01-06
fix: prevent unnecessary restarts for client-only releases

jacderida:
Previously, nodes would restart when a new release was detected, even if the new release only had a change to the client. This caused unnecessary downtime and network churn. Now we compare the `antnode` binary hash to determine if a restart is needed.

### DIFF
--- a/ant-node/src/bin/antnode/upgrade/mod.rs
+++ b/ant-node/src/bin/antnode/upgrade/mod.rs
@@ -321,13 +321,36 @@ pub async fn perform_upgrade() -> Result<()> {
         }
     };
 
-    if release_info
-        .commit_hash
-        .starts_with(ant_build_info::git_sha())
-    {
+    // Check if the antnode binary has changed by comparing SHA256 hashes.
+    let platform = ant_releases::get_running_platform()?;
+    let platform_binaries = release_info
+        .platform_binaries
+        .iter()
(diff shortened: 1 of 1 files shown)


ant-node/src/bin/antnode/upgrade/mod.rs
```diff
@@ -327,7 +327,42 @@ pub async fn perform_upgrade() -> Result<()> {
...
+        .iter()
+        .find(|b| b.name == "antnode")
+        .ok_or_else(|| {
+            UpgradeError::PlatformBinariesNotFound(format!(
+                "antnode binary not found in release for platform: {platform:?}"
+            ))
+        })?;
+
+    let current_exe_path = std::env::current_exe()?;
+    if let Ok(current_hash) = calculate_sha256(&current_exe_path) {
```
RolandSherwin: Could we log if we fail to calculate the hash? As it stands now, we want to allow upgrade even if `calculate_sha256` fails, which is fair, so adding a log line would help us to rule this case out while debugging!

maqi (changes requested):
Just wondering why not to check the antnode hash at first ?
i.e. replace the current check pkg version with check antnode hash.
so that there is no need to check the release info server twice ?

Also, the current_hash shall be populated at the very beginging of the execution, to avoid read in an updated by others exectuable ?

PR #3396, opened 2026-01-08
fix: handle deleted binary path in upgrade hash check

jacderida:
When multiple `antnode` processes run from symlinked binaries pointing to the same source, the first
node to upgrade replaces the shared binary. Subsequent nodes then see their current_exe() path with
" (deleted)" suffix appended by Linux, causing the hash calculation to fail with "No such file or
directory".

This fix:
- Strips the " (deleted)" suffix before calculating the hash
- Caches the running binary's hash using OnceCell since it's constant
  for the process lifetime
- Avoids recalculating the hash on each upgrade check

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/ant-node/src/bin/antnode/upgrade/mod.rs
+++ b/ant-node/src/bin/antnode/upgrade/mod.rs
@@ -18,6 +18,7 @@ use ant_releases::{
 use fs2::FileExt;
 use semver::Version;
 use sha2::{Digest, Sha256};
+use once_cell::sync::OnceCell;
 use std::fs::{self, File};
 use std::io::Read;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


ant-node/src/bin/antnode/upgrade/mod.rs
```diff
@@ -447,8 +477,7 @@ pub async fn perform_upgrade() -> Result<()> {
             ))
         })?;
 
-    let current_exe_path = std::env::current_exe()?;
-    let current_hash = calculate_sha256(&current_exe_path)?;
+    let current_hash = get_running_binary_hash()?;
```
maqi, line 484: the `get_running_binary_hash` shall be called at the very beginning of the node process to prevent following scenario happens:
* node_B got started up using old executable 
* node_A (which runs for a long time) got the executable upgraded shortly after node_B's startup
* node_B reaches its checking time and loading the new hash in, hence skip the upgrade

maqi (changes requested):
the get_running_binary_hash shall be called at the very beginning of the node process to prevent an edge case as detailed explained in the review comment.

maqi (approved):

PR #3397, opened 2026-01-08
feat: simpler client get closest without consensus, merkle put retrie…

grumbach:
- simpler client get closest without consensus
- merkle put retries with larger close group


maqi:
replaced by PR https://github.com/maidsafe/autonomi/pull/3397

PR #3406, opened 2026-01-13
client(feat): improve merkle upload reliability with extended timeout and Kad fallback

maqi:
## Summary

This PR improves the reliability of merkle uploads through two changes:

### 1. Extend request timeout to 120s
- Increases the client request timeout from the default to 120 seconds
- Helps with network operations that may take longer due to network conditions

### 2. Add Kad-only fallback for merkle upload retries
When merkle upload fails with verified closest peers, fall back to direct Kad query to get a different set of peers and retry. This provides more resilience for merkle uploads without affecting other upload types.

**Changes:**
- Add `get_closest_peers_kad_only()` method in networking for direct Kad query without peer verification
- Modify `upload_record_with_merkle_proof()` to use fallback on failure:
  1. First attempt uses verified closest peers
  2. If upload fails, gets fresh peers via direct Kad query and retries

This fallback mechanism is specific to merkle uploads only, as requested.

### DIFF
--- a/autonomi/src/client/merkle_payments/upload.rs
+++ b/autonomi/src/client/merkle_payments/upload.rs
@@ -219,7 +219,10 @@ impl Client {
     }
 
     /// Upload a record with a Merkle payment proof
-    pub async fn upload_record_with_merkle_proof<T: serde::Serialize>(
+    ///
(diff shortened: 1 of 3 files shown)


grumbach (approved):
LGTM

PR #3410, opened 2026-01-15
refactor(ant-node): increase request timeout to 120 seconds

mickvandijke:
- Updates the node's req/res timeout to be in line with the client req/res timeout and the KAD query timeout.
- Updates the substreams timeout for both node and client to be the same as the other request timeouts (120s).

### DIFF
--- a/ant-node/src/networking/network/init.rs
+++ b/ant-node/src/networking/network/init.rs
@@ -120,6 +120,8 @@ pub(super) fn init_driver(
         // How many nodes _should_ store data.
         .set_replication_factor(REPLICATION_FACTOR)
         .set_query_timeout(KAD_QUERY_TIMEOUT_S)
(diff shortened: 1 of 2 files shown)


ant-node/src/networking/network/init.rs
```diff
@@ -120,6 +120,8 @@ pub(super) fn init_driver(
         // How many nodes _should_ store data.
         .set_replication_factor(REPLICATION_FACTOR)
         .set_query_timeout(KAD_QUERY_TIMEOUT_S)
+        // Extend substreams timeout to allow clients with poor connections to upload large records.
```
maqi: the comment shall be updated to:
allows Node accept merkle upload requests from client, which involves certain time to complete a KAD network get_closest query.

With this parameter raised, Node will then no longer terminate the connection earlier.

maqi (approved):

PR #3419, opened 2026-01-22
feat: print information about gas costs

grumbach:
- Display gas cost to user in both merkle and regular upload flows
- fix invalid cli payment type display


mickvandijke (approved):
LGTM

PR #3421, opened 2026-01-22
ci: centralize binary build step

RolandSherwin:
This PR speeds up CI by centralizing binary builds and reusing them across workflows instead of rebuilding from scratch each time.                                                                                    

- Created reusable composite actions for common patterns: `build-binaries` for release binaries, `build-tests` for test compilation, and `local-testnet` for spinning up a test network with pre-built binaries instead building from source each time.
- Split the slow `various checks` job into three parallel jobs (`clippy`, `docs`, `feature-check`) using `cargo-hack` for per-package builds. This should cut the wall time from ~32 minutes to ~15-20 minutes.
- Upgraded `sccache` to v0.10.0 for GHA cache API v2 compatibility and made it resilient to cache service outages by testing connectivity before enabling it.

### DIFF
--- a/.github/actions/build-binaries/action.yml
+++ b/.github/actions/build-binaries/action.yml
@@ -4,25 +4,15 @@ runs:
   using: composite
   steps:
     - uses: dtolnay/rust-toolchain@stable
-    - uses: mozilla/sccache-action@v0.0.3
-    - uses: actions/github-script@v7
+    - uses: mozilla/sccache-action@v0.0.9
       with:
-        script: |
-          core.exportVariable('ACTIONS_RESULTS_URL', process.env.ACTIONS_RESULTS_URL || '');
-          core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
-    - name: Test and enable sccache
+        version: "v0.10.0"
+      continue-on-error: true
+    - name: Enable sccache
       shell: bash
       run: |
-        # Test if sccache can start successfully with GHA cache
-        export SCCACHE_GHA_ENABLED=on
-        if sccache --start-server 2>/dev/null && sccache --show-stats 2>/dev/null; then
-          echo "sccache is working, enabling it"
-          echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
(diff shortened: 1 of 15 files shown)


.github/actions/local-testnet/action.yml
```diff
@@ -0,0 +1,313 @@
...
+      shell: bash
+      run: sudo apt-get -y install jq ripgrep
+    - name: install jq and ripgrep mac
+      if: runner.os == 'macOS'
+      shell: bash
+      run: brew install jq ripgrep
+    - name: install jq and ripgrep windows
+      if: runner.os == 'Windows'
+      shell: pwsh
+      run: choco install jq ripgrep
```
maqi: It says `jq` and `ripgrep` are now already pre-installed on GitHub's `windows-latest` runners. 
The `choco install` adds unnecessary overhead (usually 1-2 minutes) here ?
This might be nessary step with old runner, but can now be saved ?

.github/actions/build-binaries/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      shell: bash
+      run: sudo apt-get update && sudo apt-get install -y mold
+    - if: runner.os == 'Linux'
+      shell: bash
+      run: echo "RUSTFLAGS=-C link-arg=-fuse-ld=mold" >> $GITHUB_ENV
+    - uses: Swatinem/rust-cache@v2
+    - shell: bash
+      run: |
+        set -euo pipefail
+        cargo build --profile=ci --bin antnode --bin ant --bin antctl --bin evm-testnet
```
maqi, line 26: The `ci` profile to be used is using `opt-level = 1` 
compared with the `release` profile (standard release i.e. `opt-level = 3`) we previous used, this new one will result in significantly slower execution.
This not only result in slower execution in normal CI runs, but aslo result in the benchmark runs not reflecting the real-world performance.

.github/actions/local-testnet/action.yml
RolandSherwin: Thanks for the comment! Updated it!

.github/actions/build-binaries/action.yml
RolandSherwin: Updated to `opt-level=3`

jacderida (approved):

PR #3422, opened 2026-01-23
Cli consistency fixes

grumbach:
- fix inconsistencies between cost and upload commands
- unify cmd options flow

```
⏺ Payment Flags for ant file cost and ant file upload                                                 
                                                                                                      
  By default, the CLI auto-selects the payment method based on chunk count:                           
  - < 64 chunks: regular payments (single-node mode)                                                  
  - >= 64 chunks: merkle tree payments                                                                
                                                                                                      
  Flags                                                                                               
                                                                                                      
  --merkle                                                                                            
  Force merkle tree payments regardless of chunk count. Best for large uploads — batches payments     
  through a smart contract.                                                                           
                                                                                                      
  --regular                                                                                           
  Force regular per-batch payments regardless of chunk count. Best for small uploads.                 
                                                                                                      
  --disable-single-node-payment                                                                       
  Use standard payment mode instead of the default single-node mode. Standard mode pays 3 nodes       
  individually, costing more in gas fees. Single-node mode (default) pays one node 3x the amount,     
  saving gas. Only applies to regular payments — has no effect on merkle.                             
                                                                                                      
  Mutual exclusivity                                                                                  
                                                                                                      
  - --merkle and --regular cannot be combined.                                                        
  - --merkle and --disable-single-node-payment cannot be combined.                                    
  - --regular and --disable-single-node-payment can be combined (regular payments with standard mode).
                                                                                                      
  Examples                                                                                            
                                                                                                      
  # Auto-select (recommended)                                                                         
  ant file cost mydir/                                                                                
  ant file upload mydir/                                                                              
                                                                                                      
  # Force merkle for a small upload                                                                   
  ant file cost --merkle mydir/                                                                       
  ant file upload --merkle mydir/                                                                     
                                                                                                      
  # Force regular for a large upload                                                                  
  ant file cost --regular mydir/                                                                      
  ant file upload --regular mydir/                                                                    
                                                                                                      
  # Regular with standard (3-node) payment mode                                                       
  ant file upload --regular --disable-single-node-payment mydir/                                      
                                                                                                      
  Resume behavior                                                                                     
                                                                                                      
  On upload, if cached payment receipts exist from a previous interrupted upload, the CLI             
  automatically resumes using the cached method — unless you explicitly override with --merkle or     
  --regular, in which case the incompatible cache is ignored.    
```

### DIFF
--- a/ant-cli/src/commands/file.rs
+++ b/ant-cli/src/commands/file.rs
@@ -16,7 +16,8 @@ use crate::wallet::load_wallet;
 use autonomi::client::PutError;
 use autonomi::client::analyze::Analysis;
 use autonomi::client::config::MERKLE_PAYMENT_THRESHOLD;
-use autonomi::client::payment::{BulkPaymentOption, PaymentOption};
+use autonomi::client::merkle_payments::MerklePaymentReceipt;
+use autonomi::client::payment::{BulkPaymentOption, PaymentOption, Receipt};
 use autonomi::files::{UploadError, estimate_directory_chunks};
 use autonomi::networking::{Quorum, RetryStrategy};
 use autonomi::{AttoTokens, Client, ClientOperatingStrategy, PaymentMode, TransactionConfig};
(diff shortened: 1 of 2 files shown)


ant-cli/src/commands/file.rs
```diff
@@ -18,154 +19,247 @@ use autonomi::client::config::MERKLE_PAYMENT_THRESHOLD;
...
+            .wrap_err("Failed to calculate cost for file")?
+    };
 
-        println!("Estimate cost to upload file: {file}");
-        println!("Total cost: {cost}");
-        println!("Method: {method_info}");
-        info!("Total cost: {cost} for file: {file}");
-    }
+    // Print results
+    println!("Estimate cost to upload file: {file}");
```
maqi, line 307: Consider retaining the threshold information for transparency
```
            format!(
                "merkle (auto-selected: ~{estimated_chunks} chunks >= {MERKLE_PAYMENT_THRESHOLD} threshold)"
            )
```

ant-cli/src/commands/file.rs
```diff
@@ -18,154 +19,247 @@ use autonomi::client::config::MERKLE_PAYMENT_THRESHOLD;
...
+                format!(
+                    "{} {action} regular payments (~{} chunks, {mode})",
+                    self.emoji, self.estimated_chunks
+                )
+            }
+        }
+    }
+}
+
+/// Determine the payment method based on flags, cached receipts, and chunk count
```
maqi: best have a detailed general scheme explaination as function comment
the cascaded if-clause is bit tricky to follow

ant-cli/src/commands/file.rs
```diff
@@ -351,61 +445,55 @@ async fn upload_dir_standard(
...
+    // Build payment option based on method and cached receipts
+    let payment_option = match &method_info.method {
+        PaymentMethod::Merkle { is_resuming } => {
+            if *is_resuming {
+                if let Some(merkle_receipt) = cached_merkle {
+                    BulkPaymentOption::ContinueMerkle(wallet.clone(), merkle_receipt)
+                } else {
+                    BulkPaymentOption::ForceMerkle(wallet.clone())
+                }
+            } else if force_merkle {
```
maqi: The logic indicated here is bit confusing that:
When returns with `PaymentMethod::Merkle` there is still chance that uploading without merkle ?

ant-cli/src/commands/file.rs
```diff
@@ -18,154 +19,247 @@ use autonomi::client::config::MERKLE_PAYMENT_THRESHOLD;
...
-            content_cost
-                .checked_add(archive_cost)
-                .ok_or_else(|| eyre!("Cost overflow when adding archive cost"))?
-        } else {
-            content_cost
-        };
-
-        println!("Estimate cost to upload file: {file}");
-        println!("Total cost: {total_cost}");
-        println!("Method: merkle (forced)");
```
maqi, line 83: This explicit printout was useful for scripting/parsing output

ant-cli/src/commands/file.rs
```diff
@@ -18,154 +19,247 @@ use autonomi::client::config::MERKLE_PAYMENT_THRESHOLD;
...
+
+/// Determine the payment method based on flags, cached receipts, and chunk count
+fn determine_payment_method(
+    estimated_chunks: usize,
+    force_merkle: bool,
+    force_regular: bool,
+    use_standard_payment: bool,
+    has_cached_regular: bool,
+    has_cached_merkle: bool,
+) -> PaymentMethodInfo {
```
maqi: Consider having determine_payment_method return a simpler enum or tuple that includes the BulkPaymentOption directly (or a builder for it), reducing the double-matching pattern

The PaymentMethodInfo struct carries emoji and estimated_chunks that could be derived (just make the derivation a unified helper function to retain the feature of centralization for the ease of future maintainance) when needed rather than stored. The pattern of creating a struct with display info, then matching on it to build BulkPaymentOption is somewhat indirect.

maqi (commented):
Best got the PR description updated to explain the changes in detail, i.e. with which input flag (or flags with combination), what behaviour/output to be expected.

ant-cli/src/commands/file.rs
grumbach: nice catch, let me simplify that

ant-cli/src/commands/file.rs
grumbach: This has been moved but is still here:
```
    println!("Estimate cost to upload file: {file}");
    println!("Total cost: {total_cost}");
    println!("Method: {method_label}");
```

ant-cli/src/commands/file.rs
grumbach: Agreed!

maqi (approved):

PR #3423, opened 2026-01-23
feat: local merkle pricing with the correct formula

grumbach:
- Fix local merkle pricing which was inconsistent with mainnet


evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -24,11 +25,38 @@ contract MerklePaymentVault is IMerklePaymentVault {
...
+    mapping(DataType => uint256) public costUnitPerDataType;
+
     // ============ Constructor ============
 
     constructor(address _antToken) {
         require(_antToken != address(0), "Invalid token address");
         antToken = IERC20(_antToken);
+
+        // Initialize cost units per data type
+        costUnitPerDataType[DataType.Chunk] = 1e18;
```
maqi: better use a named const instead

evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -24,11 +25,38 @@ contract MerklePaymentVault is IMerklePaymentVault {
     /// Number of candidates per pool (fixed)
     uint8 public constant override CANDIDATES_PER_POOL = 16;
 
+    // ============ Pricing Constants ============
+
+    /// Precision for fixed-point calculations (1e18)
+    uint256 public constant PRECISION = 1e18;
```
maqi, line 31: evm lib using the following two consts for the similar purpose:
```
/// The conversion from AttoTokens to raw value
const TOKEN_TO_RAW_POWER_OF_10_CONVERSION: u64 = 18;
/// The conversion from AttoTokens to raw value
const TOKEN_TO_RAW_CONVERSION: u64 = 1_000_000_000_000_000_000;
```
Best use them directly, or remain in the same format?

evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -189,14 +220,97 @@ contract MerklePaymentVault is IMerklePaymentVault {
...
+        // Calculate linear part: rUpper - rLower
+        uint256 linearPart = _absDiff(upperBound, lowerBound);
+
+        // Formula components:
+        // partOne = (-s/ANT) * logDiff
+        // partTwo = pMin * linearPart / PRECISION
+        // partThree = linearPart / ANT (scaled by PRECISION)
+        int256 partOne = (-int256(scalingFactor) * logDiff) / int256(ANT_PRICE * PRECISION);
+        uint256 partTwo = (linearPart * minPrice) / PRECISION;
+        uint256 partThree = (linearPart * PRECISION) / ANT_PRICE / PRECISION;
```
maqi: this is equal to `uint256 partThree = linearPart / ANT_PRICE;` ?

evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -173,7 +204,7 @@ contract MerklePaymentVault is IMerklePaymentVault {
     }
 
     /// Calculate median price from CANDIDATES_PER_POOL candidate quotes
-    function _calculateMedianPrice(CandidateNode[16] calldata candidates) internal pure returns (uint256) {
+    function _calculateMedianPrice(CandidateNode[16] calldata candidates) internal view returns (uint256) {
```
maqi, line 210: _calculateMedianPrice changes from `pure` to `view` because _getQuote now reads state.
This is necessary but increases gas cost slightly for callers?

evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -189,14 +220,97 @@ contract MerklePaymentVault is IMerklePaymentVault {
...
+
+    /// Calculate total cost unit from metrics
+    function _getTotalCostUnit(QuotingMetrics calldata metrics) internal view returns (uint256) {
+        uint256 total = 0;
+        for (uint256 i = 0; i < metrics.recordsPerType.length; i++) {
+            Record calldata record = metrics.recordsPerType[i];
+            total += costUnitPerDataType[record.dataType] * record.records;
+        }
+        // Also add the closeRecordsStored as a base contribution
+        total += metrics.closeRecordsStored * PRECISION / 1000; // Scale down close records
```
maqi: using magic number of 1000 ?

evmlib/contracts/MerklePaymentVault.sol
```diff
@@ -189,14 +220,97 @@ contract MerklePaymentVault is IMerklePaymentVault {
...
+        if (a >= b) {
+            return a - b;
+        }
+        return b - a;
+    }
+
+    /// Calculate natural logarithm using PRB Math
+    function _calculateLn(uint256 x) internal pure returns (int256) {
+        if (x == 0) {
+            return type(int256).min; // Return minimum for log(0)
```
maqi: returning `type(int256)` (i.e. `-2^255`).min could cause unexpected behaviors:
Overflow in multiplication: When computing partOne, if logDiff is extremely large (derived from type(int256).min), multiplying by scalingFactor would overflow.

Though the caller already guards against zero:
```
if (upperDiff == 0 || lowerDiff == 0) {    return minPrice;}
```
So _calculateLn(0) should never actually be called. The type(int256).min fallback is defensive code that ideally never executes.

However, if this defensive path is somehow reached (bug, future refactor), the sentinel value could cause silent arithmetic errors rather than a clear failure. 

A revert("ln(0) undefined") would be safer as defensive code.

maqi (changes requested):

evmlib/contracts/MerklePaymentVault.sol
grumbach: These share the same numeric value but represent different concepts:                                
                                                                                                      
  - PRECISION in the Solidity contract is the fixed-point scaling factor for SD59x18 arithmetic — it's
   used in ratio calculations like _getBound(), _getTotalCostUnit(), and _getQuote() for mathematical 
  precision.                                                                                          
  - TOKEN_TO_RAW_CONVERSION in Rust is about token denomination — converting between AttoTokens       
  (smallest unit) and whole tokens.                                                                   
                                                                                                      
  They're both 1e18 because Solidity's SD59x18 standard and ERC20 token decimals both use 18 decimal  
  places, but the intent is different. So I think it's best to keep it separate.

evmlib/contracts/MerklePaymentVault.sol
grumbach: This is necessary because _getQuote now reads scalingFactor, minPrice, maxCostUnit, and                  
  costUnitPerDataType from storage. Note that this is a test smart contract and that the live smart contract does much more than that!

evmlib/contracts/MerklePaymentVault.sol
grumbach: You're right, this shouldn't be here, let me remove that

maqi (approved):

PR #3425, opened 2026-01-26
node(feature): re-enable trust-based replication scoring

maqi:
## Summary
- Re-enable trust-based replication validation alongside majority consensus.
- Adjust duration scoring to map 0-5s to the full score range.
- Document trust score calculation and replication usage.


maqi:
Replaced by PR https://github.com/maidsafe/autonomi/pull/3434

PR #3433, opened 2026-01-28
feat: integrate Reachability into Antnode

RolandSherwin:
This PR integrates the `ReachabilityCheck` into `antnode`, such that, every node will first try to run the reachability check before starting the antnode swarm. 

This PR also contains some changes which would help us with the ant-node-manager integration. These include:
- Graceful shutdown of metrics server
- Critical failure logging
- Deprecating antnode rpc logic from protocol code and moving it to antnode


ant-node/src/critical_failure.rs
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+            error!("Failed to remove critical failure file");
+        }
+    }
+}
+
+fn node_error_to_reason(error: &Error) -> String {
+    match error {
+        Error::Network(network_error) => {
+            println!("Network error: {network_error}");
```
maqi: this shall be removed, logged, or logged in parallel at least.
same to the couple of other places.

ant-node/src/critical_failure.rs
```diff
@@ -0,0 +1,132 @@
...
+            let network_err_str = match network_error {
+                NetworkError::DialError(_) => "DialError".to_string(),
+                NetworkError::BootstrapError(_) => "BootstrapError".to_string(),
+                NetworkError::Io(_) => "IoError".to_string(),
+                NetworkError::KademliaStoreError(_) => "KademliaStoreError".to_string(),
+                NetworkError::TransportError(_) => "TransportError".to_string(),
+                NetworkError::ProtocolError(_) => "ProtocolError".to_string(),
+                NetworkError::EvmPaymemt(_) => "EvmPayment".to_string(),
```
maqi: typo, shall be `EvmPayment`
same to the couple of other places

ant-node/src/node.rs
```diff
@@ -177,19 +184,132 @@ impl NodeBuilder {
...
+                    }
+                }
+                Ok((ReachabilityStatus::NotReachable { reasons }, _metrics_shutdown_tx)) => {
+                    info!(
+                        "We are NOT reachable due to: {reasons:?}. Terminating the node in 120 seconds."
+                    );
+                    println!("Terminating the node in 120 seconds: {reasons:?}");
+                    let failure_error = Error::UnreachableNode;
+                    set_critical_failure(&self.root_dir, &failure_error);
+                    tokio::time::sleep(Duration::from_secs(120)).await;
```
maqi, line 292: consider using a named const, or even configurable.

Also, will 120s too long?
Users may be confused why the node hangs for 2 minutes before exiting ?

maqi (approved):
Minor Suggestions:
1. Test port conflict risk in metrics server test:
```
let port = 8081; // Use a specific test port
```
This could conflict with other tests or services. Consider using port 0 or a random port.
2. Hardcoded bootstrap constants could use documentation:
```
const CONCURRENT_DIALS: usize = 3;
const MAX_PEERS_BEFORE_TERMINATION: usize = 5;
```
Why were these specific values chosen?
3. Dead code suppression on retain_peer_id:
```
#[allow(dead_code)]
retain_peer_id: bool,
```
If this field is unused, consider removing it or documenting why it exists.
4. New metadata_extended endpoint added but seems without associated documentation about what it contains or when to use it?

ant-node/src/node.rs
RolandSherwin: Checking at the commit messages of the old branch, I can see:
```
feat(antnode): wait 120s before terminating nodes if unreachables 
```
So this was intentionally added, but I now do not remember what might be the case for this. Lemme think about this.

RolandSherwin: Looks like it might be just for the metrics server to scrape the info, if so, we could just have 10 seconds here.

PR #3434, opened 2026-01-28
node(feature): re-enable trust-based replication scoring

maqi:
Replacing PR https://github.com/maidsafe/autonomi/pull/3425

## Summary
- Re-enable trust-based replication validation alongside majority consensus.
- Adjust duration scoring to map 0-5s to the full score range.
- Document trust score calculation and replication usage.

### DIFF
--- a/ant-node/src/networking/replication_fetcher.rs
+++ b/ant-node/src/networking/replication_fetcher.rs
@@ -43,11 +43,46 @@ const PENDING_TIMEOUT: Duration = Duration::from_secs(900);
 // The time the entry will be considered as `time out` and to be cleared.
 type ReplicationTimeout = Instant;
 
+/// Manages replication fetching with trust-based and majority-based validation.
+///
+/// # Trust Scoring System
+///
+/// The ReplicationFetcher uses a dual-approach strategy for validating replication sources:
+///
+/// ## 1. Trust-Based Validation (via `peers_scores`)
+///
+/// Peers are evaluated based on their StorageChallenge response history:
+/// - Each peer maintains a history of up to 2 recent challenge results (bool: healthy/unhealthy)
+/// - A peer is considered "trustworthy" if they have 2+ scores and at least 2 are healthy
+/// - A peer is considered "untrustworthy" if they have 2+ scores but fewer than 2 healthy
+/// - Peers without sufficient history are "unknown"
+///
+/// Trust scoring is populated by the StorageChallenge mechanism in node.rs:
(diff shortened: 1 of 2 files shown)


vphongph (changes requested):
Hi, there are some fixes to bring:

#### on replication_fetcher.rs:443-455

The Some(true) branch builds result with (holder, addr, val_type) for every key, then calls initial_majority_replicates().
When majority is reached, that function returns all peers who voted (line 501-503), not just holder.
The dedup at line 452 compares full (PeerId, Addr, ValType) tuples — so (peer_A, addr_X, Chunk) passes even though (holder, addr_X, Chunk) is already present for the same key.
Same data gets queued for fetch from two peers, wasting the 5 fetch slots.

Fix: dedup on (Addr, ValType) only, ignoring PeerId.

#### on replication_fetcher.rs:110

The initial_replicates HashMap has no cleanup mechanism. Before this PR, had_enough_scoring_knowledge() called .clear() on it — that function was deleted (confirmed by the comment at line 388), with no replacement.
Entries are only removed when majority is reached (line 501). Keys that only 1 peer ever reports stay forever. Under churn or Sybil attack (single peer flooding unique fake keys), unbounded growth.

Fix: Add TTL-based pruning in prune_expired_keys_and_slow_nodes() (line 724), using PENDING_TIMEOUT (900s) as the TTL.

ant-node/src/networking/replication_fetcher.rs
```diff
@@ -667,6 +764,33 @@ impl ReplicationFetcher {
...
+        // Prune stale entries from initial_replicates that never reached majority.
+        // This prevents unbounded memory growth under churn or Sybil attacks.
+        // Use PENDING_TIMEOUT as TTL - if an entry hasn't reached majority by then,
+        // it's unlikely to ever reach it.
+        let now = Instant::now();
+        let before_count = self.initial_replicates.len();
+        self.initial_replicates
+            .retain(|(addr, _val_type), (_peers, first_seen)| {
+                let dominated_by_others = now.duration_since(*first_seen) >= PENDING_TIMEOUT;
+                if dominated_by_others {
```
vphongph, line 777: The name is a bit confusing, I think `is_expired` or `exceeded_ttl` should be more explicit

vphongph (approved):
Looks good to me! GG!

Just a minor concern about the var named `dominated_by_others`, it's a bit confusing and I think `is_expired` or `exceeded_ttl` should be more explicit!

PR #3443, opened 2026-01-31
feat(evmlib): add compact merkle payment calldata (v2)

mickvandijke:
- Add payForMerkleTree2 compact/packed calldata contract function for smaller calldata and lower gas costs (fixes reverted payments on large merkle trees)
- Add calldata verification on the node side so that the client can't lie about the quoting metrics

### Requires:

- [x] Merkle vault contract updated on arbitrum sepolia and arbitrum one
- [x] Update local test contract to support the new interface
- [x] Update contract addresses for both in evmlib


evmlib/src/merkle_batch_payment.rs
```diff
@@ -65,6 +91,105 @@ pub struct CandidateNode {
...
+    pub candidates: [CandidateNodePacked; CANDIDATES_PER_POOL],
+}
+
+/// Encode data type and total cost unit into a single U256
+///
+/// Format: `packed = (totalCostUnit << 8) | dataType`
+/// - dataType occupies bits 0-7 (lower 8 bits)
+/// - totalCostUnit occupies bits 8-255
+pub fn encode_data_type_and_cost(data_type: u8, total_cost_unit: U256) -> U256 {
+    assert!(
```
maqi: ## Potential panic/DoS in cost‑unit verification.

The packed encoding path uses assert! to enforce the 248‑bit limit, and this function is called during verification of on‑chain packed commitments. 
A malicious client could supply (colluding) signed metrics with a very large records_per_type/close_records_stored to force a panic and crash validation. 
This is a reliability/security issue because it occurs in the node’s validation path. 

Consider returning a Result and surfacing a validation error instead of panicking.

mickvandijke: Good one 👍 . Fixed in: https://github.com/maidsafe/autonomi/pull/3443/commits/94fa7662dab85710ac8c2181f615d45f8ea4875b

maqi (commented):
High Severity

* Wallet balance/allowance checking severely weakened (evmlib/src/wallet.rs)

  The cost estimation via estimateMerkleTreeCost view call is removed entirely. The balance check is reduced to wallet_balance == U256::ZERO, meaning a wallet with 1 wei would pass preflight checks but the transaction would fail on-chain, wasting gas. Users get opaque contract revert errors instead of the previous clear InsufficientTokensForQuotes error.

  The PR comment says "we can't use the existing estimate function since it expects unpacked commitments" — but adding an estimateMerkleTreeCost2 view function to the contract would solve this. Alternatively, the packed commitments could be unpacked client-side just for the estimate call.

  The allowance logic is also changed from allowance < estimated_amount to allowance < wallet_balance, which semantically means "approve max spend whenever allowance is less than total wallet balance" rather than "approve enough for this payment."

* Mandatory RPC call for every payment verification (ant-node/src/put_validation.rs)

  verify_cost_units is mandatory and requires fetching transaction calldata from the blockchain on every PUT validation. This adds latency and creates a hard dependency on RPC availability. If the RPC node is down, legitimate payments are rejected after 3 retries (each with 5-second wait = 15 seconds blocking). The security justification is sound (preventing underpayment), but the operational impact should be acknowledged — RPC downtime now causes node-wide payment verification failures.

 ---
Medium Severity

* Significant Solidity code duplication (MerklePaymentVault.sol)

 payForMerkleTree2, _calculateMedianPricePacked, and _getQuotePacked are near-copies of their v1 counterparts. The entire pricing formula (bounds, logarithms, scaling) is duplicated. Any bug fix in one function could be missed in the other. Consider refactoring shared logic into internal helpers parameterized by how data type and cost unit are extracted.

 ---
 Low Severity

* zip() in verify_cost_units is safe but not self-documenting (merkle_payment.rs:70-74)
```rust
  on_chain_winner.candidates.iter()
      .zip(self.candidate_nodes.iter())
```
  Both are fixed-size [_; CANDIDATES_PER_POOL] arrays, so zip() won't silently drop elements. But if either type is ever changed to Vec, elements could be silently skipped without verification. A defensive length assertion or comment
  would prevent future regressions.

* cost_unit_for_data_type returns U256::ZERO for unknown types — Unknown data types (>= 4) contribute zero cost, which could mask bugs if new data types are added without updating both the contract and Rust code.                     

* Retry loop structure in verify_cost_units (put_validation.rs) — The first attempt is outside the loop, then for attempt in 2..=3 runs the retries. This is correct (3 total attempts) but the structure is more convoluted than a simple `for attempt in 1..=MAX_RETRIES with the call inside the body`.

mickvandijke:
Hey @maqi , thanks for the review!

> **Wallet balance/allowance checking severely weakened (evmlib/src/wallet.rs)**

Fixed in commit: https://github.com/maidsafe/autonomi/pull/3443/commits/1a87d27b2bdfaba87cded6b2aec709a43dc4b0db

> **Mandatory RPC call for every payment verification (ant-node/src/put_validation.rs)**

This was already the case when fetching the payment info during merkle verification.

> **Significant Solidity code duplication (MerklePaymentVault.sol)**

This can be ignored. V1 will be removed in the near-future.

> **zip() in verify_cost_units is safe but not self-documenting (merkle_payment.rs:70-74)**

This won't be an issue because the array lengths will always be the same size.

maqi:
> > **Mandatory RPC call for every payment verification (ant-node/src/put_validation.rs)**
> 
> This was already the case when fetching the payment info during merkle verification.

Just as a record that original review concern was overstated — it's not a new RPC call.
It's a continuation of  the existing pattern.

Meanwhile, it is true that there is now an addiontial RPC call for the verify cost units.
However, as nodes don’t have to verify merkle payments often,
the rate limit isn't likely to be a concern.

maqi (approved):

PR #3445, opened 2026-02-02
fix(autonomi): avoid re-quoting chunks on upload resume (AUTO-848)

maqi:
## Summary

Fix the upload retry re-quoting bug (AUTO-848) where ALL chunks are re-quoted on retry even when they already exist on the network. This fix covers **both merkle and regular** upload flows.

## Problem

When an upload fails and is retried (without cached receipt), the client was:
1. Re-querying the network for ALL chunks to check existence
2. Re-quoting ALL chunks (even those known to exist)
3. Unnecessarily burdening the network

Additionally, the regular upload flow had two further issues:
- **No progressive receipt saving**: Regular payment receipts were only saved on error path (PutError::Batch), so SIGINT interrupts lost all payment proofs
- **Save/load format mismatch**: `cached_payments::save_payment()` serialized a `ChunkBatchUploadState` struct, but `load_payment_for_file()` tried to deserialize as `Receipt` (HashMap), always failing silently

### Merkle flow
Add `already_existed: HashSet<XorName>` field to `MerklePaymentReceipt` to persist which chunks were found to already exist on the network. On resume:

1. Chunks in `receipt.already_existed` - skip network check (known to exist)
2. Chunks in `receipt.proofs` - skip network check (already paid, will upload)
3. Only query network for truly unknown chunks

Progressive saving via `ClientEvent::MerkleBatchPaymentComplete` ensures the receipt is flushed to disk after each batch payment, surviving SIGINT.

### Regular flow
Add progressive saving of regular payment receipts via a new `ClientEvent::RegularBatchPaymentComplete(Receipt)` event:

1. After each batch payment, emit the event with the batch receipt
2. CLI accumulates batch receipts into a cumulative receipt and saves to disk progressively
3. On resume, `load_payment_for_file()` loads the saved `Receipt` directly
4. `pay_for_content_addrs()` with `PaymentOption::Receipt` returns the cached receipt immediately - no re-quoting

Also fix the save/load format mismatch by adding `save_regular_payment()` that writes `Receipt` directly (matching what `load_payment_for_file()` expects).

### Code Changes (autonomi crate)
- `client/mod.rs`: Add `RegularBatchPaymentComplete(Receipt)` to `ClientEvent`
- `client/merkle_payments/payments.rs`: Add `already_existed` field to `MerklePaymentReceipt`
- `client/merkle_payments/file.rs`: Update `split_existing_and_new_chunks()` to use receipt's `already_existed` set; emit progressive save events
- `client/high_level/data/helpers.rs`: Add `send_regular_batch_payment_complete()` helper; emit event after each regular batch payment
- `python.rs`: Handle new `RegularBatchPaymentComplete` variant in Python bindings

### Code Changes (ant-cli crate)
- `access/cached_payments.rs`: Add `save_regular_payment()` that saves `Receipt` directly; fix `save_payment()` to delegate to it (fixes format mismatch)
- `utils.rs`: Handle `RegularBatchPaymentComplete` event - accumulate and save receipts progressively

### CI Changes
- Extract `large_file_upload_test` from `merge.yml` to separate `large_file_upload.yml`
- Add new **"Upload retry re-quoting behaviour (merkle)"** and **"Upload retry re-quoting behaviour (regular)"** tests that:
  - Uploads a file with `--merkle` (or `--regular`), interrupts mid-way
  - Cleans up receipts, re-uploads with same `--merkle` (or `--regular`) (receipt saved progressively)
  - Resumes with same `--merkle` (or `--regular`) and cached receipt (verifies fix)
  - Fails if re-quoting detected on resume

## Testing

The CI tests verify both merkle and regular flows:
- **Step 4 (PASS)**: Verify quoting/payment happened for all chunks (expected behavior)
- **Step 6 (FAIL before fix)**: Re-quoting detected on resume (bug)
- **Step 6 (PASS after fix)**: No re-quoting on resume - cached receipt used

## Related

- Linear task: AUTO-848


mickvandijke:
**Findings (ordered by severity)**

  **1. Major: Merkle resume still re-pays already‑paid chunks when batch mixes paid + unpaid**
      - In files_put_with_merkle_payment_internal, payment is invoked for a whole batch if any chunk lacks a proof. If the batch contains a mix of already‑paid and unpaid chunks, pay_for_merkle_tree_batch will re-quote/pay for chunks
        that already have proofs, defeating the “avoid re-quoting” goal.
      - Location: autonomi/src/client/merkle_payments/file.rs (batch loop around needs_payment and pay_for_merkle_tree_batch).
      - Fix idea: filter batch_xornames to only unpaid before calling pay_for_merkle_tree_batch, or split batches into paid/unpaid and only pay for unpaid while still uploading all chunks.
      - File reference: autonomi/src/client/merkle_payments/file.rs
  **2. Major: Receipt filename is non-deterministic across progressive saves**
      - save_regular_payment derives the filename timestamp from receipt.values().next() on a HashMap. HashMap iteration order is nondeterministic, so the timestamp used for the filename can change across saves. With progressive
        saving, this can create multiple receipt files for the same upload and load_payment_for_file may return an older/stale receipt.
      - This is more likely now that receipts are saved repeatedly, not just on failure.
      - Suggestion: use a stable filename for the current upload (e.g., {file_hash}.json with mtime for expiry), or store the timestamp once and reuse it for subsequent saves.
      - File references: ant-cli/src/access/cached_payments.rs:29, ant-cli/src/access/cached_payments.rs:137
  **3. Minor (verify): Possible empty receipt caching when all chunks are free**
      - RegularBatchPaymentComplete is emitted whenever PaymentOption::Wallet is used, even if the returned receipt is empty because all chunks were free. That would save an empty receipt file, and on next run the CLI might choose
        BulkPaymentOption::Receipt and fail due to missing proofs.
      - I couldn’t confirm without tracing pay_for_content_addrs, so this is a “please double-check” rather than a confirmed bug.
      - File references: autonomi/src/client/high_level/data/helpers.rs, ant-cli/src/utils.rs

maqi:
Thanks for the review! Here is the analysis of the three concerns:

## Concern 1: Merkle resume re-pays already-paid chunks when batch mixes paid + unpaid

**Not a practical issue.** Merkle tree payment is atomic - the entire tree is paid in a single EVM transaction via `pay_for_merkle_tree_batch`. On resume, a batch either has all proofs (`needs_payment = false`, skipped entirely) or no proofs (`needs_payment = true`, paid in full). A "mixed" batch (some chunks paid, some not) can only occur if batch boundaries shift between runs (e.g., chunk count changes), which is an edge case that would at worst cause a minor overpayment for a few chunks - not a correctness bug.

The `already_existed` set tracks chunks confirmed to exist on the network (not paid-for chunks), so it correctly avoids re-querying the network for those. Chunks in `receipt.proofs` are known to have been paid and will be uploaded - also skipped from network queries. Only truly unknown chunks trigger network checks.

## Concern 2: Receipt filename non-deterministic across progressive saves - Fixed

**Valid.** `get_timestamp_from_receipt()` uses `receipt.values().next()` on a `HashMap`, which has non-deterministic iteration order. Each progressive save could pick a different entry's timestamp, creating multiple receipt files with different names. `load_payment_for_file()` would then non-deterministically pick one (potentially stale).

**Fix in f60f845:** `save_regular_payment()` now calls `remove_existing_receipts_for_hash()` to delete any existing receipt files matching the same file hash before writing the new one. This ensures only one receipt file exists at any time, always the most recent.

## Concern 3: Empty receipt caching when all chunks are free - Fixed

**Valid edge case.** When all chunks in a batch are free (already stored), `pay_for_content_addrs` returns an empty receipt. The event was being emitted and the empty receipt cached. On resume, `PaymentOption::Receipt(empty_receipt)` would be used, and `chunk_batch_upload` would skip all chunks (none present in receipt).

**Fix in f60f845:** The `RegularBatchPaymentComplete` event is now only emitted when `!receipt.is_empty()`, preventing empty receipts from being cached.

mickvandijke (approved):

PR #3452, opened 2026-02-07
fix(node-launchpad): update UI in real-time when node status transitions

jacderida:
## Summary
- Fix `update_node_items()` to check the node registry during Starting/Updating states, transitioning the UI when the registry confirms Running
- Add faster registry polling (5s instead of 180s) while any node is in a transitional state (Starting/Stopping/Updating)
- Add a `Stopping` status with spinner feedback so users see immediate visual response when stopping a node
- Suppress stale "Error starting node" dialogs when the node has already moved past the Starting state
- Ignore Terraform local state and cache files (`.terraform/`, `.terraform.lock.hcl`, `*.tfstate`). These files are generated by Terraform configurations on another branch that will soon be merged.

## Test plan
- [x] `cargo clippy --all-targets --all-features -- -Dwarnings` passes
- [x] `cargo fmt --all -- --check` passes
- [x] Unit tests pass (23/23)
- [x] Manual testing: start a node, verify UI transitions from Starting → Running without restart
- [x] Manual testing: stop a Running node, verify Stopping spinner appears and transitions to Stopped
- [x] Manual testing: verify no spurious "Error starting node" dialog after timeout

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,12 @@
 /packaged_architectures/
 /packaged_bins/
 
+# Terraform
+**/.terraform/
+**/.terraform.lock.hcl
+**/*.tfstate
+**/*.tfstate.*
+
 .claude/plugins
 # Planning files (local session artifacts)
 task_plan.md
--- a/node-launchpad/src/components/status.rs
+++ b/node-launchpad/src/components/status.rs
@@ -55,6 +55,10 @@ use tokio::sync::mpsc::UnboundedSender;
 
 pub const NODE_STAT_UPDATE_INTERVAL: Duration = Duration::from_secs(5);
 pub const NODE_REGISTRY_UPDATE_INTERVAL: Duration = Duration::from_secs(180); // 3 minutes
+/// Faster registry polling when nodes are in transitional states (Starting/Stopping/Updating).
+/// refresh_node_registry(full_refresh=false) skips the peer connection check, so this is
+/// lightweight enough for frequent polling.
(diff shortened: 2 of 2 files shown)


maqi (approved):
1. Constant name
NODE_REGISTRY_STARTING_UPDATE_INTERVAL is used for Starting, Stopping, and Updating. 
A name like NODE_REGISTRY_TRANSITION_UPDATE_INTERVAL (or “transitional”) would better match usage; optional rename.

2. .gitignore change
The only functional change there is adding resources/terraform. 
That’s unrelated to the launchpad fix. 
Prefer either:
 - Moving that to a separate commit/PR, or
 - Mentioning it in the PR description (e.g. “Also ignore resources/terraform”) so it’s clear it’s intentional.

jacderida:
> 1. Constant name
>        NODE_REGISTRY_STARTING_UPDATE_INTERVAL is used for Starting, Stopping, and Updating.
>        A name like NODE_REGISTRY_TRANSITION_UPDATE_INTERVAL (or “transitional”) would better match usage; optional rename.
> 
>     2. .gitignore change
>        The only functional change there is adding resources/terraform.
>        That’s unrelated to the launchpad fix.
>        Prefer either:
> 
> 
>     * Moving that to a separate commit/PR, or
> 
>     * Mentioning it in the PR description (e.g. “Also ignore resources/terraform”) so it’s clear it’s intentional.

I addressed these changes. Will now merge the PR. I will bypass the merge queue because I need those Terraform changes for another branch.

PR #3455, opened 2026-02-09
node(fix): persist merkle uploaded chunks in receipt for resume

maqi:
## Summary
Resolves AUTO-868. Merkle upload retry was re-uploading all chunks because the receipt did not record which chunks were successfully uploaded in previous batches.

## Changes
- Add uploaded: HashSet<XorName> to MerklePaymentReceipt (serialised with #[serde(default)] for backward compatibility).
- In split_existing_and_new_chunks, treat 
eceipt.uploaded as known-existing so resume skips them.
- After each batch (and retries), add newly uploaded chunk names to the receipt and emit send_merkle_batch_payment_complete so the CLI can persist the updated receipt.

On retry with a saved receipt, only chunks that were never successfully uploaded are uploaded again.

### DIFF
--- a/autonomi/src/client/merkle_payments/file.rs
+++ b/autonomi/src/client/merkle_payments/file.rs
@@ -206,6 +206,9 @@ impl Client {
                 "🌳 Merkle Tree {batch_num}/{num_batches}: Uploading {batch_size} chunks..."
             );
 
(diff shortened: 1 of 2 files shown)


maqi:
Replaced by PR https://github.com/maidsafe/autonomi/pull/3456

PR #3459, opened 2026-02-10
feat(node): enable automatic upgrades on Windows

jacderida:
## Summary

- Extend the automatic upgrade mechanism to support Windows using the `self-replace` crate for the rename-and-replace binary pattern
- On Windows, a running executable cannot be overwritten but can be renamed; `self-replace` handles this along with cleanup via `FILE_FLAG_DELETE_ON_CLOSE`
- Add retry logic (3 attempts, 100ms delay) for transient antivirus file locks during binary replacement
- Remove `#[cfg(unix)]` guards from the upgrade module, making auto-upgrade available on all platforms
- Add `RunNodeOutcome` enum and `StopResult::UpgradeSuccess` to distinguish upgrade stops from normal stops
- On Windows, exit with code 100 after upgrade so WinSW (`OnFailure`) restarts the service; on Unix, exit 0 so systemd (`on-success`) restarts
- Change Windows `RestartPolicy` from `Never` to `OnFailure` in node-launchpad and antctl

## Test plan

Testing was performed across all three platforms and all node deployment modes:

- [x] No regressions in service-based node upgrades on Linux
- [x] No regressions in symlink-based node upgrades on Linux
- [x] No regressions in stand-alone node upgrades on Linux
- [x] No regressions in `node-launchpad` node upgrades on Linux
- [x] Windows service-based nodes upgrade correctly
- [x] Windows stand-alone nodes upgrade correctly
- [x] Windows `node-launchpad` nodes upgrade correctly
- [x] No regressions in service-based node upgrades on macOS
- [x] No regressions in stand-alone node upgrades on macOS
- [x] No regressions in `node-launchpad` node upgrades on macOS


maqi (commented):
Concerns

  1. Missing --stop-on-upgrade in service installation args

  ant-service-management/src/node/mod.rs:build_upgrade_install_context — The --stop-on-upgrade flag is not included in the service args built here. This means service-managed nodes will default to stop_on_upgrade=false.

  With stop_on_upgrade=false, after an upgrade, the node sends NodeCtrl::Restart which calls start_new_node_process(), spawning a detached child process. The original process then exits with code 0. On Windows with OnFailure, WinSW sees
  exit 0 and does not restart — but now there's an unmanaged antnode process running outside WinSW's control. The service appears stopped from WinSW's perspective.

  For the exit-code-100 path to actually be reached, --stop-on-upgrade must be passed to service-managed nodes. I'd expect something like:
```
  // In build_upgrade_install_context, before the EVM network args:
  args.push(OsString::from("--stop-on-upgrade"));
```
  This should probably be added unconditionally (not just Windows), since on Linux with systemd Restart=on-success, the same start_new_node_process + exit 0 would cause systemd to restart the service too, resulting in duplicate nodes.

  Was --stop-on-upgrade being passed through the initial add command in testing? If so, the flag needs to be plumbed through build_upgrade_install_context for the upgrade path as well.

  2. OnFailure restart policy may cause restart loops

  ant-service-management/src/node/mod.rs:155 — With RestartPolicy::OnFailure, WinSW will restart the service on any non-zero exit code, not just 100. If the node has a persistent startup error (e.g., port conflict, corrupt data), it will
   enter a restart loop under WinSW.

  Consider whether WinSW supports limiting restart attempts or whether exit code 100 can be specifically targeted. Alternatively, a comment acknowledging this trade-off would be helpful.

  3. std::process::exit(100) skips destructors

  ant-node/src/bin/antnode/main.rs:383-388 — std::process::exit(100) does not run destructors. The _log_appender_guard in init_logging won't be dropped, which could cause the final log messages (including "Node stopped after upgrade") to
   be lost. This is after rt.shutdown_timeout() so the runtime is clean, but log flushing may be incomplete.

  A possible alternative: return a custom exit code from main() instead, e.g., via std::process::ExitCode. However, this is minor since the node is about to restart anyway.

  4. AV retry timing

  ant-node/src/bin/antnode/upgrade/mod.rs:446-448 — 3 retries with a flat 100ms delay may not be sufficient for aggressive AV software (Windows Defender real-time protection can hold locks for several hundred milliseconds). Consider:
  - Exponential backoff (e.g., 100ms, 500ms, 2000ms)
  - Or at least a longer flat delay (e.g., 500ms)

  ---
  Nits

  - ant-protocol/src/node_rpc.rs: The StopResult::UpgradeSuccess doc comment mentions WinSW specifically, but this variant is also used on Unix. Consider making the doc comment platform-neutral: "The service manager should restart the
  process so it picks up the new binary."

maqi (approved):

PR #3467, opened 2026-02-13
fix: use non-virtualized path for winsw.exe in launchpad

jacderida:
## Summary

When node-launchpad is installed via MSIX on Windows, the `%APPDATA%` path is virtualized by the
MSIX filesystem, redirecting file writes to a package-specific location. The service management
code then cannot find `winsw.exe` at the expected standard path, causing node addition to fail
with "Failed to resolve full path of the current executable".

This changes `configure_winsw()` to place `winsw.exe` at `C:\ProgramData\antctl\winsw.exe` (via
`get_node_manager_path()`) instead of the launchpad data directory (via `dirs_next::data_dir()`).
The `C:\ProgramData` path is not subject to MSIX virtualization, so it works for both MSIX and
standalone installations.

- Change is Windows-only (`#[cfg(windows)]`), no impact on other platforms
- Single file changed: `node-launchpad/src/config.rs`

## Test plan

- [x] All 23 node-launchpad unit tests pass
- [x] Clippy clean with no warnings
- [ ] Manual test: Install launchpad via MSIX on Windows, verify node addition works
- [ ] Manual test: Run standalone launchpad binary, verify node addition still works

### DIFF
--- a/node-launchpad/src/config.rs
+++ b/node-launchpad/src/config.rs
@@ -99,11 +99,19 @@ pub fn get_config_dir() -> Result<PathBuf> {
     Ok(config_dir)
 }
 
+/// Download and configure WinSW for Windows service management.
+///
+/// WinSW is placed at `C:\ProgramData\antctl\winsw.exe` (via `get_node_manager_path`) rather than
+/// the launchpad's own data directory. This is necessary because when the launchpad is installed via
+/// MSIX, Windows virtualizes `%APPDATA%` paths, redirecting file writes to an MSIX-specific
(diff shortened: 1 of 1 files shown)


jacderida:
I need to merge this into the RC branch before doing the manual testing because the issue only occurs when running the launchpad via an MSIX installation. A release candidate will be created and this will have an MSIX, at which point I can test it.

I will merge it now rather than waiting a long time for the tests to pass. The test suite will run when I merge the RC branch back in to `main`.

PR #3468, opened 2026-02-17
Various Launchpad Enhancements

jacderida:
## Summary

- **Fix MSIX winsw.exe path virtualization**: Place `winsw.exe` at `C:\ProgramData\antctl\` instead of the launchpad data directory, which is virtualized by MSIX filesystem and invisible to the service management code. Fixes node addition failing on Windows MSIX installs.
- **Fix startup delay with many nodes**: Defer the blocking `refresh_node_registry()` RPC call out of `Status::new()` so the TUI renders immediately with disk-cached state. A background refresh updates to live status afterwards. Previously caused 20+ second startup delays or prevented the app from opening on Windows MSIX.
- **Fix inverted UPnP/Custom Ports display**: Correct the `no_upnp` boolean mapping in `NodeConnectionMode::from()` which had the logic backwards, causing "Custom Ports" to show as "UPnP" on the status grid.
- **Add scrolling and scrollbar to status page node grid**: Switch the node table to `render_stateful_widget` so ratatui auto-scrolls to keep the selected row visible, and add a vertical scrollbar indicator. Previously the grid truncated when nodes exceeded the vertical space.
- **Clean up antnode.exe and antctl.exe in launchpad cleanup script**: Extend the PowerShell cleanup script to remove leftover node and control binaries.

## Test plan
- [x] Unit tests pass (23/23 for node-launchpad)
- [x] Clippy clean with no warnings
- [x] Manual testing of node grid scrolling with many nodes
- [x] Manual testing on Windows MSIX install for winsw.exe path fix

### DIFF
--- a/.claude/commands/prepare-hotfix-branch.md
+++ b/.claude/commands/prepare-hotfix-branch.md
@@ -1,0 +1,217 @@
+# Prepare Hotfix Branch for Release Candidate
+
+@release-cycle-info
+@ant-build-info/src/release_info.rs
+@CHANGELOG.md
+@.github/workflows/release.yml
+
+**CRITICAL**:
+* The current branch must not have any outstanding changes. If this is the case, stop and ask the
+  user to commit those changes.
+* The current branch name should be in the form `rc-YYYY.X.Y-hotfixZ`, where `X`, `Y` and `Z` are
+  integers. If this is not the case, warn the user and get them to create a new branch in this form
+  before proceeding.
+
+The process has the following phases:
+
+* Preparing the new release candidate
+* Determining merged PRs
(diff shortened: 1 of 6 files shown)


Nic-dorman:
**Claude Review:**

1. The scrollbar renders unconditionally - should only show when content overflows
2. Dead data_dir_path variable in configure_winsw() after the path change
3. The powershell script doesn't have -WhatIf support, which is standard PowerShell practice for destructive scripts

**No blocking issues**. The bug fixes are clearly correct, the performance improvement is well-reasoned, and the scrollbar addition is the right pattern even if the implementation could be slightly tightened.

